Issue #50, opened 2018-01-12
improve `jx get pipeline` to display nicer information


jstrachan:
so its a bit more like the jenkins console in terms of:

* display last Run number, status and duration
* dealing with Folders and Pipelines in folders
* dealing with MultiBranchProject jobs with branches + PRs
* view if a pipeline is being built and the run number if its building along with success/fail/elapsed time

PR #70, opened 2018-01-15
#50 display last Run number, status and duration

ndeloof:
require https://github.com/jenkins-x/golang-jenkins/pull/2


jstrachan:
@ndeloof looks great thanks!

PR #250, opened 2018-02-12
Add ‘jx get app’ command 

jstrachan:


michaelneale:
@rawlingsj the CI seems to be wedged here

rawlingsj:
Ah thanks @michaelneale - we ran out of disk space yesterday so this job must have got stuck then.  I've rebased and that's triggered the CI test again.

PR #255, opened 2018-02-13
bdd pipeline fixes

rawlingsj:


pkg/jx/cmd/util/factory.go
```diff
@@ -70,7 +71,7 @@ func (f *factory) SetBatch(batch bool) {
 
 // CreateJenkinsClient creates a new jenkins client
 func (f *factory) CreateJenkinsClient() (*gojenkins.Jenkins, error) {
-	url := os.Getenv("JENKINS_URL")
```
rawlingsj: This causes the jenkins auth config file to use the internal jenkins service URL when run in cluster.

PR #257, opened 2018-02-13
new command to delete pipelines

### LINKED ISSUE #123: Need to be able to delete an app/pipeline

ndeloof:
close #123

Signed-off-by: Nicolas De Loof <nicolas.deloof@gmail.com>

### DIFF
--- a/pkg/jx/cmd/delete.go
+++ b/pkg/jx/cmd/delete.go
@@ -55,6 +55,7 @@ func NewCmdDelete(f cmdutil.Factory, out io.Writer, errOut io.Writer) *cobra.Com
 	cmd.AddCommand(NewCmdDeleteEnv(f, out, errOut))
 	cmd.AddCommand(NewCmdDeleteGit(f, out, errOut))
 	cmd.AddCommand(NewCmdDeleteJenkins(f, out, errOut))
+	cmd.AddCommand(NewCmdDeletePipeline(f, out, errOut))
 	cmd.AddCommand(NewCmdDeleteRepo(f, out, errOut))
 	return cmd
 }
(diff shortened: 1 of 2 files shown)


michaelneale:
nice!

I wasn't able to see it work though: 

```
./jx delete pipeline michaelneale/demo/master
```

returned success but pipeline remains

ndeloof:
bug in golang-jenkins ?
isn't this a multibranch pipeline ? Will then be automatically re-created by branch indexing, right ?

michaelneale:
@ndeloof it is multibranch - but I would think it wouldn't recreate it instantly, and only the branch. 

I guess this probably could be replaced by "delete app" given get app command is sort-of working. 
This would cover more than the pipeline - but the whole definition, the repo etc.. so not entirely clear what the scope of it is (as it is - I think deleteJob isn't working)

jstrachan:
I wonder if we should even let folks delete pipelines inside a MultiBranchProject - as the MBP is just gonna recreate it next time it scans? Maybe we should warn the user or something?

Deleting the MBP sounds a nice idea though; for when folks want to remove projects from their jenkins

michaelneale:
@jstrachan yeah, deleting a pipeline per se is probably not the right thing to delete. 
There is a desire I have to delete *something* - ie I made a new project, imported, but I no longer want to do it. How do I clean all that up so there is no trace of it, so it isn't using resources? I can delete the project repo, the environment repos, but the pipelines and any running apps/environments will stay running? Is what I am missing more like a cleanup command? Should things be deleted when repos are deleted?

michaelneale:
Should we close this? 

There is still a need to remove things, like say you don't care about an app anymore, you will want to remove the pipeline, but also any other resources too.

PR #271, opened 2018-02-19
refactor to avoid core dump when using jx create cluster minikube

jstrachan:


pkg/jx/cmd/create_cluster_gke.go
```diff
@@ -251,42 +245,7 @@ func (o *CreateClusterGKEOptions) createClusterGKE() error {
...
-					Out:       o.Out,
-					Err:       o.Err,
-					Headless:  true,
-					BatchMode: true,
-				},
-			},
-		},
-	}
+	installOpts := &o.InstallOptions
 	err = installOpts.Run()
```
rawlingsj: Does removing this mean we still get default environments created on `jx create cluster gke` and the jenkins API token generated?  I had a hard time getting the environments created without having to add multiple common options which also set headless and batch commands.

jstrachan: yeah - seems to work! :)

PR #289, opened 2018-02-22
Fix proportions of logo in README

jlong:
Should be self explanatory. Let me know if there are questions.


rawlingsj:
great - thanks!

PR #325, opened 2018-03-01
upgrade only does platform now it seems - and does not delete

michaelneale:
Was some missing stuff to make it clearer

cc @rawlingsj

### DIFF
--- a/pkg/jx/cmd/upgrade.go
+++ b/pkg/jx/cmd/upgrade.go
@@ -4,7 +4,7 @@ import (
 	"io"
 
 	"github.com/spf13/cobra"
-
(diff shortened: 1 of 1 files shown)


michaelneale:
@rawlingsj @jstrachan not sure why the pr-merge is failing (what is that job about anyway?)

jstrachan:
@michaelneale looks like Jenkins X is out of disk again - lemme fix & retrigger

jstrachan:
@michaelneale phew thats looking better - fingers crossed it works ;)

jstrachan:
@michaelneale using `gofmt` in your editor sounds good btw - I also do `make fmt` from time to time also

PR #408, opened 2018-03-15
fix(install): avoid logging api token fixes #405

### LINKED ISSUE #405: jx install shows API keys for GitHub and Jenkins
# Description 

During `jx install` the input lines that ask for API Key show the value input rather than hiding it as I would expect for passwords: 

# Example

See the `? API Token:` lines below.

## GitHub

```
Lets set up a git username and API token to be able to perform CI / CD

To be able to create a repository on GitHub we need an API Token
Please click this URL https://github.com/settings/tokens/new?scopes=repo,read:user,user:email,write:repo_hook

Then COPY the token and enter in into the form below:

? GitHub user name: bitwiseman
? API Token: 415c5a9ad6e14dbffd3b0f12c4ecd7ce3af9a185
Cloning the Jenkins X cloud environments repo to /Users/bitwiseman/.jx/cloud-environments
```

## Jenkins

```
	Getting Jenkins API Token
Please go to http://jenkins.jx.192.168.99.100.nip.io/me/configure and click Show API Token to get your API Token
Then COPY the token and enter in into the form below:

? API Token: 8081036810181178236625439f0ab31d
Created user admin API Token for Jenkins server  at http://jenkins.jx.192.168.99.100.nip.io
```

rawlingsj:


bitwiseman:
@rawlingsj 
It is not just the logging but showing them when they are input.

michaelneale:
@bitwiseman so this fixes the logging of the Jenkins one and the github one, but yeah, not the input - so almost there. 

UNfortunately my knowledge of go here fails, but it looks not easy to do: something like https://stackoverflow.com/questions/2137357/getpasswd-functionality-in-go is needed when asking for the github personal token.

rawlingsj:
@michaelneale yeah we use something called https://github.com/AlecAivazis/survey#password which has a password type that masks input which works well.

PR #454, opened 2018-03-20
Remove unnecessary newlines from create cluster command help message

gtseres:
Removes the extra newlines from the `jx create cluster --help` command output:

```
  * draft (CLI that makes it easy to build applications that run on kubernetes) \n
  * minikube (single-node Kubernetes cluster inside a VM on your laptop ) \n
  * virtualisation drivers (to run minikube in a VM) \n
```


jstrachan:
@gtseres many thanks!

PR #460, opened 2018-03-21
fixes link to command line reference doc

aaronwalker:
The current link redirects to the old docs which I guess isn't where it should go :)


jstrachan:
@aaronwalker thanks!

PR #475, opened 2018-03-22
Sanitize user name label when creating GKE cluster

jlordiales:
Hi there!

Amazing project, really looking forward to see what I can contribute with.

This tries to fix an issue I had today while trying out the `jx create cluster gke` command. My user name is "firstname.lastname" on mac so the command failed with 

`ERROR: (gcloud.container.clusters.create) ResponseError: code=400, message=invalid field 'cluster.resource_labels.value': "jose.ordiales". It must begin with a lowercase character ([a-z]), end with a lowercase alphanumeric ([a-z0-9]) with dashes (-), and lowercase alphanumeric ([a-z0-9]) between.`

The sanitize function just replaces any non-allowed characters with `-` but I guess it could be made smarter if there's a need.

I only applied this to the "created-by" label because the user doesn't have a lot of control over that one. The ones passed through the CLI are set explicitly so maybe it's ok for the command to fail when those are not valid.

### DIFF
--- a/pkg/jx/cmd/create_cluster_gke.go
+++ b/pkg/jx/cmd/create_cluster_gke.go
@@ -19,6 +19,7 @@ import (
 	"github.com/jenkins-x/jx/pkg/util"
 	"github.com/spf13/cobra"
 	"gopkg.in/AlecAivazis/survey.v1"
+	"regexp"
 )
 
 // CreateClusterOptions the flags for running crest cluster
@@ -67,6 +68,7 @@ var (
 		jx create cluster gke
 
 `)
+	disallowedLabelCharacters = regexp.MustCompile("[^a-z0-9-]")
 )
 
 // NewCmdGet creates a command object for the generic "init" action, which
@@ -223,7 +225,7 @@ func (o *CreateClusterGKEOptions) createClusterGKE() error {
 	labels := o.Flags.Labels
 	user, err := os_user.Current()
 	if err == nil && user != nil {
-		username := user.Username
+		username := sanitizeLabel(user.Username)
 		if username != "" {
 			sep := ""
 			if labels != "" {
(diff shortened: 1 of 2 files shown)


jlordiales:
Ok, I can see that the build failed but can't really see the details why (can't login to https://jenkins.build.cd.jenkins-x.io/job/jenkins-x/job/jx/job/PR-475/1). `make build test` works locally though

rawlingsj:
this is great thanks!  

ok sorry about the locked down CI at the moment, soon we'll have integration with jenkins credentials and kubernetes secrets but for now to make sure no secrets can be leaked in the console we've turned off anonymous read.

the error looks like a missing dependency..

https://github.com/jenkins-x/jx/pull/475/commits/0b016a2ed0c1366dafa955dbfe0594759f0f73fd#diff-b82437227a0d24642d60c26dd219329fR4

```
# github.com/jenkins-x/jx/pkg/jx/cmd

pkg/jx/cmd/create_cluster_gke_test.go:4:2: cannot find package "github.com/magiconair/properties/assert" in any of:

	/home/jenkins/go/src/github.com/jenkins-x/jx/vendor/github.com/magiconair/properties/assert (vendor tree)

	/usr/local/go/src/github.com/magiconair/properties/assert (from $GOROOT)

	/home/jenkins/go/src/github.com/magiconair/properties/assert (from $GOPATH)

FAIL	github.com/jenkins-x/jx/pkg/jx/cmd [setup failed]

```

jlordiales:
Ah! wrong import... will fix now

rawlingsj:
LGTM - many thanks @jlordiales!

rawlingsj:
@jlordiales 1.1.21 should be available in ~20 mins with you change in

PR #478, opened 2018-03-22
Align list of valid resource types for create

jlordiales:
Sorry, just a small OCD thing.

```
jx create options
Creates a new resource.

  Valid resource types include:

  * archetype
  * cluster
  * env
  * git
    * spring (aka 'springboot')
```

The first 4 lines were tab indented and the last one space indented :)


rawlingsj:
:) thanks!

PR #486, opened 2018-03-23
Always include user account as part of organisation list

jlordiales:
This fixes an issue I was discussing yesterday on the Slack channel with @jstrachan, where if the user belongs to just 1 organisation then that organisation was always used by default to create repos.

### DIFF
--- a/pkg/gits/provider.go
+++ b/pkg/gits/provider.go
@@ -11,8 +11,12 @@ import (
 	"gopkg.in/AlecAivazis/survey.v1"
 )
 
-type GitProvider interface {
+type OrganisationLister interface {
 	ListOrganisations() ([]GitOrganisation, error)
+}
+
+type GitProvider interface {
+	OrganisationLister
 
 	ListRepositories(org string) ([]*GitRepository, error)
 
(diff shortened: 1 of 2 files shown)


jstrachan:
@jlordiales thats awesome thanks! Love the extra test too; we're badly need more tests :) 

BTW any chance you could rebase your PR - we merged a few PRs this morning? Failing that we could try the `Update branch` button

pkg/gits/provider.go
```diff
@@ -12,8 +12,12 @@ import (
 	"gopkg.in/AlecAivazis/survey.v1"
 )
 
-type GitProvider interface {
+type OrganisationLister interface {
 	ListOrganisations() ([]GitOrganisation, error)
+}
+
+type GitProvider interface {
```
jlordiales: I'm hoping to get some feedback from people with more experience in Go here.

Coming from Java the GitProvider interface looks quite standard. But interfaces with lots of methods seem to be considered an anti-pattern in Go (https://blog.chewxy.com/2018/03/18/golang-interfaces/, https://nathanleclaire.com/blog/2015/10/10/interfaces-and-composition-for-effective-unit-testing-in-golang/), mainly because they make testing a lot harder.
In this particular example, `PickOrganisation` only needs something that knows how to list organisations. And by passing this new interface it is possible to easily mock and write unit test (see the added test in this PR).

What are people views on this? Interface composition would still allow to keep the current GitProvider interface and  have smaller interfaces to use where needed.

jlordiales:
I messed the rebase. Fixing

pkg/gits/provider.go
jstrachan: sounds good to me!

jstrachan:
thanks!

PR #490, opened 2018-03-23
Add ability to output terraform configuration to a specific directory

whoisjake:
This gives jx the ability to spit out [terraform](https://www.terraform.io/) configuration when creating a cluster on aws via kops. This is extremely useful for using configuration management of an infrastructure.

```
$ jx create cluster aws --terraform /path/to/output
```


jstrachan:
@whoisjake many thanks!

Issue #36, opened 2018-01-11
jx import: should support creating new repositories on bitbucket.org


jstrachan:
we have a GitProvider interface https://github.com/jenkins-x/jx/blob/master/pkg/gits/provider.go#L14

we need to implement that for bitbucket

wbrefvem:
@jstrachan I'd like to take a crack at this. At a glance, there's no official go client for bitbucket but there is a swagger/OAI spec and some third-party clients (none of which, as far as I can tell, are based on the spec). So the implementation will differ slightly from the github provider. My feeling is that the bitbucket provider should be based on the auto-generated client. Here's what it looks like: https://github.com/wbrefvem/go-bitbucket

Thoughts?

jstrachan:
@wbrefvem sounds awesome thanks! Let us know if you need a hand with anything!

I'm just fixing up a few gremlins right now with the support of multiple git providers to make the UX a little more slick - getting closer to `gitea` being a peer of github - so hopefully you won't hit any issues with the rest of jenkins x by the time you get that far ;)

Would be good to have a test suite for git providers really too one of these days ;)

As an aside, we've a little [chart for gitea](https://github.com/jenkins-x/charts/tree/gitea/incubator/gitea) so it can be installed via [jx create addon gitea](http://jenkins-x.io/commands/jx_create_addon_gitea/); I wonder if we should try hack a bitbucket chart for cases where folks want to install it inside the kubernetes cluster too?

The license key stuff adds a bit of extra complication though; so lets not worry about that yet (e.g. so we can point jenkins X at an external bitbucket server or hosted service etc). But at some point we may want to add bitbucket as a helm chart + addon etc. Looks like there's a docker image already:
https://bitbucket.org/atlassian/docker-atlassian-bitbucket-server

jstrachan:
in terms of test kits, I guess we could maybe just reuse the current [BDD tests](https://github.com/jenkins-x/godog-jx) to ensure we can create/import/PR/release/raise issues in Jenkins X and have a flavour of the tests for each git provider. We could just then enhance the BDD tests to

* validate the release notes are generated fine
* perform a PR after the project is built and verify CI runs on it and that when its merged it re-releases a new version

aairey:
side-note: I would be more interested in [BitBucket Server](https://developer.atlassian.com/server/bitbucket/how-tos/command-line-rest/) for our setup.
Should we create a separate issue for this as the API differs from bitbucket.org?

[...]

PR #497, opened 2018-03-24
More implementations of Bitbucket GitProvider methods

wbrefvem:
ref #36


jstrachan:
@wbrefvem many thanks!

PR #503, opened 2018-03-26
More work on #36 plus vendoring

wbrefvem:


jstrachan:
looks like a compile error:
```
pkg/gits/bitbucket.go:258:19: cannot use commit["hash"] (type interface {}) as type string in assignment: need type assertion
```

wbrefvem:
Whoops. :)

wbrefvem:
@jstrachan This is what I'm seeing locally from `make test`:

```
pkg/jx/cmd/common_issues.go:8:2: cannot find package "github.com/jenkins-x/jx/pkg/issues" in any of:
	GOPATH/src/github.com/wbrefvem/jx/vendor/github.com/jenkins-x/jx/pkg/issues (vendor tree)
	/usr/local/Cellar/go/1.10/libexec/src/github.com/jenkins-x/jx/pkg/issues (from $GOROOT)
	$GOPATH/src/github.com/jenkins-x/jx/pkg/issues (from $GOPATH)
```

I can reset to the tip of `jenkins-x:master` and still get the error, where I didn't before. No errors and tests pass when I test individual packages like `pkg/gits/` that similarly depend on packages under `jenkins-x/jx`, which is neither vendored nor locally cloned. 

I'm stumped. Any thoughts?

jstrachan:
@wbrefvem phew - I was stumped there too ;) 

BTW when I'm working on my fork of jenkins-x/jx I still clone my fork to `$GOROOT/src/github.com/jenkins-x/jx` to avoid issues with imports etc

wbrefvem:
Aha, that makes sense. That's a much cleaner solution!

michaelneale:
he I was thinking of suggesting that, but I figured it was the lack of my Go foo that made me put it in $GOROOT

PR #581, opened 2018-04-05
Enhance help message with note for WIP quickstarts

gtseres:
"WIP-" quickstarts are excluded from the shown quickstarts, but this is not something currently depicted in the help message of the command.


gtseres:
I have put the message in the `LongDesc`. Is it the right place?

jstrachan:
@gtseres looks great - thanks!

PR #587, opened 2018-04-07
Implements remaining BitbucketCloudProvider methods, closes #36

### LINKED ISSUE #36: jx import: should support creating new repositories on bitbucket.org

wbrefvem:
Some things aren't supported by the Bitbucket Cloud API, and as such the following methods throw errors:

* `AddPRComment`
* `CreateIssueComment`
* `UpdateRelease`

The first two can be done by POSTing to the same URLs that the web interface uses. This method doesn't support app passwords, however, so we'd have to get the user's web password. 2FA would then have to be dealt with. Releases are simply not supported on Bitbucket Cloud. There is a notion of downloads, but they don't support tagging, release notes, change logs, or really any of the fields of a `GitRelease`

### DIFF
--- a/pkg/gits/bitbucket.go
+++ b/pkg/gits/bitbucket.go
@@ -446,12 +446,83 @@ func (b *BitbucketCloudProvider) CreateWebHook(data *GitWebHookArguments) error 
 	return nil
 }
 
+func BitbucketIssueToGitIssue(bIssue bitbucket.Issue) *GitIssue {
+	id := int(bIssue.Id)
+	ownerAndRepo := strings.Split(bIssue.Repository.FullName, "/")
+	owner := ownerAndRepo[0]
+
+	var assignee GitUser
+
+	if bIssue.Assignee != nil {
+		assignee = GitUser{
+			URL:   bIssue.Assignee.Links.Self.Href,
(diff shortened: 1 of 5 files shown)


jstrachan:
Adding comments to issues & PRa is not essential & we could use a chat bot instead. It’s just some extra feedback really on a preview or promotion.

Generating releases is similar really - for that we could longer term just deploy the markdown file into a git repo / branch so at least the data is somewhere.

Eg maybe for git providers that don’t support release notes we use the convention that we lazily create a repo called `release-notes` in the same user/org as the repo being released - then in that repo we add a dir for each repo name & add each release note as a `$version.md` file.

Folks would then have a git wiki of release notes etc & a URZl to view them?

For now though these are all nice to have feedback - it’s much more important for folks to be able to do PRs, releases & promotions - so it’d be good not to throw errors if these 3 methods cannot be implemented- a warning is fine

wbrefvem:
@jstrachan duly noted. I'll probably get to that next, but I've opened up an issue in case I forget. ;)

Issue #622, opened 2018-04-12
jx import should query the new pipeline git secrets to find the correct credential


jstrachan:
if you have multiple git servers in an installation, there can be different credentials. So `jx import` should query the `Secrets` for the pipeline git credential for the right git server URL and use that by default - if none is found default to `jenkins-x-git`. 

Then we'll use the correct secret for the imported projects git URL automatically

jstrachan:
Yeah - we kinda do that now when creating new tokens for git - we create a secret with labels & annotations which define the git server kind, name, url etc so it’s easy to know what credential name we should use

PR #630, opened 2018-04-15
fix: jx import creates Jenkins git credentials with ID

wbrefvem:
see #622 

@jstrachan not sure if this is what you had in mind for 622 so I left it open, but I think this fixes it.

### DIFF
--- a/pkg/jx/cmd/import.go
+++ b/pkg/jx/cmd/import.go
@@ -616,6 +616,11 @@ func (o *ImportOptions) DoImport() error {
 	if jenkinsfile == "" {
 		jenkinsfile = jenkins.DefaultJenkinsfile
 	}
+
+	if o.Credentials == "" {
+		// e.g. user-github or user-bitbucket
+		o.Credentials = authConfigSvc.Config().DefaultUsername + "-" + gitProvider.Label()
+	}
 	return jenkins.ImportProject(o.Out, o.Jenkins, gitURL, o.Dir, jenkinsfile, o.BranchPattern, o.Credentials, false, gitProvider, authConfigSvc, false, o.BatchMode)
 }
 


jstrachan:
Thanks! 

I was thinking of actually querying all the available pipeline Secrets of kind `git` and find the one for the same git server URL - if one is not found then prompt the user to add one. 

Lemme take a stab at this one as I know the secret loading code pretty well...

wbrefvem:
Sure thing. Wasn't totally sure I understood what you were getting at, so I'm interested to see how you approach this.

jstrachan:
@wbrefvem still polishing, but a heads up: https://github.com/jstrachan/jx/blob/malarkey/pkg/jx/cmd/common_import.go#L60-L62

which then calls:
https://github.com/jstrachan/jx/blob/malarkey/pkg/jx/cmd/common_import.go#L196-L212

This branch also works if you try something like:
```
jx import --url https://some-new-repo.com/foo/bar.got
```
where the user has never called `jx create git server ....` for `some-new-repo.com` and so incrementally sets up the server, auth + credentials on the fly

wbrefvem:
@jstrachan that makes a lot more sense. IIUC, you're using k8s as a store for secrets not necessarily used by k8s, right?

jstrachan:
yeah; we use this plugin: https://github.com/jenkinsci/kubernetes-credentials-provider-plugin so that we can use `Secrets` to manage the available `Credentials` in Jenkins; which means we can use GitOps / Helm to manage them etc.

Then the advantage of a `Secret` is we can add extra labels and annotations to it. e.g to indicate a `Secret/Credential` is for git and if so what git kind, server name and server URL is it for

wbrefvem:
Okay, I think I grok this now. Thanks for the help!

PR #700, opened 2018-04-23
Add more information for preview environments

kzantow:
For some use cases, we need additional information stored in K8s resources pertaining to the users and pull requests where preview environments were created, this seems to suffice to provide enough data.

@jstrachan PTAL

### DIFF
--- a/pkg/gits/bitbucket.go
+++ b/pkg/gits/bitbucket.go
@@ -9,7 +9,9 @@ import (
 
 	"github.com/jenkins-x/jx/pkg/util"
 
+	"github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/auth"
+	"github.com/jenkins-x/jx/pkg/jx/cmd/log"
 	"github.com/wbrefvem/go-bitbucket"
 )
 
@@ -364,6 +366,8 @@ func (b *BitbucketCloudProvider) UpdatePullRequestStatus(pr *GitPullRequest) err
 	}
 
 	pr.State = &bitbucketPR.State
+	pr.Title = bitbucketPR.Title
+	pr.Body = bitbucketPR.Summary.Raw
 
 	if bitbucketPR.MergeCommit != nil {
 		pr.MergeCommitSHA = &bitbucketPR.MergeCommit.Hash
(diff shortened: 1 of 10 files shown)


pkg/jx/cmd/preview.go
```diff
@@ -236,6 +237,41 @@ func (o *PreviewOptions) Run() error {
...
+
+	if err != nil {
+		log.Warn("Unable to get statuses for PR " + prName + "\n")
+	}
+
+	buildStatus := ""
+	buildStatusUrl := ""
+	if len(statuses) > 0 {
+		status := statuses[len(statuses)-1]
+		buildStatus = status.State
```
kzantow: It's unclear what to do here, none of the data returned _seems_ to be a reasonable way to actually identify the status (E.g. "failed" "building" etc.)

jstrachan: each URL has its own status as folks could add a jenkins + sonarqube check for example and each one can be triggered N times  etc. All we can know is that for each unique URL they all pass or one of them fails

kzantow: @jstrachan it's actually more complicated than that. we could retrigger a build manually due to some environmental failure, which results in success but the old statuses remain in the history, so it would always show failure despite being passing if we did that

pkg/gits/bitbucket.go
```diff
@@ -389,6 +393,16 @@ func (b *BitbucketCloudProvider) UpdatePullRequestStatus(pr *GitPullRequest) err
 	return nil
 }
 
+func (p *BitbucketCloudProvider) GetPullRequest(owner, repo string, number int) (*GitPullRequest, error) {
```
jstrachan: not a biggie but seen as GetPullRequest is a simple helper method wrapping UpdatePullRequestStatus we could just add it as a generic helper method - not a biggie though

jstrachan:
@kzantow looks great - many thanks!

pkg/gits/bitbucket.go
kzantow: This is your baby, it just seems much more intuitive to have a `GetPullRequest`?

jstrachan: I was just leaning towards the poor GitProvider implementor having more work to do ;) Not a biggie though

PR #721, opened 2018-04-26
Added ability to use g1-small nodes in GKE

garethjevans:
When using the GCE trial account, i'd like the "free" allowance to last a bit longer, g1-small's should be about as low in terms of available memory/cpu as you can go - I think f1-micros are too small.


rawlingsj:
whoop!

PR #746, opened 2018-04-30
Refactoring too generic function argument name

hekonsek:
Refactored `arg` => `addon` to make code more readable.


hekonsek:
Many thanks, Rob!

PR #753, opened 2018-05-02
Added support for installing generic Helm Charts add-ons

hekonsek:
Hi,

Right now jx comes with a hard-coded list [1] of add-ons that can be installed into cluster. Can you consider adding some more generic ways of installing new stuff into jx-managed cluster?

In particular it will be useful to be able to indicate Helm Chart you want to install as addon, instead of being limited to Grafana and Prometheus. I was thinking about something like...

    jx create addon helm:stable/kibana

...which allows me to install any chart from standard charts repository [1].

I love idea of being able to switch between clusters in jx and installing various add-ons depending on the needs, but I believe that add-ons mechanism should be more similar to plugins mechanism, then to predefined hardcoded list.

What do you think?

[1] https://github.com/jenkins-x/jx/blob/ce05417f3fdb9f02167be8cfba700b82b3200aed/pkg/kube/constants.go#L140
[2] https://github.com/kubernetes/charts

### DIFF
--- a/pkg/jx/cmd/create_addon.go
+++ b/pkg/jx/cmd/create_addon.go
@@ -8,6 +8,7 @@ import (
 	"github.com/jenkins-x/jx/pkg/kube"
 	"github.com/jenkins-x/jx/pkg/util"
 	"github.com/spf13/cobra"
+	"strings"
 )
 
 // CreateAddonOptions the options for the create spring command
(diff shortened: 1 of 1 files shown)


rajdavies:
We are actually going to extend the addons in the same way draft do plugins - see https://github.com/Azure/draft/blob/master/docs/reference/dep-005.md - which will fix this issue in a more generic way. Would you be able to wait until this is done ?

hekonsek:
I wanted to propose supporting add-ons defined in external GitHub repo as well, so it is great to hear that you plan to implement plugins this way. +1

Yes, I can wait. I will close this pull request and wait for more generic approach. BTW Any estimate on when do you plan to introduce those plugins?

PR #768, opened 2018-05-07
Running commands should not be verbose by default

hekonsek:
Hi,

Currently `jx` is kinda "chatty" ;) . For example creating new minikube cluster generates the following output:

```
Starting local Kubernetes v1.7.3 cluster...
Starting VM...
oving files into cluster...
Setting up certs...
Starting cluster components...
Connecting to cluster...
Setting up kubeconfig...
Kubectl is now configured to use the cluster.
```

While by default the following should be more than enough:

```
Creating Minikube cluster...
Minikube cluster created.
```

Jx already comes with common `--verbose` flag set to `false` by default, so I propose to start respecting it more.

In particular I would like to start by changing `CommonOptions#runCommand` not to redirect std out and err in non-verbose mode and limit it only to displaying error message. Std out and err should be redirected to console only in verbose mode.

From what I know I'm not the only who thinks that `jx` tells us too much by default and it will be better to make it less chatty.

What do you think?

### DIFF
--- a/pkg/jx/cmd/common_commands.go
+++ b/pkg/jx/cmd/common_commands.go
@@ -23,8 +23,10 @@ func (o *CommonOptions) runCommandFromDir(dir, name string, args ...string) erro
 
 func (o *CommonOptions) runCommand(name string, args ...string) error {
 	e := exec.Command(name, args...)
-	e.Stdout = o.Out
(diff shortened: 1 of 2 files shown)


hekonsek:
Yay! Thanks! :)

Issue #470, opened 2018-03-22
Suggestion: `jx console` defaulting to BlueOcean


mumoshu:
## User story

`jx console` as of today seems to open e.g. `http://jenkins.jx.192.168.99.100.nip.io/` which opens the dashboard of the traditional Jenkins UI.

As being a total Jenkins newb myself(not actually as I do have been using it in my paid jobs for years but I personally had not tended to tweak admin settings thru UI not to break things...) and also a company-internal ambassador of jx, I'd like everyone including me in my organization to get used to Jenkins through a friendly UI like so-called "modern" CI services like Travis, CircleCI, and so on.

## Suggested solution

I'm really hesitant to say things like this but honestly - can we just default it to `http://jenkins.jx.192.168.99.100.nip.io/blue/pipelines/blue/pipelines/` so that it opens up BlueOcean? 😃

## Acknowledge

Regardless, thank you very much for starting a great project!

I had been struggling long about how to get people adapt to Jenkins due to my inability to tell people how to use it correctly(in my own definition - like "don't modify settings at any level via UI whenever possible", "code everything from pipelines to Jenkins cluster settings", "use container images for build reproducibility everywhere").
And I have ended up combining two or more CI systems in multiple occasions - like Jenkins for veterans and others for new-comers.

It has been a frustration to me though - even though I LOVE all CI systems I've personally adopted for their distinct sweet-spots, for longer-term productivity throughout the whole organization, I still believed that converging to a single versatile CI is important.

`jx` seems to be the way to go towards my goal. Please keep up the great work, and please let me know if I can help you in any way to accelerate this evolution! Thx.

jstrachan:
@mumoshu great idea thanks!

gtseres:
I could give this a try. Is the [GetServiceUrl](https://github.com/jenkins-x/jx/blob/287cb34e775835daf1150d49aeef2d76fbf309a5/pkg/kube/services.go#L91) that should be updated?

jstrachan:
@gtseres ooh that'd be awesome yeah, thanks!

I guess we should tweak this function to take an optional extra path within the service:
https://github.com/jenkins-x/jx/blob/master/pkg/jx/cmd/console.go#L71

then after the URL is found for the service (which typically has no path associated with it) then we should append the path if there is one. For most services there won't be - but for Jenkins we can then use nicer default paths etc.

[...]

PR #771, opened 2018-05-08
Default to BlueOcean

bardec:
This PR addresses #470 - which asks to use BlueOcean as a default for `jx console`. From @jstrachan's idea, I went ahead and added a `--classic` for people that would still like to use the original skin (sadists 😜). 

I decided to keep this option without modifying `Open` as was discussed, as I felt this was currently a pretty specific use case, and would have involved opening a pretty big can of worms (at what level
 should something know if it has a path?). `Open` gets used a lot, so I decided to just keep this in the `console` command. The downside of this is that `jx open jenkins` does not open Blue, but the classic skin. It might get messy supporting this - as open would have to allow a special call for `jx open jenkins` :/

Let me know if anyone has ideas here!

### DIFF
--- a/pkg/jx/cmd/console.go
+++ b/pkg/jx/cmd/console.go
@@ -32,7 +32,10 @@ var (
 		jx console
 
 		# Print the Jenkins X console URL but do not open a browser
-		jx console -u`)
+		jx console -u
+		
+		# Open the Jenkins X console in a browser using the classic skin
+		jx console --classic`)
 )
 
 func NewCmdConsole(f cmdutil.Factory, out io.Writer, errOut io.Writer) *cobra.Command {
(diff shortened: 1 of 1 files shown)


bardec:
we should probably also update corresponding redirects for the GitHub Check details (and whatever else there is). If anyone could point me in the right direction that would be awesome :)

jstrachan:
@bardec many thanks - this is awesome!

On the GitHub Check stuff, I think thats probably configured inside Jenkins somewhere - haven't spotted where yet ;) Probably something in the config.xml needs tweaking I guess https://github.com/jenkins-x/charts/blob/jenkins/stable/jenkins/templates/config.yaml#L9 - will have to experiment where that URL comes from

bardec:
Awesome, I'll dig into it - no promises but maybe I can 🔭 some config values.

jstrachan:
if you find a config I can help figure out the helm / ConfigMap stuff to get it switched

PR #772, opened 2018-05-08
Minikube: Prompt for VM driver should not be invoked if driver option has been used

hekonsek:
When creating Minikube cluster prompt for VM driver should not be invoked if driver option has been used. 

Basically if I execute `jx create cluster minikube --vm-driver=kvm2` I don't want jx to prompt me again for VM driver, but with `kvm2` selected as default. I just want jx to use `kvm2` instead.

PR #789, opened 2018-05-09
Minikube cpu/memory cmd options should not invoke prompt

hekonsek:
This is continuation of #772. Choosing cpu and memory via cmd options should not invoke input prompt. In particular executing command like `jx create cluster minikube --cpu=4 --memory=8000 --vm-driver=kvm2` should work in a kinda "batch mode".


hekonsek:
Thanks, James!

PR #795, opened 2018-05-10
Ignore vendor and editor files

sanketjpatel:


jstrachan:
we've been purposely checking in the vendor files; so its easier for folks developing on jx and add a bit of stability to the versions of things we're all building from - as we often use the latest of some dependencies etc. Not sure we want to exclude them?

jstrachan:
btw I've added the `.vs-code .gitignore` as part of this PR https://github.com/jenkins-x/jx/pull/797

I'm still a fan of including the vendoring stuff; its just a little easier for folks to develop jx at the cost of a larger git clone/pull ;)

hekonsek:
I agree with James that we should vendor dependencies in our Git repository. However I recommend to do it via govendor - #800 .

sanketjpatel:
Agree with @jstrachan and @hekonsek 
Closing this PR

PR #800, opened 2018-05-11
Added govendor support

hekonsek:
Hi,

When looking at #795 I realized that jx doesn't use govendor [1]. I believe that vendoring dependencies is a way to go in GoLang, but I highly recommend doing this with govendor. The main reasons to do so:
- govendor keeps Git revision you vendored (in `vendor/vendor.json` file). For example - https://github.com/jenkins-x/jx/compare/master...hekonsek:govendor?expand=1#diff-bd290170e2912d3e8694db1a151066e5R8 . This is particularly useful if you are vendoring from master and allows you to keep history of your library updates.
- it removes tests, Travis, gitignore and other non-useful files from the libraries you vendor
- adding/updating vendored dependency is as simple as `govendor fetch github.com/spf13/cobra`

In this particular PR I have registered spf/13cobra as govendor dependencies. As you can see we now know which revision has been vendored from master. And it removed some garbage from vendored library making jx Git repo smaller.

What do you think?

[1] https://github.com/kardianos/govendor


sanketjpatel:
Hi @hekonsek 
Thank you for suggesting `govendor`.
I added all the dependencies to `vendor.json` here.
https://github.com/jenkins-x/jx/pull/804
Can you please take a look?

hekonsek:
@rawlingsj Actually this PR is "safer" version of #804. In my opinion we should not migrate to govendor in a single big PR. Instead we should add initial `vendor.json` file and then vendor every dependency one by one. PR #804 modifies almost 10000 files in a single commit - IMHO we should try to avoid such massive changes in a single PR.

The great thing about govendor is that it allows you to migrate existing projects with many dependencies gradually, you don't have to migrate every dependency in a one go. jx should take advantage of this.

If you guys merge this PR, then I (and maybe @sanketjpatel) could help you with migrating other dependencies in subsequent PRs.

hekonsek:
Any updates on this? Anything else I could do to help merge it? :)

rawlingsj:
I'm wondering if we're considering to move away from Glide if we should look at `dep` as I've heard folks suggest this is what people are moving towards https://github.com/golang/dep.  It would be good to understand the pros/cons.

PR #828, opened 2018-05-16
[WIP] #818 Add Users to K8s resources with best-guess Email

### LINKED ISSUE #818: Consistent user information across preview / issues / commits
Although it may not be possible in all cases due to random authors in git commits, these should have a consistent shape, including information like avatar, web link, name, etc.

kzantow:
This should resolve #818  when complete by adding UserDetails as an independent K8s resource type, with some smarts to get the user's email populated. This will be kept up-to-date as the user submits pull requests, currently, and should probably be the source of truth for all user data instead of duplicating it across other K8s resource types.


pkg/gits/git.go
```diff
@@ -173,10 +173,10 @@ func GitCommitIfChanges(dir string, message string) error {
 	if !changed {
 		return nil
 	}
-	return GitCommit(dir, message)
+	return GitCommitDir(dir, message)
 }
 
-func GitCommit(dir string, message string) error {
```
kzantow: Renamed this since I think it made more sense as the commit object

PR #830, opened 2018-05-17
feat: moved to dep for dependency management

rajdavies:


rawlingsj:
@rajdavies this looks great - over 2 million lines removed!

@hekonsek dep is looking pretty good TBH and seems to have a lot of traction.  Also there's 3.5k verses 9k github stars.  I'm thinking we go with this PR rather than https://github.com/jenkins-x/jx/pull/800, WDYT?

Issue #704, opened 2018-04-24
Errors while trying to use locally deployed Gitlab with Jenkins-X


mpwaldhorst:
I was trying out Jenkins-X to investigate use with a locally hosted Gitlab server.
For this I was following the guide on http://jenkins-x.io/developing/git/.
When trying to use the `jx create git token` command as described on http://jenkins-x.io/developing/git/

I received the following error:
```
Failed to create GitService with name gitlab: namespaces "jx" not found
```

After this I tried editing the `.jx/gitAuth.yaml` file and copied the way github credentials are stored like this:
```
servers:
- url: https://gitlab.server.url.com/
  users:
    - username: myUserName
      apitoken: 'gitlabAccessToken'
      bearer: ""
  name: gitlab
  kind: gitlab
  currentuser: "myUserName"
defaultusername: ""
currentserver: https://gitlab.server.url.com/
```

After this I tried running `jx create cluster minikube`.
I was asked to select a organisation, the only option I received was the username I had provided in the `gitAuth.yaml` file.
After selecting this organisation I received the following error:
```
Creating repository myUserName/environment-armfeather-staging
Creating git repository myUserName/environment-armfeather-staging
error creating cluster POST https://gitlab.com/api/v4/projects: 401 {message: 401 Unauthorized}
```

The command seems to try to post to the gitlab.com servers instead of the local servers.

I hope this helps you guys.

ccojocar:
dup of #1021

PR #847, opened 2018-05-20
Temporary hack to use self hosted gitlab base urls

binarix:
Temporary fix for https://github.com/jenkins-x/jx/issues/704 pending https://github.com/xanzy/go-gitlab/pull/403 to prevent leaking implementation detail of api version.

### DIFF
--- a/pkg/gits/gitlab.go
+++ b/pkg/gits/gitlab.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"time"
+	"net/url"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


michaelneale:
how temporary will it be ? if the links change merges will it still work but then this can be undone later?

ccojocar:
@binarix What's the status of this PR? Could you provide a more permanent fix now? It seems that your PR to `go-gitlab` was closed.

ccojocar:
@binarix Alright, I will close this then.

Issue #874, opened 2018-05-23
auto add ksync to path when running sync


michaelneale:
User has to add it to path by hand, with unconviently named warnings.

PR #875, opened 2018-05-23
add ksync to the path so users don't have to

michaelneale:
See issue: https://github.com/jenkins-x/jx/issues/874


michaelneale:
hey @rawlingsj @jstrachan apologies about my appaling golang, but this may work - uses cross platform path separator... 

(was my first use of VS Code too - it is quite good)

PR #886, opened 2018-05-23
User should be briefly notified about helm update attempt

hekonsek:
User should be aware that `jx` is attempting to update Helm repository.


hekonsek:
Thank you, sir ;)

PR #890, opened 2018-05-23
Initial draft of an upgrade cluster command

### LINKED ISSUE #889: Add an 'jx upgrade cluster' command that will update the kubernetes cluster
It would nice to be able to perform CVE patches to the kubernetes clusters on GKE via the JX command.

garethjevans:
Fixes #889


rawlingsj:
Looks great, the only thing I wonder is we probably want to restrict this for GKE users initially and then over time we can do the same for AKS, EKS etc.  Let's merge this and maybe add a check or survey prompt for the cloud provider in a later PR.

PR #895, opened 2018-05-24
Prompting that feature is experimental when upgrading a cluster

garethjevans:
I've also added a check to see if the gcloud command exists, and am being more verbose when running the upgrade to give the user feedback on what is happening.

### DIFF
--- a/pkg/jx/cmd/upgrade_cluster.go
+++ b/pkg/jx/cmd/upgrade_cluster.go
@@ -63,6 +63,24 @@ func NewCmdUpgradeCluster(f cmdutil.Factory, out io.Writer, errOut io.Writer) *c
 
 // Run implements the command
 func (o *UpgradeClusterOptions) Run() error {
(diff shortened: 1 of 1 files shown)


pkg/jx/cmd/upgrade_cluster.go
```diff
@@ -63,6 +63,24 @@ func NewCmdUpgradeCluster(f cmdutil.Factory, out io.Writer, errOut io.Writer) *c
 
 // Run implements the command
 func (o *UpgradeClusterOptions) Run() error {
+	confirm := false
```
jstrachan: LGTM! We could use `if !o.BatchMode` to avoid prompting for confirmation if folks use batch mode (`-b`) - not all commands have a batch mode but its handy if folks wanna use `jx` inside shells scripts etc. Maybe we could add a `--yes` CLI argument also to avoid the confirm?

PR #917, opened 2018-05-28
small typo

michaelneale:
self explanatory!

PR #921, opened 2018-05-28
Fix typo in help text

Cervator:
Noticed by chance while upgrading and thought I'd offer a quick fix.

Coincidentally doing the CLI upgrade updated the version at `/root/jx/bin/jx` while the one in use was at `/usr/local/bin/jx` - had to copy it over manually then all was fine. Not sure if that's a quirk in my environment setup or a legit bug. Using root since random throwaway cloud workspace.


Cervator:
Just noticed I'm not the only one who spotted it - closing in favor of #917, sorry about that :-)

michaelneale:
@Cervator no worries!

PR #929, opened 2018-05-29
feat(rsh) prefer bash over dash

yahavi:
In Jenkins pod, `/bin/sh` is a link to `/bin/dash` which is good only for non-interactive script execution.
Bash, on the other hand, is more human friendly and should be used by default.


jstrachan:
the tricky thing is knowing if a pod/container has `sh` or `bash` available or not. I wonder if there's a way we can figure that out or run a command to use whichever one is found.

I guess we could also have a CLI argument to enable `bash v sh` etc

ianwalter:
@jstrachan Can you just look in `/etc/shells` to see if bash is available?

ccojocar:
Closing this in favour of #1214.

Issue #950, opened 2018-05-31
jx create devpod --sync does not add ksync to the PATH


michaelneale:
~/.jx/bin needs to be in the path (or ksync installed to the system) or else create devpod --sync will not work

michaelneale:
FIXED

PR #951, opened 2018-05-31
add ksync to the path here also

michaelneale:
see https://github.com/jenkins-x/jx/issues/950

simple enough cc @jstrachan


jstrachan:
Looks good! I did wonder if we should add this part expression into all of the various `runCommand*` helper methods but there is quite a lot of those ;) as a similar issue exists for helm/kubectl/kops etc

michaelneale:
yes - wouldn't hurt to sprinkle it everywhere TBH - it modifies existing path but only for currently running process. Just not sure of test coverage on other platforms or at all.

PR #954, opened 2018-05-31
Dynamically fetch GKE zones fixes #952

### LINKED ISSUE #952: Dynamically fetch GKE zones
Currently the GKE zones are hardcoded in [the code](https://github.com/jenkins-x/jx/blob/master/pkg/jx/cmd/gke/helper.go#L3). It would be nice if we dynamically fetched the zones from GKE.

gtseres:


gtseres:
@jstrachan please let me know if it makes sense to do the same thing for GCP machine types as well. My only concern is if are all of them supported for GKE.

pkg/jx/cmd/gke/helper.go
```diff
@@ -1,55 +1,28 @@
...
-		"northamerica-northeast1-c",
-
-		"southamerica-east1-a",
-		"southamerica-east1-a",
-		"southamerica-east1-a",
+func GetGoogleZones() []string {
+	var zones []string
+	out, err := exec.Command("gcloud", "compute", "zones", "list").Output()
+	if err != nil {
+        fmt.Println("error occured")
```
rawlingsj: let's return the error if we cannot list the zones as we wont be able to proceed

rawlingsj:
looks good - just the one review comment

PR #978, opened 2018-06-06
Clean up the charts of each environment when uninstall the jenkins platform

ccojocar:
In addition removes also the jx namespace in order to clean up all orphan
resources such as secrets.


rawlingsj:
nice!

PR #1051, opened 2018-06-14
feat: Add gc for helm configmaps

tdcox:
Checks for versions of ConfigMaps in kube-system created by Helm for each release and prunes the number of versions retained to a value defined by --revision-history-limit (defaults to 10).

### DIFF
--- a/pkg/jx/cmd/gc.go
+++ b/pkg/jx/cmd/gc.go
@@ -22,6 +22,7 @@ const (
 
     * previews
     * activities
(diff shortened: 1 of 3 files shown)


jstrachan:
the CI fails with a compile error:
```
pkg/jx/cmd/gc_helm_test.go:90:8: undefined: strings.Builder
```
I hope we can get our hosted Jenkins X to report build errors on the actual PRs soon...

PR #1060, opened 2018-06-15
Using Organisation name for draft replacement instead of currentUser

ysaakpr:
Using organisation name instead of current user while doing string replacement in draft


ysaakpr:
Thank you

Issue #919, opened 2018-05-28
Addon for OWASP ZA Proxy scanner (baseline scan)


michaelneale:
One handy thing that could be done, and I am not sure at what stage in the pipeline, but probably on a preview app, and probably as an addon, is the OWASP "Zed Attack Proxy" scanner that looks for common vulnerabilities. https://github.com/zaproxy/zaproxy/ in particular: https://github.com/zaproxy/zaproxy/wiki/ZAP-Baseline-Scan (a more dangerous active scan could also be done on preview environments possible)

This can be quite simple to run. It can be stateless (one shot) or if you mount a volume it could keep sessions (in reality we will probably want to compare a PR with staging environment - to see if it makes things worse, vs sessions). Ie if a change makes more errors than existed before, or warnings, perhaps fail the build. 

eg to run a baseline test against any URL: 

`docker run -t owasp/zap2docker-stable zap-baseline.py -t https://jenkins.io`
(this could be a preview app) will generate results like: 

```
docker run -t owasp/zap2docker-stable zap-baseline.py -t https://jenkins.io
_XSERVTransmkdir: ERROR: euid != 0,directory /tmp/.X11-unix will not be created.
May 28, 2018 9:41:54 AM java.util.prefs.FileSystemPreferences$1 run
INFO: Created user preferences directory.
Total of 1048 URLs
PASS: Cookie No HttpOnly Flag [10010]
PASS: Cookie Without Secure Flag [10011]
PASS: Content-Type Header Missing [10019]
PASS: Information Disclosure - Sensitive Informations in URL [10024]
PASS: Information Disclosure - Sensitive Information in HTTP Referrer Header [10025]
PASS: HTTP Parameter Override [10026]
PASS: Information Disclosure - Suspicious Comments [10027]
PASS: Viewstate Scanner [10032]
PASS: Weak Authentication Method [10105]
PASS: Absence of Anti-CSRF Tokens [10202]
PASS: Private IP Disclosure [2]
PASS: Session ID in URL Rewrite [3]
PASS: Script Passive Scan Rules [50001]
PASS: Insecure JSF ViewState [90001]
PASS: Charset Mismatch [90011]
PASS: Application Error Disclosure [90022]
PASS: Loosely Scoped Cookie [90033]
WARN-NEW: Incomplete or No Cache-control and Pragma HTTP Header Set [10015] x 480 
	https://jenkins.io/
	https://jenkins.io
	https://jenkins.io/sitemap.xml
[...]

PR #1068, opened 2018-06-18
write the preview app url to the workspace so other cli tools can use it

michaelneale:
Could help with: 
https://github.com/jenkins-x/jx/issues/919

Seemed a simple way to pump out the temporary/preview url so other tools could use it. 

Thoughts @jstrachan @rawlingsj ?


rawlingsj:
Yeah seems fine to me, longer term I guess we could use a canonical place in the workspace to store things like this but I think this is fine.

michaelneale:
@rawlingsj yeah - not exactly elegant but could be useful. Not sure if this pattern of using files to pass on state will be useful, will see how it goes... (the one thing I like about it is that it is dumb simple... and when gluing things together via shell scripting it is useful to have files - and easy to diagnose)

Issue #1100, opened 2018-06-21
jx create cluster gke fatal Authentication failed Invalid username or password


davecore82:
Hi,

I'm using cloud shell in GCP and doing:

```
curl -L https://github.com/jenkins-x/jx/releases/download/v1.3.14/jx-linux-amd64.tar.gz | tar xzv
sudo mv jx /usr/local/bin
jx create cluster gke --skip-login
```

It fails with:

```
$ jx create cluster gke --skip-login --verbose

[...]
About to create repository environment-bearsharp-staging on server https://github.com with user <github user>
? Which organisation do you want to use? <github user>

Creating repository <github user>/environment-bearsharp-staging
Creating git repository <github user>/environment-bearsharp-staging
Cloning into '/home/<gcp user>/.jx/environments/<github user>/environment-bearsharp-staging'...
remote: Counting objects: 67, done.
remote: Total 67 (delta 0), reused 0 (delta 0), pack-reused 67
Unpacking objects: 100% (67/67), done.
[master 3b43956] Use correct namespace for environment
 1 file changed, 1 insertion(+), 1 deletion(-)
[master 52e0bbe] Add environment configuration
 1 file changed, 9 insertions(+)
remote: Invalid username or password.
fatal: Authentication failed for 'https://<github user>:3fe99df9baa856d577170a6a878e0525c47fe74f%20@github.com/<github user>/environment-bearsharp-staging.git/'
error creating cluster failed to invoke git push -u origin master in /home/<gcp user>/.jx/environments/<github user>/environment-bearsharp-staging due to exit status 128
error: failed to invoke git push -u origin master in /home/<gcp user>/.jx/environments/<github user>/environment-bearsharp-staging due to exit status 128
```

jstrachan:
I wonder if the github token you setup to access your git repo `environment-bearsharp-staging` is not valid? 

If you edit `~/.jx/gitAuth.yaml` by hand and fix up the API token or use the `jx create git token -n GitHub myUserName` to do it for you - then you could try create the environment again via `jx create env staging`

davecore82:
Hi James,

You were right. I had an extra space at the end of my apitoken in ~/.jx/gitAuth.yaml

It works now. Thanks a lot!

PR #1103, opened 2018-06-21
fix: triming whitespace around api token

garethjevans:


garethjevans:
Should help fix #1100

PR #1157, opened 2018-06-27
Pass repo type as parameter to the library

paveq:
In some cases Mercurial repo is created, while we intend to create Git repo. I suspect this has something to do with organization settings in Bitbucket.

As we currently support only Git type of repositories anyway, it should be ok to hardcode it here for now.

Note: there's separate bug in go_bitbucket library, which does not seem to pass these options to the API at all. I've notified author of the library about this, and investigating a fix there.


jstrachan:
@paveq that was fast - many thanks!

paveq:
Related PR in go_bitbucket: https://github.com/paveq/go-bitbucket/pull/1

paveq:
Related PR in go_bitbucket for reference: https://github.com/wbrefvem/go-bitbucket/pull/1

PR #1182, opened 2018-07-01
add docker registry prefix options

### LINKED ISSUE #843: Biggest network problem for me, because of the GFW
As the title, if the process of jx install can pull images from a register, that will be very good. And more Chinese people can learn and use JenkinsX.

### LINKED ISSUE #940: support pulling Jenkins X images from other than dockerhub
```
jx install --docker-registry fooio
```
For folks that have issues pulling from dockerhub i.e. in China with the great firewall

When someone specifies this flag we could generate a new helm values.yaml that overrides every docker image registry.  I.e. something like this:

```
nexus:
  image:
    repository: fooio/nexus
chartmuseum
  image:
    repository: fooio/chartmuseum
jenkins:
  Agent:
    PodTemplates:
      Maven:
        Containers:
          Maven:
            Image: fooio/builder-maven:0.0.255
```

LinuxSuRen:
fixes #843 
fixes #940 

@jstrachan @rawlingsj I just add docker registry for builder-pod and Jenkins master, other components like nexus, chart-museum still keep the origin. And I feel my solution maybe are not very well.
Please help to review this PR.
Thanks.

### DIFF
--- a/pkg/config/docker-registry.go
+++ b/pkg/config/docker-registry.go
@@ -1,0 +1,97 @@
+package config
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+
+	"gopkg.in/yaml.v2"
+)
+
+type DockerRegistryConfig struct {
+	DockerRegistry DockerRegistry `yaml:"docker-registry,omitempty"`
+}
+
+type DockerRegistry struct {
+	Prefix string `yaml:"prefix,omitempty"`
+}
+
+type DockerRegistryService struct {
+	DockerRegistry DockerRegistryConfig
+	Flags          DockerRegistryFlags
+}
+
+type DockerRegistryFlags struct {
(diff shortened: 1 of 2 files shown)


LinuxSuRen:
If I can specify a local docker registry, it will be faster to pulling these images. So, I just need to mirror part of these images.

jstrachan:
more here: https://docs.docker.com/registry/recipes/mirror/

yeah I'm just saying rather than forking every upstream helm chart both used by jenkins X and not used by jenkins X (e.g. prometheus, istio et al) to add in a docker registry prefix - its maybe better to fix this by enabling actual docker registry mirroring on the docker daemons. docker daemons can already be configured with a number of docker registry mirrors which means we won't have to fork all known helm charts we want to reuse to add a custom prefix in front - but could just enable registry mirroring for any docker image at all

LinuxSuRen:
Thanks for your tips. I will close this PR for now.

PR #1191, opened 2018-07-03
refactor:(helm) extract the helm actions behind an interface and provide an implementation for helm CLI

ccojocar:


jstrachan:
@ccojocar OK the BDD tests are working again; wanna rebase and see if this works?

PR #1221, opened 2018-07-06
update oce to oke

hhuiliu:
per oracle’s product managers’ correction,  use “oke” instead of “oce”. 
once the code is merged ,we will do doc pull-request. Thanks a lot


carlossg:
@hhuiliu there are some issues with JX on OKE, are you the right person to talk to or can point to somebody on the OKE team ?

PR #1265, opened 2018-07-11
fixes for AWS

jstrachan:


ijcd:
I think this introduced a bug that requires the region to be `us-west-1`

https://github.com/jenkins-x/jx/blob/8d7e1284ffe77536c007b3559b0e92243fddbc74/pkg/jx/cmd/create_cluster_aws.go#L184

PR #1273, opened 2018-07-12
fix:(helm) run the helm commands in verbose mode

ccojocar:


jstrachan:
BDD tests worked fine - they just timed out

PR #1277, opened 2018-07-13
Create owasp addon

michaelneale:
This is really just some sugar for useing a post preview job for https://github.com/jenkins-x/jx/issues/919

Opened for discussion - is this reasonable to do or better to leave it as a raw post preview job?


michaelneale:
it copies a bunch of code from create_post_preview_job.go - perhaps it should just invoke it and reuse it with some minor refactoring?

jstrachan:
@michaelneale yeah we should be able to just invoke the create_post_preview_job with some arguments?

michaelneale:
@jstrachan did you mean to merge this? (before refactoring to reuse post preview?) no biggie, can refactor it at any point, just wanted to make sure it was deliberate!

psiinon:
Bit late I'm afraid, but if might be better to call this 'owasp-zap' or something similar, rather than just 'owasp'.
[OWASP](https://www.owasp.org) is the organisation and ZAP is just one of the OWASP tools. At a future date someone might want to create an add-on for another OWASP tool (like [Dependency check](https://www.owasp.org/index.php/OWASP_Dependency_Check)) and that could get confusing

michaelneale:
@psiinon not too late - can adjust with a future PR, will put it on my todo (I wanted to shift to using the weekly image anyway sometime). Never too late!

PR #1289, opened 2018-07-15
when running many jx commands in parallel some collisions occur. Use …

i0n:
…a file based mutex to lock local maven install. Other procs can do the same.


rawlingsj:
The CI tests are failing with 
```

pkg/jx/cmd/common_install.go:15:2: url redeclared as imported package name

	previous declaration at pkg/jx/cmd/common_install.go:7:2

```

i0n:
I'll submit a new PR for this once my other PR is merged.

PR #1291, opened 2018-07-15
switch to bdd-jx and update jx-docs after test run

i0n:


ccojocar:
@i0n What's the deal with this PR? Are we ready for the switch?

i0n:
Should be good to merge, not sure why but looks like builds are timing out.

PR #1310, opened 2018-07-18
multiple fixes and code tidy up

rawlingsj:


rawlingsj:
The last test failed because we were github rate limited when checking if the environment PR was merged https://github.com/jenkins-x-bot/environment-hideplaid-staging/pull/1.  Not sure if we can switch to GHE for the BDD environment repos too?  Anyway manually merging as all tests passed.

PR #1313, opened 2018-07-18
fix:(compliance) gracefully terminate the compliance results after the archive was read

ccojocar:


ccojocar:
@i0n Thanks! I triggered again the build.

PR #1329, opened 2018-07-19
Fix/1328

### LINKED ISSUE #1328: RFE `jx gc gke`: Fix for no fw rules or disks and be more verbose about result
### Summary 

`jx gc gke`: Fix for no fw rules or disks and be more verbose about result

### Jx version

The output of `jx --version` is: `1.3.106`

### Expected behavior

* I would like to get an information that a local script file (`gc_gke.sh`) was created
* If no firewall rules exist the command creates a buggy delete line (rule names are missing)
* If no disks are available for deletion a hint should created

ascheman:
Some small improvements to fix #1328


rawlingsj:
awesome - thanks!

PR #1350, opened 2018-07-24
feat:(step_verify) add a commands which verifies the state of the pods of an application after deployment

ccojocar:


i0n:
This looks awesome. Could use this in the tests.

ccojocar:
We can use it. I am testing it right now within a pipeline.

ccojocar:
I got it working within a pipeline. See the definition here the https://github.com/jenkins-x/sso-operator/blob/master/Jenkinsfile.

rawlingsj:
code lgtm, I've not got a chance to test it now though if that's what you want before merging @ccojocar - if it's been tested then merge away else I can take a look early next week.

ccojocar:
@rawlingsj I have been testing it in the https://github.com/jenkins-x/sso-operator/blob/master/Jenkinsfile since 2 days. I have done couple of deployments with this step enabled. It seems to work fine so far.

PR #1365, opened 2018-07-25
Gc preview app

angoothachap:


rawlingsj:
Once CI passes we can merge, we'll need to tie in the chart release to the Jenkinsfile but we can do that separately.  Think I'll need to do that as only admins can change the pipeline.

PR #1403, opened 2018-07-31
feat(prow) initial spike to enable prow for promotion when using a cluster created by Terraform

rawlingsj:


rawlingsj:
Still needs a bit of polishing but will do over subsequent PRs

Issue #1420, opened 2018-08-03
provide a nice developer UX above Jenkinsfiles + Knative Builds


jstrachan:
it would be great to let folks use simple YAML knative-style pipeline steps which can then be converted into knative Build resources on a PR / merge to master or turned into Jenkinsfiles if need be.

we already have an optional `jenkins-x.yaml` file which lets you do things like associate issue trackers to a project etc.
https://github.com/jenkins-x/jx/blob/e4914f484b3443132e93d5933a22b766a80e2ab1/pkg/config/project_config.go#L18

So what if we allowed folks to specify Pipelines in there? We could also let folks specify just the bare minimum, rather than the full Pod/Container stuff with secrets/volumes and the like and use templates (like PodTemplates) to enrich the YAML?

e.g. something like this for `jenkins-x.yml`:

```yaml
buildPack: maven
builds:
  pullRequest:
    steps:
    - name: Run tests
      args:
      - mvn
      - test
  release:
    steps:
    - name: Perform Release
      args:
      - jx 
      - step
      - release
```

then rather like the current PodTemplates in Jenkins, we could use team sepecific default builds/templates and reuse things like its default image for the build pack and add things like docker socket, secrets and whatnot.

Also we can then define logical pipelines (release, pipeline, feature and others) then map those to physical branches at a team level using wildcards. e.g. 

``` yaml
branchPatterns:
  master: master
  pullRequest: pr-.*
  feature: feature.*
```

We could also add a way to add pre-post steps so that we could reuse the, say, 7 steps in a release but just add custom steps before/after the pipelines in the build packs?

Then when we import project we could generate the Jenkinsfile from this; or could turn it into a Build resource dynamically in Prow on each push to a branch.

For Prow we can then setup separate Prow jobs for each kind of branch; for Jenkins we can either let Jenkins figure that out from a universal Jenkinsfile - or we could do all this in Prow and just create/update/delete vanilla Pipeline jobs in Jenkins

jstrachan:
we could then even make the `jenkins-x.yml` file totally optional and each push to a branch would re-evaluate the build pack and reuse it from the build packs.

Or you could include a minimal `jenkins-x.yml` just specifying what build pack to run:

```yaml
buildPack: maven
```

or maybe just specifying one pipeline and reusing the others:

```yaml
buildPack: maven
builds:
  pullRequest:
    steps:
    - name: Run tests
      args:
      - mvn
      - test
```

jstrachan:
[...]

PR #1426, opened 2018-08-06
initial support for generating knative builds

jstrachan:
using the `jenkins-x.yml` file to simplify knative build authoring

#1420

### DIFF
--- a/pkg/tests/files.go
+++ b/pkg/tests/files.go
@@ -1,6 +1,7 @@
 package tests
 
 import (
+	"fmt"
 	"io/ioutil"
 	"strings"
 	"testing"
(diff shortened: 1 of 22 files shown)


pkg/tests/files.go
```diff
@@ -53,3 +54,28 @@ func AssertFilesExist(t *testing.T, expected bool, paths ...string) {
 		}
 	}
 }
+
+func AssertEqualFileText(t *testing.T, expectedFile string, actualFile string) error {
```
ccojocar: I think, we can replace this file test helpers with `https://godoc.org/gotest.tools/fs`.

jstrachan: awesome, I didn't know about those - will do thanks!

jstrachan: lemme refactor the unit tests to reuse gotest.tools in a separate PR...

PR #1458, opened 2018-08-09
fix: lets ensure the web dependency is added to spring boot apps

jstrachan:
otherwise the resulting microservice will typically fail


rawlingsj:
> @jstrachan: you cannot LGTM your own PR.

I think I can disable that rule

rajdavies:
approve

Issue #1413, opened 2018-08-01
jx create cluster eks does not accept --zones flag


gossnj:
### Summary 
`jx create cluster eks` does not accept `--zones` flag. This is a flag in eksctl, and as noted in README (https://eksctl.io/), frequently necessary for creation us-east1 at the moment. 

### Steps to reproduce the behavior

`jx create cluster eks --region=us-east-1` will frequently fail, producing an UnsupportedAvailabilityZoneException and recommending specifying availability zones. However, there is no way to do this from jx create cluster eks. 

Help text makes reference to EKS_AVAILABILITY_ZONES env var, but as far as I can tell, that is unused.

### Jx version

The output of `jx version` is:

```
1.3.144
```

### Kubernetes cluster

What kind of Kubernetes cluster are you using & how did you create it?

Attempting to create EKS cluster, 1.10

### Operating system / Environment

Mac OS X 10.13.4, EKS

### Expected behavior

`--zones` is accepted and passed on to eksctl. 

### Actual behavior

Error/help text is shown, and there is no way to create cluster in specific availability zones.

jstrachan:
the fix should be released within the hour

PR #1461, opened 2018-08-09
Add zones option to create cluster eks

polothy:
Ran into this issue https://github.com/weaveworks/eksctl/issues/118 which was fixed in this PR https://github.com/weaveworks/eksctl/pull/131

tl;dr: `eksctl` picks a random'ish availability zone, but if you run out of capacity in that availability zone, it errors.  So, you have to keep trying till it randomly'ish picks a zone that has capacity.  Solution, pass on a `--zones` option to `eksctl`.

I modeled the fix after `jx create cluster aws` but I was not aggressive about requiring zones since `eksctl` can work without the option being passed.  Let me know if it would work like the `aws` command and prompt the user for zones.

The changes here are untested too.

### DIFF
--- a/pkg/jx/cmd/create_cluster_eks.go
+++ b/pkg/jx/cmd/create_cluster_eks.go
@@ -26,6 +26,7 @@ type CreateClusterEKSFlags struct {
 	NodesMin            int
 	NodesMax            int
 	Region              string
+	Zones               string
 	Profile             string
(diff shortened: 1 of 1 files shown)


polothy:
Looks like this fixes #1413

jstrachan:
@polothy awesome thanks!

Issue #664, opened 2018-04-18
bitbucket cloud cannot query pull request status


jstrachan:
I tried doing promotion to a bitbucket cloud based environment and the PR is created fine but it then fails querying the PR status:
```
+ jx promote -b --all-auto --timeout 1h --version 0.0.2
Promoting app demo5 version 0.0.2 to namespace jx-staging-bb2
Using pipeline: jstrachan/demo5/master build: #2
Cloning into '/home/jenkins/.jx/environments/jstrachan/environment-jx-staging-bb2'...
Found remote branch names master, HEAD -> origin/master, master
Switched to branch 'promote-demo5-0.0.2'
[promote-demo5-0.0.2 8d8c2af] Promote demo5 to version 0.0.2
 1 file changed, 8 insertions(+), 5 deletions(-)
remote: 
remote: Create pull request for promote-demo5-0.0.2:        
remote:   https://bitbucket.org/jstrachan/environment-jx-staging-bb2/pull-requests/new?source=promote-demo5-0.0.2&t=1        
remote: 
To https://bitbucket.org/jstrachan/environment-jx-staging-bb2.git
 * [new branch]      HEAD -> promote-demo5-0.0.2
Created Pull Request: https://bitbucket.org/jstrachan/environment-jx-staging-bb2/pull-requests/1

Using pipeline: jstrachan/demo5/master build: #2
error: Failed to query the Pull Request status for https://bitbucket.org/jstrachan/environment-jx-staging-bb2/pull-requests/1 Status: 404 Not Found, Body: {"type": "error", "error": {"message": "Resource not found", "detail": "There is no API hosted at this URL.\n\nFor information about our API's, please refer to the documentation at: https://developer.atlassian.com/bitbucket/api/2/reference/"}}
```

I wonder if there's a way to query the status for bitbucket? Maybe we have to use Jenkins itself to query the status if the bitbucket REST API doesn't support it?

jstrachan:
Incidentally the Activity tab does show activity on a PR: 
https://bitbucket.org/jstrachan/environment-jx-staging-bb2/pull-requests/1/demo5-to-002/activity

maybe we can query that? It looks like Jenkins tries to update the activity on a PR /  build too:

jstrachan:
just tried with the latest greatest code and got the same:
```
EXITCODE   0Using pipeline: jstrachan/demo6/master build: #4
[...]

PR #1473, opened 2018-08-13
fix: Correct bitbucket Author, Org, and Repo.

johnmshields:
The issue in jenkins-x/jx#664 is caused by some mismatching of the PR's
Author and the "username" in the various Bitbucket APIs. In the case
where the repository is part of an organization the "username" is
actually the organization name. Likewise, if the PR author does not
match the repository username the same issue would occur.

This commit updates the `bitbucket_cloud.go` provider to use the
organization name as the PR `Owner` and the short repository name as the
`Repo`. It also adds the `Author` to the returned PR and uses the
proviers `UserInfo` method to obtain the full author information.

I'll add that I don't have access to a Bitbucket Server instance at the moment to I can't confirm that these changes are needed there as well. My guess is that they are similar. I'd be happy to add changes to `bitbucket_server.go` if someone can confirm they are needed and/or test them.


wbrefvem:
@jstrachan I tested the changes locally, and at this point, IMO, the failures are due to regressions. However, @johnmshields I'd love to hear your thoughts, and in any case thanks for the PR!

wbrefvem:
Here's what I'm getting when I run `make test`:

```--- FAIL: TestBitbucketCloudProviderTestSuite (60.06s)
    --- FAIL: TestBitbucketCloudProviderTestSuite/TestListCommitStatus (0.00s)
    	require.go:765:
    			Error Trace:	bitbucket_cloud_test.go:312
    			            				bitbucket_cloud_test.go:332
    			Error:      	Expected nil, but got: &errors.errorString{s:"Status: 404 Not Found, Body: 404 page not found\n"}
    			Test:       	TestBitbucketCloudProviderTestSuite/TestListCommitStatus
    --- FAIL: TestBitbucketCloudProviderTestSuite/TestMergePullRequest (0.00s)
    	require.go:765:
    			Error Trace:	bitbucket_cloud_test.go:344
    			Error:      	Expected nil, but got: &errors.errorString{s:"Status: 404 Not Found, Body: 404 page not found\n"}
    			Test:       	TestBitbucketCloudProviderTestSuite/TestMergePullRequest
    --- FAIL: TestBitbucketCloudProviderTestSuite/TestPullRequestLastCommitStatus (0.00s)
    	require.go:765:
    			Error Trace:	bitbucket_cloud_test.go:306
    			Error:      	Expected nil, but got: &errors.errorString{s:"Status: 404 Not Found, Body: 404 page not found\n"}
    			Test:       	TestBitbucketCloudProviderTestSuite/TestPullRequestLastCommitStatus
```

johnmshields:
Yes, let me take a look. I tried running last night with the clean master from jenkins-x/jx and got similar errors. I wanted to see what happened on the CI server... But maybe my environment wasn't fully clean. I'll take a look and let you know.

pkg/gits/bitbucket_cloud.go
```diff
@@ -366,8 +367,8 @@ func (b *BitbucketCloudProvider) UpdatePullRequestStatus(pr *GitPullRequest) err
 	prID := int32(*pr.Number)
 	bitbucketPR, _, err := b.Client.PullrequestsApi.RepositoriesUsernameRepoSlugPullrequestsPullRequestIdGet(
 		b.Context,
-		b.Username,
-		strings.TrimPrefix(pr.Repo, pr.Owner+"/"),
```
wbrefvem: The logic behind this and other instances of the same is that there were cases where a the `repo` argument was of the form `owner/repo` and it was necessary to trim the `owner/` portion. In any case, `strings.TrimPrefix` is idempotent and so this shouldn't need to be change (and doing so may produce regressions).

johnmshields: I can re-review it, but what I found was that the reason it was coming in as `owner/repo` was due to the original PR creation logic that used `FullName` as the `pr.Repo` value. The updated code just uses the `Name` value to set the `pr.Repo`.

Everywhere I could find that this method was used it was getting the value from the `pr.Repo`.

wbrefvem: Ah okay, that makes sense.

PR #1478, opened 2018-08-13
Register domain in AWS when using the 'domain' option

polothy:
We were seeing different behavior when using the `--domain` option in `jx create cluster eks` vs not using it and manually entering it when prompted.  With the option, the domain is not registered with a CNAME whereas when using the prompt, the domain does get a CNAME.

Assuming that the behavior should be the same regardless of where the domain comes from.


jstrachan:
@polothy great stuff thanks!

PR #1495, opened 2018-08-17
Upgrade ingress improvements

ccojocar:
- Improve the cert-manager detection
- Set as default the prod `ACME` service, the staging returns fake certificates


rawlingsj:
Added the put on hold label so we can discus when I'm back but in short the details from slack:

> Cosmin Cojocar [Yesterday at 3:33 PM]
I think the `letsencrypt staging` is useless, and we have it like default value in `jx upgrade ingress`

> James Rawlings [11 hours ago]
it’s not useless :slightly_smiling_face: it creates a self signed cert used to validate letsencrypt can use the config provided and validate the http challenge.  If for any reason like not having a public web endpoint or incorrect domain is supplied then the strict rate limiting by letsencrypt will block IPs and others methods for 1-2 weeks from memory.  So to start with while it’s a new feature it’s best to stick with staging as default.  The comment the `jx upgrade ingress` does try to explain some of this but can always be improved.

pkg/kube/crds.go
```diff
@@ -318,3 +319,28 @@ func CleanCertmanagerResources(c kubernetes.Interface, ns string, config Ingress
...
+	if err != nil {
+		// Failed to create the test issuer, most likely no cert-manager is available
+		return false
+	}
+	_, err = c.CoreV1().RESTClient().Delete().RequestURI(fmt.Sprintf("/apis/certmanager.k8s.io/v1alpha1/namespaces/%s/issuers", ns)).Name(CertmanagerIssuerTesting).DoRaw()
+	if err != nil {
+		// cert-manager is there but somehow it failed to delete the fake issuer
+		return true
+	}
+	return true
```
munnerz: Just a thought: you can probably use the discovery API here instead of manually attempting to create/delete a (Cluster)Issuer resource. This should let you know if the `certmanager.k8s.io/v1alpha1` API group is available 😄

ccojocar: @munnerz Thanks! This a good suggestion. I'll give it a try.

rawlingsj:
@ccojocar sorry about not replying sooner!  Back now and just catching up.  The `jx` question about staging / production should have described it so maybe we should improve the message to explain the self signed cert https://github.com/jenkins-x/jx/blob/c098069/pkg/jx/cmd/upgrade_ingress.go#L262.  That said it's not going to keep me awake at night so I'll leave it up to you, feel free to keep `prod` as the default in the PR.

ccojocar:
still needs some work

ccojocar:
Closing this since there is not anymore a conflict between the cert-manager installed by `jx` and the one installed by `jenkinsio/dex` chart. This dependency is removed right now from the dex chart.

Issue #763, opened 2018-05-04
Allow a team to be configured to create private repos


jstrachan:
It would be nice to be able to configure a Team to use a default of repositories being private rather than public

darthcabs:
Having `jx` alert that the repo is public, and instruct the user to go ahead and make it private (if possible) before deploying any apps. Once I put sensitive information as environment variables in values.yaml and commited before realising the repo was public. It wasn't terribly smart of me, I admit, but could happen to everyone of us haha

PR #1499, opened 2018-08-17
Add --git-private flag to make private repositories

polothy:
This allows for the creation of private git repositories.

I tested creating private repositories with Bitbucket Cloud via `jx create cluster eks`.  I realize there are several other commands that also get this flag and other git providers, but I haven't tested those individually.

I was a little surprised by the change I had to make in `pkg/jx/cmd/install.go`, unsure if other options need to be mapped over?  Like Username or ApiToken?


polothy:
Maybe fixes #763 ?

PR #1503, opened 2018-08-20
Split Test suite into 3 sections with update to docs

i0n:


docs/contributing/hacking.md
```diff
@@ -70,17 +70,51 @@ $ make build      # runs dep and builds `jx`  inside the build/
...
+} else {
+	// Slow test goes here...
+}
+```
+Slows tests can (and should) still include `t.Parallel()`
+
+Adding unencapsulated tests is not encouraged. If you absolutely must add a test that is not encapsulated, it needs to be wrapped like so:
 ```
+if os.Getenv("RUN_UNENCAPSULATED_TESTS") == "true" {
+	// Unencapsulated test goes here...
```
ccojocar: I think, it would be nice to define somewhere all the dependencies of an `unencapsulated test`. Even informally with some text would be nice.

pkg/gits/commits_test.go
```diff
@@ -36,91 +39,95 @@ func assertParseCommit(t *testing.T, input string, expected *CommitInfo) {
...
 func TestChangelogMarkdown(t *testing.T) {
-	releaseSpec := &v1.ReleaseSpec{
-		Commits: []v1.CommitSummary{
-			{
-				Message: "some commit 1\nfixes #123",
-				SHA:     "123",
-				Author: &v1.UserDetails{
-					Name:  "James Strachan",
-					Login: "jstrachan",
+	if os.Getenv("RUN_UNENCAPSULATED_TESTS") == "true" {
```
ccojocar: I would wrap this if into a closure just to make things more readable:

```Go
type UnencapsulatedFunc func()

func RunUnencapsulated(t *testing.T, func UnencapsulatedFunc) {
if os.Getenv("RUN_UNENCAPSULATED_TESTS") == "true" {
    func()
} else {
  t.Skip("skipping TestChangelogMarkdown; RUN_UNENCAPSULATED_TESTS not set")
}
```

and then in tests

```Go
RunUnencapsulated(t, func() {
// test code
})
```

i0n:
@ccojocar I've changed the mechanism for running integration or unencapsulated tests. Now you can include them in separate files with a build directive at the top, so you don't need to wrap them in a function to keep them out of the main unit test suite. I've updated the docs to reflect this change. I've also broken the tests out into _test packages as this convention will help us define our public APIs and make things a lot cleaner. It's a big commit but functionally things should remain the same, I'm quite keen to get this merged ASAP because rebasing will be a real pain on this one.

PR #1507, opened 2018-08-21
Customizing service jxing-nginx-ingress-controller

msvticket:
Allow values to be overriden in myvalues.yaml when creating service jxing-nginx-ingress-controller

### DIFF
--- a/pkg/jx/cmd/init.go
+++ b/pkg/jx/cmd/init.go
@@ -493,7 +493,7 @@ func (o *InitOptions) initIngress() error {
 		}
 
 		values := []string{"rbac.create=true" /*,"rbac.serviceAccountName="+ingressServiceAccount*/}
-		valuesFiles := []string{}
+		valuesFiles := []string{"./myvalues.yaml"}
 		if o.Flags.Provider == AWS || o.Flags.Provider == EKS {
 			// we can only enable one port for NLBs right now
 			enableHttp := "false"


pkg/jx/cmd/init.go
```diff
@@ -493,7 +493,7 @@ func (o *InitOptions) initIngress() error {
 		}
 
 		values := []string{"rbac.create=true" /*,"rbac.serviceAccountName="+ingressServiceAccount*/}
-		valuesFiles := []string{}
+		valuesFiles := []string{"./myvalues.yaml"}
 		if o.Flags.Provider == AWS || o.Flags.Provider == EKS {
```
rawlingsj: we should probably check if the file exists and also use a non platform file separator, i.e. `filepath.Join(".", "myvalues.yaml")` not `./myvalues.yaml`

ccojocar:
@msvticket Please could you address the review comments? Thanks

pkg/jx/cmd/init.go
msvticket: That makes sense. But I get a bit confused with how it's handled in install.go. First valuesFiles is initialized as:
```go
valueFiles := []string{"./myvalues.yaml", "./secrets.yaml", secretsFileName, adminSecretsFileName, configFileName}
```

Then myvalues.yaml is appended to this list as described by a comment:

	// Overwrite the values with the content of myvaules.yaml files from the current folder if exists, otherwise
	// from ~/.jx folder also only if is present

Is this a mistake and "./myvalues.yaml" should be removed from the initialization?

How about I extract the code implementing the above comment to a separate method and use in both places?

msvticket:
How about this? I now started using it just in init and install. Or should the new function in helm_helpers.go be put in a separate pull request first since it is referenced via github?

PR #1533, opened 2018-08-25
refactor: extract into a helper function the code which sets the myvalues.yaml file

ccojocar:


ccojocar:
@i0n I saw that you reverted the original PR. I re-worked it a bit and use the helper function also for upgrade platform command.

PR #1540, opened 2018-08-27
All Git options should be copied for 'jx install'

hekonsek:
Hi,

Right now jx doesn't copy username and token Git options when creating new environment. That causes `jx install --provider=eks --git-api-token=mySecret --git-username=hekonsek --git-provider-url=https://bitbucket.org` command to fail.

We should simply copy whole `GitRepositoryOptions` struct to avoid this issue.


hekonsek:
Awesome, thank you!

PR #1555, opened 2018-08-29
fix:(jenkins api token) fix the api token for new Jenkins version

ccojocar:


ccojocar:
This works great if already a token exists for an user. In that case, the command `jx create jenkins token -p <PASSWORD> --browser <USER>` is able to generate successfully another token, and to parse it from the HTML page.

Instead, it does not work if the token table is empty. It seems that the token input form is dynamically inserted into the DOM, and `chromedp` package is not able to detect this new node into the DOM tree. As a result, the `Generate` button does not show up into the DOM nodes, therefore the command gets stuck waiting for the button to be visible.

ccojocar:
@jstrachan @rawlingsj With this refactoring, I think we can upgrade to latest version of `jenkins-x-image` which contains the latest Jenkins and plugins. 

I tested, the installation, create quickstart, create spring and promotion. I haven't found any issue so far.

Issue #1556, opened 2018-08-29
Bitbucket Cloud's archive endpoint is different from Github's (Quickstart fails to pull)


cburch-bb:
### Summary 
Bitbucket Cloud's archive endpoint is different from Github's (Quickstart fails to pull).  If you try to `jx create quickstart` pointing to a bitbucket cloud repo, the pull archive request will fail
`Downloading Quickstart source zip from https://bitbucket.org/myusername/myrepo/archive/master.zip with basic auth for user: myusername
Unable to connect to the server: stopped after 10 redirects`

### Steps to reproduce the behavior
1. Add a quickstartlocation pointing to a bitbucketcloud account `jx create quickstartlocation`
2. Create a quickstart pointing to a repo from that bitbucketcloud account `jx create quickstart`
The process will error when trying to pull the archive because the url has been hard coded to a github style archive endpoint https://github.com/jenkins-x/jx/blob/9fd776419c4044fb5edf0f4cddd9230d6475fe60/pkg/quickstarts/model.go#L19 
bitbucket uses a slightly different style endpoint: https://bitbucket.org/OWNER/REPOSITORY/get/master.zip

### Jx version

The output of `jx version` is:

```
NAME               VERSION
jx                 1.3.200
jenkins x platform 0.0.2137
kubernetes cluster v1.9.6
kubectl            1.10.5
helm client        2.10.0-rc.1+gaa98e7e
helm server        2.10.0-rc.1+gaa98e7e
git                git version 2.9.0.windows.1
```

### Kubernetes cluster

What kind of Kubernetes cluster are you using & how did you create it? n/a

### Operating system / Environment

n/a

### Expected behavior

Bitbucket cloud server quickstarts should use the bitbucket style archive endpoints.

### Actual behavior

Github endpoints are used and therefore the request fails, quickstart create fails

cburch-bb:
I've tested with the latest build (1.3.211) after @polothy 's merge, works for bitbucket cloud now, thanks!

PR #1565, opened 2018-08-30
fix: allow for downloading quickstarts from other git providers

polothy:
Looking for feedback on implementation.  I went with an interface because not all git providers may have a way to download (though I don't have an example of one right now).  The interface allows for checking if the download can even happen before querying for repositories.

An alternative implementation would be to add a method to the `GitProvider` interface and the default implementation would be to return an empty string and log a warning.  Elsewhere when the quickstart is created, it does check for a non-empty download URL, so it would error out there.

Tested all three of these providers: GitHub, Bitbucket Cloud and Bitbucket Server.

### DIFF
--- a/pkg/gits/bitbucket_cloud.go
+++ b/pkg/gits/bitbucket_cloud.go
@@ -830,6 +830,10 @@ func (b *BitbucketCloudProvider) ServerURL() string {
 	return b.Server.URL
 }
 
+func (b *BitbucketCloudProvider) BranchArchiveURL(org string, name string, branch string) string {
+	return util.UrlJoin(b.ServerURL(), org, name, "get", branch+".zip")
+}
+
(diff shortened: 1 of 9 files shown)


polothy:
This fixes #1556

polothy:
@ccojocar Thanks for the feedback!

> I think, it would be nice to have an implementation of the interface also for gitlab and gitea.
Agreed, I was going to look at other providers once I got some feedback on implementation.  In regards to `gitea`, do you know what the URL would be? I cannot find a running instance.

> In this case, probably is a good option to embed the interface into the Provider type
Thanks, I'll refactor it.

polothy:
Pushed up a new commit, only outstanding issue is Bitbucket Server `1.0` vs `latest` I think.

PR #1574, opened 2018-09-03
ECR should take AWS_DEFAULT_REGION into account

hekonsek:
Hi,

Right now when you create new application on AWS EKS, jx reads only `AWS_REGION` environment variable. At the same time `jx install --provider=eks` respects `AWS_DEFAULT_REGION`. As a result if you have only `AWS_DEFAULT_REGION=us-east-1` set on your machine, jx will create EKS cluster in `us-east-1`, but new repositories will be created in `us-west-1` (which is current default). Also application build respects original `AWS_DEFAULT_REGION`, so eventually application build fail on Jenkins as it tries to push repository in wrong region (west instead of east).

According to AWS CLI documentation `AWS_DEFAULT_REGION` is the proper env to set to indicate AWS region. However it is also common for tools to use `AWS_REGION`.

I propose then that `ecr#NewAwsSession` should respect both `AWS_DEFAULT_REGION` and `AWS_REGION`. This simple fix solves the problem.

### DIFF
--- a/pkg/jx/cmd/create_cluster.go
+++ b/pkg/jx/cmd/create_cluster.go
@@ -56,8 +56,7 @@ const (
     * minikube (single-node Kubernetes cluster inside a VM on your laptop)
 	* minishift (single-node OpenShift cluster inside a VM on your laptop)
 	* openshift for installing on 3.9.x or later clusters of OpenShift
-    * coming soon:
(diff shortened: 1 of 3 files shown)


hekonsek:
Sorry guys, I've reset my master branch and automatically closed PR by this ;) Created new one in dedicated branch - #1580. Sorry!

PR #1578, opened 2018-09-03
EKS should use ecr-login

hekonsek:
Hi,

Recently jx deployed into EKS cannot push images into ECR (see #1550). In order to fix this behavior, you have to execute the following steps:

```
echo config.json > '{ "credsStore": "ecr-login" }'
kubectl delete secret jenkins-docker-cfg
kubectl create secret generic jenkins-docker-cfg --from-file=./config.json
```

This patch performs this operation automatically when EKS is used as provider (i.e. for `jx install --provider=eks` and `jx create cluster eks`).

IMHO This is essential to perform such configuration for EKS out of the box to provide smooth user experience.


hekonsek:
Are you sure your CI test is correct? I can build project from this branch. Can you run CI checks again?

jstrachan:
@hekonsek are you sure you installed Jenkins X via the `eks` provide? 

This should be setup by default:
https://github.com/jenkins-x/cloud-environments/blob/master/env-eks/myvalues.yaml#L6

jstrachan:
@hekonsek just wondered why the helm values config model has failed

hekonsek:
@jstrachan  100% sure, James. Tested several times. Also I'm not the only one who experiences this (see #1550).

If https://github.com/jenkins-x/cloud-environments/blob/master/env-eks/myvalues.yaml#L6 is supposed to be copied, then we should try to investigate why it is not. I can try to help you with that, as having smooth EKS support for jx is important for me now.

The question is if we should merge this PR as a workaround until we find the root cause of the issue. Right now people run into EKS/ECR failures with the latest jx if they try to use EKS and it would be awesome to make it work. If you merge this PR and #1580 then EKS support will be working. So IMHO we should:
- merge this PR (as a workaround)
- merge #1580
- celebrate because EKS support is now working again :) 
- find root cause of this configuration (https://github.com/jenkins-x/cloud-environments/blob/master/env-eks/myvalues.yaml#L6) not being copied
- fix root cause issue and remove this workaround

You decide :) .

jstrachan:
@hekonsek cool thanks - we all want the same thing, OOTB awesome on EKS. So yeah, lets merge this - but would really love to figure out why its broken.

hopefully the next CI build will work...

hekonsek:
Awesome, James! Many thanks. I will also investigate now what is the root cause of the issue and why env specific settings are not picked up.

jstrachan:
@hekonsek btw thanks for all your PRs lately! Its great to work with you again!

jstrachan:
hope to get them all merged ASAP...

hekonsek:
@jstrachan Same here, James! I want my new employer to take advantage of Jenkins X awesomeness, so expect more stuff coming from me, especially in the area of AWS/EKS. :)

PR #1582, opened 2018-09-04
Added --skip-installation option to 'create cluster' command

hekonsek:
Hi,

I'd like our devOps team to use Jenkins X to provision Kubernetes cluster on EKS, but I'd like to do it in a dedicated step, without installing actual Jenkins X platform into it. I'd like to achieve something like:

```
jx create cluster eks
jx install --provider=eks
```

I'd like to use such approach, because it is easier to re-provision Jenkins X if needed with Ansible playbooks if we have dedicated playbook for cluster provisioning and Jenkins X installation. 

Would you consider adding `--skip-installation` flag to `create cluster` command? That would solve our issue:

```
jx create cluster eks --skip-installation
jx install --provider=eks
```

Thanks!

### DIFF
--- a/pkg/jx/cmd/create_cluster.go
+++ b/pkg/jx/cmd/create_cluster.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/jenkins-x/jx/pkg/log"
 	"io"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


hekonsek:
Awesome, thanks Cosmin!

PR #1585, opened 2018-09-04
Fixes for git auth in pipelines

i0n:


affixalex:
I think this could potentially help me to address an issue I'm having in one of my pipelines. It seems like `jx step git version` isn't working quite as I'd expect.

PR #1598, opened 2018-09-06
CLI upgrade should link to new binary instead of replacing old one

hekonsek:
Hi,

The current approach for upgrading CLI via `jx upgrade cli` is broken. When I try to upgrade cli (on Fedora 28) I see something like:

```
$ jx upgrade cli
Downloading https://github.com/jenkins-x/jx/releases/download/v1.3.221/jx-linux-amd64.tar.gz to /home/hekonsek/.jx/bin/jx.tgz...
Downloaded /home/hekonsek/.jx/bin/jx.tgz
error: open /home/hekonsek/.jx/bin/jx: text file busy
```

The reason for this error is that you cannot replace/remove binary you're currently executing from the very same binary. As a result `jx upgrade cli` only downloads tarball, but I have to extract it manually after `jx upgrade cli` returns with error.

I propose the following approach to solve this issue:
- download tarbar of jx version `a.b.c`
- extract tarball as `~/.jx/jx-a.b.c` binary
- create symlink `~/.jx/jx-a.b.c => ~/.jx/jx`
- clean old binaries from `~/.jx`

With this approach you will have up to two jx binaries max in your `~/.jx` directory and latest of them symlinked into `~/.jx/jx`. Basically this is a trick taking advantage of the fact that you cannot replace/remove binary you're currently executing from the very same binary, BUT you can remove symlink to this binary.

What do you think?


rawlingsj:
> The reason for this error is that you cannot replace/remove binary you're currently executing from the very same binary

I've not had a chance to look properly yet, will try tomorrow but I was pretty sure in __go__ you can replace the same binary that you're executing from.  Am sure we did this on minishift before.

hekonsek:
I've found an another workaround to fix this problem. Closing this PR on the behalf of #1608.

PR #1604, opened 2018-09-06
feat(prow) handle env vars when triggered by prow not MBP

rawlingsj:


pkg/jx/cmd/common.go
```diff
@@ -768,3 +768,30 @@ func (o *CommonOptions) copyCertmanagerResources(targetNamespace string, ic kube
...
+	job := os.Getenv("JOB_NAME")
+	if job != "" {
+		return job
+	}
+	owner := os.Getenv("REPO_OWNER")
+	repo := os.Getenv("REPO_NAME")
+	branch := os.Getenv("BRANCH_NAME")
+
+	if owner != "" && repo != "" && branch != "" {
+		return fmt.Sprintf("%s/%s/%s", owner, repo, branch)
```
jglick: This is incorrect in general as the `branch` may need to be URL-encoded.

PR #1624, opened 2018-09-10
JXBinaryLocation() function should attempt to lookup for result in env variable

hekonsek:
Right now `jx` binary location lookup is practically limited to `which jx` invocation. This is fine for most cases, however I'd like to be able specify `jx` location via env variable. This feature would be primarily useful when user didn't add jx to classpath but still would like to upgrade binary for example. It will also make testing easier, as right now unit testing `CommonOptions#installJx()` is impossible without affecting your existing installation.


jstrachan:
@hekonsek I wonder if specifying a binary dir might be better - then if `jx` installs any binary (jx / helm etc) it'd use that dir?

jstrachan:
mind you I guess folks can already specify `$JX_HOME` and the install of binaries defaults to `$JX_HOME/bin`

PR #1637, opened 2018-09-11
Initial work to add theia

pmuir:
* basically working

### DIFF
--- a/pkg/jx/cmd/delete_devpod.go
+++ b/pkg/jx/cmd/delete_devpod.go
@@ -112,6 +112,13 @@ func (o *DeleteDevPodOptions) Run() error {
 		if err != nil {
 			return err
 		}
+		// Also delete Theia
+
+		theiaServiceName := name + "-theia"
+		err = client.CoreV1().Services(ns).Delete(theiaServiceName, &metav1.DeleteOptions{})
+		if err != nil {
+			return err
+		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/jx/cmd/delete_devpod.go
```diff
@@ -112,6 +112,13 @@ func (o *DeleteDevPodOptions) Run() error {
 		if err != nil {
 			return err
 		}
+		// Also delete Theia
+
+		theiaServiceName := name + "-theia"
+		err = client.CoreV1().Services(ns).Delete(theiaServiceName, &metav1.DeleteOptions{})
```
ccojocar: You can configure the `OwnerReference` to be the DevPod when you create the service. As a result, Kubernets will garbage collect the service when the pod is deleted.
An example from `sso-operator`:

```
svc.SetOwnerReferences(append(svc.GetOwnerReferences(), ownerRef(sso)))
``` 
```
func ownerRef(sso *apiv1.SSO) metav1.OwnerReference {
	controller := true
	return metav1.OwnerReference{
		APIVersion: apiv1.SchemeGroupVersion.String(),
		Kind:       apiv1.SSOKind,
		Name:       sso.Name,
		UID:        sso.UID,
		Controller: &controller,
	}
}
```

pmuir: Nice will add

Issue #1537, opened 2018-08-27
jx step link services not working


hamzapurra:
### Summary 

Running `jx step link services --from-namespace fromNamespace --target-namespace targetNamespace` does not seem to create services in the target namespace

Having inspected the codebase, there are a few problems that I saw:
- In the `step_link_services.go` file, the line
`lookedUpServiceFromTargetNamespace, _ := kubeClient.CoreV1().Services(targetNamespace).Get(service.GetName(), metav1.GetOptions{})`
always seems to return a value even if the service does not exist in the target namespace

- Changing the line
`kubeClient.CoreV1().Services(targetNamespace).Update(&targetService)`
to
`kubeClient.CoreV1().Services(targetNamespace).Create(&targetService)`
threw a couple of errors since it tries to create a service in the target namespace with the same `ClusterIP` as the service in the from namespace

- Adding the following lines seemed to create services in the target namespace
`
targetService.ResourceVersion = ""
targetService.Spec.ClusterIP = ""
`

- The newly created services don't have any endpoints as it uses pod selectors to point to pods that don't actually exist in the target namespace

- I feel the proper way to solve these problems would be to create services in the target namespace using the type `ExternalName` as mentioned here https://stackoverflow.com/a/44329470

### Jx version

The output of `jx version` is:

```
jx                 1.3.191
jenkins x platform 0.0.2082
kubernetes cluster v1.9.8
kubectl            v1.9.8
helm client        v2.10.0-rc.1+gaa98e7e
helm server        v2.10.0-rc.1+gaa98e7e
git                git version 2.15.2 (Apple Git-101.1)
```

### Kubernetes cluster

What kind of Kubernetes cluster are you using & how did you create it?

AWS cluster created using kops

### Expected behavior

Running `kubectl get service --namespace targetNamespace` does not show the new services

angoothachap:
asserts added in PR https://github.com/jenkins-x/jx/pull/1658

PR #1658, opened 2018-09-12
Bug fix - issue#1635

angoothachap:


angoothachap:
@i0n @jstrachan Added some asserts as part of issue #1537 investigation

PR #1663, opened 2018-09-12
Added binaries' version management

hekonsek:
Hi James,

Here is the proposal for keeping downloaded binaries versions under control and upgrade them together with jx cli. In particular we would like to avoid situation when jx downloaded helm or eksctl six months earlier while the latest cli assumes that those dependencies should be more up to date. 

TL;DR; I have added `jx upgrade binaries` command which traverses over the binaries in `~/.jx/bin` directory, checks if binaries installed there have to upgraded and upgrades them if necessary.

I have created initial version of generic `installOrUpdateBinary` function which has been tested right now against eksctl (and this is the only binary we check now - I can add all the others in subsequent PRs). Purpose of `installOrUpdateBinary` is to provide unified way of installing and controlling version of fetched binaries. In particular this function maintains `~/.jx/binaries.yml` files which keeps track of the version of dependency which is currently installed. Ideally all dependencies should be installed by invocation of this function similar to the following:

```
func (o *CommonOptions) installEksCtl(skipPathScan bool) error {
	return o.installOrUpdateBinary("eksctl",
		 "weaveworks",
		"https://github.com/weaveworks/eksctl/releases/download/{{.version}}/eksctl_{{.os}}_{{.arch}}.{{.extension}}",
		"0.1.1", skipPathScan)
}
```

...where we specify version of the binary which should be used with current version of jx and use GoLang templates to define download URL.

Here is how `jx upgrade binaries` works in action if it detects old version of dependency:

```
$ jx upgrade binaries
Downloading https://github.com/weaveworks/eksctl/releases/download/0.1.1/eksctl_linux_amd64.tar.gz to /home/hekonsek/.jx/bin/eksctl.tar.gz...
Downloaded /home/hekonsek/.jx/bin/eksctl.tar.gz
```

If you're OK with this approach, then I can start managing other dependencies downloaded by jx using this solution.

CC @jstrachan


jstrachan:
That all sounds good with me, thanks!

One thought was; folks sometimes install binaries by hand or have them in their `$PATH` already before running `jx`. We already kinda find the value of lots of binaries today in `jx version` - so I wonder if we could try to find out the actual version from the binary itself - if we know how to do that (which for helm / kubectl we do) - and if we can't find the version, then fall back on this values yaml file?

We should be able to reuse/refactor the code in `jx version` to find the versions for `helm & kubectl` at least? Some others are go based and tend to have a way to find their version too (and we have `jx --version` too)

hekonsek:
Yes, I was thinking about it as well.

The way my PR works now is - if there is binary detected in PATH during installation process, then we skip installation to `~/.jx/bin` and we don't manage the binary version via jx. The assumption here is that if user installed binary by herself then we cannot manage it anyway (because it is probably installed by root and because we should not mess with binaries we didn't install in the first place). However it would be nice to inform user about binary version mismatch even if binary is installed in PATH outside of ~/.jx/bin. Something like "hey, we see that you have helm v2.9.0 installed in /usr/bin/local/helm, but we suggest you to upgrade it to v2.10.0".

What we could add in the next PRs is:
- adding possibility of passing optional function argument to `installOrUpdateBinary` which can be used to fetch version from existing binary. This function could rely on a version detection code that we already have in `jx version`. binaries.yml file would be a fallback then for binaries which version cannot be easily extracted or should not be extracted for various reasons (for example `some-binary --version` could make slow call to the internet to fetch some data). binaries.yml would be then a safe generic default that works for all the binaries if needed.
- when we detect binary in PATH (outside ~/.jx/bin) during installation process and its version is different from version requested by `installOrUpdateBinary` then we should print warning alert suggesting the user that she should upgrade binary to the proper version.
- I also think that for reading versions of binaries we installed into ~/.jx/bin we should rely on binaries.yml instead of fetching version from binary execution, because it is just faster and is less likely to break over the time if `foo --version` output changes and cannot be parsed anymore.

My proposal to proceed would be: 
- let's start with merging this PR, as it already provides value for eksctl binary management
- In next PR I will add support to `installOrUpdateBinary` to be able to provide parsing logic for detecting version of binary from command output (also starting from here jx is gonna to suggest user to upgrade binary installed outside ~/.jx/bin if the version is different from one used in jx)
- Then in next PR I will migrate helm and kubectl to use `installOrUpdateBinary` and logic from `jx version` for detecting runtime changes
- Then I could migrate all the other binaries to be managed like that.

What do you think?

hekonsek:
@jstrachan Hi James. Any updates on this? Is there anything I could do to help to get it merged?

pkg/jx/cmd/common_install.go
```diff
@@ -196,6 +199,106 @@ func (o *CommonOptions) downloadFile(clientURL string, fullPath string) error {
 	return nil
 }
 
+func (o *CommonOptions) installOrUpdateBinary(binary string, gitHubOrganization string, downloadUrlTemplate string, version string, skipPathScan bool) error {
+	binaries := map[string]string{}
+	configDir, err := util.ConfigDir()
+	if err != nil {
+		return err
+	}
+	binariesConfiguration := configDir + "/binaries.yml"
```
ccojocar: Please could you use `filepath.Join` here is more portable?

pkg/jx/cmd/common_install.go
```diff
@@ -1069,65 +1172,11 @@ func (o *CommonOptions) installAws() error {
...
-	err = os.Remove(tarFile)
-	if err != nil {
-		return err
-	}
-	return os.Chmod(fullPath, 0755)
+func (o *CommonOptions) installEksCtl(skipPathScan bool) error {
+	return o.installOrUpdateBinary("eksctl",
+		 "weaveworks",
+		"https://github.com/weaveworks/eksctl/releases/download/{{.version}}/eksctl_{{.os}}_{{.arch}}.{{.extension}}",
+		"0.1.1", skipPathScan)
```
ccojocar: I think, you should pass in an empty version to really get the latest version from GitHub. Otherwise, it will always download `0.1.1`.

pkg/jx/cmd/common_install.go
hekonsek: Good point. I will change it :)

pkg/jx/cmd/common_install.go
hekonsek: Actually this is for purpose. Given version of jx should use the fixed version of binary dependency, for example eksctl. This way you can always be sure for example `jx 0.15 => eksctl 0.1` and `jx 0.16 => eksctl 0.3`. And you know that jx has downloaded version of binary dependency which has been used and tested by jx team. jx should control which version of dependency it is downloading, the same way you specify exact version of jar dependencies in Maven project POM instead of using the latest one.

ccojocar: Got it. At least would be nice to keep the versions of all binaries in a central place, instead of hard-code it into the `installEksCtrl` method. We can have something like: `installEksCtrl(version string, skipPathScan bool). For instance, in this case you can provide the fixed version from upgrade binaries command.

ccojocar:
@hekonsek It seems that the unit test fails: 

```
--- FAIL: TestJXBinaryLocationFailure (0.00s)

panic: When() requires an argument which has to be 'a method call on a mock'. [recovered]

	panic: When() requires an argument which has to be 'a method call on a mock'.



goroutine 26 [running]:

testing.tRunner.func1(0xc00014a900)

	/usr/local/go/src/testing/testing.go:792 +0x387

panic(0x754320, 0xc0001dc090)

	/usr/local/go/src/runtime/panic.go:513 +0x1b9

github.com/jenkins-x/jx/vendor/github.com/petergtz/pegomock/internal/verify.Argument(0xc000042f00, 0x7edb1e, 0x46, 0x0, 0x0, 0x0)

	/home/jenkins/go/src/github.com/jenkins-x/jx/vendor/github.com/petergtz/pegomock/internal/verify/verify.go:7 +0xb3

github.com/jenkins-x/jx/vendor/github.com/petergtz/pegomock.When(0xc000042f78, 0x2, 0x2, 0x0)

	/home/jenkins/go/src/github.com/jenkins-x/jx/vendor/github.com/petergtz/pegomock/dsl.go:366 +0x89

github.com/jenkins-x/jx/pkg/util_test.TestJXBinaryLocationFailure(0xc00014a900)

	/home/jenkins/go/src/github.com/jenkins-x/jx/pkg/util/dirs_test.go:27 +0x13e

testing.tRunner(0xc00014a900, 0x7f2340)

	/usr/local/go/src/testing/testing.go:827 +0xbf

created by testing.(*T).Run

	/usr/local/go/src/testing/testing.go:878 +0x353

FAIL	github.com/jenkins-x/jx/pkg/util	0.250s

?   	github.com/jenkins-x/jx/pkg/util/mocks	[no test files]

ok  	github.com/jenkins-x/jx/pkg/version	0.127s	coverage: 40.0% of statements

?   	github.com/jenkins-x/jx/pkg/workflow	[no test files]

```

hekonsek:
@ccojocar It fails on your machine or on CI? I tried `make test` and running from IDE and it always passed OK. Let's try to nail why it fails in your environment.

ccojocar:
@hekonsek The build fails on this PR. You can run the `make test-slow-integration` to reproduce it.

hekonsek:
@ccojocar `make test-slow-integration` works well for me. Are you sure this is not an another CI build issue? It sometimes fails from time to time from what I see. Can you try to rerun build on CI and see if this issue is reproducible?

ccojocar:
@hekonsek It seems so, now is green. Please could do one small fix and inject the version in `installEksCtrl` function? We can define the version as a const in a file such as binary_versions.go, where we can keep all the versions. Then you can inject it from upgrade binaries command.

hekonsek:
@ccojocar Many thanks for the merge! I will add version to the `installEksCtrl` in the next PR.

PR #1673, opened 2018-09-13
Added `jx delete aws` command

hekonsek:
Hi,

I've added `jx delete aws` command which is extremely useful when testing AWS deployments of jx. It is very common to create test installation in AWS by creating VPC dedicated for it. After you mess with your test deployment you usually would like to clean it up, which is a nightmare as VPC has references to subnets, elastic load balancers, network interfaces, internet gateways and so forth. You need to detach those resources from VPC and then delete all of them. Actually this is something that doesn't work well even for AWS web GUI or eksctl, so usually you end up by cleaning this mess by yourself manually.

This command can be executed like...

    jx delete aws --vpc-id=84777a7f8e

...and it will take care of the cleaning heavy-lifting for you.


hekonsek:
Yay! Thank you :) .

PR #1680, opened 2018-09-13
fix: create devpods

jstrachan:
lets always reuse the devpods correctly when not syncning and always output the URLs/IDE stuff


pkg/jx/cmd/create_devpod.go
```diff
@@ -549,21 +550,24 @@ func (o *CreateDevPodOptions) Run() error {
...
+	if err != nil {
+		return err
+	}
 	if theiaServiceURL != "" {
-		if err != nil {
-			return err
-		}
 		pod, err = client.CoreV1().Pods(curNs).Get(name, metav1.GetOptions{})
 		pod.Annotations["jenkins-x.io/devpodTheiaURL"] = theiaServiceURL
+		//pod.Annotations["jenkins-x.io/devpod/theia-url"] = theiaServiceURL
```
pmuir: Looks like a typo?

jstrachan: whoops - lemme zap thanks!

Issue #1601, opened 2018-09-06
newer jx cli doesn't respect --environment-git-owner parameter for staging environment


rezmuh:
### Summary 
I've been using --environment-git-owner parameter in the past but as of recent versions of jx cli, it doesn't pay attention to the parameter for staging environment anymore. production environment is still created correctly on the correct --environment-git-owner.
  
### Steps to reproduce the behavior

```
jx create cluster aws \
--kubernetes-version 1.10.5 \
--cluster-name cdub.tech \
--domain cdub.tech \
--http false \
--tls-acme true \
--default-environment-prefix cdub \
--environment-git-owner coverwatch \
--zones ap-southeast-1a,ap-southeast-1b,ap-southeast-1c \
--master-size m4.large \
--nodes 2 \
--node-size m4.large
```

### Jx version

The output of `jx version` is:

```
NAME               VERSION
jx                 1.3.221
jenkins x platform 0.0.2267
kubernetes cluster v1.10.5
kubectl            v1.11.2
helm client        v2.10.0+g9ad53aa
helm server        v2.10.0+g9ad53aa
git                git version 2.18.0

```

### Kubernetes cluster

What kind of Kubernetes cluster are you using & how did you create it?

It is on aws and was created with kops through `jx create cluster aws`

### Operating system / Environment

Arch Linux

### Expected behavior

Per the above command, two environment repos should be on coverwatch/environment-cdub-staging and coverwatch/environment-cdub-production

### Actual behavior
Repo for staging environment is created in rezmuh/environment-cdub-staging (my GitHub username) while the production is still correct, it is at coverwatch/environment-cdub-production

affixalex:
Big ol' +1 to this. I may dig into this further today but a fix would be glorious.

affixalex:
On a perhaps related note, `jx create env` doesn’t seem to be allowing me to select an organization and I think think is likely the underlying issue with not respecting `--environment-git-owner` - a bug in the code for interacting with the GitHub API. I'm not sure about the specifics here, but I'm going to look into it right now.

Salaboy:
I've reported this a while ago.. with a very cryptic title: https://github.com/jenkins-x/jx/issues/1516 is this the same behaviour?

i0n:
Hey there! Perhaps I can lend a hand/get this fixed for you guys.

There has been a bit of an un-going refactor recently around `jx install` as we straighten out a security model for user credentials. 

[...]

PR #1701, opened 2018-09-17
Ensure owner flag is respected by install

Lazyshot:
Currently, given the order of setting the `CreateEnvOptions` configs, the Owner option is overwritten and therefore not respecting the `environment-git-owner` flag.

This is explained more in #1601

### DIFF
--- a/pkg/jx/cmd/install.go
+++ b/pkg/jx/cmd/install.go
@@ -704,17 +704,15 @@ func (options *InstallOptions) Run() error {
 			}
 
 			log.Info("Creating default staging and production environments\n")
+			// Common CreateEnv Options
+			options.CreateEnvOptions.GitRepositoryOptions = options.GitRepositoryOptions
+			options.CreateEnvOptions.GitRepositoryOptions.Owner = options.Flags.EnvironmentGitOwner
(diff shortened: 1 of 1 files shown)


pkg/jx/cmd/install.go
```diff
@@ -704,17 +704,15 @@ func (options *InstallOptions) Run() error {
...
 
 			log.Info("Creating default staging and production environments\n")
-			options.CreateEnvOptions.Options.Name = "staging"
-			options.CreateEnvOptions.Options.Spec.Label = "Staging"
-			options.CreateEnvOptions.Options.Spec.Order = 100
+			// Common CreateEnv Options
+			options.CreateEnvOptions.GitRepositoryOptions = options.GitRepositoryOptions
 			options.CreateEnvOptions.GitRepositoryOptions.Owner = options.Flags.EnvironmentGitOwner
 			options.CreateEnvOptions.Prefix = options.Flags.DefaultEnvironmentPrefix
-			if options.BatchMode {
```
ccojocar: Just wondering, why is the batch mode removed?

Lazyshot: I removed them with the intent of moving them up to the "common options" area, but forgot to paste it. Update incoming.

PR #1735, opened 2018-09-20
fix: gc-activities needs permission to get deployments to detect prow

polothy:
Hopefully fixes https://github.com/jenkins-x/jx/issues/1705

**WARNING: this change is just a guess, I do not know how to test/verify it!**

`jx gc activities` now calls `kubectl get deployments` here:

https://github.com/jenkins-x/jx/blob/8d7538980cb4887bc9866152732fafd7a62a5830/pkg/kube/namespaces.go#L100

This currently fails due to permissions, example:
```
❯ kubectl logs --tail=100 jenkins-x-gc-activities-1537207200-zwb8m -n jxmt
error: deployments.apps "prow-build" is forbidden: User "system:serviceaccount:jxmt:jenkins-x-gc-activities" cannot get deployments.apps in the namespace "jxmt"
```

### DIFF
--- a/charts/jx/values.yaml
+++ b/charts/jx/values.yaml
@@ -84,6 +84,12 @@ clusterrole:
     - get
     - delete
     - list
+  - apiGroups:
+    - ""
+    resources:
+    - deployments
+    verbs:
+    - get
 
 activeDeadlineSeconds: 300
 backoffLimit: 5


polothy:
To reviewers: just randomly noticed these:
https://github.com/jenkins-x/jx/blob/8d7538980cb4887bc9866152732fafd7a62a5830/charts/service-linker/Makefile#L3

https://github.com/jenkins-x/jx/blob/8d7538980cb4887bc9866152732fafd7a62a5830/charts/service-linker/values.yaml#L1

Should those be renamed to `service-linker`?  Every chart uses the directory name there except for this `service-linker` chart.  If yes, I can make a PR if you like.

hekonsek:
We desperately need #1705 to be fixed. It brings EKS cluster down due to the k8s 1.10.0 bug. My dirty workaround for this one is to disable `jenkins-x-gc-activities` cronjob.

rawlingsj:
> Should those be renamed to service-linker? Every chart uses the directory name there except for this service-linker chart. If yes, I can make a PR if you like.

@polothy yeah good spot I think you're right, a PR would be great.  And many thanks for this one too!

PR #1736, opened 2018-09-20
`create eks cluster` command should support --node-type flag

hekonsek:
`create eks cluster` command should be able to pass `--node-type` flag to `eksctl`.


hekonsek:
Thanks!

PR #1739, opened 2018-09-21
Added AWS region resolving logic

hekonsek:
Added AWS region resolving logic to `jx delete aws command`, so user don't have to always specify it via command line flag.


hekonsek:
Thank you!

PR #1741, opened 2018-09-21
installChart function should accept struct parameter

hekonsek:
Hi,

Function `installChartAt` should accept not only `setValues`, but also `valueFiles` parameter. I've added this parameter support in this PR, however I also believe that this is a good opportunity to improve design of `installChartAt` function.

As GoLang doesn't support function overloading, we end up with functions with terribly long list of parameters which requires us to refactor many places in the codebase when we add new parameter to the list (for example right now when I'm add `valueFiles` parameter). Pattern to address this kind of issues is to replace list arguments with struct which holds the list of possible arguments. For example:

```
func Foo(a string, b string)

=>

type FooOptions struct {
  a string 
  b string
}
func Foo(fooOptions FooOptions)
```

I have added function `func (o *CommonOptions) installChartOptions(options InstallChartOptions) error` which takes such struct instead of list of parameters. I refactored existing `installChart` and `installChartAt` functions to use this new `installChartOptions` function. Eventually we should replaces all `installChart` and `installChartAt` calls with `installChartOptions` invocations.


hekonsek:
Thank you!

PR #1764, opened 2018-09-24
eksctl binary version should be extracted into const

hekonsek:
As agreed in #1663 I've moved eksctl version into constant and also provided version of function which could take custom version as argument.


hekonsek:
Thank you, James!

PR #1766, opened 2018-09-24
fix: service-linker chart names

polothy:
See https://github.com/jenkins-x/jx/pull/1735#issuecomment-423260358 and https://github.com/jenkins-x/jx/pull/1735#issuecomment-423766524

Not sure how to test this change and only making it because @rawlingsj thought it was a good idea.

Cheers!


rawlingsj:
> My approval is just based on @rawlingsj misspelling ;-)

:)

PR #1772, opened 2018-09-25
Migrated installChart() calls to installChartOptions()

hekonsek:
Migrated a couple of `installChart()` calls to `installChartOptions()` function introduced in #1741. We should prefer using `installChartOptions()` as it doesn't require refactoring when we add new arguments to the function definition.


hekonsek:
I've closed this PR on behalf of #1780.

PR #1780, opened 2018-09-25
Migrated installChart() calls to installChartOptions()

hekonsek:
Migrated a couple of `installChart()` calls to `installChartOptions()` function introduced in #1741. We should prefer using `installChartOptions()` as it doesn't require refactoring when we add new arguments to the function definition.

This is the same PR as #1772 but created from dedicated branch.


hekonsek:
Thank you, James!

PR #1783, opened 2018-09-25
Added `jx get eks` command

hekonsek:
I've added `jx get eks` command which can be used to list EKS clusters. This is pretty useful addition as with it you don't have to switch between `jx` and `eksctl` when working with EKS clusters. Also this command relies on our common `amazon.ResolveRegion()` util, so you don't have to remember about passing AWS region when listing EKS clusters.


jstrachan:
the BDD tests actually worked - there was just a timeout

PR #1793, opened 2018-09-26
fix:(sso) prompt only the organisations which have the user as member

### LINKED ISSUE #1769: Can't use user as org during jx create addon cloudbees --sso
### Summary 

If you select a user as an org during the `jx create addon cloudbees --sso` then you get an error when you try to login using github saying (something like) 'User data cannot be retrieved'.

Select a user (rather than a real org) at the question:

```
? Select GitHub organizations to authorize users from:
```

And then complete the install. Then try to log in to the web UI using GitHub.

### Jx version

✔ pmuir:~ ➭ jx version
Using helmBinary helm, with noTiller false
NAME               VERSION
jx                 1.3.310
jenkins x platform 0.0.2544
kubernetes cluster v1.9.7-gke.6
kubectl            v1.11.2
helm client        v2.10.0+g9ad53aa
helm server        v2.10.0+g9ad53aa
git                git version 2.18.0

### Kubernetes cluster

GKE

### Expected behavior

The user accounts shouldn't appear in the org list if they don't work.

ccojocar:
fixes #1769


jstrachan:
@ccojocar wonder if you could rebase this please?

PR #1794, opened 2018-09-26
Fixes to adding pipeline user as collaborator

msvticket:
Add pipeline user as collaborator in organisation so it works if it's different from user. Previously 404 was returned by github.

Skip accepting invitation as pipeline user if credentials are missing. Previously we got "panic: runtime error: invalid memory address or nil pointer dereference"
If pipeline user is part of organisation the acceptance step isn't needed anyway.


jstrachan:
@msvticket sorry about that - we're hoping to get the CI jobs to report back to the PRs with fail logs. Been having some issues with our production cluster running out of resources when doing lots of releases+PRs concurrently. Will try get it merged ASAP

PR #1796, opened 2018-09-26
Added AWS session util

hekonsek:
Right now jx doesn't fully respect AWS cli contract for detecting region and credentials. In particular jx should respect the following when detecting region and credentials:
- `AWS_*` environment variables
- `~/.aws/config`
- `~/.aws/credentials`

Also `AWS_PROFILE` environment variable and profiles defined in `~/.aws/config` and `~/.aws/credentials` should be taken into account.

In general we would like jx users to reuse their existing AWS configurations seamlessly. This PR provides a centralized way of creating AWS session (`NewAwsSession` function) which should be used by all AWS-related code in the jx to ensure that we're respecting all the constrains listed above.


jstrachan:
@hekonsek ah looks like we've a merge conflict - fancy taking a peek?

hekonsek:
Sure, done!

jstrachan:
@hekonsek damn more conficts ;) you're going too fast for us to keep up!

hekonsek:
OK. Resolved, merged and pushed ;) .

PR #1797, opened 2018-09-26
refactor: use the Jenkins cookie after login to generate the Jenkins API token

ccojocar:


ccojocar:
@jstrachan This version works also with headless mode of Chrome. I think the tests will no pass because the structure of the `login window` is different in the latest version of Jenkins. I had to adapt some query params used by chromedp.  

I tested with the following version of Jenkins, myvalues.yaml:
```yaml
jenkins:
  Master:
    ImageTag: "0.0.33"
```

The installation worked fine for me when run in batch+headless mode and also interactive.

ccojocar:
A new version of jenkins-x-image `0.0.34` is about to be released. 

https://jenkins.jx.cd.jenkins-x.io/blue/organizations/jenkins/jenkins-x%2Fjenkins-x-image/detail/master/4/pipeline/

ccojocar:
@jstrachan The build is failing because the `/me/descriptorByName/jenkins.security.ApiTokenProperty/generateNewToken` endpoint is not available in the old Jenkins version. 

Can we run the BDD tests with version 0.0.34 of Jenkins image when is released?

jstrachan:
@ccojocar we could for sure - I wonder can this code work on both old + new Jenkins versions? Just wondering if we should patch the code to work with both versions; or is the new version only required for the BDD tests?

To get the BDD tests working on this version I think we can tweak this file to override the jenkins image to use for the test:
https://github.com/jenkins-x/jx/blob/master/myvalues.yaml.template
something like this:
```yaml
jenkins:
  Master:
    ImageTag: 0.0.34
```

ccojocar:
> we could for sure - I wonder can this code work on both old + new Jenkins versions? Just wondering if we should patch the code to work with both versions; or is the new version only required for the BDD tests?

I think we should keep the existing code for old jenkins and use the new implementation for new Jenkins. Maybe we can retrieve the Jenkins version dynamically, keep the old code until we do the release and then remove it. 
 
We cannot use the same approach with the old Jenkins because the `/me/descriptorByName/jenkins.security.ApiTokenProperty/generateNewToken` is not available.

> https://github.com/jenkins-x/jx/blob/master/myvalues.yaml.template
something like this:

I added the latest Jenkins version in myvalues.yaml file. Let's see if the BDD tests pass.

ccojocar:
@jstrachan BDD tests passed with jenkins image 0.0.34. Should I try to make backward compatible?

jstrachan:
BTW the BDD tests did just work - the test timed out again before waiting for it to complete properly

ccojocar:
Do you know what is causing this timeout?

jstrachan:
Am afraid not. We were having issues last week around github API calls; the bdd tests now all run on GHE so we don't get hit by API call timeouts any more. Sometimes things can go a little slow when concurrent tests are running as the cluster needs to auto-scale.

Am hoping a larger timeout period can provide a workaround

jstrachan:
BTW if we merged this change, would `jx` break with older jenkins versions? wonder if we need the code to be smart enough to detect which jenkins server is running?

ccojocar:
Can we detect somehow dynamically the Jenkins version? We can keep the old code for older versions of Jenkins.

jstrachan:
if you do `curl -v $jenkins/api` you get the jenkins version in the response...

```
< HTTP/1.1 403 Forbidden
...
<X-Jenkins: 2.126
```

ccojocar:
cool! Let me try to work something out.

pkg/jx/cmd/create_jenkins_token.go
```diff
@@ -138,12 +153,23 @@ func (o *CreateJenkinsUserOptions) Run() error {
...
+		version, err := o.getJenkinsVersion(apiUrl)
 		if err != nil {
-			log.Warnf("unable to automatically find API token with chromedp using URL %s\n", tokenUrl)
+			return errors.Wrap(err, "getting Jenkins version")
+		}
+		if version.LT(JenkinsReferenceVersion) {
+			err = o.tryFindAPITokenFromBrowserOlderJenkins(tokenUrl, userAuth)
+		} else {
+			newTokenUrl := jenkins.JenkinsNewTokenURL(server.URL)
+			err = o.tryFindAPITokenFromBrowser(tokenUrl, newTokenUrl, userAuth)
```
jglick: This looks like overkill @ccojocar. See note in #2459.

PR #1800, opened 2018-09-26
feat(cwp) added build detection for custom war packager

garethjevans:


jstrachan:
BTW the BDD tests worked fine, they just timed out due to overloading of the test cluster

PR #1804, opened 2018-09-27
Upgraded eksctl version

hekonsek:
Upgraded eksctl version to 1.1.3. You can upgrade eksctl version managed by jx by executing `jx upgrade binaries`.

@jstrachan This is so cool to start having binaries version under control finally. I can now go to out devOps team and tell them to execute `jx upgrade binaries` as soon as this PR is merged. This is exactly what we needed. I will proceed with migrated other binaries to be controlled via this mechanism.

### DIFF
--- a/pkg/binaries/binaries.go
+++ b/pkg/binaries/binaries.go
@@ -5,6 +5,8 @@ import (
 	"runtime"
 )
 
+const EksctlVersion = "0.1.3"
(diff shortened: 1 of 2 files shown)


jstrachan:
BDD tests currently suffering from github API rate limiting - they eventually worked though!

PR #1820, opened 2018-09-28
AWS commands should respect --region and --profile flags

hekonsek:
Right now some AWS-related commands respect `--region` and `--profile` flags and some not. It confuses AWS users, as it is not clear where jx searches for AWS configuration among different commands.

This PR adds support for those flags for `jx get eks` and `jx get aws info` commands.


jstrachan:
BDD tests timed out again - they were really fine

PR #1835, opened 2018-10-01
Expose tiller namespace to extraValues

akihikokuroda:
The tiller namespace specified in the jx install command doesn't reach to the PODs.  The PODs still use the tiller in kube-system namespace.  This change expose the tiller namespace to the extraValues.yaml file (and configMap) to allow the jenkins config and pod templates make use of it.


a-roberts:
👍 

This is a useful PR for supporting multiple tiller installs on a cluster - in a realistic scenario one could have a secure tiller somewhere or an out of date tiller (let's say below version 2.9 or 2.10), and we want to set up a new tiller just to try jx, without interfering with our out of date/secure one.

This goes hand in hand with 
- https://github.com/jenkins-x/jenkins-x-platform/pull/4052
- https://github.com/jenkins-x/charts/pull/4

Issue #1857, opened 2018-10-02
Allow users to specify pull secrets for service accounts when creating an environment


a-roberts:
### Summary 
In the case of using our own private registry with Jenkins X, deployments in created namespaces (e.g. jx-production or jx-staging) need access to images from said registry.

Deployments will be run using the default service account in the namespace(s) that get created when you create an environment.

I propose that `jx create env` be augumented such that we can pass in a space separated list of pull secrets, e.g. to `jx create env --pull-secrets mysecret1 mysecret2 mysecret3`.

Upon receiving said secret list, we ensure that once we create the namespace and the service account, it has

- imagePullSecrets:
mysecret1
mysecret2
mysecret3

all defined upfront already.

### Steps to reproduce the behavior

Manual steps are currently required in order for pods in created environments to be able to use a private registry.

### Jx version

The output of `jx version` is:

```
jx                 1.3.354
jenkins x platform 0.0.2709
Kubernetes cluster v1.11.1+icp-ee
kubectl            v1.11.1
helm client        v2.10.0+g9ad53aa
helm server        v2.10.0+g9ad53aa
git                git version 2.15.2 (Apple Git-101.1)
```

### Kubernetes cluster

What kind of Kubernetes cluster are you using & how did you create it?

I'm using IBM Cloud Private which has its own Docker registry.

### Expected behavior

I'd like for the service account created to have the image pull secrets so I needn't do manual patching

### Actual behavior

Need to manually patch service account

a-roberts:
I'm working on this, my branch and changes are [here](https://github.com/jenkins-x/jx/compare/master...a-roberts:serviceaccountpatch?expand=1).

I would like this to be a common option but still want it prompted for when someone does a `jx create env`, not sure how to do this yet.

I may be stuck because I don't yet have a namespace to do the patching of the service account for. 

How can I make sure its defined upfront somewhere in some sort of template like we would when defining service accounts through Helm? 🤔

PR #1885, opened 2018-10-04
Add ability to patch service accounts upfront through create env

a-roberts:
See the issue https://github.com/jenkins-x/jx/issues/1857 for background

I would like to add a few tests as well: 
- can it handle no secrets correctly
- can it handle one
- can it handle many

Creating the PR here so folks can see the code and what we are thinking: 
credits to @johnmcollier for the patching code, John is interested in doing the same for dev pods.

```
build/jx create env --pull-secrets="foo bar bas"
... extra info omitted for brevity ...

Created environment kfc
Namespace jx-kfc created
 Service account "default" in namespace "jx-kfc" configured to use pull secret(s) [foo bar bas]
Pull secret(s) must exist in namespace jx-kfc before deploying your applications in this environment
Created Jenkins Project: http://jenkins.jx.9.20.201.31.nip.io/job/a-roberts/job/environment-jx-kfc/

Note that your first pipeline may take a few minutes to start while the necessary images get downloaded!

Creating GitHub webhook for a-roberts/environment-jx-kfc for url http://jenkins.jx.9.20.201.31.nip.io/github-webhook/
```

I verified the service account in the target namespace is configured to use all the pull secrets. This is my first implementation PR so let's see if the tests work here!


pkg/jx/cmd/common.go
```diff
@@ -285,6 +289,11 @@ func (o *CommonOptions) TeamAndEnvironmentNames() (string, string, error) {
 	return kube.GetDevNamespace(kubeClient, currentNs)
 }
 
+func (o *CommonOptions) ParseImagePullSecrets() []string {
+	pullSecrets := strings.Fields(o.PullSecrets)
+	return pullSecrets
+}
+
```
pmuir: Should be GetImagePullSecrets() to be consistent with other method names

a-roberts: Agreed, thanks

pmuir: Nice. Unortunately that's as far as I can provide a sensible review ;-)

a-roberts:
Hi there, anything I need to do for this? Updating branch every so often but don't want to forever

rawlingsj:
My slight concern with this is we're updating the default service account in a namespace.  I guess I'm wondering what happens if a pod uses a specific serviceaccount and not default?  Also should we instead be creating a new serviceaccount and updating that?  Just trying to think through this PR.

jstrachan:
happy to merge and keep improving - but yeah its probably better to have a SA per environment and use that instead of the `default`?

rawlingsj:
OK lets merge and update a separate SA in another PR.

a-roberts:
Thanks all, yeah very valid concerns, when creating an environment I noticed several service accounts actually, and if the pods are being run under the service account "default" it made sense to patch this. 

Agreed, it may be better to create a different service account that we patch, and then we pass in this service account name as part of the pod template for pipelines (so that we know the pipeline gets run under that service account, which has access to the registry for pushing to). 

Deployed applications in this environment would need to run under the same service account to have that secret as well to pull the image. So it would be a dual-use service account really, perhaps "pipeline-sa" would be a useful name for it.

PR #1895, opened 2018-10-05
Created Prometheus addon

hekonsek:
This PR adds minimal proper Prometheus support via addon.

In particular it exposes Prometheus Server web UI via Ingress and secures access to it via Nginx controller Basic Authentication (with password stored in Kubernetes secret). As a result after executing `jx create addon prometheus --password mySecret` I can go straight to `http://prometheus.jx.mydomain.com`, use admin/mySecret credentials and enjoy using Prometheus web UI.


pkg/jx/cmd/create_addon_prometheus.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return err
+	}
+	prometheusIngressConfig  := path.Join("/tmp", "prometheusIngressConfig_" + uuid.New())
+	err = ioutil.WriteFile(prometheusIngressConfig, valuesBytes, 0644)
+	if err != nil {
+		return err
+	}
+
+	err = o.installChartOptions(InstallChartOptions{
```
ccojocar: We also configure the `SetValues` just in case someone wants to overwrite some variables:
```
	setValues := strings.Split(o.SetValues, ",")

	and then in the struct:

        SetValues: setValues
```

hekonsek: Good point. Flag added :) .

hekonsek:
Let me know if there is anything I can do to help merge this PR.

PR #1897, opened 2018-10-05
fix:(quickstart) configure the current git server and user before importing the project

ccojocar:


pmuir:
I think the underlying issue was caused by #1838 so I've put in #1904 to revert it and reopened the issue

ccojocar:
The create quickstart seems to be fixed now. I'm closing this.

PR #1918, opened 2018-10-08
fix: update prow image version to 0.1.3

garethjevans:


rawlingsj:
dont think we need this now that tags are in the build templates

PR #1934, opened 2018-10-09
Removed duplicated logic from binary installation checking code

hekonsek:
I've removed some duplicated logic from `binaryShouldBeInstalled` function, which should be defined in a single place, in particular in `shouldInstallBinary` function.

I've also removed `binDir string` parameter from `shouldInstallBinary`, as we always pass results of `util.JXBinLocation()` invocation as its value anyway.

This is refactoring step towards having binaries checking/downloading/installation logic cleaned and extracted into `binaries` package.


hekonsek:
@jstrachan It looks like a false alarm for me. Can we try to rerun tests?

PR #1936, opened 2018-10-09
Don't try and re-add an alias

daveconde:


jstrachan:
@daveconde still learning Prow! I'd missed `/ok-to-test` :)

PR #1943, opened 2018-10-10
Fixes for AWS (kops) cluster creation

polothy:
1. The `runKops` function was doing an infinite loop.
2. The `kops validate cluster` command needs to go through `runKops` to get the `--state` flag.


polothy:
Closing, this isn't working yet.  Will re-open shortly.

PR #1952, opened 2018-10-11
Avoid prompting when creating from quickstart in batch mode

rajivnathan:
Proposed fix for https://github.com/jenkins-x/jx/issues/1953

The `jx create quickstart` command prompts for selection even if the batch mode `-b` option is specified. This causes scripts to hang. This PR simply adds a check for batch mode and returns an error to avoid prompting.


rajivnathan:
Rebased on the latest from master in case there was something that was causing the tests to fail.

PR #1956, opened 2018-10-11
feat: add --values flag to jx create addon, closes #1954

### LINKED ISSUE #1954: Add --values to jx create addon
### Summary 

Currently we support Helm's `--set` flag for setting variables during addon creation. It would be nice to support `--values` for passing in values files. Something like the following:

`jx create addon foo --values myValues.yaml`
`jx create addon foo -f myValues.yaml`
`jx create addon foo -f bar.yaml,baz.yaml,qux.yaml`

wbrefvem:


wbrefvem:
@rawlingsj Any details on what's failing? I ran the `jx-spring` BDD test locally and it passed.

PR #1971, opened 2018-10-12
feat: extract lastCommitSha from prow

pmuir:


pmuir:
/assign @rawlingsj 

Can you see if this makes sense? I don't really know if it will work or not ;-)

rawlingsj:
@pmuir yeah makes sense
/approve

PR #2056, opened 2018-10-22
Don't register CRDs on import, quickstart or get activities

msvticket:
A jenkins-x user administered with `jx edit userroles` and `jx controller role` doesn't have the permissions to register CRDs.
By the time jx import, quickstart or get activities are run they should have been registered by an admin (using `jx install` for example), in a jx step or controller anyway.

Without this fix a user gets errors like:
`error: customresourcedefinitions.apiextensions.k8s.io is forbidden: User "foo" cannot create customresourcedefinitions.apiextensions.k8s.io at the cluster scope`


ccojocar:
@msvticket Have you doubled checked if all CRDs are registered by the `install`? I think we need to make sure first that they are installed before removing their registration here.

ccojocar:
@msvticket Thanks for clarification. In general, they should be already registered up to this point, but I just want to double check that we don't break things by remove it the CRD registration.

@jstrachan Please could you have a look if it is safe to remove the CRD registration here? Thanks

msvticket:
This pull request has been superseded by #2246

PR #2074, opened 2018-10-24
standardize to uppercase Prow

darinpope:


darinpope:
@pmuir @ccojocar does the `needs-ok-to-test` label need to be removed?

PR #2076, opened 2018-10-24
Bitbucket Server integration: Wrong states are used, promote fails

wenzlaff:
promote.go requires states (PR, commits) that are different from the ones used in Bitbucket Server. SUCCESSFUL vs success for example.

The Bitbucket Cloud implementation already takes care of this using a statemap.

This is a hotfix that enables basic functionality of the Bitbucket Server integration, a major refactoring of Bitbucket Cloud and Server integrations seems to a good idea.


go.mod
```diff
@@ -149,24 +215,34 @@ require (
...
+	gopkg.in/testfixtures.v2 v2.5.0 // indirect
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7
 	gopkg.in/warnings.v0 v0.1.2
 	gopkg.in/yaml.v2 v2.2.1
 	k8s.io/api v0.0.0-20180628040859-072894a440bd
 	k8s.io/apiextensions-apiserver v0.0.0-20180621085152-bbc52469f98b
 	k8s.io/apimachinery v0.0.0-20180621070125-103fd098999d
 	k8s.io/client-go v8.0.0+incompatible
-	k8s.io/code-generator v0.0.0-20180904193909-8c97d6ab64da
+	k8s.io/code-generator v0.0.0-20181017053441-40b92a94e6ad
```
ccojocar: Please could you revert this to old version?

ccojocar:
@wenzlaff Please could your rebase and also revert to the old version of code-generator dependency?

go.mod
wenzlaff: @ccojocar is it sufficient to just delete go.sum and go.mod in the github diff view?

ccojocar: I would revert them, rebase from `master` and run a `make test`.

PR #2092, opened 2018-10-25
feat(install): Prevent against accidental uninstall from wrong cluster

### LINKED ISSUE #1647: Protect against accidental `jx uninstall` on wrong cluster
It's a bit too easy to accidentally uninstall Jenkins X when attached to the wrong cluster.  In general it's good practice to use the `jx shell` command to use an isolated shell, and `jx prompt` so you can see which context you are attached to.

For extra protection two great suggestions from @agentgonzo:

1) Take the github approach to deleting repos - once you issue `jx uninstall`, it should prompt you to enter the name of the cluster/context you are wanting to uninstall jx from. If this doesn't  match the current context, it doesn't do it for you. Simple and effective

2) Take the AWS approach to "Termination protection". It's nothing more than a metadata boolean flag against the resource that you can set/unset through a new command. "Destructive" commands like jx unininstall would check to see whether this protection is enabled and if so, would refuse to do the action. So for "important" resources/clusters, you turn protection on as a manual step after installing. Then to uninstall it's a 2-step process ("unprotect" and then "uninstall"). Again, simple and effective.

agentgonzo:
User must now enter the name of the current context to uninstall as a manual step.

Fixes #1647


pkg/kube/config.go
```diff
@@ -17,37 +17,20 @@ const (
...
-	config, err := po.GetStartingConfig()
-	if err != nil {
-		return nil, po, fmt.Errorf("Could not load the kube config file %s due to %s", po.GetDefaultFilename(), err)
-	}
-	return config, po, err
+// KubeConfig implements kube interactions
+type KubeConfig struct{}
+
+// NewKubeConfig creates a new KubeConfig struct to be used to interact with the underlying kube system
+func NewKubeConfig() *KubeConfig {
```
ccojocar: You can use `Kuber` instead of a pointer to the struct.

pkg/jx/cmd/context.go
```diff
@@ -116,11 +116,13 @@ func (o *ContextOptions) Run() error {
 		if err != nil {
 			return fmt.Errorf("Failed to update the kube config %s", err)
 		}
-		fmt.Fprintf(o.Out, "Now using namespace '%s' from context named '%s' on server '%s'.\n", info(ctx.Namespace), info(newConfig.CurrentContext), info(kube.Server(config, ctx)))
+		fmt.Fprintf(o.Out, "Now using namespace '%s' from context named '%s' on server '%s'.\n",
+			info(ctx.Namespace), info(newConfig.CurrentContext), info(kube.Server(config, ctx)))
```
amuniz: `o.Kube()` instead of `kube` here?

pkg/jx/cmd/delete_context.go
```diff
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	"github.com/jenkins-x/jx/pkg/jx/cmd/templates"
-	"github.com/jenkins-x/jx/pkg/kube"
```
amuniz: Is this ok? I think `o.Kube().LoadConfig()` still needs this package.

pkg/jx/cmd/context.go
agentgonzo: not necessary (also not exported). The only methods that get exposed by the interface are `LoadConfig` and `UpdateConfig`. All the others (like this) just act on the config object passed in so no _need_ to extract them to the interface.

pkg/jx/cmd/delete_context.go
agentgonzo: It compiles so I guess so!
`o.Kube().LoadConfig()` pulls in this package transitively I guess.

PR #2104, opened 2018-10-27
feat: add binary plugins

pmuir:
Binary plugins in jx behave exactly like binary plugins in kubectl, see 
https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/


go.mod
```diff
@@ -162,11 +164,13 @@ require (
 	k8s.io/api v0.0.0-20180628040859-072894a440bd
 	k8s.io/apiextensions-apiserver v0.0.0-20180621085152-bbc52469f98b
 	k8s.io/apimachinery v0.0.0-20180621070125-103fd098999d
+	k8s.io/cli-runtime v0.0.0-20181026155151-1ee5ba10d7e3 // indirect
 	k8s.io/client-go v8.0.0+incompatible
 	k8s.io/code-generator v0.0.0-20181017053441-8c97d6ab64da
 	k8s.io/gengo v0.0.0-20180718083919-906d99f89cd6
 	k8s.io/helm v2.7.2+incompatible
 	k8s.io/kube-openapi v0.0.0-20180719232738-d8ea2fe547a4
+	k8s.io/kubernetes v1.12.2 // indirect
```
ccojocar: Probably we want to keep this dependency to v1.11.3?

pkg/jx/cmd/plugin.go
```diff
@@ -0,0 +1,265 @@
...
+
+func (o *PluginListOptions) Complete() error {
+	o.Verifier = &CommandOverrideVerifier{
+		root:        o.Cmd.Root(),
+		seenPlugins: make(map[string]string, 0),
+	}
+	return nil
+}
+
+func (o *PluginListOptions) Run() error {
```
ccojocar: It would be nice to have a test for this.

pmuir:
/hold

Putting this explicitly on hold as I need to come back to this in a couple of days :-D

go.mod
pmuir: Fixed

PR #2119, opened 2018-10-30
test(git provider): add tests for git providers creation from URL

ccojocar:


rawlingsj:
ok this is building now

ccojocar:
@rawlingsj Thanks! The tests seem to failed because of the secret masking:
```
expected: auth.UserAuth{Username:"test", ApiToken:"test", BearerToken:"", Password:""} actual : auth.UserAuth{Username:"****", ApiToken:"****", BearerToken:"", Password:""}
```
Any idea how can I disable it during tests?

rawlingsj:
> Thanks! The tests seem to failed because of the secret masking:
Any idea how can I disable it during tests?

Ah I see, hmm I'm afraid I dont, @jtnord might have an idea?

PR #2132, opened 2018-10-31
feat: add support for helm managed extension repos

pmuir:
* allows extension repository in use to be configured in a config map
* this allows us to password protect them
* added a jx edit extensionsrepository command to set the extensions repo
  in use
* Use gitProvider.GetContent to allow access to private repos for extension
  definitions


pkg/helm/helm_cli.go
```diff
@@ -257,6 +257,28 @@ func (h *HelmCLI) InstallChart(chart string, releaseName string, ns string, vers
 	return h.runHelm(args...)
 }
 
+// Fetch a Helm Chart
+func (h *HelmCLI) FetchChart(chart string, version *string, untar bool, untardir string) error {
```
ccojocar: Please could you add a test for this in `helm_cli_test.go`?

PR #2146, opened 2018-11-01
feat(vault): encode to HCL the vault path rule and add some tests

ccojocar:
#### Submitter checklist

#### Description

feat(vault): encode to HCL the vault path rule and add some tests

### DIFF
--- a/pkg/vault/rule_test.go
+++ b/pkg/vault/rule_test.go
@@ -3,6 +3,7 @@ package vault_test
 import (
 	"testing"
 
+	"github.com/hashicorp/hcl"
 	"github.com/jenkins-x/jx/pkg/vault"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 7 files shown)


pkg/vault/rule_test.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+		"marshal nil policy with error": {
+			rule: nil,
+			err:  true,
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			output, err := tc.rule.String()
```
agentgonzo: Don't compare strings for things like this. 
1) It causes a lot of unnecessary headaches (eg, if we change a library implementation at some point, or change whitespace formatting, the tests will fail). Also, testing like this broke on Windows where a newline is CR-LF but in unix it's just LF
2) Ultimately, we don't care about the `string` representation, we care about the `json`/`yaml` representation.

keep `tc.rule` as a struct (or a JSON/YAML object) and deserialize the the `want` to the same type, then do a comparison on the typed objects.

ccojocar:
@agentgonzo I modified the test to use the HCL parser, and now it compares the structs instead of the strings. Please could you have a look? Thanks

PR #2152, opened 2018-11-02
Fix: don’t unmarshal non-yams

pmuir:


pmuir:
FWIW keeps erroring in BDD tests trying to setup Jenkins

PR #2155, opened 2018-11-02
Update OWNERS file

jtnord:
#### Submitter checklist

- [x] Change is code complete and matches issue description.
- [ ] Change is covered by existing or new tests.

#### Description

Updates the owners file to add new members.

#### Which issue this PR fixes

N/A

PR #2156, opened 2018-11-02
Update OWNERS file

jtnord:
#### Submitter checklist

#### Description

Update the owners file so all contributers can approve without having to constantly keep this in sync.

This may or may not be required and conflicts with #2155 so the correct approach should be taken.
This will include more people than 2155 (by virtue of the team and owners becoming out of sync)

N/A


jtnord:
/hold
we should decide if this is a better approach than #2155

agentgonzo:
/approve
I prefer this over #2155

rawlingsj:
lets go with #2155 as we know it works

PR #2159, opened 2018-11-02
fix: complete the role controller

jstrachan:
added unit test to show its correct behaviour as environments are created,
    users added or removed from EnvironmentRoleBindings


pmuir:
Closed/opened to get tide to kick in after GitHub outage

PR #2167, opened 2018-11-05
feat(vault): Vault CLI integration

agentgonzo:
#### Submitter checklist

#### Description
Added ability get the `vault` CLI config to be able to use the official vault client

`jx get vault-config` will create a login and print out the `export`s for `VAULT_ADDR` and `VAULT_TOKEN` for the user.


agentgonzo:
/hold
@ccojocar - this is a WIP - can you take a look over it whilst I'm still sorting it out and writing tests to give me some early feedback on potential design/issues

pkg/jx/cmd/get.go
```diff
@@ -95,6 +95,8 @@ func NewCmdGet(f Factory, in terminal.FileReader, out terminal.FileWriter, errOu
 	cmd.AddCommand(NewCmdGetUser(f, in, out, errOut))
 	cmd.AddCommand(NewCmdGetWorkflow(f, in, out, errOut))
 	cmd.AddCommand(NewCmdGetVault(f, in, out, errOut))
+	//cmd.AddCommand(NewCmdGetSecret(f, in, out, errOut))
```
agentgonzo: This one is a bit hacky and unnecessary ATM - will get to it later.

pkg/jx/cmd/get_secret.go
```diff
@@ -0,0 +1,82 @@
+package cmd
```
agentgonzo: This isn't well thought-out at the moment and does a bad job. Don't worry about it at the moment. Will fix later.

pkg/jx/cmd/common/interfaces.go
```diff
@@ -0,0 +1,11 @@
+package common
+
+import (
+	"github.com/banzaicloud/bank-vaults/operator/pkg/client/clientset/versioned"
+	"k8s.io/client-go/kubernetes"
+)
+
+type NewCommonOptionsInterface interface {
+	KubeClient() (kubernetes.Interface, string, error)
+	VaultOperatorClient() (versioned.Interface, error)
```
ccojocar: I would move this in the factory.go which is used as a facade for cmd.

pkg/jx/cmd/get_vault_config.go
```diff
@@ -0,0 +1,81 @@
...
+	clientFactory := vault.VaultClientFactory{
+		Options: o,
+	}
+	client, err := clientFactory.NewVaultClient(o.Namespace)
+	if err != nil {
+		return err
+	}
+
+	// Echo the client config out to the command line to be piped into bash
+	if runtime.GOOS == "windows" {
```
ccojocar: I would use `os.Setenv` function instead of handling specific per OS.

pkg/jx/cmd/vault/interfaces.go
```diff
@@ -0,0 +1,17 @@
+package vault
```
ccojocar: I would move this file to `pkg/vault`. We should keep all the vault specific things in that package.

pkg/jx/cmd/vault/vault.go
```diff
@@ -0,0 +1,25 @@
+package vault
+
+import (
+	"fmt"
+	"github.com/hashicorp/vault/api"
+)
+
+// GetSecrets returns a list of secrets from Vault
+// It's a bit hacky at the moment
+func GetSecrets(client *api.Client) ([]string, error) {
```
ccojocar: I would design the interface a bit differently in order to hide`api.Client` underneath, and then you can generate a fake interface.

```Go
type Vaulter interface {
    Secrets() (map[string]string, error)
    Secret(name string) (string, error)
...
}

type JxVaulter struct {
     client *api.Client
}

func NewJxVaulter(client *api.Client) {
}

func (v *JxVaulter) Secrets() (map[string]string, error) {
}
```

pkg/jx/cmd/vault/interfaces.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"github.com/hashicorp/vault/api"
+	"github.com/jenkins-x/jx/pkg/kube"
+)
+
+// Vaulter is an interface for creating new vault clients
+type Vaulter interface {
```
ccojocar: See my comment regarding this interface below.

pkg/jx/cmd/vault/vaultfactory.go
```diff
@@ -0,0 +1,108 @@
+package vault
```
ccojocar: I would move this to `pkg/vault`.

pkg/jx/cmd/vault/vaultfactory.go
```diff
@@ -0,0 +1,108 @@
...
+	token, err := getTokenFromVault(serviceAccount.Name, jwt, vaultClient)
+	vaultClient.SetToken(token)
+	return vaultClient, nil
+}
+
+func (v *VaultClientFactory) getServiceAccountFromVault(vault *kube.Vault) (*v1.ServiceAccount, error) {
+	return v.kubeClient.CoreV1().ServiceAccounts(vault.Namespace).Get(vault.AuthServiceAccountName, meta_v1.GetOptions{})
+}
+
+func (v *VaultClientFactory) getSecretFromServiceAccount(sa *v1.ServiceAccount, namespace string) (*v1.Secret, error) {
```
ccojocar: There is a helper function for this in `kube/service_accounts.go`. I would use that function.

pkg/jx/cmd/vault/vaultfactory.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil, err
+	}
+	v := vaultSelectorImpl{
+		vaultOperatorClient: operator,
+		kubeClient:          kubeclient,
+	}
+	return v, nil
+}
+
+func (v vaultSelectorImpl) GetVault(namespace string) (*kube.Vault, error) {
```
ccojocar: Isn't there a function in `kube/vault.go` for this which get the vault by name?

pkg/jx/cmd/vault/vaultfactory.go
```diff
@@ -0,0 +1,108 @@
...
+	vaults, err := kube.GetVaults(v.kubeClient, v.vaultOperatorClient, namespace)
+	if err != nil {
+		return nil, err
+	} else if len(vaults) == 0 {
+		return nil, errors.New(fmt.Sprintf("no vaults found in namespace '%s'", namespace))
+	}
+	if len(vaults) > 1 {
+		return selectVault(vaults)
```
ccojocar: I would inject this as a closure in the `GetVault` function if you want to have a complicated selector on the vault params. I think for now it is enough to get it by name.

pkg/jx/cmd/get_vault_config.go
agentgonzo: That's not quite the same. `os.Setenv` will set the env parameter for the current (`jx`) process, which will get discarded when the process terminates. This needs to be able to set the user's environment for use with the official `vault` CLI. Unfortunately (given environment variables get dicarded when the current running process exits) the only way to do this is to export a string that the user can pipe into the current terminal. 
This is the way that AWS tools do `docker` config (and the easiest way to do it) - see https://docs.aws.amazon.com/cli/latest/reference/ecr/get-login.html

pkg/jx/cmd/vault/vaultfactory.go
agentgonzo: Grr... yes. I was looking for `GetVault` and could only see `GetVaults`. It's actually called `FindVault` - thanks for pointing it out.

pkg/jx/cmd/common/interfaces.go
agentgonzo: Unfortunately, `factory.go` is still inside the `cmd` package. Given that parts of the `cmd` package need to import the `vault` package, I needed to put this interface _outside_ the `cmd` package to prevent a circular dependency.

pkg/jx/cmd/vault/vaultfactory.go
agentgonzo: Nope - `FindVault` just returns a bool :-/

agentgonzo:
Err... Goland was a bit over-zealous in refactoring `service` --> `services`. I'll do a rebase now and fix it.

pkg/vault/jxvaulter.go
```diff
@@ -0,0 +1,45 @@
...
+	client, err := clientFactory.NewVaultClient(o.VaultName(), o.VaultNamespace())
+	return newJxVaulter(client), err
+}
+
+func (v *JxVaulter) Config() (vaultUrl url.URL, vaultToken string, err error) {
+	parsed, err := url.Parse(v.client.Address())
+	return *parsed, v.client.Token(), err
+}
+
+func (v *JxVaulter) Secrets() ([]string, error) {
```
ccojocar: Wasn't be better to return a `map[string]string` where the key of the map is the secret name and the value is the value of the secret. We will definitely need this.

pkg/vault/jxvaulter.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	out := make([]string, 0)
+	for key, value := range secrets.Data {
+		_ = key
+		_ = value
+		out = append(out, fmt.Sprintf("%v", value))
+		out = append(out, fmt.Sprintf("%T", value))
```
ccojocar: I don't get why you are appending twice the value. It would go with a map to return the both values.

pkg/vault/vaultfactory.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	vaultClient, err := api.NewClient(config)
+	token, err := getTokenFromVault(role, jwt, vaultClient)
+	vaultClient.SetToken(token)
+	return vaultClient, nil
+}
+
+// GetConfigData generates the information necessary to configure an api.Client object
+// Returns the api.Config object, the JWT needed to create the auth user in vault, and an error if present
+func (v *VaultClientFactory) GetConfigData(name string, namespace string) (config *api.Config, jwt string, saName string, err error) {
```
ccojocar: I think we can make this function private to the package by renaming to `getConfigData`.

pkg/vault/vaultfactory.go
```diff
@@ -0,0 +1,78 @@
...
+	token, err := serviceaccount.GetServiceAccountToken(v.kubeClient, namespace, serviceAccount.Name)
+
+	return &api.Config{Address: vlt.URL}, token, serviceAccount.Name, err
+}
+
+func (v *VaultClientFactory) getServiceAccountFromVault(vault *Vault) (*v1.ServiceAccount, error) {
+	return v.kubeClient.CoreV1().ServiceAccounts(vault.Namespace).Get(vault.AuthServiceAccountName, meta_v1.GetOptions{})
+}
+
+func getTokenFromVault(role string, jwt string, vaultClient *api.Client) (string, error) {
```
ccojocar: maybe turn this into a method => `func (v *VaultClientFactory) getTokenFromVault(...`

pkg/vault/vaultselector.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (v vaultSelectorImpl) selectVault(vaults []*Vault) (*Vault, error) {
+	surveyOpts := survey.WithStdio(v.In, v.Out, v.Err)
+	vaultMap, vaultNames := make(map[string]*Vault, len(vaults)), make([]string, len(vaults))
+	for i, vault := range vaults {
+		vaultMap[vault.Name] = vault
+		vaultNames[i] = vault.Name
+	}
+	prompts := &survey.Select{
```
ccojocar: I would move this `survey` into cmd. I think it belongs to the command no to the vault package. I know that we have this in various places but we should start to separate the command specific logic.

pkg/jx/cmd/get_vault_config.go
```diff
@@ -0,0 +1,97 @@
...
+	// Echo the client config out to the command line to be piped into bash
+	if o.terminal == "" {
+		if runtime.GOOS == "windows" {
+			o.terminal = "cmd"
+		} else {
+			o.terminal = "sh"
+		}
+	}
+	if o.terminal == "cmd" {
+		_, _ = fmt.Fprintf(o.Out, "set VAULT_ADDR=%s\nset VAULT_TOKEN=%s\n", url.String(), token)
```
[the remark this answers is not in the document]
agentgonzo: That's not quite the same. 

`os.Setenv` will set the env parameter for the current (`jx`) process, which will get discarded when the process terminates. This needs to be able to set the user's environment for use with the official vault CLI. Unfortunately (given environment variables get discarded when the current running process exits) the only way to do this is to `export` a string that the user can pipe into the current terminal.
This is the way that AWS tools do docker config (and the easiest way to do it) - see https://docs.aws.amazon.com/cli/latest/reference/ecr/get-login.html

pkg/vault/jxvaulter.go
```diff
@@ -0,0 +1,45 @@
...
+	// Change this when we decide what kind of schema/pattern we should use for storing secrets (which will be better
+	// understood when we start to store secrets)
+	secrets, err := v.client.Logical().List("secret")
+	if err != nil {
+		return nil, err
+	}
+
+	out := make([]string, 0)
+	for key, value := range secrets.Data {
+		_ = key
```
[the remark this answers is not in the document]
agentgonzo: I think a remnant of debugging/dev.

pkg/vault/jxvaulter.go
agentgonzo: The underlying vault store can support more than just a simple string-->string mapping.
ie, any given secret can have any format supported by JSON. That's why I left it as just the list of secret names for the moment. We can expand/fix this when we start storing secrets in vault. I didn't want to guess ahead of time how we may store secrets without knowing what secrets we need to storre

pkg/vault/jxvaulter.go
agentgonzo: Errr... leftovers from debugging :-(

pkg/kube/serviceaccount/serviceaccounts.go
```diff
@@ -1,4 +1,4 @@
-package kube
+package serviceaccount
```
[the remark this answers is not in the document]
agentgonzo: It already is: pkg/kube/serviceaccount/serviceaccounts.go ;-)

pkg/vault/vaultfactory.go
agentgonzo: I could, but it doesnt' need anything from the VaultClientFactory struct to do it's job, so I don't see the benefit. We'd still have to pass in the `api.Client` object because I don't like the idea of the VaultClientFactory maintaining any clients that it has created.

pkg/vault/vaultfactory.go
agentgonzo: It's used in the _test package, so it can't be :-(

pkg/vault/jxvaulter.go
ccojocar: Is it the returned type a string, even though the content type is some serialised JSON. Or is it an interface{}?

agentgonzo: The returned type is a `string` (being the name of the secret). This method doesn't return the underlying data.

PR #2168, opened 2018-11-05
fix: get `jx create addon istio` working with recent releases

jstrachan:
looks like git clone doesn't work so you need to download a tarball/zip of a release to be able to install via helm


rawlingsj:
This just failed with an OOM in the build pod.

PR #2170, opened 2018-11-05
fix: Use HA URL for chartmuseum

pmuir:
#### Submitter checklist

#### Description
Use HA URL for chartmuseum

fixes #

### DIFF
--- a/pkg/jx/cmd/init.go
+++ b/pkg/jx/cmd/init.go
@@ -73,7 +73,7 @@ const (
 	// INGRESS_SERVICE_NAME service name for ingress controller
 	INGRESS_SERVICE_NAME = "jxing-nginx-ingress-controller"
 	// DEFAULT_CHARTMUSEUM_URL default URL for Jenkins X ChartMuseum
-	DEFAULT_CHARTMUSEUM_URL = "https://chartmuseum.jx.cd.jenkins-x.io"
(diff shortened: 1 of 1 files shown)


pkg/jx/cmd/init.go
```diff
@@ -73,7 +73,7 @@ const (
 	// INGRESS_SERVICE_NAME service name for ingress controller
 	INGRESS_SERVICE_NAME = "jxing-nginx-ingress-controller"
 	// DEFAULT_CHARTMUSEUM_URL default URL for Jenkins X ChartMuseum
-	DEFAULT_CHARTMUSEUM_URL = "https://chartmuseum.jx.cd.jenkins-x.io"
+	DEFAULT_CHARTMUSEUM_URL = "http://chartmuseum.jenkins-x.io"
```
agentgonzo: why not https?

pmuir: It's a GCS bucket and it's quite hard to add a cert to a GCS bucket (/cc @i0n). It would be nice though, so create an issue?

pmuir: Well to be more direct, we don't have a valid cert right now :-)

agentgonzo: We should probably get one ;-). How do we get HTTPS certs? letsencrypt or another provider?

PR #2178, opened 2018-11-06
feat(preview): Default to batch mode when running in pipeline.

markawm:
When running inside pipeline, use batch mode by default. If the user specifies the --batch-mode
command line option, then always honour that.

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/jx/cmd/preview.go
+++ b/pkg/jx/cmd/preview.go
@@ -106,7 +106,7 @@ func NewCmdPreview(f Factory, in terminal.FileReader, out terminal.FileWriter, e
 			options.Cmd = cmd
 			options.Args = args
 			//Default to batch-mode when running inside the pipeline (but user override wins).
-			if !cmd.Flag("batch-mode").Changed {
+			if !cmd.Flag(optionBatchMode).Changed {
 				options.BatchMode = options.Factory.IsInCDPIpeline()
 			}
 			err := options.Run()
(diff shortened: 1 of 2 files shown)


pkg/jx/cmd/preview.go
```diff
@@ -105,6 +105,10 @@ func NewCmdPreview(f Factory, in terminal.FileReader, out terminal.FileWriter, e
 		Run: func(cmd *cobra.Command, args []string) {
 			options.Cmd = cmd
 			options.Args = args
+			//Default to batch-mode when running inside the pipeline (but user override wins).
+			if !cmd.Flag("batch-mode").Changed {
```
pmuir: Having added optionBatchMode as a constant to `common.go` it would make sense to use it here (and also ask people to start using it) - I've been wanting to make this change for a while!

Otherwise 👍

markawm: D'oh! That is _totally_ what I meant to do and the reason I made them all constants. I just managed to miss out the other half of it! :-(

pkg/jx/cmd/preview.go
```diff
@@ -105,6 +105,10 @@ func NewCmdPreview(f Factory, in terminal.FileReader, out terminal.FileWriter, e
 		Run: func(cmd *cobra.Command, args []string) {
 			options.Cmd = cmd
 			options.Args = args
+			//Default to batch-mode when running inside the pipeline (but user override wins).
+			if !cmd.Flag(optionBatchMode).Changed {
+				options.BatchMode = options.Factory.IsInCDPIpeline()
```
jglick: Not apparently introduced here, but FYI a typo (s/`I`/`i`/)

markawm: cheers for spotting & fixing.

PR #2197, opened 2018-11-07
feat: recursively delete extension children

pmuir:
#### Submitter checklist

fixes #

### DIFF
--- a/pkg/extensions/management.go
+++ b/pkg/extensions/management.go
@@ -1,0 +1,27 @@
+package extensions
+
+import (
+	jenkinsv1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	"github.com/jenkins-x/jx/pkg/log"
+)
+
+// GetAndDeduplicateChildrenRecursively will walk a tree of extensions rooted at ext and add them to the flattened
+// set exts. The lookupByUUID map is used to resolve UUID references to Extensions,
+// and a warning will be emitted if the extension is not present in the lookupByUUID map.
+func GetAndDeduplicateChildrenRecursively(ext jenkinsv1.Extension, lookupByUUID map[string]jenkinsv1.Extension,
+	exts map[string]*jenkinsv1.Extension) error {
+	// Add the ext
+	exts[ext.Spec.UUID] = &ext
+	// Recursively find exts
+	for _, childUUID := range ext.Spec.Children {
+		if child, ok := lookupByUUID[childUUID]; ok {
+			err := GetAndDeduplicateChildrenRecursively(child, lookupByUUID, exts)
+			if err != nil {
+				return err
+			}
+		} else {
+			log.Warnf("Unable to find child %s of %s\n", childUUID, ext.Spec.FullyQualifiedName())
+		}
+	}
+	return nil
+}
--- a/pkg/jx/cmd/delete_extension.go
+++ b/pkg/jx/cmd/delete_extension.go
@@ -127,10 +127,12 @@ func (o *DeleteExtensionOptions) Run() error {
 	}
 
 	names := make([]string, 0)
-	lookup := make(map[string]jenkinsv1.ExtensionSpec)
+	lookup := make(map[string]jenkinsv1.Extension)
+	lookupByUUID := make(map[string]jenkinsv1.Extension)
 	for _, e := range exts.Items {
-		lookup[e.Spec.FullyQualifiedKebabName()] = e.Spec
+		lookup[e.Spec.FullyQualifiedKebabName()] = e
 		names = append(names, e.Spec.FullyQualifiedKebabName())
+		lookupByUUID[e.Spec.UUID] = e
 	}
 	if o.All {
 		args = names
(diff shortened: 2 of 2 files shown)


pkg/extensions/management.go
```diff
@@ -0,0 +1,25 @@
+package extensions
+
+import (
+	jenkinsv1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	jenkinsclientv1 "github.com/jenkins-x/jx/pkg/client/clientset/versioned/typed/jenkins.io/v1"
+	"github.com/jenkins-x/jx/pkg/log"
+)
+
+func GetChildrenRecursively(ext jenkinsv1.Extension, configLookup map[string]jenkinsv1.ExtensionConfig, lookupByUUID map[string]jenkinsv1.Extension, extensionsClient jenkinsclientv1.ExtensionInterface, ns string, verbose bool) ([]jenkinsv1.Extension, error) {
```
agentgonzo: NIT: wrap lines when >120 chars

agentgonzo: Needs goDoc (as to what the different parameters mean)

pkg/jx/cmd/delete_extension.go
```diff
@@ -144,26 +146,48 @@ func (o *DeleteExtensionOptions) Run() error {
...
+		extensionsToDeleteMap[e.Spec.UUID] = e
+	}
+
+	extensionsToDeleteStrings := make([]string, 0)
+	for _, e := range extensionsToDeleteMap {
+		extensionsToDeleteStrings = append(extensionsToDeleteStrings, e.Spec.FullyQualifiedName())
+	}
+
+	// TODO display the extensions to delete in a tree view
+	if !o.BatchMode && !util.Confirm(fmt.Sprintf("You are about to delete the Extensions: %s", strings.Join(extensionsToDeleteStrings, ", ")), false, "The list of Extensions to be deleted", o.In, o.Out, o.Err) {
```
agentgonzo: Do you want some kind of `-y` protection in batch mode? That would be analagous to saying 'yes' to the prompt in interactive mode

agentgonzo: Can you split this across multiple lines for readability?

pkg/jx/cmd/delete_extension.go
```diff
@@ -172,12 +196,12 @@ func (o *DeleteExtensionOptions) Run() error {
 				return err
 			}
 		}
-		err := extensionsClient.Delete(name, &metav1.DeleteOptions{})
+		err := extensionsClient.Delete(ext.ObjectMeta.Name, &metav1.DeleteOptions{})
 		if err != nil {
```
agentgonzo: Whilst you're in this area, this will break out of the loop early if an extension fails to delete, which is normally non-ideal.
Eg, if you have `a`, `b`, `c` to delete and `b` fails, `a` will have been deleted, but it won't even attempt to delete `c` once `b` fails.
If this is what you want, then that's OK (but it's normally not)
You probably want to add `err` to a list and return that list-of-errors (or combine them into one error summarising which ones failed to uninstall) and then return that at the end of the `for` loop

pkg/extensions/management.go
```diff
@@ -0,0 +1,25 @@
...
+	result := make([]jenkinsv1.Extension, 0)
+	// Recursively find children
+	for _, childUUID := range ext.Spec.Children {
+		if child, ok := lookupByUUID[childUUID]; ok {
+			children, err := GetChildrenRecursively(child, configLookup, lookupByUUID, extensionsClient, ns, verbose)
```
agentgonzo: NIT: I would rename `children` to `grandchildren` here

pkg/extensions/management.go
```diff
@@ -0,0 +1,25 @@
+package extensions
+
+import (
+	jenkinsv1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	jenkinsclientv1 "github.com/jenkins-x/jx/pkg/client/clientset/versioned/typed/jenkins.io/v1"
+	"github.com/jenkins-x/jx/pkg/log"
+)
+
+func GetChildrenRecursively(ext jenkinsv1.Extension, configLookup map[string]jenkinsv1.ExtensionConfig, lookupByUUID map[string]jenkinsv1.Extension, extensionsClient jenkinsclientv1.ExtensionInterface, ns string, verbose bool) ([]jenkinsv1.Extension, error) {
+	result := make([]jenkinsv1.Extension, 0)
```
agentgonzo: I would suggest that you make `result` a set (well, a `map`) and make this method return as set (map) to return unique children, rather than making the caller do the de-duplication.

pkg/extensions/management.go
pmuir: Do you know how to make GoLand do this for me?

pkg/jx/cmd/delete_extension.go
pmuir: I don't think we generally do that?

pkg/jx/cmd/delete_extension.go
pmuir: Makes sense.

pkg/extensions/management.go
pmuir: Good idea.

pkg/extensions/management.go
agentgonzo: image

pmuir: Very nice!

pkg/jx/cmd/delete_extension.go
agentgonzo: It's an open question. If this is the correct behaviour I'm fine with it - just wanted to ask to make sure

pmuir:
Yeah, tests need adding, I know ;-)

pkg/jx/cmd/delete_extension.go
```diff
@@ -144,40 +146,56 @@ func (o *DeleteExtensionOptions) Run() error {
...
+		}
+	}
+
+	extensionsToDeleteStrings := make([]string, 0)
+	for _, e := range extensionsToDelete {
+		extensionsToDeleteStrings = append(extensionsToDeleteStrings, e.Spec.FullyQualifiedName())
+	}
+
+	// TODO display the extensions to delete in a tree view
+	if !o.BatchMode && !util.Confirm(fmt.Sprintf("You are about to delete the Extensions: %s",
```
agentgonzo: Nit: Autoformatting isn't always as good as humans. Try to keep different methods/parameter-sets on the same line, eg:
```
if !o.BatchMode && !util.Confirm(
        fmt.Sprintf("You are about to delete the Extensions: %s", strings.Join(extensionsToDeleteStrings, ", ")), 
        false, list of Extensions to be deleted", o.In, o.Out, o.Err) {
```
or some variant thereof reads easier to me

pmuir: I hate multi-line statements, so tbh it's greek to me ;-)

PR #2220, opened 2018-11-09
feat: load jx-install-config from encrypted secrets stored in git

garethjevans:

fixes #


pkg/jx/cmd/upgrade_platform.go
```diff
@@ -271,5 +272,24 @@ func (o *UpgradePlatformOptions) Run() error {
...
 	}
 
-	return o.Helm().UpgradeChart(o.Chart, o.ReleaseName, ns, &targetVersion, false, nil, false, false, values, valueFiles)
+	err = o.Helm().UpgradeChart(o.Chart, o.ReleaseName, ns, &targetVersion, false, nil, false, false, values, valueFiles)
+	if err != nil {
+		return errors.Wrap(err, "unable to upgrade helm chart")
+	}
+
+	if o.Flags.CleanupTempFiles {
+		err = os.Remove(secretsFileName)
```
jtnord: is this really fatal enough to warrant an error (rather than a warning?) after all the upgrade completed as far I I understand, just that the files failed to be cleaned up.

Also shouln;t this be removed even if helm failed to upgrade (it sounds like it is pretty "secrety" to leave lying around?

garethjevans: This code has been taken from the install command which deals with the tmp files in the same way.

PR #2228, opened 2018-11-09
improvement: Add convenient way to get HTTP Client with sensible timeouts

mikecirioli:
Add convenience methods for getting http.Client with sensible default timeouts, then use the new methods to fix some points in the current code base where we need those timeouts.

- ~~[WIP]~~ unit tests
- ~~[WIP]~~ migrate trouble spots in current codebase to use new methods

#### Submitter checklist

fixes #


pkg/util/http_utils.go
```diff
@@ -0,0 +1,42 @@
...
+var jxDefaultTransport http.RoundTripper = &http.Transport{
+	DialContext: (&net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+		DualStack: true,
+	}).DialContext,
+	MaxIdleConns:          100,
+	IdleConnTimeout:       90 * time.Second,
+	TLSHandshakeTimeout:   10 * time.Second,
+	ExpectContinueTimeout: 1 * time.Second,
```
rawlingsj: can we make these configurable with env vars maybe, using the values above as defaults?

mikecirioli: sure thing

pkg/util/http_utils_test.go
```diff
@@ -0,0 +1,74 @@
+package util
+
+import (
+	"github.com/jenkins-x/jx/src/github.com/stretchr/testify/assert"
```
rawlingsj: shouldn't this be "github.com/stretchr/testify/assert"?

mikecirioli: possibly, although my IDE provided multiple suggestions and it was not clear which one to choose.  I'll try the one you have suggested.

rawlingsj: yeah worth a try

mikecirioli: that did the trick, thanks!

pmuir: Otherwise you are using a transitive dependency rather than a direct dependency

PR #2225, opened 2018-11-09
Suppressing call to addTeamSettingsCommandsFromTags since it slows down every jx command

jglick:
I noticed that `jx shell` paused for around a second before giving me every prompt when using my GKE cluster. Clearly `$PS1` was responsible, and in fact it turned out that `jx prompt` pauses. If I switch to an EKS cluster, the pause is a little longer, and when I switch to a Minikuke cluster with the Minikuke VM switched off, it pauses for a couple of _minutes_ until it gets a timeout. A bunch of `strace` tuning later, I found a lot of stack traces like

```
[pid 12737] 09:02:38.003019 <... epoll_ctl resumed> ) = -1 EPERM (Operation not permitted)
 > ~/.jx/bin/jx(runtime.epollctl+0x18) [0x5a048]
 > ~/.jx/bin/jx(internal/poll.runtime_pollClose+0x60) [0x277e0]
 > ~/.jx/bin/jx(internal/poll.(*pollDesc).init+0xb7) [0x84e97]
 > ~/.jx/bin/jx(internal/poll.(*FD).Init+0x5f) [0x85b1f]
 > ~/.jx/bin/jx(os.newFile+0xed) [0x9008d]
 > ~/.jx/bin/jx(os.openFileNolog+0x1aa) [0x9039a]
 > ~/.jx/bin/jx(os.OpenFile+0x5f) [0x8e31f]
 > ~/.jx/bin/jx(os.Open+0x46) [0x8e206]
 > ~/.jx/bin/jx(io/ioutil.ReadFile+0x52) [0xfaa12]
 > ~/.jx/bin/jx(k8s.io/client-go/transport.dataFromSliceOrFile+0x50) [0xb26b50]
 > ~/.jx/bin/jx(k8s.io/client-go/transport.loadTLSFiles+0xcb) [0xb2699b]
 > ~/.jx/bin/jx(k8s.io/client-go/transport.tlsConfigKey+0x6e) [0xb2397e]
 > ~/.jx/bin/jx(k8s.io/client-go/transport.(*tlsTransportCache).get+0x81) [0xb233e1]
 > ~/.jx/bin/jx(k8s.io/client-go/transport.New+0x139) [0xb262a9]
 > ~/.jx/bin/jx(k8s.io/client-go/rest.TransportFor+0x68) [0xb36a28]
 > ~/.jx/bin/jx(k8s.io/client-go/rest.RESTClientFor+0xf3) [0xb2d043]
[...]

PR #2234, opened 2018-11-12
Fix: Avoiding unnecessary initialization just to get struct members

msvticket:


jglick:
Supersedes #2225. Verified in 1.3.560.

PR #2246, opened 2018-11-14
Fix: Allow non admin to execute various jx commands

msvticket:
A jenkins-x user administered with `jx edit userroles` and `jx controller role` doesn't have the
permissions to register CRDs. But the CRDs should have been registerad during install or upgrade anyway.
So the following commands don't register CRD anymore: `create devpod`, `create quickstart`, `import` and all `get`.

#### Submitter checklist


jstrachan:
@msvticket ah cool - just double checking the code. Actually I got the wrong step - its `jx step env apply` not `jx step post install` :) - I can sort that out now and double check the `RegisterAllCRDs()` works with the new `--gitops` mode for installing

jstrachan:
@msvticket ignore all the above - this PR should fix those issues https://github.com/jenkins-x/jx/pull/2248 (the code was pretty fresh in my head as I've just been hacking on the new `--gitops` mode)

msvticket:
So, should I revert my change in `step post install`?

PR #2252, opened 2018-11-14
Vault install

agentgonzo:
#### Submitter checklist

#### Description
a `vault-operator` is installed by default when using either of the following two commands:
* `jx install`
* `jx create cluster gke`
In conjunction with either of the two flags:
* `--gitops`
* `--vault` (new flag)

Please feel free to tell me if this is all wrong as I don't fully understand the codebase yet

### DIFF
--- a/pkg/jx/cmd/create_addon_vault.go
+++ b/pkg/jx/cmd/create_addon_vault.go
@@ -77,11 +77,12 @@ func NewCmdCreateAddonVault(f Factory, in terminal.FileReader, out terminal.File
 
 // Run implements the command
 func (o *CreateAddonVaultOptions) Run() error {
-	_, _, err := o.KubeClient()
-	if err != nil {
-		return fmt.Errorf("cannot connect to Kubernetes cluster: %v", err)
-	}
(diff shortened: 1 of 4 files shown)


pkg/jx/cmd/create_addon_vault.go
```diff
@@ -77,11 +77,12 @@ func NewCmdCreateAddonVault(f Factory, in terminal.FileReader, out terminal.File
 
 // Run implements the command
 func (o *CreateAddonVaultOptions) Run() error {
-	_, _, err := o.KubeClient()
```
agentgonzo: This call to `o.KubeClient()` seemed to be a bit pointless

ccojocar: This is just to check if there is a kuberntes connection when running the create addon. It can be ignored.

ccojocar:
We would need to create also a vault after the operator is installed. It's better to send this in a different PR.

agentgonzo:
>We would need to create also a vault after the operator is installed. It's better to send this in a different PR.

Yep, that was my plan!

PR #2257, opened 2018-11-14
Fail the build if there are uncommitted modifications

jglick:
Copied from https://github.com/jenkinsci/jenkins/commit/9d269168a59f2863467bf17ebe1ed8800da04bd2. This would have failed on #2167, for example, as building that touches `go.mod` + `go.sum` to register new modules. The build succeeds, but is presumably nondeterministic as Go is just guessing at the versions to use each time. We want the author of the PR to commit these files and have that commit be tested.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,6 +38,7 @@ pipeline {
                     sh "echo building Pull Request for preview ${TEAM}"
 
                     sh "make linux"
+                    sh 'test `git status --short | tee /dev/stderr | wc --bytes` -eq 0'
                     sh "make test-slow-integration"
                     sh "./build/linux/jx --help"
 


agentgonzo:
@jstrachan - Is this worth putting in the 'default' `Jenkinsfile` for go projects  (Is this a draft pack?) so that all new projects benefit from this change?

jglick:
> Is this worth putting in the 'default' `Jenkinsfile` for go projects

It could potentially be useful in _any_ `Jenkinsfile`; as I noted in the description, I originally wrote this trick for Jenkins core, when I found in https://github.com/jenkinsci/jenkins/pull/2860 that a test was creating a nonignorable file.

PR #2274, opened 2018-11-16
chore add missing entries from go.sum

jtnord:
#### Submitter checklist

#### Description

Add some entries that where missing from `go.sum`


jglick:
Relation to #2257?

jtnord:
@jglick that doesn't fully work.  the makefile and build won;t pull uneeded dependencies it seems and some things may be conditionally compiled (haven;t checked)  but on my windows system if I run
`go list ./...` which is used in the `Makefile` bootstrap then it will modify the files (presumably as I am on windows and it pulls those files.
If i run `set GOOS=linux & go list ./...` then it does not update the files.

I expected others to be able to reproduce this on linux/mac by doing `export GOOS=windows & go list ./...` but that apparantly does not show the issue.

https://kubernetes.slack.com/archives/C9LTHT2BB/p1542375114244400?thread_ts=1542116244.174600&cid=C9LTHT2BB

jglick:
@jtnord maybe `go mod tidy` is what you were looking for? Just guessing.

PR #2282, opened 2018-11-19
bug(controller buildnumbers): Encode PipelineActivity names to avoid collisions

markawm:
#### Submitter checklist

#### Description

Current impl. converts any name-illegal chars into -, which could cause collisions and
therefore two pipelines to share the same build number ID space.
(there is also a bug in the current impl. that won't properly support upper case owner/repo/branch names).

fixes #

### DIFF
--- a/pkg/kube/activity_test.go
+++ b/pkg/kube/activity_test.go
@@ -240,11 +240,11 @@ func TestPipelineID(t *testing.T) {
 
 	// Upper case allowed in our ID, but not in the K8S 'name'.
 	pID = kube.NewPipelineID("OwNeR1", "rEpO1", "BrAnCh1")
-	validatePipelineID(t, pID, "OwNeR1/rEpO1/BrAnCh1", "owner1-repo1-branch1")
+	validatePipelineID(t, pID, "OwNeR1/rEpO1/BrAnCh1", ".ow.ne.r1-r.ep.o1-.br.an.ch1")
 
 	//Punctuation other than '-' and '.' not allowed in K8S 'name'.
 	pID = kube.NewPipelineID("O/N!R@1", "therepo", "thebranch")
-	validatePipelineID(t, pID, "O/N!R@1/therepo/thebranch", "o-n%21r%401-therepo-thebranch")
+	validatePipelineID(t, pID, "O/N!R@1/therepo/thebranch", ".o-.n.21.r.401-therepo-thebranch")
 }
 
 func validatePipelineID(t *testing.T, pID kube.PipelineID, expectedID string, expectedName string) {
--- a/pkg/kube/activity_test.go
+++ b/pkg/kube/activity_test.go
@@ -232,3 +232,22 @@ func TestCreatePipelineDetails(t *testing.T) {
 		}
 	}
 }
+
+func TestPipelineID(t *testing.T) {
+	// A simple ID.
+	pID := kube.NewPipelineID("o1", "r1", "b1")
+	validatePipelineID(t, pID, "o1/r1/b1", "o1-r1-b1")
+
+	// Upper case allowed in our ID, but not in the K8S 'name'.
+	pID = kube.NewPipelineID("OwNeR1", "rEpO1", "BrAnCh1")
+	validatePipelineID(t, pID, "OwNeR1/rEpO1/BrAnCh1", "owner1-repo1-branch1")
+
+	//Punctuation other than '-' and '.' not allowed in K8S 'name'.
+	pID = kube.NewPipelineID("O/N!R@1", "therepo", "thebranch")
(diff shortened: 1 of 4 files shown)


pkg/util/k8s.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// EncodeKubernetesName takes a string and turns it into a form suitable for use as a Kubernetes name.
+// K8S names are lower case, numbers, '-', and '.'. Invalid characters are percent-encoded in a _similar_ way to URL
+// encoding, with a percent followed by the numerical character code. Conversion is:
+//   Upper case letters -> lower case.
+//   'a-z', '0-9', '-', '.' -> left as-is.
+//   Any other characters: percent-encoded ('%' + rune hex code).
+func EncodeKubernetesName(name string) string {
+	name = strings.ToLower(name)
```
jtnord: what about collisions?
I can have `org/repo/branch` and `org/repo/BrAnCh` and they are different things.

pkg/util/k8s.go
```diff
@@ -0,0 +1,31 @@
...
+	encodedName := strings.Builder{}
+
+	for _, ch := range name {
+		if allowedInK8SName(ch) {
+			encodedName.WriteRune(ch)
+		} else {
+			encodedName.WriteString(fmt.Sprintf("%%%X", int(ch)))
```
jtnord: do we need a function to "UnEncode" these (e.g. jx get builds) ?  what does the tools show (hopefully not the percent encoded version?)

markawm: Not that I can see atm.
tbh, I'm not entirely sure where we display these atm, but yes tools could well show them. For the values that currently work, the tools will show the same stuff. For the values that currently don't work at all, the tools should display some weird stuff. But at least it'll work. :-)

pkg/util/k8s_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestEncodeKubernetesNameNoEncoding(t *testing.T) {
+	original := "abcdefghijklmnopqrstuvwxyz-.0123456789"
```
jtnord: `T.parallel()`

pkg/util/k8s_test.go
```diff
@@ -0,0 +1,33 @@
...
+	assert.Equal(t, original, original, "Didn't expect _any_ encoding to happen.")
+}
+
+func TestEncodeKubernetesNameUpper(t *testing.T) {
+	original := "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
```
jtnord: `T.parallel()`

pkg/util/k8s_test.go
```diff
@@ -0,0 +1,33 @@
...
+	expected := strings.ToLower(".a.b.c.d.e.f.g.h.i.j.k.l.m.n.o.p.q.r.s.t.u.v.w.x.y.z")
+	assert.Equal(t, expected, EncodeKubernetesName(original), "Expected upper->.lower conversion.")
+}
+
+func TestEncodeKubernetesPunctuationString(t *testing.T) {
+	//Note _valid_ - and . hidden in the string below.
```
jtnord: `T.parallel()`

pkg/util/k8s_test.go
```diff
@@ -0,0 +1,33 @@
...
+	original := "!\"£$%^&*()-=_+¬`;'#:@~,./<>?[]{}\\|€"
+	expected := ".21.22.A3.24.25.5E.26.2A.28.29-.3D.5F.2B.AC.60.3B.27.23.3A.40.7E.2C..2F.3C.3E.3F.5B.5D.7B.7D.5C.7C.20AC"
+	assert.Equal(t, expected, EncodeKubernetesName(original), "Expected punctuation to be period-encoded.")
+}
+
+func TestEncodeKubernetesNameExtendedRunes(t *testing.T) {
+	original := "⌘日本語"
```
jtnord: `T.parallel()`

pkg/util/k8s.go
markawm: Good point, that _was_ the original point of this change, after all!
Upper case are now converted to lower case (to keep things as readable as possible), and given a prefix to avoid upper/lower collisions like this.
I also realised another key requirement hadn't been met with the original encoding, as % isn't a valid character in a K8S name. So the encoding is as-before with codes, but now uses period in place of percent.

pkg/kube/activity_test.go
```diff
@@ -232,3 +232,22 @@ func TestCreatePipelineDetails(t *testing.T) {
 		}
 	}
 }
+
+func TestPipelineID(t *testing.T) {
+	// A simple ID.
```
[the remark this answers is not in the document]
markawm: True... will add.

PR #2291, opened 2018-11-20
feat(jx) add the OS details to `jx version` output

jtnord:
#### Submitter checklist

#### Description

It is useful to know the OS version of the client in reports so this
change addes an "Operating System" column to the details listed by
`jx version`

tested on `Windows 10 Pro 1803`, `OSX` (some version), `Bash on Ubuntu on Windows` and `centos`, `debian` and `alpine` docker images.

### DIFF
--- a/pkg/util/system/version_darwin.go
+++ b/pkg/util/system/version_darwin.go
@@ -1,0 +1,45 @@
+// +build darwin
+
+package system
+
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+// returns a human friendly string of the current OS
+// in the case of an error this still returns a valid string for the details that can be found.
+func GetOsVersion() (string,  error) {
+	retVal := "unknown OSX version"
+
+	// you can not run the command once to get all this :-(
+	pn, err := runCommand("sw_vers", "-productName")
+	if err != nil {
+		return retVal, err
+	}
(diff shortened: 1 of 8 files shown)


pkg/util/system/version_darwin.go
```diff
@@ -0,0 +1,45 @@
...
+
+	pb, err := runCommand("sw_vers", "-buildVersion")
+	if err != nil {
+		return fmt.Sprintf("%s unknown build", retVal), err
+	}
+	return fmt.Sprintf("%s build %s", retVal, pb), nil
+}
+
+
+func runCommand(command string, args ...string) (string, error) {
```
markawm: This is duplicated code - assume you've added both because of the conditional building, which means not dup'd in runtime, but still in src control.
There is already a very similar function in cmd.CommonOptions.RunCommand (though I think the way it interacts with the verbose flag is not ideal). I actually starting refactoring some of this into the util package (just util.RunCommandBackground). How about adding your version (i.e. one that cares about the output) as util.RunCommand()?

jtnord: So having thought about this some more, I can potentially see cases where we may want to get the OS version from the `util` package and that would cause circular dependencies.
I could export this function in the system package and refactor all other usage from it but that seems a bit too much work for a this small piece of code.

PR #2305, opened 2018-11-22
WIP bug (options): Make global CommonOptions available on all commands.

markawm:
#### Submitter checklist

#### Description

Add CommonOptions to the root command so that they are available (inherited) in all sub-commands
that are created. This also means that 'jx options' actually returns something, because it finds the
options inherited from the base command.

This PR has got pretty big... unavoidable really because the refactoring touches any usage of `CommonOptions`, which is obviously a lot!
Whilst it feels like the right fix, I have some 'fear' about the impact of this change, and this PR is just looking for thoughts & feedback atm.

Don't expect anyone to read all of it - in summary the change is:
 - Each of the specific Options (e.g. `AddOptions`) now contains a _pointer_ to `CommonOptions` rather than its own `CommonOptions`.
 - `cmd.NewJXCommand()` (which creates all the commands) now creates _one_ CommonOptions and passes it into the sub-commands it creates.
 - `cmd.NewJXCommand()` adds the common options once to the root command, so they are inherited by all sub-commands.

There is a conflict on the `-b` short option (batch-mode from common, and build from some specific commands). I have therefore **removed** the batch-mode short option. But concerned about any pipeline steps for e.g. that might break...
(I tried to remove the common batch-mode one when the build one is in use (as discussed on slack), but couldn't find a way to do this.)

fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ test-integration-report-html: get-test-deps test-integration
 	@gocov convert cover.out | gocov-html > cover.html && open cover.html
 
 test-slow-integration:
-	@CGO_ENABLED=$(CGO_ENABLED) $(GO) test -v -v -memprofile mem.out -count=1 -tags=integration -coverprofile=cover.out ./...
+	@CGO_ENABLED=$(CGO_ENABLED) $(GO) test -v -count=1 -tags=integration -coverprofile=cover.out ./...
 
 test-slow-integration-report: get-test-deps test-slow-integration
 	@gocov convert cover.out | gocov report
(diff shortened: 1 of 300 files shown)


markawm:
Thanks for the comments, @garethjevans - I made the changes you suggested: `-b` now means batch-mode universally, and there is no longer a short option for the few other commands that were using `-b` (generally to identify a build). This has just been held up by various build failures - now that is solved by the build is failing with an out of memory error. I'll keep trying it...

ccojocar:
@markawm Are you planning to move forward with this PR? It seems to need some rebase.

markawm:
@ccojocar, I was getting build failures running out of memory - since other builds are working now then I guess I've somehow introduced a memory leak (though not sure where/how...).
I'll rebase and try it again, but I was considering abandoning as I'm not sure its worth the effort to track it down. :-(

ccojocar:
@markawm Let's close this PR and you can open another one.

markawm:
ok, will do.

PR #2334, opened 2018-11-26
fix:  Add domain name validation when configuring ingress to loadbalancer

mikecirioli:
#### Submitter checklist

#### Description
Adds simple domain name validation (domain names must not contain any whitespace) when configuring LB ingress during installation.  Also adds unit test for `NoWhiteSpaceValidator`

fixes #


pkg/jx/cmd/init.go
```diff
@@ -925,3 +925,19 @@ func (o *CommonOptions) GetDomain(client kubernetes.Interface, domain string, pr
 
 	return domain, nil
 }
+
+//NoWhiteSpaceValidator is an input validator for the survey package that disallows any whitespace in the input
+func NoWhiteSpaceValidator() survey.Validator {
```
ccojocar: I would be nice if you could move this function in the `util/survey.go`, that can be reused in other parts of the code.

mikecirioli: that was actually something i was going to ask about this morning :)

PR #2362, opened 2018-11-29
feat(vault): apply the --vault flag to all commands

agentgonzo:
the `--vault` flag now applies to all commands.
When applied, the factory will use vault to store (some - still a WIP) secrets in vault rather than the local filesystem

### DIFF
--- a/pkg/jx/cmd/cmd.go
+++ b/pkg/jx/cmd/cmd.go
@@ -21,14 +21,18 @@ const (
     `
 )
 
+var _factory Factory
+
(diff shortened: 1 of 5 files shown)


pkg/jx/cmd/cmd.go
```diff
@@ -21,14 +21,18 @@ const (
     `
 )
 
+var _factory Factory
```
agentgonzo: I'm not sure if this is the best way to pass the `factory` object over to the `preRun` function, but I couldn't find an easy way to do it without making it a variable at the package level.

PR #2375, opened 2018-11-29
feat(vault): Store a flag in the jx-install-config to denote if secrets are to be stored in vault

agentgonzo:
If `jx install --vault` is run, secrets get stored in vault. With this PR, it will now update the `jx-install-config` configmap with a new entry: `useVaultForSecrets: "true"` that tells the client that secrets are stored in vault (as opposed to on the file systme).

This can (and is) queried by the client to determine whether to load the secrets from vault.

### DIFF
--- a/pkg/io/secrets/secret_locations.go
+++ b/pkg/io/secrets/secret_locations.go
@@ -9,31 +9,51 @@ import (
 
 const vaultSecretsMarker = "useVaultForSecrets"
 
-var usingVault *bool // use a tri-state boolean. nil means uninitialised (so need to lookup from cluster)
+type SecretLocation interface {
+	// InVault returns whether secrets are stored in Vault
+	InVault() bool
+	// SetInVault sets whether secrets are stored in Vault or not
+	SetInVault(useVault bool)
+}
+
+type secretLocation struct {
+	kubeClient kubernetes.Interface
+	namespace  string
+	usingVault *bool // use a tri-state boolean. nil means uninitialised (so need to lookup from cluster)
+}
+
+// NewSecretLocation creates a SecretLocation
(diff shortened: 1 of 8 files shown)


pkg/io/secrets/secret_locations.go
```diff
@@ -0,0 +1,50 @@
...
+		usingVault = newBool(useVault)
+		return nil
+	}, nil)
+	if err != nil {
+		logrus.Errorf("Error saving configmap %s: %v", kube.ConfigMapNameJXInstallConfig, err)
+	}
+}
+
+// UsingVaultForSecrets returns true if the cluster has been configured to store secrets in vault
+func UsingVaultForSecrets(kubeClient kubernetes.Interface, namespace string) bool {
```
ccojocar: These methods name are confusing. You just need a method to set the vault in the config map and one to get the flag from the configmap. Both in a stateless fashion. E.g. if in get not flag is set you return false by default.

pkg/io/secrets/secret_locations.go
```diff
@@ -0,0 +1,65 @@
...
+		kubeClient: kubeClient,
+		namespace:  namespace,
+	}
+}
+
+// UsingVaultForSecrets returns true if the cluster has been configured to store secrets in vault
+func (s *secretLocation) InVault() bool {
+	if s.usingVault == nil {
+		configMap := getInstallConfigMap(s.kubeClient, s.namespace)
+		b := configMap[vaultSecretsMarker] != ""
```
ccojocar: I think you need to handle nil. It case the configmap does not exists, the `getInstallConfigMap` will only log a message and return nil.

agentgonzo: good call

ccojocar: Great! Thanks

PR #2378, opened 2018-11-29
fix(vault): ensure that the created vault has a service account and a proper role binding

ccojocar:
#### Submitter checklist

#### Description

Ensure that the created vault has a service account and a proper role binding

Starting with kubernetes 1.10, a cluster role binding is required for service account token review. Vault
uses this feature to authenticate with a Kubernetes service account.

fixes #

### DIFF
--- a/pkg/kube/roles.go
+++ b/pkg/kube/roles.go
@@ -239,31 +239,45 @@ func DeleteClusterRoleBinding(kubeClient kubernetes.Interface, name string) erro
 	return nil
 }
 
+// IsClusterRole checks if a cluster role exists
+func IsClusterRole(kubeClient kubernetes.Interface, name string) bool {
+	_, err := kubeClient.RbacV1().ClusterRoles().Get(name, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return false
+	}
+	return true
+}
+
 // CreateClusterRole creates a new cluster role
 func CreateClusterRole(kubeClient kubernetes.Interface, namesapce string, name string,
 	apiGroups []string, resources []string, verbs []string) error {
-	_, err := kubeClient.RbacV1().ClusterRoles().Get(name, metav1.GetOptions{})
+	role := &rbacv1.ClusterRole{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       clusterRoleKind,
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Rules: []rbacv1.PolicyRule{
+			rbacv1.PolicyRule{
+				APIGroups: apiGroups,
+				Resources: resources,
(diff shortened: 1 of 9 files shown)


pkg/kube/roles.go
```diff
@@ -180,3 +187,79 @@ func userRolesFor(userKind string, userName string, envRoles map[string]*v1.Envi
...
+				Kind:     clusterRoleKind,
+				Name:     clusterRoleName,
+			},
+		}
+
+		_, err := kubeClient.RbacV1().ClusterRoleBindings().Create(rb)
+		if err != nil {
+			return errors.Wrap(err, "creating cluster role binding")
+		}
+		return nil
```
agentgonzo: this line is unnecessary is it will just fall through to the next `return` at the bottom of the function

pkg/vault/vault.go
```diff
@@ -91,10 +95,20 @@ func VaultGcpServiceAccountSecretName(vaultName string, clusterName string) stri
...
 
 // CreateVault creates a new vault backed by GCP KMS and storage
-func CreateVault(vaultOperatorClient versioned.Interface, name string, ns string,
+func CreateVault(kubeClient kubernetes.Interface, vaultOperatorClient versioned.Interface, name string, ns string,
 	gcpServiceAccountSecretName string, gcpConfig *GCPConfig, authServiceAccount string,
 	authServiceAccountNamespace string, secretsPathPrefix string) error {
 
+	err := createVaultServiceAccount(kubeClient, ns, name)
+	if err != nil {
+		return err
```
agentgonzo: wrap the error with a message

pkg/kube/roles.go
```diff
@@ -180,3 +187,79 @@ func userRolesFor(userKind string, userName string, envRoles map[string]*v1.Envi
 	}
 	return answer
 }
+
+// CreateClusterRoleBinding creates or extends a cluster role binding in a given namespace for a service account
+func CreateClusterRoleBinding(kubeClient kubernetes.Interface, namespace string, name string,
```
agentgonzo: This doesnt' actually do what the method/godoc implies (in that if it already exists, it just silently exists and does nothing).
I suggest either modifying the name/documentation or splitting into 2 methods, one to check if it exists and one to always Create it

agentgonzo: If a clusterrolebinding with the specified `name` already exists, then this method gives no feedback to the caller that it didn't create the cluster. Moreover, the namespace, serviceAccoundName and clusterRoleName passed into the method could differ from the `ClusterRoleBinding` already defined in the cluster, leading to confusion on behalf of the caller when the data in the clusterrolebinding differs from what they have called this method with (because it doesn't create anything).

The return value needs to indicate whether the cluster role was created or not. Either returning a boolean, or maybe even returning the ClusterRoleBinding (that either pre-existed or was created) to the user (or both)

PR #2397, opened 2018-12-03
feat: initial support for chart repo username/password

pmuir:
#### Submitter checklist

#### Description
* This works for non-GitOps but won’t work for GitOps mode as that requires vault support

Will add tests once we have GitOps support

fixes #


pmuir:
Out of date

PR #2405, opened 2018-12-04
chore() prevent external use of `KubeClientCached`

jtnord:
#### Submitter checklist

#### Description

`KubeClientCached` can be changed by the running code so it should not be relied upon externally.  Ideally it would not be relied upon internally either but that is a bigger change, so at least prevent external misuse of this.
The replacement is to use `KubeClient()` and check for any error.

Please check the changes to the test code is ok.

### DIFF
--- a/pkg/jx/cmd/upgrade_ingress_test.go
+++ b/pkg/jx/cmd/upgrade_ingress_test.go
@@ -19,10 +19,16 @@ import (
 type TestOptions struct {
 	cmd.UpgradeIngressOptions
 	Service *v1.Service
+	// for testing so we can return the same client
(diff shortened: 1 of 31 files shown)


pkg/jx/cmd/upgrade_ingress_test.go
```diff
@@ -20,12 +21,13 @@ type TestOptions struct {
 	Service *v1.Service
 }
 
+func (o *TestOptions) KubeClient() (kubernetes.Interface, string, error) {
```
jtnord: *REVIEWERS* please pay special attention to this file

ccojocar:
Close it favour of #2608

PR #2417, opened 2018-12-05
refactor(install): break the install command in smaller methods and make the cloud providers configuration more self contained

ccojocar:
#### Submitter checklist

#### Description

- Breaks the install command in smaller methods in oder to understand the flow
- Make the cloud provider configuration during the install more self contained in dedicated method
- Extract the helm values configuration and helm values file creation
- Separate the secrets files from helm values files
- Extract the GitOps logic in separate method
- Ensure that the GitOps secrets.yaml file is stored in Vault
- Ensure that secrets.yaml is fetch from vault when environment chart is installed

fixes #

### DIFF
--- a/pkg/jx/cmd/install.go
+++ b/pkg/jx/cmd/install.go
@@ -773,17 +773,17 @@ func (options *InstallOptions) installPlatformGitOpsMode(gitOpsEnvDir string, gi
 	valuesFile := filepath.Join(gitOpsEnvDir, helm.ValuesFileName)
 	err := helm.SaveRequirementsFile(requirementsFile, requirements)
 	if err != nil {
-		return errors.Wrapf(err, "Failed to save GitOps helm requirements file %s", requirementsFile)
+		return errors.Wrapf(err, "failed to save GitOps helm requirements file %s", requirementsFile)
 	}
 
 	err = configStore.Write(chartFile, []byte(GitOpsChartYAML))
 	if err != nil {
-		return errors.Wrapf(err, "Failed to save file %s", chartFile)
+		return errors.Wrapf(err, "failed to save file %s", chartFile)
 	}
 
 	err = helm.CombineValueFilesToFile(secretsFile, secretsFiles, JenkinsXPlatformChartName, nil)
 	if err != nil {
-		return errors.Wrapf(err, "Failed to generate %s by combining helm Secret YAML files %s", secretsFile, strings.Join(secretsFiles, ", "))
+		return errors.Wrapf(err, "failed to generate %s by combining helm Secret YAML files %s", secretsFile, strings.Join(secretsFiles, ", "))
 	}
 
 	if options.Flags.Vault {
@@ -803,7 +803,7 @@ func (options *InstallOptions) installPlatformGitOpsMode(gitOpsEnvDir string, gi
 	}
 	err = helm.CombineValueFilesToFile(valuesFile, valuesFiles, JenkinsXPlatformChartName, extraValues)
 	if err != nil {
-		return errors.Wrapf(err, "Failed to generate %s by combining helm value YAML files %s", valuesFile, strings.Join(valuesFiles, ", "))
+		return errors.Wrapf(err, "failed to generate %s by combining helm value YAML files %s", valuesFile, strings.Join(valuesFiles, ", "))
 	}
 
 	gitIgnore := filepath.Join(gitOpsDir, ".gitignore")
--- a/pkg/jx/cmd/install.go
+++ b/pkg/jx/cmd/install.go
@@ -128,7 +128,7 @@ const (
 
 	GitOpsChartYAML = `name: env
 version: 0.0.1
-description: GitOps Environment for this Environment 
+description: GitOps Environment for this Environment
 maintainers:
   - name: Team
 icon: https://www.cloudbees.com/sites/default/files/Jenkins_8.png
@@ -476,6 +476,11 @@ func (options *InstallOptions) Run() error {
 		err = options.createSystemVault(client, ns)
 	}
 
+	err = options.configureDockerRegistry(client, ns)
+	if err != nil {
+		return errors.Wrap(err, "configuring the docker registry")
+	}
+
 	helmConfig := &options.CreateEnvOptions.HelmValuesConfig
 	if helmConfig.ExposeController.Config.Domain == "" {
 		helmConfig.ExposeController.Config.Domain = options.InitOptions.Flags.Domain
@@ -483,54 +488,6 @@ func (options *InstallOptions) Run() error {
 	domain := helmConfig.ExposeController.Config.Domain
 	if domain != "" && addon.IsAddonEnabled("gitea") {
 		helmConfig.Jenkins.Servers.GetOrCreateFirstGitea().Url = "http://gitea-gitea." + ns + "." + domain
-	}
-	if options.Flags.Prow {
-		t := true
-		f := false
-		helmConfig.ControllerBuild.Enabled = &t
-		helmConfig.Jenkins.Enabled = &f
-	}
-
-	dockerRegistry, err := options.dockerRegistryValue()
-	if err != nil {
-		return errors.Wrap(err, "failed to get the docker registry value")
-	}
-
-	kubeConfig, _, err := options.Kube().LoadConfig()
-	if err != nil {
-		return err
-	}
-	if options.Flags.Provider == AKS {
-		/**
-		 * Assign ACR role to AKS
-		 */
-		server := kube.CurrentServer(kubeConfig)
-		azureCLI := aks.NewAzureRunner()
-		resourceGroup, name, cluster, err := azureCLI.GetClusterClient(server)
-		if err != nil {
-			return errors.Wrap(err, "failed to get cluster from Azure")
-		}
-		registryID := ""
-		helmConfig.PipelineSecrets.DockerConfig, dockerRegistry, registryID, err = azureCLI.GetRegistry(resourceGroup, name, dockerRegistry)
-		if err != nil {
-			return errors.Wrap(err, "failed to get registry from Azure")
-		}
-		azureCLI.AssignRole(cluster, registryID)
-		log.Infof("Assign AKS %s a reader role for ACR %s\n", util.ColorInfo(server), util.ColorInfo(dockerRegistry))
-	}
-
-	if options.Flags.Provider == IKS {
-		dockerRegistry = iks.GetClusterRegistry(client)
-		helmConfig.PipelineSecrets.DockerConfig, err = iks.GetRegistryConfigJSON(dockerRegistry)
-	}
-	if dockerRegistry != "" {
-		if helmConfig.Jenkins.Servers.Global.EnvVars == nil {
-			helmConfig.Jenkins.Servers.Global.EnvVars = map[string]string{}
-		}
-		helmConfig.Jenkins.Servers.Global.EnvVars["DOCKER_REGISTRY"] = dockerRegistry
-		if isOpenShiftProvider(options.Flags.Provider) && dockerRegistry == "docker-registry.default.svc:5000" {
-			options.enableOpenShiftRegistryPermissions(ns, helmConfig, dockerRegistry)
-		}
 	}
 
 	initOpts := &options.InitOptions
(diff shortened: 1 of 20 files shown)


pkg/jx/cmd/install.go
```diff
@@ -381,151 +383,229 @@ func (options *InstallOptions) Run() error {
...
+	if err != nil {
+		return errors.Wrap(err, "configure the kubectl")
+	}
+
+	options.Flags.Provider, err = options.GetCloudProvider(options.Flags.Provider)
+	if err != nil {
+		return errors.Wrapf(err, "retrieving cloud provider '%s'", options.Flags.Provider)
+	}
+
+	err = options.setMinikubeFromContext()
```
agentgonzo: Do we still support minikube?

pkg/jx/cmd/install.go
```diff
@@ -381,151 +383,229 @@ func (options *InstallOptions) Run() error {
...
+	}
+	providerEnvDir := filepath.Join(cloudEnvDir, fmt.Sprintf("env-%s", strings.ToLower(options.Flags.Provider)))
+	valuesFiles, secretsFiles, temporaryFiles, err := options.getHelmValuesFiles(configStore, providerEnvDir)
+	if err != nil {
+		return errors.Wrap(err, "getting the helm value files")
+	}
+
+	log.Infof("Installing Jenkins X platform helm chart from: %s\n", providerEnvDir)
+
+	options.Verbose = true
```
agentgonzo: This seems odd. Shouldn't it be up to the user to specify whether to be Verbose or not? (or is it required to have output from subsequent methods? - if so, i think this could do with a comment as to _why_ it is being done)

pkg/jx/cmd/install.go
```diff
@@ -381,151 +383,229 @@ func (options *InstallOptions) Run() error {
...
+	if err != nil {
+		return errors.Wrap(err, "selecting the Jenkins installation type")
+	}
+
+	err = options.configureAndInstallProw(ns)
+	if err != nil {
+		return errors.Wrap(err, "configuring and installing Prow")
+	}
+
+	err = options.verifyTiller(client, ns)
```
agentgonzo: Aren't we running without Tiller now?

pkg/jx/cmd/install.go
```diff
@@ -381,151 +383,229 @@ func (options *InstallOptions) Run() error {
...
-			options.helm = helm.NewHelmTemplate(helmCli, helmCli.CWD, client, originalNs)
-		} else {
-			helmTemplate, ok := helmer.(*helm.HelmTemplate)
-			if ok {
-				options.helm = helmTemplate
-			} else {
-				log.Warnf("Helm facade is not a *helm.HelmCLI or *helm.HelmTemplate: %#v\n", helmer)
-			}
-		}
+	err = options.configureTillerInDevEnvironment()
```
agentgonzo: Aren't we running without tiller now?

pkg/jx/cmd/install.go
```diff
@@ -381,151 +383,229 @@ func (options *InstallOptions) Run() error {
...
-			if err != nil || flag == false {
-				return errors.New("Existing tiller must be uninstalled first in order to use the jx in tiller less mode")
-			}
-			// Uninstall helm and tiller first to avoid using some older version
-			err = options.UninstallBinary(binDir, "tiller")
-			if err != nil {
-				return errors.Wrap(err, "uninstalling existing tiller binary")
-			}
-		}
+	err = options.configureHelm3(ns)
```
agentgonzo: helm3? How does this differ from Helm? Comment needed, or rename the method to `configureHelm` if we _only_ install helm?

pkg/jx/cmd/install.go
```diff
@@ -381,151 +383,229 @@ func (options *InstallOptions) Run() error {
...
-	dependencies = append(dependencies, helmBinary)
-	err = options.installRequirements(options.Flags.Provider, dependencies...)
+
+	err = options.applyGitOpsDevEnvironmentConfig(gitOpsEnvDir, ns)
 	if err != nil {
-		return errors.Wrap(err, "failed to install the platform requirements")
+		return errors.Wrap(err, "applying the GitOps development environment config")
 	}
 
+	log.Success("\nJenkins X installation completed successfully\n")
```
agentgonzo: the non-`f` log methods (like `log.Success` here as opposed to `log.Successf`) will automatically append the `\n`, so no need to put the final `\n` in the message. See https://github.com/fatih/color/blob/master/color.go#L444-L446 (which is what gets eventually called by this method)

pkg/jx/cmd/install.go
```diff
@@ -381,151 +383,229 @@ func (options *InstallOptions) Run() error {
...
+
+	options.logAdminPassword()
+
+	log.Infof("\nYour Kubernetes context is now set to the namespace: %s \n", util.ColorInfo(ns))
+	log.Infof("To switch back to your original namespace use: %s\n", util.ColorInfo("jx ns "+originalNs))
```
agentgonzo: for documentation purposes, we should use the fully-named `jx namespace` rather than the shorthand `jx ns`

pkg/jx/cmd/install.go
```diff
@@ -665,251 +693,209 @@ func (options *InstallOptions) Run() error {
...
+	if err != nil {
+		return errors.Wrapf(err, "Failed to save GitOps helm requirements file %s", requirementsFile)
 	}
 
-	// lets add any GitHub Enterprise servers
-	gitAuthCfg, err := options.CreateGitAuthConfigService()
+	err = configStore.Write(chartFile, []byte(GitOpsChartYAML))
 	if err != nil {
-		return errors.Wrap(err, "failed to create the git auth config service")
+		return errors.Wrapf(err, "Failed to save file %s", chartFile)
```
agentgonzo: nit: I think errors in go are supposed to be lowercase (I don't really mind anyway)

pkg/jx/cmd/install.go
```diff
@@ -665,251 +693,209 @@ func (options *InstallOptions) Run() error {
...
 	}
-	err = options.addGitServersToJenkinsConfig(helmConfig, gitAuthCfg)
+
+	err = helm.CombineValueFilesToFile(secretsFile, secretsFiles, JenkinsXPlatformChartName, nil)
 	if err != nil {
-		return errors.Wrap(err, "failed to add the Git servers to Jenkins config")
+		return errors.Wrapf(err, "Failed to generate %s by combining helm Secret YAML files %s", secretsFile, strings.Join(secretsFiles, ", "))
```
agentgonzo: nit: I think errors in go are supposed to be lowercase (I don't really mind anyway)

pkg/jx/cmd/install.go
```diff
@@ -665,251 +693,209 @@ func (options *InstallOptions) Run() error {
...
-	if username != "" && originalGitToken != "" && originalGitServer != "" {
-		err = gitAuthCfg.SaveUserAuth(originalGitServer, &auth.UserAuth{
-			ApiToken: originalGitToken,
-			Username: username,
-		})
+
+		err = util.DestroyFile(secretsFile)
 		if err != nil {
-			return err
+			return errors.Wrapf(err, "destrying the secrets file '%s' after storing it in Vault", secretsFile)
```
agentgonzo: typo:
```suggestion
			return errors.Wrapf(err, "destroying the secrets file '%s' after storing it in Vault", secretsFile)
```

pkg/jx/cmd/install.go
```diff
@@ -923,321 +909,815 @@ func (options *InstallOptions) Run() error {
...
+	if helmConfig.ExposeController.Config.Domain == "" {
+		helmConfig.ExposeController.Config.Domain = options.InitOptions.Flags.Domain
 	}
 
-	err = options.Helm().UpdateRepo()
-	if err != nil {
-		return errors.Wrap(err, "failed to update the helm repo")
+	domain := helmConfig.ExposeController.Config.Domain
+	if domain != "" && addon.IsAddonEnabled("gitea") {
+		helmConfig.Jenkins.Servers.GetOrCreateFirstGitea().Url = "http://gitea-gitea." + namespace + "." + domain
```
agentgonzo: why `gitea-gitea` and not just `gitea`?

pkg/jx/cmd/install.go
```diff
@@ -923,321 +909,815 @@ func (options *InstallOptions) Run() error {
...
+	dir, err := util.ConfigDir()
+	if err != nil {
+		return valuesFiles, secretsFiles, temporaryFiles,
+			errors.Wrap(err, "failed to create a temporary config dir for Git credentials")
+	}
+	gitSecretsFileName := filepath.Join(dir, GitSecretsFile)
+	err = configStore.Write(gitSecretsFileName, []byte(gitSecrets))
+	if err != nil {
+		return valuesFiles, secretsFiles, temporaryFiles,
+			errors.Wrap(err, "failed to write the git secrets in the secrets file")
```
agentgonzo: For this and subsequent errors, it is probably worth putting the name of the file in the error

pkg/jx/cmd/install.go
```diff
@@ -923,321 +909,815 @@ func (options *InstallOptions) Run() error {
...
+	}
+
+	adminSecretsFileName := filepath.Join(dir, AdminSecretsFile)
+	err = configStore.WriteObject(adminSecretsFileName, adminSecrets)
+	if err != nil {
+		return valuesFiles, secretsFiles, temporaryFiles,
+			errors.Wrap(err, "failed to write the admin config file")
```
agentgonzo: ditto

pkg/jx/cmd/install.go
```diff
@@ -923,321 +909,815 @@ func (options *InstallOptions) Run() error {
...
+	}
+
+	extraValuesFileName := filepath.Join(dir, ExtraValuesFile)
+	err = configStore.WriteObject(extraValuesFileName, helmConfig)
+	if err != nil {
+		return valuesFiles, secretsFiles, temporaryFiles,
+			errors.Wrap(err, "failed to write the helm config file")
```
agentgonzo: ditto

pkg/jx/cmd/install.go
```diff
@@ -923,321 +909,815 @@ func (options *InstallOptions) Run() error {
...
+	}
+	log.Infof("Generated helm values %s\n", util.ColorInfo(extraValuesFileName))
+
+	err = options.modifySecrets(helmConfig, adminSecrets, gitSecrets)
+	if err != nil {
+		return valuesFiles, temporaryFiles, secretsFiles, err
```
agentgonzo: wrap the error?

pkg/jx/cmd/install.go
ccojocar: yes

pkg/jx/cmd/install.go
ccojocar: This was original in the code in this place. I haven't tired to remove it. I think we should remove it and let the use to set the `Verbose` options? @jstrachan What do you think?

pkg/jx/cmd/install.go
ccojocar: We have both.

pkg/jx/cmd/install.go
ccojocar: I think, we configure helm 3 before running the addon. I don't what to do a deeper refactoring here. @jstrachan Do we still need helm3 for addons?

pkg/jx/cmd/install.go
ccojocar: We support both.

pkg/jx/cmd/install.go
ccojocar: This is the original code. I don't want to modify in the PR. It can be refactoedr later.

pkg/jx/cmd/install.go
jstrachan: a while back someone tried to make the commands not verbose; but when installing when things go wrong all kinds of useful error message from tools like git, helm, gcloud et al get gobbled up without verbose. 

So we should default verbose to true for `jx install / jx create cluster ...` for sure

pkg/jx/cmd/install.go
jstrachan: we still need helm3 - docs on [no tiller](https://jenkins-x.io/news/helm-without-tiller/) and [helm3](https://jenkins-x.io/architecture/helm3/)

when helm3 is sufficiently GA so most users are happy to switch we'll be able to make it the default and remove the tiller/helm2 stuff but thats a way off

PR #2424, opened 2018-12-05
fix: Repos might be nil

pmuir:
#### Submitter checklist

fixes #


pmuir:
/hold

Putting on hold for @mikecirioli to test out - @mikecirioli can you do /hold cancel when you have tested?

amuniz:
Maybe a test covering the change so it does not happen again?

PR #2437, opened 2018-12-06
fix: validating git at the start of the create terraform command

garethjevans:

fixes #


garethjevans:
Fixes #2437

PR #2444, opened 2018-12-07
fix: ensure git user + email is setup

jstrachan:
lets refactor and add a reusable function too

### DIFF
--- a/pkg/gits/helpers.go
+++ b/pkg/gits/helpers.go
@@ -1,0 +1,42 @@
+package gits
+
+import (
+	"github.com/pkg/errors"
+	"os"
+	"os/user"
+)
+
+// EnsureUserAndEmailSetup returns the user name and email for the gitter
+// lazily setting them if they are blank either from the environment variables
+// `GIT_AUTHOR_NAME` and `GIT_AUTHOR_EMAIL` or using default values
+func EnsureUserAndEmailSetup(gitter Gitter) (string, string, error) {
+	userName, _ := gitter.Username("")
+	userEmail, _ := gitter.Email("")
+	if userName == "" {
+		userName = os.Getenv("GIT_AUTHOR_NAME")
+		if userName == "" {
+			user, err := user.Current()
+			if err == nil && user != nil {
+				userName = user.Username
+			}
+		}
+		if userName == "" {
+			userName = "jenkins-x-bot"
+		}
+		err := gitter.SetUsername("", userName)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/jx/cmd/controller_build.go
```diff
@@ -101,6 +101,12 @@ func (o *ControllerBuildOptions) Run() error {
 	o.EnvironmentCache = kube.CreateEnvironmentCache(jxClient, ns)
 
 	if o.InitGitCredentials {
+		// lets validate we have git configured
+		_, _, err = gits.EnsureUserAndEmailSetup(o.Git())
+		if err != nil {
+		  return err
```
ccojocar: I think would be nice to wrap this error with a message.

pkg/jx/cmd/controller_team.go
```diff
@@ -92,28 +91,9 @@ func (o *ControllerTeamOptions) Run() error {
...
-		if userEmail == "" {
-			userEmail = "jenkins-x@googlegroups.com"
-		}
-		err = gitter.SetEmail("", userEmail)
-		if err != nil {
-			return errors.Wrapf(err, "Failed to set the git email to %s", userEmail)
-		}
+	_, _, err = gits.EnsureUserAndEmailSetup(co.Git())
+	if err != nil {
+	  return err
```
ccojocar: also here

pkg/jx/cmd/controller_build.go
jstrachan: all the error conditions are already wrapped inside the `EnsureUserAndEmailSetup()` function so didn't see any need to wrap again

pkg/jx/cmd/controller_team.go
jstrachan: ditto to reply in above conversation

pkg/gits/helpers.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"github.com/pkg/errors"
+	"os"
+	"os/user"
+)
+
+// EnsureUserAndEmailSetup returns the user name and email for the gitter
+// lazily setting them if they are blank either from the environment variables
+// `GIT_AUTHOR_NAME` and `GIT_AUTHOR_EMAIL` or using default values
+func EnsureUserAndEmailSetup(gitter Gitter) (string, string, error) {
```
agentgonzo: would it make more sense to make this a receiver function on `Gitter`?

PR #2456, opened 2018-12-07
test(uninstall): make uninstall command test state independent

ccojocar:
#### Submitter checklist

- [ x] Change is code complete and matches issue description.
- [ x] Change is covered by existing or new tests.

fixes #

### DIFF
--- a/pkg/jx/cmd/uninstall.go
+++ b/pkg/jx/cmd/uninstall.go
@@ -213,6 +213,9 @@ func (o *UninstallOptions) cleanupConfig() error {
 	if err != nil {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


pkg/jx/cmd/uninstall.go
```diff
@@ -213,6 +213,9 @@ func (o *UninstallOptions) cleanupConfig() error {
 	if err != nil {
 		return nil
 	}
+	if authConfigSvc == nil {
```
agentgonzo: you could just put this as 
```
if err != nil || authConfigSvc == nil {
```

PR #2458, opened 2018-12-07
fix: lets use the teams StorageLocation settings in `jx step collect`

jstrachan:


jstrachan:
@pmuir does the above all seem OK with you? its a little bit of a change to what you had before

jstrachan:
btw we have `gits.ParseGitURL(text)` to avoid using a regex

Issue #2451, opened 2018-12-07
`jx delete team cheese` trashes the admin namespace


jtnord:
### Summary 

when deleting a team via `jx delete team <teamname>` the command completes but it also trashes the team in the admin namespace.  (the admin namespace is still present but a lot of controllers are now gone),

### Steps to reproduce the behavior

```
jx install
jx create team foo
<wait>
jx delete team foo
```

### Jx version

The output of `jx version` is:

not sure as my cluster was trashed... it was create a day before this so probably platform was 3019

```
jx version
NAME               VERSION
jx                 [32m1.3.637[0m
jenkins x platform [32m0.0.3027[0m
Kubernetes cluster [32mv1.9.7-gke.11[0m
kubectl            [32mv1.10.3[0m
helm client        [32mv2.11.0+g2e55dbe[0m
helm server        [32mv2.11.0+g2e55dbe[0m
git                [32mgit version 2.16.2.windows.1[0m
Operating System   [32mWindows 10 Pro 1803 build 17134[0m
```

### Kubernetes cluster

What kind of Kubernetes cluster are you using & how did you create it?

GKE  `jx create terraform`

### Operating system / Environment

OS X / Windows / N/A

### Expected behavior

unrelated teams from that being deleted are unaffected

### Actual behavior

the admin namespace is affected.

PR #2469, opened 2018-12-10
bug(create team): Update status on the Team CRD in the admin namespace

markawm:
#### Submitter checklist

#### Description

The code tries to use the admin namespace as returned by kube.GetAdminNamespace(), but this appears to just return
the current namespace rather than finding the admin namespace - i.e. it can change!
Fixed by being explicit at all times about the admin NS to use. But I think there is another todo to sort out the current
'TODO' in kube.GetAdminNamespace():
`// TODO find the admin namespace via a label on the current dev namespace - or use current?`

fixes #

### DIFF
--- a/pkg/jx/cmd/delete_team.go
+++ b/pkg/jx/cmd/delete_team.go
@@ -161,7 +161,8 @@ func (o *DeleteTeamOptions) deleteTeam(name string) error {
 	}
 	o.changeNamespace(name)
 
-	err = o.ModifyTeam(name, func(team *v1.Team) error {
+	//TODO: will be wrong admin ns here.
+	err = o.ModifyTeam(ns, name, func(team *v1.Team) error {
 		team.Status.ProvisionStatus = v1.TeamProvisionStatusDeleting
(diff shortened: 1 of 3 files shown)


pkg/jx/cmd/delete_team.go
```diff
@@ -161,7 +161,8 @@ func (o *DeleteTeamOptions) deleteTeam(name string) error {
 	}
 	o.changeNamespace(name)
 
-	err = o.ModifyTeam(name, func(team *v1.Team) error {
+	//TODO: will be wrong admin ns here.
```
jtnord: ahh... #2451 ?

markawm: i'm just turning my attention to the delete one now - but yes, I suspect that is the cause of wrecking the admin namespace!

markawm:
Created task to follow-up with James S re. kube.GetAdminNamespace behaviour as a possible better fix for this.
Wasn't able to get the build logs for the first failure; tests pass locally, and /retest was clean.

Issue #2062, opened 2018-10-23
jx uses api.github.com and gets rate limited


frznlogic:
### Summary 
When setting up jx for the first time it will try to download helm using api to download latest helm etc. Sitting in a network together with 20 other people trying to do a jenkins-x introduction we ran into rate limiting. According to https://developer.github.com/v3/#rate-limiting you get 60 accesses per hour when not authenticated, and 5000 when authenticated. I've not been able to find how to authenticate yet (using jx). A documented and/or flag to use OAuth values should be nice perhaps, or perhaps some other way?

### Steps to reproduce the behavior
Run jx create cluster gke --prow approximately 60 times and it should happen. 

### Jx version

The output of `jx version` is:

09:58 osan@work9:~$ jx --version
1.3.464
### Kubernetes cluster

What kind of Kubernetes cluster are you using & how did you create it?

NA
### Operating system / Environment

Ubuntu 18.04
### Expected behavior
Getting the helm env installed I guess? :)

### Actual behavior
```

10:03 osan@work9:~$ jx create cluster gke --prow
? Missing required dependencies, deselect to avoid auto installing:  [Use arrows to move, type to fil? Missing required dependencies, deselect to avoid auto installing: helm
error: error installing helm: Unable to get latest version for github.com/kubernetes/helm GET https://api.github.com/repos/kubernetes/helm/releases/latest: 403 API rate limit exceeded for 95.141.107.254. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) [rate reset in 10m33s]

```

johnmcollier:
@frznlogic Can you try checking if `GH_TOKEN` token is set on your local machine? And if not, try setting it? JX should be configured to use it for any Github API calls (see https://github.com/jenkins-x/jx/blob/3b004378693bdf2f43be4b7c4fc987fcab356222/pkg/util/downloads.go#L76)

jstrachan:
Its a PITA you can't query the latest release of projects on github without hitting github rate limiting!

we could start automating keeping track of what version of dependencies like helm/kubectl/eksctl and the like on a Jenkins X website page maybe?

agentgonzo:
It's possible to do without using the API as https://github.com/helm/helm/releases/latest will issue a redirect to the tag - so we can do a simple get with that and grab the `Location` header

PR #2471, opened 2018-12-10
feat: use brew if available to install helm

garethjevans:

fixes #


garethjevans:
This won't fix #2062 but should help reduce the load against github.com

PR #2493, opened 2018-12-12
feat: add git provider user to User

pmuir:
#### Submitter checklist

fixes #


pmuir:
@ccojocar can we auto-populate this new field when the user logs in using Dex if they use GitHub to login?

ccojocar:
@pmuir This would be nice but unfortunately this information is not provided by dex which intermediates the authentication with GitHub. 

dex issues a new ID token which is independent of the downstream IdP, this token has the following claims:
 https://github.com/jenkins-x/dex/blob/432a5da2252c3f22b5f882b8dcc1c4ed8b1c0678/server/oauth2.go#L242.

pmuir:
That's a shame - oh well :-(

PR #2309, opened 2018-11-22
 Using PATCH instead of PUT to update PipelineActivity 

jtnord:
#### Submitter checklist

- [ ] Change is code complete and matches issue description.
- [ ] Change is covered by existing or new tests.

#### Description

subsumes #2110 where I resolved a merge conflict badly and do not have rights on deniedboardings repo to push the fix.  (appologies @msvticket)

This is to avoid http status 409 when other process has updated activity.
Also simplifying a bit by updating once in any case and not zero, one or two times depending on. The zero case didn't happen anyway as far as I could see.

Fixes #1890

#### Which issue this PR fixes

fixes #

PR #2110, opened 2018-10-29
 Using PATCH instead of PUT to update PipelineActivity 

msvticket:
This is to avoid http status 409 when other process has updated activity.
Also simplifying a bit by updating once in any case and not zero, one or two times depending on. The zero case didn't happen anyway as far as I could see.

Fixes #1890

PR #2494, opened 2018-12-12
Using PATCH instead of PUT to update PipelineActivity

### LINKED ISSUE #1890: Possible timing issue between jx promote and jx-resource jenkins plugin
### Summary 

Running a PR build for the jx project I got

```
 Created Pull Request: https://github.beescloud.com/****/environment-pr-1888-2-staging/pull/1
        
        Failed to update PipelineActivity: Operation cannot be fulfilled on pipelineactivities.jenkins.io "jenkins-x-tests-bdd-test-spring-1538663830-master-1": the object has been modified; please apply your changes to the latest version and try again
        error: Operation cannot be fulfilled on pipelineactivities.jenkins.io "jenkins-x-tests-bdd-test-spring-1538663830-master-1": the object has been modified; please apply your changes to the latest version and try again
```

Unknown

msvticket:
#### Submitter checklist

#### Description

This is to avoid http status 409 when other process has updated activity.
Also simplifying a bit by updating once in any case and not zero, one or two times depending on. The zero case didn't happen anyway as far as I could see.

This pull request was first made as #2110. Then James Nord created a new (#2309) one subsuming the first. But that pull request was basically abandoned. So now I try again...

A problem is that FakePipelineActivities.Patch doesn't support other than strategic merge patch which at least is the main reason that some tests fail. See k8s.io/client-go@v8.0.0+incompatible/testing/fixture.go:140
Later versions of k8s.io/client-go supports JSON Patch as well. I tried upgrading but had to give up when I found out that heptio/sonobuoy doesn't support later versions of k8s.io/client-go. It isn't obvious to me how to resolve this.

fixes #1890

### DIFF
--- a/pkg/kube/activity.go
+++ b/pkg/kube/activity.go
@@ -3,7 +3,7 @@ package kube
 import (
 	"encoding/json"
 	"fmt"
-	"reflect"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"strconv"
 	"strings"
 	"time"
@@ -182,24 +182,19 @@ func (k *PipelineActivityKey) GetOrCreate(activities typev1.PipelineActivityInte
 	}
 	a, err := activities.Get(name, metav1.GetOptions{})
 	if err != nil {
-		create = true
-		a = defaultActivity
+		if errors.IsNotFound(err) {
+			create = true
+			a = defaultActivity
+		} else {
+			return defaultActivity, create, err
+		}
 	}
-	oldSpec := a.Spec
-	spec := &defaultActivity.Spec
+	spec := &a.Spec
 	updateActivitySpec(k, spec)
 	if create {
 		answer, err := activities.Create(a)
 		return answer, true, err
 	} else {
-		if !reflect.DeepEqual(&a.Spec, &oldSpec) {
-			patch, err := json.Marshal(defaultActivity)
-			if err != nil {
-				return defaultActivity, create, nil
-			}
-			answer, err := activities.Patch(a.Name, types.MergePatchType, patch)
-			return answer, false, err
-		}
 		return a, false, nil
 	}
 }
@@ -413,6 +408,29 @@ func (k *PromoteStepActivityKey) GetOrCreatePromoteUpdate(activities typev1.Pipe
 	return a, s, p, p.Update, created, err
 }
 
+
+// PatchRow is a json patch structure
+type PatchRow struct {
+	Op    string `json:"op"`
+	Path  string `json:"path"`
+	Value interface{} `json:"value"`
+}
+
+// Almost making a full Update, but using Patch to work around problem with sometimes getting http status 409 back
+func patchSpec(activities typev1.PipelineActivityInterface, activity *v1.PipelineActivity) (err error) {
+	things := make([]PatchRow, 1)
+	things[0].Value = activity.Spec
+	things[0].Op = "add"
+	things[0].Path = "/spec"
+
+	patchBytes, err := json.Marshal(things)
+	if err != nil {
+		return
+	}
+	_, err = activities.Patch(activity.Name, types.JSONPatchType, patchBytes)
+	return
+}
+
 func (k *PromoteStepActivityKey) OnPromotePullRequest(activities typev1.PipelineActivityInterface, fn PromotePullRequestFn) error {
 	if !k.IsValid() {
 		return nil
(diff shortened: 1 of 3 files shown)


pkg/kube/activity.go
```diff
@@ -407,6 +410,29 @@ func (k *PromoteStepActivityKey) GetOrCreatePromoteUpdate(activities typev1.Pipe
...
+// Almost making a full Update, but using Patch to work around problem with sometimes getting http status 409 back
+func patchSpec(activities typev1.PipelineActivityInterface, activity *v1.PipelineActivity) (err error) {
+	things := make([]PatchRow, 1)
+	things[0].Value = activity.Spec
+	things[0].Op = "add"
+	things[0].Path = "/spec"
+
+	patchBytes, err := json.Marshal(things)
+	if err != nil {
+		return
```
ccojocar: I would be nice to wrap the error.

msvticket: OK, will do

pkg/kube/activity.go
```diff
@@ -407,6 +407,22 @@ func (k *PromoteStepActivityKey) GetOrCreatePromoteUpdate(activities typev1.Pipe
 	return a, s, p, p.Update, created, err
 }
 
+// Almost making a full Update, but using Patch to work around problem with sometimes getting http status 409 back
```
jglick: So this is not really fixing the race condition which the K8s API server helpfully alerted you to with the 409 error. It is just masking it, by clobbering any other changes to the `spec` that happened to come in from another client. To my mind that is worse. At least without this patch, you can sometimes recover _and_ behave correctly simply by retrying the entire operation.

The real fix would be to review every use of a `PUT` call, which seems to be wrapped in a generated `Update` method, and try to replace it with a `PATCH`. [Here](https://github.com/jenkins-x/jx/blob/d2b2e75f1764b76c1a769e223f6c0c3601744e9a/pkg/jx/cmd/step_collect.go#L248:L259) is a simple example.

The trouble is that the generated `Patch` methods, as opposed to `Update` methods, lack type safety: you need to [manually construct a JSON patch body](https://github.com/jenkins-x/jx/blob/d2b2e75f1764b76c1a769e223f6c0c3601744e9a/pkg/kube/serviceaccount/serviceaccounts.go#L15:L22) before [applying it](https://github.com/jenkins-x/jx/blob/d2b2e75f1764b76c1a769e223f6c0c3601744e9a/pkg/kube/serviceaccount/serviceaccounts.go#L38:L50), which is a lot of work and looks error-prone. In fact the example I linked to seems to be the only such case in this whole repo so far! Compare the manually defined JSON structure

```go
type JsonPatch struct {
	ImagePullSecret *[]ImagePullSecret `json:"imagePullSecrets"`
}
type ImagePullSecret struct {
	Name string `json:"name"`
}
```

to the definition of the struct field in the official API class:

```go
ImagePullSecrets []LocalObjectReference `json:"imagePullSecrets,omitempty" protobuf:"bytes,3,rep,name=imagePullSecrets"`
// …
type LocalObjectReference struct {
	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
}
```

jglick: I suppose a less cumbersome approach would be to make `Update` methods take an old and new object, to be called like this:

```diff
diff --git a/pkg/jx/cmd/step_collect.go b/pkg/jx/cmd/step_collect.go
index 9d7a90f5..96d020de 100644
--- a/pkg/jx/cmd/step_collect.go
+++ b/pkg/jx/cmd/step_collect.go
@@ -249,11 +249,12 @@ func (o *StepCollectOptions) collectGitURL(sourceURL string) (err error) {
 		if err != nil {
 			return err
 		}
+		oldA := *a
 		a.Spec.Attachments = append(a.Spec.Attachments, jenkinsv1.Attachment{
 			Name: classifier,
 			URLs: urls,
 		})
-		_, err = client.JenkinsV1().PipelineActivities(ns).Update(a)
+		_, err = client.JenkinsV1().PipelineActivities(ns).Update(&oldA, a)
 		if err != nil {
 			return err
 		}
```

The implementation could optimistically send a `PUT` request, assuming the API server can be relied on to check `resourceVersion`s, and if it gets a 409 then use reflection to compute the minimal diff between the two trees and generate a `PATCH` request.

Alternately, we could just avoid `PATCH` alttogether and retry the whole block that failed in a `PUT`:

```diff
diff --git a/pkg/jx/cmd/step_collect.go b/pkg/jx/cmd/step_collect.go
index 9d7a90f5..48b11c69 100644
--- a/pkg/jx/cmd/step_collect.go
+++ b/pkg/jx/cmd/step_collect.go
@@ -245,6 +245,7 @@ func (o *StepCollectOptions) collectGitURL(sourceURL string) (err error) {
 				Build:    build,
 			},
 		}
+		for {
 			a, _, err := key.GetOrCreate(activities)
 			if err != nil {
 				return err
@@ -254,10 +255,15 @@ func (o *StepCollectOptions) collectGitURL(sourceURL string) (err error) {
 				URLs: urls,
 			})
 			_, err = client.JenkinsV1().PipelineActivities(ns).Update(a)
-		if err != nil {
+			if err == nil {
+				break
+			} else if strings.Contains(err.Error(), "409") {
+				continue
+			} else {
 				return err
 			}
 		}
+	}
 	return nil
 }
```

though I assume there is some more type-safe way of accessing the error code. (Seems there is a `StatusError` and you can check for `StatusReasonConflict` or something.)

msvticket: I think it's much better to, as you say, clobber the CRD than fail. Especially since at times  promotions has failed repeatedly on 409.

I did start a fix that did minimal patches, but I realised that the fix would become unreasonably big and complex.

ccojocar:
@msvticket Are you planning to still work on this PR?  Should we close it for now and try to address the issue in another PR?

msvticket:
I was planning to work on it, but nobody has addressed the issues I raised in "Special notes for the reviewer(s)"

ccojocar:
@msvticket I am closing this PR. Please could you send the new solution in a different PR? Thank you

PR #2505, opened 2018-12-14
Added verbose flag to `eks get` command

hekonsek:
Added verbose flag to `eks get` command.


hekonsek:
I guess we have to re-run tests.

PR #2511, opened 2018-12-14
EKS cluster creation should deal with rolled back stacks

hekonsek:
If EKS cluster creation process is interrupted or if cluster with given name already exists, there will be CloudFormation stack in ROLLBACK_COMPLETE state left. Such dead stack prevents eksctl from creating cluster with the same name. This is common activity then to remove stacks like this and this function performs this action.

This PR detects such dead stacks and deletes them before creating the cluster with eksctl.

Also we should try to detect that cluster with given name exists before running eksctl, otherwise we end up with CloudFormation stack being initialized and failing after some time. If we know that cluster with given name already exists, we should just detect this fact and inform user about it as soon as possible. This PR implements this as well.

### DIFF
--- a/pkg/cloud/amazon/eks.go
+++ b/pkg/cloud/amazon/eks.go
@@ -1,0 +1,74 @@
+package amazon
+
+import (
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


hekonsek:
False negative? :) Maybe we should re-run tests.

PR #2520, opened 2018-12-15
feat: add a top level command to create new projects

jstrachan:


pkg/jx/cmd/create_project.go
```diff
@@ -0,0 +1,158 @@
...
+	"os"
+)
+
+const (
+	createQuickstartName = "Create new application from a Quickstart"
+	createCamelName      = "Create new Apache Camel microservice"
+	createSpringName     = "Create new Spring Boot microservice"
+	importDirName        = "Import existing code from a directory"
+	importGitName        = "Import code from a git repository"
+	importGitHubName     = "Import code from a github repository"
```
ccojocar: I wondering if these shouldn't be more generic than just GitHub.

pkg/jx/cmd/create_project.go
```diff
@@ -0,0 +1,158 @@
...
+			CommonOptions: CommonOptions{
+				Factory: f,
+				In:      in,
+				Out:     out,
+				Err:     errOut,
+			},
+		},
+	}
+
+	cmd := &cobra.Command{
```
ccojocar: We should maybe retie the other commands for simplicity and always instruct the user to use the `create project` command instead of `create quickstart` or `import`.

jstrachan: though I do like the specific commands too - it also allows much more powerful CLI options and completion

jstrachan: the top level `jx create project` acts as a handy wizard to get folks going until they learn the lower level richer commands I guess?

pkg/jx/cmd/create_project.go
jstrachan: yeah - we should be able to support importing from any git repository I think; we should probably fix that in `jx import` too

PR #2523, opened 2018-12-17
fix(Jenkinsfile): cleanup the old github credential from Jenkinsfile

ccojocar:
#### Submitter checklist

fixes #

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,6 +30,7 @@ pipeline {
             steps {
                 dir ('/home/jenkins/go/src/github.com/jenkins-x/jx') {
                     checkout scm
+                    sh "jx step git credentials"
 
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -30,12 +30,14 @@ pipeline {
             steps {
                 dir ('/home/jenkins/go/src/github.com/jenkins-x/jx') {
                     checkout scm
+                    sh "git config --global credential.helper store"
                     sh "jx step git credentials"
```
rawlingsj: I think we can remove these now that we use knative build creds init container.  Credentials get automatically mounted now so I would have thought we could remove all 
```
sh "git config --global credential.helper store"
sh "jx step git credentials"
```

ccojocar: alright

PR #2546, opened 2018-12-20
It should be possible to specify Git server type when creating new environment

hekonsek:
Right now when executing `jx create env` Git server kind is auto-detected from Git provider server URL. It works OK for cloud solutions like GitHub or BitBucketCloud, but doesn't work well for self-hosted BitBucket or GitLab.

It should be possible to enforce Git provider kind using command line option.


hekonsek:
Closing for a while - https://kubernetes.slack.com/messages/C9LTHT2BB/convo/C9LTHT2BB-1545296444.247900

PR #2597, opened 2019-01-04
fix(login) : login now works on systems without io-notify

jtnord:
#### Submitter checklist

#### Description

when tailing the netlog if the system did not support `ionotify` then we
would not see any new changes to the file.  witching ot polling on all
systems as the overhead of doing this is minimal.

fixes #2453


mikecirioli:
/approve

LGTM, and i validated that this works for me locally on mac os/x

PR #2605, opened 2019-01-07
fix(ingress): github webhooks weren't being updated on `jx upgrade ingress`

### LINKED ISSUE #2018: jx upgrade ingress should update the environment webhooks
when we run `jx upgrade ingress` we typically change from http <-> https or change the domain which can break old webhooks on the environment git repositories.

So lets update the environment repo webhooks to the new URLs

agentgonzo:
`jx upgrade ingress` wasn't correctly setting the webhooks in github.
It was incorrectly checking whether the webhook matched the _new_ URL (that you were trying to set it _to_)

fixes #2018

### DIFF
--- a/pkg/jx/cmd/update_webhooks.go
+++ b/pkg/jx/cmd/update_webhooks.go
@@ -154,14 +154,13 @@ func (options *UpdateWebhooksOptions) updateRepoHook(git gits.GitProvider, repoN
 					Repo: &gits.GitRepository{
 						Name: repoName,
 					},
-					URL: webhookURL,
+					URL:         webhookURL,
+					ExistingURL: options.PreviousHookUrl,
 				}
 
 				if isProwEnabled {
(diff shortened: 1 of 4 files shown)


pkg/jx/cmd/update_webhooks.go
```diff
@@ -154,15 +154,14 @@ func (options *UpdateWebhooksOptions) updateRepoHook(git gits.GitProvider, repoN
...
-					URL: webhookURL,
+					URL:         webhookURL,
+					ExistingURL: options.PreviousHookUrl,
 				}
 
 				if isProwEnabled {
 					webHookArgs.Secret = string(hmacToken.Data["hmac"])
 				}
 
-				log.Infof("Updating WebHook with new args\n")
```
agentgonzo: This lead to confusion when nothing actually happened (see @jtnord 's comments) and is better removed as we have sufficient logging in the `UpdateWebHook` method

PR #2610, opened 2019-01-07
Add --ssh-public-key flag for create cluster aws command to use custo…

StephenWangasg:
Add --ssh-public-key flag to "jx create cluster aws" command to use a pub key from a location rather the default ~/.ssh/id_rsa.pub

### DIFF
--- a/pkg/jx/cmd/create_cluster_aws.go
+++ b/pkg/jx/cmd/create_cluster_aws.go
@@ -44,6 +44,7 @@ type CreateClusterAWSFlags struct {
 	NodeSize               string
 	MasterSize             string
 	State                  string
+	SshPublicKey           string
 	Tags                   string
 }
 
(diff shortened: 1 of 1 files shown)


pkg/jx/cmd/import.go
```diff
@@ -488,14 +488,14 @@ func (options *ImportOptions) DraftCreate() error {
 		return err
 	}
 
-	provider, err := gits.CreateProvider(options.GitServer, options.GitUserAuth, options.Git())
```
aiqdevops: 1. options.GitProvider is already created before calling DraftCreate(). No need to create provider here.
2. In case you run import via 'jx import --org orgname  --github' and call the import in 2nd pass, options.GitUserAuth is nil, this ends up a crash here.

pkg/jx/cmd/import.go
```diff
@@ -488,14 +488,14 @@ func (options *ImportOptions) DraftCreate() error {
...
+	if options.GitUserAuth == nil {
+		userAuth := options.GitProvider.UserAuth()
+		options.GitUserAuth = &userAuth
 	}
-
+	
 	if options.Organisation == "" {
 		gitUsername := options.GitUserAuth.Username
-		options.Organisation, err = gits.GetOwner(options.BatchMode, provider, gitUsername, options.In, options.Out, options.Err)
+		options.Organisation, err = gits.GetOwner(options.BatchMode, options.GitProvider, gitUsername, options.In, options.Out, options.Err)
```
aiqdevops: Since options.GitProvider is for sure created before you call DraftCreate(), just use it here.

StephenWangasg:
Any updates on this?

rawlingsj:
sorry, we were having some infra issues, think that's all resolved now so fingers crossed this will go through now

pkg/jx/cmd/create_cluster_aws.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/jenkins-x/jx/pkg/log"
 	"github.com/jenkins-x/jx/pkg/util"
 	"github.com/spf13/cobra"
-	"gopkg.in/AlecAivazis/survey.v1"
+        "gopkg.in/AlecAivazis/survey.v1"
```
ccojocar: wrong alignment

StephenWangasg: Changed

PR #2629, opened 2019-01-09
fix(gc) lets clean up stale pull request activities too

rawlingsj:


pkg/jx/cmd/gc_activities.go
```diff
@@ -118,6 +121,17 @@ func (o *GCActivitiesOptions) Run() error {
 	activityBuilds := make(map[string][]int)
 
 	for _, a := range activities.Items {
+		// if the activity is a PR and has completed over a week ago lets GC it
+		if strings.Contains(a.Name, "-pr-") {
```
amuniz: Isn't there a less flaky way to identify a PR activity? Maybe a new attribute in the Spec would be useful (and not only for this change).

PR #2641, opened 2019-01-10
Users1

pmuir:
#### Submitter checklist

fixes #


pkg/users/git.go
```diff
@@ -0,0 +1,161 @@
...
+	GitProvider gits.GitProvider
+	JXClient    jenkninsv1client.Interface
+	Namespace   string
+}
+
+// GitUserAsUser will convert the GitUser to a Jenkins X user and attempt to complete the user info by:
+// * checking the user custom resources to see if the user is present there
+// * making a call to the gitProvider
+// as often user info is not complete in a git response
+func (r *GitUserResolver) GitUserAsUser(user *gits.GitUser) (*jenkinsv1.User, error) {
```
ccojocar: It would be nice to have a test for this method since has a lot of complex logic. You can use the fake client.

pkg/apis/jenkins.io/v1/types_users.go
```diff
@@ -35,15 +35,20 @@ type UserList struct {
...
-	SlackUser         string       `json:"slackUser,omitempty"  protobuf:"bytes,8,opt,name=slackUser"`
-	GitProviderUser   string       `json:"gitProviderUser,omitempty"  protobuf:"bytes,9,opt,name=gitProviderUser"`
+	Login             string             `json:"login,omitempty"  protobuf:"bytes,1,opt,name=login"`
+	Name              string             `json:"name,omitempty"  protobuf:"bytes,2,opt,name=name"`
+	Email             string             `json:"email,omitempty"  protobuf:"bytes,3,opt,name=email"`
+	CreationTimestamp *metav1.Time       `json:"creationTimestamp,omitempty" protobuf:"bytes,4,opt,name=creationTimestamp"`
+	URL               string             `json:"url,omitempty"  protobuf:"bytes,5,opt,name=url"`
+	AvatarURL         string             `json:"avatarUrl,omitempty"  protobuf:"bytes,6,opt,name=avatarUrl"`
+	ServiceAccount    string             `json:"serviceAccount,omitempty"  protobuf:"bytes,7,opt,name=serviceAccount"`
+	Accounts          []AccountReference `json:"accountReference,omitempty"  protobuf:"bytes,8,opt,name=accountReference"`
```
jtnord: @pmuir I am not up to speed on all the internals of CRDs - but won't this break existing clients that are using the User crd?

pmuir: (1) "this" being the removal of the `SlackUser` and `GitProviderUser` fields?
(2) And existing clients being clients other than the go client we generate as part of the `jx` project?

If so, (1) yes it would but AFAIK the only people using those fields are me, so it should be safe - it's really a case of tidiness vs safety, and I chose tidiness, perhaps I should deprecate if you know of uses? I only added `GitProviderUser` about 3 weeks ago, so I think it should be safe, and `SlackUser` was added for the slackbot which shouldn't - let me make sure I PR'd removing it.

(2) is a problem, but generally not something we have addressed yet.

jtnord: > (1) "this" being the removal of the SlackUser and GitProviderUser fields?

yes.
not so much directly using those fields, but users using the struct elsewhere at an older version?
e.g. `core-backend` retrieves users as does a local `jx` install (at an version older than whenever this gets merged).

> (2) And existing clients being clients other than the go client we generate as part of the jx project?

well not really as it could be an old jx client.  (see point 1).

basically I don't know of anyone using those fields - so it's not that those break per say - but more that will this break clients compiled against the old code (old jx versions, core-backend) if they see a change in the data structure.  (IIUC adding fields is ok, removing them is more of an issue?).  This is where my understanding gets fuzzy and I may just be completely wrong :)

> (2) is a problem, but generally not something we have addressed yet.

in which case given core-backend is not GA and upgrading a local `jx` install is pretty easy then that is ok, I guess it just needs to be release noted somewhere and we need to make sure core-backend is updated to use the newer version of the jx code.

pmuir: For (2) I'll update the commit message to use semantic commit format to make sure this goes into the changelog (got to remember how!)

PR #2642, opened 2019-01-10
chore: Add a CRD for storing application('project') metadata

agentgonzo:
Add a new CRD for storing metadata about applications (also known as projects?). Data in this CRD will be used to implement better deletion of projects/applications. This PR implements CRD creation and basic generated client code.

This is a tidyup (due to merge conflicts from master) of #2588


agentgonzo:
All comments on #2588 have been addressed with the exception of the naming of 'Application' vs 'App' vs 'project'... :-/

pkg/apis/jenkins.io/v1/types_extensions.go
```diff
@@ -438,6 +438,40 @@ type AppList struct {
...
+type ApplicationList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+
+	Items []Application `json:"items"`
+}
+
+// ApplicationSpec provides details of the metadata for an App
+type ApplicationSpec struct {
+	Name        string
```
jtnord: what is name here?  (or how would it be distinct from the actual resource name)? (or maybe I misunderstand this)

pkg/apis/jenkins.io/v1/types_extensions.go
```diff
@@ -438,6 +438,40 @@ type AppList struct {
...
+	Description string
+	Org         string
```
jtnord: do / should you also not have a git server (id whatever) here?  (I think you can have multiple - as `jx create git server` exists (presumably for some reason)?

pkg/apis/jenkins.io/v1/types_extensions.go
agentgonzo: Yeah, not entirely sure what I was thinking. I can't see how it would be different from `Repo` in my mind.

pkg/apis/jenkins.io/v1/types_extensions.go
agentgonzo: yeah, good point.

PR #2646, opened 2019-01-10
fix: add caching into the build number generator

jstrachan:
so we don't have to wait to List all PipelineActivity resources every time we need a build number; instead we only do it on startup of the build number generator

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/kube/pipeline_cache.go
+++ b/pkg/kube/pipeline_cache.go
@@ -1,0 +1,87 @@
+package kube
+
+import (
+	"github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	"github.com/jenkins-x/jx/pkg/client/clientset/versioned"
+	"github.com/jenkins-x/jx/pkg/log"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/client-go/tools/cache"
+	"sort"
+	"time"
+)
+
+// PipelineNamespaceCache caches the pipelines for a single namespace
+type PipelineNamespaceCache struct {
+	pipelines map[string]*v1.PipelineActivity
+}
+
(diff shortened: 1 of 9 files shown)


pkg/kube/pipeline_cache.go
```diff
@@ -0,0 +1,96 @@
...
+
+// PipelineNamespaceCache caches the pipelines for a single namespace
+type PipelineNamespaceCache struct {
+	pipelines map[string]*v1.PipelineActivity
+}
+
+// NewPipelineCache creates a cache of pipelines for a namespace
+func NewPipelineCache(jxClient versioned.Interface, ns string) *PipelineNamespaceCache {
+	pipeline := &v1.PipelineActivity{}
+	stop := make(chan struct{})
```
ccojocar: It seems that nobody is closing the `stop` channel. Maybe it should be returned to the caller which might  decide when to stop the controller/cache.

pkg/kube/pipeline_cache.go
```diff
@@ -0,0 +1,96 @@
...
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/client-go/tools/cache"
+	"sort"
+	"time"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// PipelineNamespaceCache caches the pipelines for a single namespace
+type PipelineNamespaceCache struct {
+	pipelines map[string]*v1.PipelineActivity
```
ccojocar: Maybe considering using a sync.Map https://golang.org/pkg/sync/#Map since the callbacks can be called concurrently to the `Pipelines` call.

jstrachan:
@ccojocar changes applied thanks!

PR #2651, opened 2019-01-11
[feat] Get Jenkins API token using REST calls rather than Chrome screen-scraping

jglick:
- [X] for legacy security realm with no crumb (current result of `jx install`)
- [ ] handle modern security realms
- [ ] handle crumbs

Makes

```sh
jx create jenkins token --browser -p '…' admin
```

(when `~/.jx/jenkinsAuth.yaml` does not yet define `.servers[0].users[0].apitoken`) much faster (10–20s → ~100ms running against a local server), and probably more reliable as well.

Note that the `--headless` option is no longer interpreted by this command.


pkg/jx/cmd/create_jenkins_token.go
```diff
@@ -208,7 +197,8 @@ func (o *CreateJenkinsUserOptions) Run() error {
 	return nil
 }
 
-func (o *CreateJenkinsUserOptions) tryFindAPITokenFromBrowser(tokenUrl string, newTokenUrl string, userAuth *auth.UserAuth) error {
+func (o *CreateJenkinsUserOptions) getAPITokenFromREST(serverURL string, userAuth *auth.UserAuth) error {
+	newTokenURL := jenkins.JenkinsNewTokenURL(serverURL)
 	var ctx context.Context
 	var cancel context.CancelFunc
 	if o.Timeout != "" {
```
ccojocar: You can create the context with a timeout.

```Go
ctx, cancel := context.WithTimeout(context.Background(), timeout)
defer cancel()
```
This will ensure that the context will be canceled automatically after the timeout expires.

ccojocar:
I am happy to see this happening. Probably we should also remove the chrome from the base image after this PR is merged and validated. It will reduce the image size quite a bit. 

https://github.com/jenkins-x/jenkins-x-serverless/blob/5da670d05cd603a4bb3f181084d336438e1f9e85/Dockerfile.base#L14

Chrome can be only added to the nodejs image for selenium/frontend tests.

pkg/jx/cmd/create_jenkins_token.go
jstrachan: just added that line

jglick:
> we should also remove the chrome from the base image

It is still used in `create_git_api_token`, but I suppose you mean that this code is not used from automated tests (only from user desktop systems) so it need not be in the serverless image?

The bot seems to have removed the label I put in place to remind people that this PR would introduce a regression for anyone running a server customized to use either a non-legacy security realm or CSRF defense. I will work on a follow-up PR restoring those functions.

pkg/jx/cmd/create_jenkins_token.go
```diff
@@ -207,6 +207,7 @@ func (o *CreateJenkinsUserOptions) getAPITokenFromREST(serverURL string, userAut
 			return errors.Wrap(err, "parsing the timeout value")
 		}
 		ctx, cancel = context.WithTimeout(context.Background(), duration)
+		defer cancel()
```
jglick: Not sure I understand why this was added. There was already a `defer cancel()` a few lines below.

(And the code being discussed was not touched in my PR. Did you think there was something wrong with it before?)

pkg/jx/cmd/create_jenkins_token.go
```diff
@@ -96,8 +90,8 @@ func NewCmdCreateJenkinsUser(f Factory, in terminal.FileReader, out terminal.Fil
 	options.ServerFlags.addGitServerFlags(cmd)
 	cmd.Flags().StringVarP(&options.ApiToken, "api-token", "t", "", "The API Token for the user")
 	cmd.Flags().StringVarP(&options.Password, "password", "p", "", "The User password to try automatically create a new API Token")
-	cmd.Flags().StringVarP(&options.Timeout, "timeout", "", "", "The timeout if using browser automation to generate the API token (by passing username and password)")
-	cmd.Flags().BoolVarP(&options.UseBrowser, "browser", "", false, "Use a Chrome browser to automatically find the API token if the user and password are known")
+	cmd.Flags().StringVarP(&options.Timeout, "timeout", "", "", "The timeout if using REST to generate the API token (by passing username and password)")
+	cmd.Flags().BoolVarP(&options.UseBrowser, "browser", "", false, "Use REST calls to automatically find the API token if the user and password are known")
```
jglick: Note that the naming of the option is now misleading, but I was not sure how to deprecate the old option nicely.

(I am not exactly sure why this option needs to be passed at all—I do not see a clear use case for a command that just waits for you to get a token by hand—nor I am clear on why `--password` needs to be specified when `~/.jx/jenkinsAuth.yaml` already stores the password. But these behaviors are orthogonal to my change.)

pkg/jx/cmd/create_jenkins_token.go
ccojocar: right, I haven't seen that. It can be removed.

PR #2668, opened 2019-01-14
fix(preview): when deleting the preview, also purge the helm release

vbehar:
#### Submitter checklist

#### Description

Instead of just deleting the Environment resource and the associated namespace,
deleting a preview should also delete (purge) the helm release:
- allows to run pre/post-delete helm hooks
- if using std helm - without the template or no-tiller flag - it avoids leaving broken releases behind

Our use-case is that we have charts using hooks to register and cleanup cloud resources. the registration part works well when the chart is deployed in a preview environment, but the delete hooks are never called, because right now the helm release is never deleted.

Calling helm in the gcpreviews jobs also has impacts on the RBAC setup, because if the team setting is to use a local tiller, then the job will need to have sufficient rights to create/delete resources.
To handle this, I added a new feature in the jx chart: setting a custom service account name. This allows to run a pod with an existing service account, without creating a new one.
I opened another PR in jenkins-x-platform to run the gcpreviews with the jenkins service account (which is cluster-admin): https://github.com/jenkins-x/jenkins-x-platform/pull/4652

### DIFF
--- a/pkg/jx/cmd/delete_preview.go
+++ b/pkg/jx/cmd/delete_preview.go
@@ -98,6 +98,24 @@ func (o *DeletePreviewOptions) Run() error {
 }
 
 func (o *DeletePreviewOptions) deletePreview(name string) error {
+	jxClient, ns, err := o.JXClient()
+	if err != nil {
+		return err
+	}
+
+	environment, err := kube.GetEnvironment(jxClient, ns, name)
+	if err != nil {
+		return err
+	}
+	releaseName := kube.GetPreviewEnvironmentReleaseName(environment)
+	if len(releaseName) > 0 {
+		log.Infof("Deleting helm release: %s\n", util.ColorInfo(releaseName))
+		err = o.Helm().DeleteRelease(ns, releaseName, true)
+		if err != nil {
+			return err
+		}
+	}
+
 	log.Infof("Deleting preview environment: %s\n", util.ColorInfo(name))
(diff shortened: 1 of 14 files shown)


pmuir:
@jstrachan probably one for you to review?

pmuir:
Or @markawm

pkg/jx/cmd/delete_preview.go
```diff
@@ -98,6 +98,25 @@ func (o *DeletePreviewOptions) Run() error {
...
+	if err != nil {
+		return err
+	}
+
+	environment, err := kube.GetEnvironment(jxClient, ns, name)
+	if err != nil {
+		return err
+	}
+	if environment != nil {
+		if releaseName := environment.Annotations[AnnotationReleaseName]; len(releaseName) > 0 {
```
ccojocar: Maybe move this check in a function in `pkg/kube/env.go`.  I think `func IsPreview(env *v1.Environment) bool` sounds like a good name.

vbehar: there was already an `IsPreviewEnvironment` func, and I needed to get back the release name, so I created a `GetPreviewEnvironmentReleaseName(env) string` func

vbehar:
thanks @ccojocar ! can you please also get https://github.com/jenkins-x/jenkins-x-platform/pull/4652 merged, to get the right permissions for the gcpreview job? thanks

PR #2669, opened 2019-01-14
fix: replace pr.Author with GitUserResolver

pmuir:
#### Submitter checklist

#### Description
* Replace direct uses of `pr.Author` with user 
resolution via `users.GitUserResolver`. 
* Consolidate user related functions in user package
* Remove `provider.UserInfo` lookups from `getPullRequest`
* Improve matching algorithm safety

fixes #


pkg/users/users.go
```diff
@@ -30,13 +32,17 @@ func GetUsers(jxClient versioned.Interface, ns string) (map[string]*v1.User, []s
 }
 
 // CreateUser creates a new default User
-func CreateUser(ns string, login string, name string, email string) *v1.User {
-	user := &v1.User{
+func CreateUser(ns string, login string, name string, email string) *jenkinsv1.User {
+	id := login
+	if email != "" {
+		id = kube.EmailToK8sID(email)
```
jtnord: @pmuir why was this done?  we need this to be the login id as that is mostly stable.  the email for a login is definitely not stable for github auth.  Whilst the github auth is currently using whacky identifiers it is on my list of things todo to change to using the github username (e.g. jtnord for me).
It is also likely to collide if you want to allow both native accounts (gcloud) and github - as I both cases these can very likely be the same.

jtnord: also I think this does not work for what you are trying to do except in the narrow case where the user uses the same email in commits as they do in the auth system (which for github is oft not the case to reduce spam)

pmuir: I was trying to NOT change existing names. I think previously there were two approaches:
* https://github.com/pmuir/jx/blob/2c8ddc28e4ddb95fb8bc15c9764ed0f0418aa64b/pkg/kube/user_details.go#L42
* https://github.com/pmuir/jx/blob/2c8ddc28e4ddb95fb8bc15c9764ed0f0418aa64b/pkg/kube/users.go#L33

And I think I picked the wrong one to go with. I think the best thing to do is change this to be based on the id always?

jtnord: > I think the best thing to do is change this to be based on the id always

yes - I will get a PR filed for this.  
We also (whilst digging) found [FindByEmail](https://github.com/jenkins-x/jx/blob/24824dba093b5d16b323dd4932fa27b0dcab6ac8/pkg/users/user_details.go#L27-L37) which seems to have no usage and would not have worked for users created with the old CreateUser.  Not sure if I should fix that or revert it.

pmuir: If no one is using it, I'll remove it.

Thanks for the PR!

jtnord: https://github.com/jenkins-x/jx/pull/2740

PR #2670, opened 2019-01-14
fix(buildnum): Services fails on repos with underscores

agentgonzo:
Trying to build https://github.com/jenkins-x/oauth2_proxy, it fails

eg:
```
time="2019-01-14T14:38:19Z" level=error msg="Unable to get next build number for pipeline jenkins-x/oauth2_proxy/master" error="PipelineActivity.jenkins.io \"jenkins-x-oauth2_proxy-master-1\" is invalid: metadata.name: Invalid value: \"jenkins-x-oauth2_proxy-master-1\": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')"
```

### DIFF
--- a/pkg/kube/activity.go
+++ b/pkg/kube/activity.go
@@ -63,12 +63,14 @@ type PipelineID struct {
 // The string identifier is expected to follow the format `<owner>/>repository>/<branch>`, though this isn't actually
 // validated/mandated here.
 func NewPipelineIDFromString(id string) PipelineID {
+	sanitisedName := strings.Replace(strings.ToLower(id), "/", "-", -1)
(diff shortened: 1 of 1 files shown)


markawm:
As discussed - fixes immediate problem of '_', but plenty of other repos names could break things. I'd suggest a displayName label that can be used by commands, UIs, etc. and then we have more freedom w.r.t the Name field which is quite restricted, and can generate a value or come up with a proper encoding scheme.

PR #2692, opened 2019-01-15
feat(teams) store the admin namespace in the team

### LINKED ISSUE #2662: add the ability to locate the 'admin' namespace
### Summary 

From [https://kubernetes.slack.com/archives/C9LTHT2BB/p1547215177297500](slack)
Certain parts of `jx` need to know where the admin namespace is.  currently there is no easy way to find this.
It would be great if there was an easy way to find this (e.g. a label) so when performing actions like `jx create team` `jx create user` the CRDs will end up in the correct namespace.

Currently there is nothing to prevent a user performing a full install into distinct namespaces, as part of this the second install could be rejected and the user directed to create a team instead.  Or the user could be allowed to do this - in which case finding the admin namespace would likely resort to the first admin namespace found (which would likely cause confusion, and not be as good.) or more work would be to introduce a new parameter `--adminnamespace` for the case of multiple installs.

N/A

### Jx version

jx 1.3.702

### Kubernetes cluster

What kind of Kubernetes cluster are you using & how did you create it?

N/a

N/A

### Expected behavior

code can easily identify the admin namespace.

### Actual behavior

code has to jump through some hoops to guess the admin namespace.

jtnord:
#### Submitter checklist

#### Description

Various commands need to know where the admin namespace is, this change
now records the admin namespace for a team as an annotation on the team

It will also update any existing teams with the annotation when the team
controller is run.

fixes #2662

### DIFF
--- a/pkg/jx/cmd/controller_team.go
+++ b/pkg/jx/cmd/controller_team.go
@@ -306,6 +306,11 @@ func (o *ControllerTeamOptions) onTeamChange(obj interface{}, kubeClient kuberne
 			}
 		}
 
+		err = kube.SetAdminNamespace(kubeClient, teamNs, adminNs)
+		if err != nil {
+			log.Errorf("Unable set admin namespace on team %s to %s - %s", util.ColorInfo(teamNs), util.ColorInfo(adminNs), err)
+			return
+		}
 		err = oc.ModifyTeam(adminNs, team.Name, func(team *v1.Team) error {
 			team.Status.ProvisionStatus = v1.TeamProvisionStatusComplete
 			team.Status.Message = "Installation complete"
@@ -315,7 +320,14 @@ func (o *ControllerTeamOptions) onTeamChange(obj interface{}, kubeClient kuberne
 			log.Errorf("Unable to update team %s to %s - %s", util.ColorInfo(teamNs), v1.TeamProvisionStatusComplete, err)
 			return
 		}
+	} else if v1.TeamProvisionStatusComplete == team.Status.ProvisionStatus {
+		err := kube.SetAdminNamespace(kubeClient, teamNs, adminNs)
+		if err != nil {
+			log.Errorf("Unable set admin namespace on team %s to %s - %s", util.ColorInfo(teamNs), util.ColorInfo(adminNs), err)
+			return
+		}
 	}
+
 }
 
 // LoadProwOAuthConfig returns the OAuth Token for Prow
--- a/pkg/kube/teams.go
+++ b/pkg/kube/teams.go
@@ -11,7 +11,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-const admin_namespace_annotation = "jenkins-x.io/admin-namespace"
+const adminNamespaceAnnotation = "jenkins-x.io/admin-namespace"
 
 // GetAdminNamespace tries to find the admin namespace that corresponds to this team.
 // in other words this is the namespace where the team CRD was initially created when this team was created,
@@ -21,7 +21,7 @@ func GetAdminNamespace(kubeClient kubernetes.Interface, teamNs string) (string, 
 	if err != nil {
 		return "", errors.Wrapf(err, "Failed to obtain the team namespace (%s) when getting the admin namespace", teamNs)
 	}
-	adminNs := namespace.Annotations[admin_namespace_annotation]
+	adminNs := namespace.Annotations[adminNamespaceAnnotation]
 	if adminNs != "" {
 		return adminNs, nil
 	}
@@ -35,12 +35,12 @@ func SetAdminNamespace(kubeClient kubernetes.Interface, teamNs string , adminNs 
 	if err != nil {
 		return errors.Wrapf(err, "Failed to update the obtain the namespace (%s) when updating the admin namespace", teamNs)
 	}
-	oldAdminNs := namespace.Annotations[admin_namespace_annotation]
+	oldAdminNs := namespace.Annotations[adminNamespaceAnnotation]
 	if oldAdminNs == adminNs {
 		// nothing to do
 		return nil
 	}
-	namespace.Annotations[admin_namespace_annotation] = adminNs
+	namespace.Annotations[adminNamespaceAnnotation] = adminNs
 	// TODO use patch
 	_, err = kubeClient.CoreV1().Namespaces().Update(namespace)
 	return err
--- a/pkg/kube/teams.go
+++ b/pkg/kube/teams.go
@@ -19,7 +19,7 @@ const adminNamespaceAnnotation = "jenkins-x.io/admin-namespace"
 func GetAdminNamespace(kubeClient kubernetes.Interface, teamNs string) (string, error) {
 	namespace, err := kubeClient.CoreV1().Namespaces().Get(teamNs, metav1.GetOptions{})
 	if err != nil {
-		return "", errors.Wrapf(err, "Failed to obtain the team namespace (%s) when getting the admin namespace", teamNs)
+		return "", errors.Wrapf(err, "obtaining the team namespace (%s) when getting the admin namespace", teamNs)
 	}
 	adminNs := namespace.Annotations[adminNamespaceAnnotation]
 	if adminNs != "" {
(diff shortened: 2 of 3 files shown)


pkg/jx/cmd/controller_team.go
```diff
@@ -306,6 +306,11 @@ func (o *ControllerTeamOptions) onTeamChange(obj interface{}, kubeClient kuberne
 			}
 		}
 
+		err = kube.SetAdminNamespace(kubeClient, teamNs, adminNs)
+		if err != nil {
+			log.Errorf("Unable set admin namespace on team %s to %s - %s", util.ColorInfo(teamNs), util.ColorInfo(adminNs), err)
```
agentgonzo: error printing should be `%v`

pkg/jx/cmd/controller_team.go
```diff
@@ -315,7 +320,14 @@ func (o *ControllerTeamOptions) onTeamChange(obj interface{}, kubeClient kuberne
 			log.Errorf("Unable to update team %s to %s - %s", util.ColorInfo(teamNs), v1.TeamProvisionStatusComplete, err)
 			return
 		}
+	} else if v1.TeamProvisionStatusComplete == team.Status.ProvisionStatus {
+		err := kube.SetAdminNamespace(kubeClient, teamNs, adminNs)
+		if err != nil {
+			log.Errorf("Unable set admin namespace on team %s to %s - %s", util.ColorInfo(teamNs), util.ColorInfo(adminNs), err)
```
agentgonzo: `%v` for errors

pkg/kube/teams.go
```diff
@@ -10,11 +11,39 @@ import (
...
-	return ns, nil
+const adminNamespaceAnnotation = "jenkins-x.io/admin-namespace"
+
+// GetAdminNamespace tries to find the admin namespace that corresponds to this team.
+// in other words this is the namespace where the team CRD was initially created when this team was created,
+// or the current team namespace for the case where this team was just created with a standalone `jx install`
+func GetAdminNamespace(kubeClient kubernetes.Interface, teamNs string) (string, error) {
+	namespace, err := kubeClient.CoreV1().Namespaces().Get(teamNs, metav1.GetOptions{})
+	if err != nil {
+		return "", errors.Wrapf(err, "Failed to obtain the team namespace (%s) when getting the admin namespace", teamNs)
```
agentgonzo: go style for wrapping errors is to omit the "Failed to ..." prefix for errors, and just put the reason - eg: "obtaining the team...". Run something and force an error and you'll see it in the output

pkg/kube/teams.go
```diff
@@ -10,11 +11,39 @@ import (
...
+	}
+	return teamNs, nil
+}
+
+// SetAdminNamespace annotates the given namespace with a backlink to the admin namespace.
+// it does not make any changes if the current annotation points to the same admin namespace
+func SetAdminNamespace(kubeClient kubernetes.Interface, teamNs string , adminNs string) error {
+	namespace, err := kubeClient.CoreV1().Namespaces().Get(teamNs, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "Failed to update the obtain the namespace (%s) when updating the admin namespace", teamNs)
```
agentgonzo: same here

pkg/jx/cmd/controller_team.go
agentgonzo: looks like they are equivalent for errors

pkg/kube/teams.go
agentgonzo: ditto

pkg/jx/cmd/controller_team.go
jtnord: it appears to make no difference and I was copying the rest of the code.

```
	log.Errorf("this is a test error with percent s  : %s", err)
	log.Errorf("this is a test error with percent v  : %v", err)
	log.Errorf("this is a wrapped error with percent s  : %s", errors.Wrap(err, "a wrapped message"))
	log.Errorf("this is a wrapped error with percent v  : %v", errors.Wrap(err, "a wrapped message"))
```
produces
```
this is a test error with percent s  : Failed to obtain the team namespace (myteamBBBAD) when getting the admin namespace: namespaces "myteamBBBAD" not found
this is a test error with percent v  : Failed to obtain the team namespace (myteamBBBAD) when getting the admin namespace: namespaces "myteamBBBAD" not found
this is a wrapped error with percent s  : a wrapped message: Failed to obtain the team namespace (myteamBBBAD) when getting the admin namespace: namespaces "myteamBBBAD" not found
this is a wrapped error with percent v  : a wrapped message: Failed to obtain the team namespace (myteamBBBAD) when getting the admin namespace: namespaces "myteamBBBAD" not found
```

pmuir:
/hold
Testing PRs atm, please see #jenkins-x-dev

pmuir:
Was a infra error

pkg/kube/teams.go
```diff
@@ -10,11 +11,39 @@ import (
...
+
+// GetAdminNamespace tries to find the admin namespace that corresponds to this team.
+// in other words this is the namespace where the team CRD was initially created when this team was created,
+// or the current team namespace for the case where this team was just created with a standalone `jx install`
+func GetAdminNamespace(kubeClient kubernetes.Interface, teamNs string) (string, error) {
+	namespace, err := kubeClient.CoreV1().Namespaces().Get(teamNs, metav1.GetOptions{})
+	if err != nil {
+		return "", errors.Wrapf(err, "obtaining the team namespace (%s) when getting the admin namespace", teamNs)
+	}
+	adminNs := namespace.Annotations[adminNamespaceAnnotation]
```
ccojocar: nit: 
```Go
adminNs, ok := ...
if ok && adminNs != "" {
...
}
```

pkg/kube/teams.go
```diff
@@ -10,11 +11,39 @@ import (
...
+}
+
+// SetAdminNamespace annotates the given namespace with a backlink to the admin namespace.
+// it does not make any changes if the current annotation points to the same admin namespace
+func SetAdminNamespace(kubeClient kubernetes.Interface, teamNs string , adminNs string) error {
+	namespace, err := kubeClient.CoreV1().Namespaces().Get(teamNs, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "obtaining the namespace (%s) when updating the admin namespace", teamNs)
+	}
+	oldAdminNs := namespace.Annotations[adminNamespaceAnnotation]
```
ccojocar: ditto

pkg/kube/teams_test.go
```diff
@@ -0,0 +1,108 @@
...
+	err = SetAdminNamespace(client, "myteam", "my-admin-namespace")
+	assert.NoError(t, err, "error occurred calling SetAdminNamespace")
+	adminNs, err := GetAdminNamespace(client, "myteam")
+	assert.NoError(t, err, "error occurred calling GetAdminNamespace")
+	assert.Equal(t, "my-admin-namespace" ,adminNs)
+}
+
+func createMockClient(objDef string) (kubernetes.Interface, error) {
+	decode := scheme.Codecs.UniversalDeserializer().Decode
+	obj, _, err := decode([]byte(objDef), nil, nil)
```
ccojocar: Why not to construct directly the namespace object instead of mashaling YAML? If someone, modifies the YAML content by mistake the tests will fail.

pkg/kube/teams_test.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not decode data")
+	}
+
+	mockKubeClient := kube_mocks.NewSimpleClientset(obj)
+	//ki, bool := mockKubeClient.(kubernetes.Interface)
```
ccojocar: remove this line

pkg/kube/teams_test.go
jtnord: in constructing the object I can make the same mistake twice (as I am querying the object later), so doing it this way means I can catch those mistakes.
As for if someone modified the yaml and the tests fail then that is no different to them changing an object and causing the test to fail - the yaml is part of the contract of the test so I do not understand that comment.

pkg/kube/teams_test.go
jtnord: the PR is is now in the tide mergepool so I will do that in a follow up.

pkg/kube/teams_test.go
ccojocar: What about upgrading the kuberntes client library, which might introduce an incompatibility in the schema which is less likely but yeah. IMHO is a bit fragile.

pkg/kube/teams_test.go
jtnord: tide lied and a normal build somehow kicked off for an unknown reason so fixed here.

pkg/kube/teams_test.go
jtnord: k8s can't just change the schema it will break all the things,  everytones deployment.yaml etc etc the world will fall down (and if they do change it it would be a new version) otherwise it will break and everything.
the internal object representaion is more likely to change than the interchange format  - so I think to say this is more fragile is actually incorrect - it is less fragile.

pkg/kube/teams.go
jtnord: given I need to treat an empty string the same as not present for this case I don;t think it is needed.

pkg/kube/teams.go
jtnord: ditto :)

PR #2716, opened 2019-01-16
chore: implement lazy creation of sourcerepository objects

mikecirioli:
#### Submitter checklist

#### Description
`jx platform` installs that don't have a `sourcerepositories` crd will not have all their projects available after an upgrade.  This PR addresses that by implementing a "lazy" update on pipeline activity which will create the missing `sourcerepository` entry if it doesn't exist.

This required refactoring pipeline activity methods to take a `jxClient` (versioned.Interface) instead of the more specialized `PipelineActivityInterface` they previously used.

- [x] Fix broken test 

fixes #


pkg/kube/activity.go
```diff
@@ -161,8 +163,23 @@ func GenerateBuildNumber(activities typev1.PipelineActivityInterface, pipelines
 	return build, answer, nil
 }
 
+func lazyUpdateSourceRepositoryCRD(jxClient versioned.Interface, ns string, activity *v1.PipelineActivity) error {
+	//FIXME: this relies on a bad assumption that the name of the sourcerepositories CRD entry is always equal to the repository name
+	_, err := sourcerepository.NewSourceRepositoryService(jxClient, ns).GetSourceRepository(activity.RepositoryName())
```
ccojocar: You can construct the object once and store it a variable, instead of call it twice here and in the if. e.g:

```
srs := sourcerepository.NewSourceRepositoryService(jxClient, ns)
```

pkg/kube/activity.go
```diff
@@ -161,8 +163,23 @@ func GenerateBuildNumber(activities typev1.PipelineActivityInterface, pipelines
...
+
+	if err != nil {
+		// sourcerepository for this project does not exist yet, so create one
+		err := sourcerepository.NewSourceRepositoryService(jxClient, ns).CreateSourceRepository(
+			activity.RepositoryName(), activity.Spec.GitOwner,activity.Spec.GitURL)
+		if err != nil {
```
ccojocar: Maybe wrap the error instead of logging here, and let the caller to log the error.

pkg/kube/activity.go
```diff
@@ -415,15 +446,16 @@ func (k *PromoteStepActivityKey) GetOrCreatePromoteUpdate(activities typev1.Pipe
 	return a, s, p, p.Update, created, err
 }
 
-func (k *PromoteStepActivityKey) OnPromotePullRequest(activities typev1.PipelineActivityInterface, fn PromotePullRequestFn) error {
+func (k *PromoteStepActivityKey) OnPromotePullRequest(jxClient versioned.Interface, ns string, fn PromotePullRequestFn) error {
 	if !k.IsValid() {
 		return nil
```
ccojocar: Isn't this an error case? Why not error is returned?

pkg/kube/activity.go
```diff
@@ -161,8 +163,23 @@ func GenerateBuildNumber(activities typev1.PipelineActivityInterface, pipelines
 	return build, answer, nil
 }
 
+func lazyUpdateSourceRepositoryCRD(jxClient versioned.Interface, ns string, activity *v1.PipelineActivity) error {
```
agentgonzo: the 'laziness' is the way that this method is called, rather than anything to do with the method itself. I suggest dropping 'lazy' from the function name

pkg/kube/activity.go
```diff
@@ -161,8 +163,23 @@ func GenerateBuildNumber(activities typev1.PipelineActivityInterface, pipelines
...
+	//FIXME: this relies on a bad assumption that the name of the sourcerepositories CRD entry is always equal to the repository name
+	_, err := sourcerepository.NewSourceRepositoryService(jxClient, ns).GetSourceRepository(activity.RepositoryName())
+
+	if err != nil {
+		// sourcerepository for this project does not exist yet, so create one
+		err := sourcerepository.NewSourceRepositoryService(jxClient, ns).CreateSourceRepository(
+			activity.RepositoryName(), activity.Spec.GitOwner,activity.Spec.GitURL)
```
agentgonzo: Nit: Space after the comment in parameter lists
```suggestion
			activity.RepositoryName(), activity.Spec.GitOwner, activity.Spec.GitURL)
```

pkg/kube/activity.go
```diff
@@ -171,23 +188,36 @@ func (k *PipelineActivityKey) GetOrCreate(activities typev1.PipelineActivityInte
...
+	a, err := activitiesClient.Get(name, metav1.GetOptions{})
 	if err != nil {
 		create = true
 		a = defaultActivity
 	}
 	oldSpec := a.Spec
+	oldLabels := a.Labels
+
+	// lazy create missing sourcerepository entries
+	if a.Labels[v1.LabelSourceRepository] == "" {
```
agentgonzo: 🐛 If there are no labels applied, then `a.Labels` will be nil, meaning that this will throw a nil reference exception (I'm about 99% sure of this, but _may_ be wrong)

pkg/kube/activity.go
```diff
@@ -171,23 +188,36 @@ func (k *PipelineActivityKey) GetOrCreate(activities typev1.PipelineActivityInte
...
-		answer, err := activities.Create(a)
+		answer, err := activitiesClient.Create(a)
 		return answer, true, err
 	} else {
 		if !reflect.DeepEqual(&a.Spec, &oldSpec) {
-			answer, err := activities.Update(a)
+			answer, err := activitiesClient.Update(a)
+			return answer, false, err
+		}
+		if !reflect.DeepEqual(&a.Labels, &oldLabels) {
```
agentgonzo: The contents of these if blocks are the same, so to my eyes it reads easier if you have `!reflect.DeepEqual(&a.Spec, &oldSpec) || !reflect.DeepEqual(&a.Labels, &oldLabels)` and a single conditional

PR #2726, opened 2019-01-17
`delete aws` command should wait for gateway address unmapping

hekonsek:
Right now `delete aws` command sometimes fails on detaching gateway, before unmapping public address from that gateway takes too long to complete.

After you attempt to detach internet gateway, AWS might need some time to unmap public address from that gateway. We should give AWS some time to complete this operation and retry detach operation several time before giving up and throwing an error.


pkg/jx/cmd/delete_aws.go
```diff
@@ -95,13 +97,27 @@ func (o *DeleteAwsOptions) Run() error {
 	}
 	for _, internetGateway := range internetGateways.InternetGateways {
 		if len(internetGateway.Attachments) > 0 {
-			_, err = svc.DetachInternetGateway(&ec2.DetachInternetGatewayInput{InternetGatewayId: internetGateway.InternetGatewayId, VpcId: aws.String(vpcid)})
-			if err != nil {
-				return err
+			detachAttemptsLeft := 11
```
ccojocar: why `11` :)?

ccojocar: I would recommend to use this retry mechanism https://github.com/jenkins-x/jx/blob/0ccc1304ed636279ea5b3860ee6a271c8ca79315/pkg/jx/cmd/common.go#L540

pkg/jx/cmd/delete_aws.go
```diff
@@ -95,13 +97,27 @@ func (o *DeleteAwsOptions) Run() error {
...
+			for ; detachAttemptsLeft > 0; {
+				_, err = svc.DetachInternetGateway(&ec2.DetachInternetGatewayInput{InternetGatewayId: internetGateway.InternetGatewayId, VpcId: aws.String(vpcid)})
+				fmt.Printf("Detaching internet gateway %s from VPC %s...\n", *internetGateway.InternetGatewayId, vpcid)
```
ccojocar: change `fmt` to `log`

pkg/jx/cmd/delete_aws.go
```diff
@@ -95,13 +97,27 @@ func (o *DeleteAwsOptions) Run() error {
...
+				if err != nil {
+					if strings.Contains(err.Error(), "Please unmap those public address(es) before detaching the gateway") {
+						detachAttemptsLeft--
+						fmt.Printf("Waiting for public address to be unmapped from internet gateway. Detach attempts left: %d\n", detachAttemptsLeft)
```
ccojocar: ditto

pkg/jx/cmd/delete_aws.go
```diff
@@ -95,13 +97,27 @@ func (o *DeleteAwsOptions) Run() error {
...
+					}
+				} else {
+					detachAttemptsLeft = 0
+				}
 			}
+			fmt.Printf("Internet gateway %s detached successfully from VPC %s...\n", *internetGateway.InternetGatewayId, vpcid)
 		}
 
 		_, err = svc.DeleteInternetGateway(&ec2.DeleteInternetGatewayInput{InternetGatewayId: internetGateway.InternetGatewayId})
+		fmt.Printf("Deleting internet gateway %s...\n", *internetGateway.InternetGatewayId)
```
ccojocar: ditto

pkg/jx/cmd/delete_aws.go
```diff
@@ -95,13 +97,27 @@ func (o *DeleteAwsOptions) Run() error {
...
+						fmt.Printf("Waiting for public address to be unmapped from internet gateway. Detach attempts left: %d\n", detachAttemptsLeft)
+						time.Sleep(10 * time.Second)
+					} else {
+						return err
+					}
+				} else {
+					detachAttemptsLeft = 0
+				}
 			}
+			fmt.Printf("Internet gateway %s detached successfully from VPC %s...\n", *internetGateway.InternetGatewayId, vpcid)
```
ccojocar: ditto

rawlingsj:
[APPROVALNOTIFIER] This PR is **NOT APPROVED**

This pull-request has been approved by:
To fully approve this pull request, please assign additional approvers.
We suggest the following additional approver: **jstrachan**

If they are not already assigned, you can assign the PR to them by writing `/assign @jstrachan` in a comment when ready.

The full list of commands accepted by this bot can be found [here](https://go.k8s.io/bot-commands).

The pull request process is described [here](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process)

<details open>
Needs approval from an approver in each of these files:

- **[OWNERS](https://github.com/jenkins-x/jx/blob/master/OWNERS)**

Approvers can indicate their approval by writing `/approve` in a comment
Approvers can cancel approval by writing `/approve cancel` in a comment
</details>

pkg/jx/cmd/delete_aws.go
hekonsek: Done.

pkg/jx/cmd/delete_aws.go
hekonsek: Good point. I've converted "magic number" to constant :) . And it is 10 attempts, because it is reasonable number from my tests against dead VPC instances :) .

hekonsek: When it comes to `retry()` it doesn't really handle my case. The thing is that if AWS return any error other than unmapping error, I'd like command to fail immediately. I want to retry only for certain type of error and fail immediately for all the others.

If we could consider extending `retry(..., call func() error) (err error)` signature to be more like `retry(..., call func() (err error, fatalError error)) (err error, fatalError error)` then it will cover my use case, because it would be possible to distinguish between recoverable callback error which should be retried and fatal error which should make retry to stop immediately and return error.

I propose to keep my retry logic as it is right now and then maybe create another PR in which we could extend `retury` to support fatal errors?

ccojocar: I got your point. You can define a new retry function like this:

```Go
type FataError struct {
     E error
}

func (err *FataError) Error() string {
     return fmt.Sprintf("fatal error: %s", err.E.Error())
}

func (o *CommonOptions) retryUntilFatalError(attempts int, sleep time.Duration, call func() error) (err error) {
	for i := 0; ; i++ {
		err = call()
		if err == nil {
			return
		}
                
                ok, err := err.(FatalError)
                if ok {
                      return err
                }

		if i >= (attempts - 1) {
			break
		}

		time.Sleep(sleep)

		log.Infof("retrying after error:%s\n", err)
	}
	return fmt.Errorf("after %d attempts, last error: %s", attempts, err)
}
```

And then re-use it for you case. 

I am pretty sure that we will have this scenario in future.

hekonsek: I'm just wondering, because we have only 12 places in jx codebase where `retry` is used, maybe we could just refactor existing function? That new version of function can cover all the existing use cases as it is superset of the old retry function.

Maybe a plan to go would be to create `retryUntilFatalError` function now and use it in my case, but also mark `retry` as deprecated? And migrate old codecase to use `retryUntilFatalError` instead of `retry` in next PRs?

ccojocar: sounds good!

hekonsek: Pushed changes. Can you take a look at those now? Thanks!

hekonsek:
Can we try to merge this one now? :)

PR #2739, opened 2019-01-18
wip: use vault for secrets if available

pmuir:
#### Submitter checklist

#### Description
feat: use vault for secrets if available
* also fixes bug where versions for charts were not getting passed
* also use a template for secrets if no vault

@ccojocar does this look approximately correct? If you can validate the basic approach, I'll work on adding tests

fixes #


pkg/jx/cmd/add_app.go
```diff
@@ -224,83 +266,82 @@ func (o *AddAppOptions) Run() error {
...
-						o.ValueFiles = []string{
-							valuesFile.Name(),
+						allSecrets := map[string][]*surveyutils.GeneratedSecret{
+							appsGeneratedSecretKey: secrets,
+						}
+						ybs, err := json.Marshal(allSecrets)
+						if err != nil {
+							return err
 						}
+						valuesYaml = append(valuesYaml, ybs...)
```
ccojocar: Is the `ybs` file which contains the secret going to be removed from the local file systems?

pmuir:
> How I understand the flow:
> 
> * a new secret is randomly generated or received from user
> * it is stored  in vault

Yes

> * an new k8s secret is created which contains the value of the vault secret
> * the k8s secret is used when installing the chart

Right now, no. The code you see for creating the secret is is handling the case that vault is not installed.

> I think this approach looks fine to me for now. In the mid-term, we would like to get rid of the duplicated secrets and use a CSI Driver or Flex Volume to mount the secrets directly from vault into the Pod. This will improve the secrets security, rotation, and avoid stale secrets, but this will require some change to the current approach.

How do I handle this (getting the secret back out of the vault when the helm chart install actually happens) in the meantime? I guess in `jx step env apply`? Does `jx step env apply` do this already.

/cc @jstrachan 

> Also, you need to handle the secret rotation when an app is updated. The user most likely will want to change the secret in vault and also propagate this change to the application. The easiest way would be to update the k8s secret from vault when the app is updated.

Yeah. In general I haven't written the code to handle upgrades to any configuration, including secrets. Going to do that next.

> Also is not clear to me in which namespace are the secrets created. How is going to play with team/user access control? Could potentially a secret which is only accessible to an admin in vault be leaked to a user in k8s, which does not have the admin rights, but have access to the k8s namespace where the application is installed.

For the non-gitops case, the secrets are created in the namespace the user chooses (by default the dev namespace). For gitops, this is handled by `jx step env apply` (also @jstrachan is this correct?)

pkg/jx/cmd/add_app.go
pmuir: No, thanks.

jstrachan:
@pmuir I think we should vault secret handling to the `jx step helm apply` which is invoked via `jx step env apply` - so that if we do a `jx step env apply` or `jx step helm apply`  in an environment pipeline we do whatever vault stuff is required to pass secrets into helm

To @ccojocar's point around avoiding the need to make `Secret` resources along with handling namespaces and stuff; we could try just use vault to generate the `secrets.yaml` file thats passed into helm on each environment pipeline without needing to Secrets and leaving the helm chart to make any secrets required.

Handling rotation could be tricky but we may need to just ensure we trigger environment pipelines a few times per rotation period to see if anything has changed?

pmuir:
Ok, I've fixed the issue @ccojocar found. I'm going to do the work for `jx step helm apply` in another PR.

pmuir:
This is really failing. I need some help from @ccojocar to work out how to handle the vault stuff in tests

pmuir:
/test this
BDD timeout

pmuir:
/hold
Just testing batch PRs

PR #2740, opened 2019-01-18
chore(users) users need to be identified by their login not email

jtnord:
#### Submitter checklist

#### Description

this changes the behaviour introduced in #2669 which before that change
was incositent depending on if you where calling Create or
CreateOrUpdate.

Additionally FindByEmail would not have worked correctly previously due
to the fact that some users where based on their login and some based on
their email.  I found no useage of this code, so rather than fixing it I
have removed it.  If it transpires there are users of this in some
private repo then we can reinstate a working version.

see https://github.com/jenkins-x/jx/pull/2669#discussion_r24863822 for
the discussion

see discussion at https://github.com/jenkins-x/jx/pull/2669#discussion_r24863822


pkg/users/users.go
```diff
@@ -35,10 +35,7 @@ func GetUsers(jxClient versioned.Interface, ns string) (map[string]*jenkinsv1.Us
 
 // CreateUser creates a new default User
 func CreateUser(ns string, login string, name string, email string) *jenkinsv1.User {
-	id := login
-	if email != "" {
-		id = kube.EmailToK8sID(email)
-	}
+	id := kube.ToValidName(login)
```
agentgonzo: Is `login` being empty here _ever_ a possibility?

markawm:
(fingers crossed now...)

pkg/jx/cmd/preview_test.go
```diff
@@ -141,8 +140,6 @@ preview:
 // functions).
 // TODO: Refactor the implementation & test so the various stages of creating a preview env. can be tested individually.
 func TestRun_CreateNewPreviewEnv(t *testing.T) {
-	t.Parallel()
```
jtnord: the code was using EnvironmentVariables which are global and inherently not suitable for parallelism - so removed this whilst I was here.

pkg/users/users.go
jtnord: *everything* is a possibility and it is a global function.  we will find out if anyone does pass a `nil` login soon enough - but this was how the old code worked.

pkg/users/user_details.go
```diff
@@ -96,7 +85,7 @@ func (this *UserDetailService) CreateOrUpdateUser(u *v1.UserDetails) error {
 			ObjectMeta: metav1.ObjectMeta{
 				Name: id,
 			},
-			User: *u,
+			Spec: *u,
```
jtnord: User is deprecated, so switched to Spec whilst I was here.

Issue #2741, opened 2019-01-18
Tests are creating 'stuff' in the cluster


markawm:
### Summary 

Happened to notice that when I _didn't_ have a valid K8S context, tests would fail, i.e. _some_ tests are using a real connection to a cluster. Hence any 'stuff' they create/delete/update can be real.
Possibly other case, but observed right now: `ConfigureTestOptions` in `test_helpers.go` is setting up fake jx/k8s clients and setting these into the cached copies in CommonOptions, which is all great. Any code that calls CommonOptions.JXClient() will get that cached one returned. But it is also putting a _real_ `Factory` into the CommonOptions. Then, if any test or production code calls `CommonOptions.CreateJXClient()`, this will ignore the cached copy and create a new client using the (real) Factory.

### Steps to reproduce the behavior
Run unit tests, e.g. `make test` or just `/test this` on a PR.

### Jx version

The output of `jx version` is:

```
COPY OUTPUT HERE
```

### Kubernetes cluster

What kind of Kubernetes cluster are you using & how did you create it?

ccojocar:
@markawm Are these unit tests or integration tests? Some tests are annotated with the `//+build integration` build tag, because they require a real cluster and jenkins-x installation.

When is this happening? When you  are running `make test` or `make test-slow-integration`?

jstrachan:
might be an idea to force env vars to point to dummy context just in case?

we do that here.. https://github.com/jenkins-x/jx-quickstarts-bdd-test-job/blob/master/jx/scripts/release.sh#L29-L33

PR #2744, opened 2019-01-18
fix: never create a new kube client or jx client

pmuir:
This means the mocks get used

#### Submitter checklist

#### Description
Continuation of work for #2741 

This one is less safe as I didn't write all the involved code :-/

fixes #


markawm:
Change itself LGTM - much happier to see everything going through CommonOptions. My only hesitation with making this kind of change was whether _some_ of the new clients were actually valid and genuinely needed to be new because, for example, two namespaces were going to be accessed at the same time. Anyone know if this is necessary, or even possible??? @jstrachan ?

PR #2745, opened 2019-01-18
Make bdd reports optional

davidcurrie:
#### Submitter checklist

#### Description

Not all BDD tests that use `jx step bdd` may be creating a `reports` directory and the command currently fails if it does not exist.


pmuir:
/hold

jstrachan:
@davidcurrie sorry about that, we've a conflict - any chance you can rebase your branch please?

davidcurrie:
Apologies for the force push but it was easier to just start again. Less of a problem now anyway given that the failure of the copy doesn't fail the build.

PR #2746, opened 2019-01-18
EKS resources should be tagged as created by jx by default

hekonsek:
EKS resources should be tagged as created by jx by default.


pmuir:

pmuir:
/retest
GitHub was returning 503s

PR #2748, opened 2019-01-18
add missing CLI args for jx stash

jstrachan:
also add a few more doc links


pmuir:
/hold

PR #2753, opened 2019-01-19
fix: lets not fail if there is no `.gitignore` when importing a project

jstrachan:


pmuir:
Manually merging for now despite test failures.

PR #2767, opened 2019-01-21
improve install config capture + refactor bucket create code

jstrachan:


pkg/kube/constants.go
```diff
@@ -244,8 +244,20 @@ const (
 
 	AUTH = "auth"
 
-	Region = "Region"
+	// KubeProvider stores the kubernetes provider used to create the cluster or install
+	KubeProvider = "kubeProvider"
```
ccojocar: Should we remove in future the cloud provider from team settings? I think is more a cluster config.

jstrachan: agreed

PR #2773, opened 2019-01-21
fix(upgrade ingress): get the upgrade ingress command working in batch mode with and without TLS

ccojocar:
#### Submitter checklist

fixes #


pmuir:
Just whilst we're debugging the tide problem

PR #2794, opened 2019-01-22
Initialize jxclient

fiunchinho:

#### Description
We need to make sure we have a valid jx client, otherwise we may get a nil pointer dereference

### DIFF
--- a/pkg/jx/cmd/common_install.go
+++ b/pkg/jx/cmd/common_install.go
@@ -1691,12 +1691,12 @@ func (o *CommonOptions) createWebhookProw(gitURL string, gitProvider gits.GitPro
 
 func (o *CommonOptions) isProw() (bool, error) {
 	ns := o.devNamespace
-	if ns == "" {
-		ns = o.currentNamespace
-	}
-	jxClient, _, err := o.JXClient()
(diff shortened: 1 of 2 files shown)


pkg/jx/cmd/common_install.go
```diff
@@ -1694,7 +1694,11 @@ func (o *CommonOptions) isProw() (bool, error) {
 	if ns == "" {
 		ns = o.currentNamespace
 	}
-	env, err := kube.GetEnvironment(o.jxClient, ns, "dev")
+	jxClient, _, err := o.JXClient()
```
ccojocar: You can configure the jx client and set the dev environment in one call with `JXClientAndDevNamespace`.

jstrachan:
@ccojocar I tweaked the PR to use `JXClientAndDevNamespace()`

pkg/jx/cmd/common_install.go
jstrachan: I've fixed this

fiunchinho:
👍

fiunchinho:
I guess someone needs to /ok_to_test ?

pmuir:
/test this
Missing build log

PR #2797, opened 2019-01-22
feat(ingress): define a new flag in the upgrade ingress command which allows to specify the namespace where the ingress config is stored

ccojocar:
#### Submitter checklist

#### Description

It defines a new flag in the upgrade ingress command which allows to specify where the ingress config map is stored. This option is required 
for BDD tests to be able to run the upgrade ingress in a self-contained namespace without modifying the entire cluster configuration.

fixes #


pmuir:
/test this
Build got lost in some kind of upgrade or outage

PR #2804, opened 2019-01-22
feat(create user): add new flag to User CRD to enable serviceaccount.

markawm:
#### Submitter checklist

#### Description

New command line flag --create-service-account that gets stored in User.createServiceAc **(boolean).**

fixes #

### DIFF
--- a/pkg/apis/jenkins.io/v1/types_users.go
+++ b/pkg/apis/jenkins.io/v1/types_users.go
@@ -45,6 +45,7 @@ type UserDetails struct {
 	AvatarURL         string             `json:"avatarUrl,omitempty"  protobuf:"bytes,6,opt,name=avatarUrl"`
 	ServiceAccount    string             `json:"serviceAccount,omitempty"  protobuf:"bytes,7,opt,name=serviceAccount"`
 	Accounts          []AccountReference `json:"accountReference,omitempty"  protobuf:"bytes,8,opt,name=accountReference"`
+	CreateServiceAc   bool               `json:"createServiceAc,omitempty"  protobuf:"bytes,9,opt,name=createServiceAc"`
 }
 
 // AccountReference is a reference to a user account in another system that is attached to this user
--- a/pkg/apis/jenkins.io/v1/types_users.go
+++ b/pkg/apis/jenkins.io/v1/types_users.go
@@ -11,6 +11,11 @@ import (
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +k8s:openapi-gen=true
 
+const (
+	UserTypeLocal    = "User"
+	UserTypeExternal = "ServiceAccount"
+)
+
 // User represents a git user so we have a cache to find by email address
 type User struct {
 	metav1.TypeMeta `json:",inline"`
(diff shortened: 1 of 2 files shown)


pmuir:
This doesn't feel right as CRDs are supposed to represent state rather than proceedural logic.

markawm:
Agreed... is the worry just about naming, then? The state is that this user _should_ have a service account, and a controller will ensure the system state matches that.

pkg/apis/jenkins.io/v1/types_users.go
```diff
@@ -45,6 +45,7 @@ type UserDetails struct {
 	AvatarURL         string             `json:"avatarUrl,omitempty"  protobuf:"bytes,6,opt,name=avatarUrl"`
 	ServiceAccount    string             `json:"serviceAccount,omitempty"  protobuf:"bytes,7,opt,name=serviceAccount"`
 	Accounts          []AccountReference `json:"accountReference,omitempty"  protobuf:"bytes,8,opt,name=accountReference"`
+	CreateServiceAc   bool               `json:"createServiceAc,omitempty"  protobuf:"bytes,9,opt,name=createServiceAc"`
```
ccojocar: I would use full words for this for clarity since is an API definition, e.g `CreateServiceAccount`.

markawm: First comment - done.
Second comment - I'm not sure I follow: the `ServiceAccount` property will _always_ be empty when the User is first created, until the controller creates the `ServiceAccount` then updates the `User` to set this reference. i.e. the controller would _always_ create the ServiceAccount, rather than selected during user creation.

ccojocar: Not sure then, If I understand the purpose of `CreateServiceAccount` flag. What's going to happen if it is `false`? Will the controller not create the user?

pmuir:
> state is that this user _should_ have a service account

To me, that isn't state, that's logic. To me the sate is the user has a service account, or doesn't have a service account - this is the same thing that @ccojocar is asking about 

> You can also check the ServiceAccount field. If that field is empty, the controller should create that service account and populate the name of the service account in the CRD

pkg/apis/jenkins.io/v1/types_users.go
markawm: Yes, that's right: if `CreateServiceAccount==false`, then the controller won't do anything. Only if `CreateServiceAccount==true` will it create the `ServiceAccount` and set the ref. into `User.ServiceAccount`.

markawm:
I _think_ I see what you mean: its a flag that is driving logic, which isn't quite right for CRDs. The bit I'm missing is: aren't many bits of CRDs state... that we then use to drive logic? e.g. "If this is a type _X_ User then do _this_, but if its a type _Y_ User then do _that_". This is why I was asking about the naming, would it make more sense if instead modelled as a `userType` field or something?

(Or have I missed the point completely?! Do you have any suggestions that would better model this?)

pkg/apis/jenkins.io/v1/types_users.go
ccojocar: What is going to happen with the users that do not have a service account? Is the main purpose of this flag to delay the service account creation until an admin comes and set it to `true`? TBH I don't have an overview of the entire use case. 

My only concern is that we will have a user resource with status `created` but with inconsistent metadata. Some users will have service accounts while others not. Probably, we should define also different statuses for user resource e.g. init, ready etc. 

Normally, a controller should watch when a user is created, create the service account and update the status of the users as e.g. `provisioned/ready`.

jtnord: users who are known about by the k8s platform do not need to have service accounts.
for example if using gke and you have a basic role (enough to be able to use gcloud login) to access the cluster then jenkins x needs the User in order to be able to work out who to apply the roles to.
in the case of a Google Cloud User they do not need a service account.  For non cloud users (ie those without access to the k8s cluster) they need the service account.

//cc @swashbuck1r

jtnord: I think we are just getting confused about the name of this.  to me is a boolean flag to indicate the type of user - perhaps `nativeUser` would be more correct - and that would be true for users that auth with the cluster and false for ones that need some other form of auth (like the service account)

pkg/apis/jenkins.io/v1/types_users.go
```diff
@@ -37,14 +37,15 @@ type UserList struct {
...
-	Accounts          []AccountReference `json:"accountReference,omitempty"  protobuf:"bytes,8,opt,name=accountReference"`
+	Login                string             `json:"login,omitempty"  protobuf:"bytes,1,opt,name=login"`
+	Name                 string             `json:"name,omitempty"  protobuf:"bytes,2,opt,name=name"`
+	Email                string             `json:"email,omitempty"  protobuf:"bytes,3,opt,name=email"`
+	CreationTimestamp    *metav1.Time       `json:"creationTimestamp,omitempty" protobuf:"bytes,4,opt,name=creationTimestamp"`
+	URL                  string             `json:"url,omitempty"  protobuf:"bytes,5,opt,name=url"`
+	AvatarURL            string             `json:"avatarUrl,omitempty"  protobuf:"bytes,6,opt,name=avatarUrl"`
+	ServiceAccount       string             `json:"serviceAccount,omitempty"  protobuf:"bytes,7,opt,name=serviceAccount"`
+	Accounts             []AccountReference `json:"accountReference,omitempty"  protobuf:"bytes,8,opt,name=accountReference"`
+	CreateServiceAccount bool               `json:"createServiceAccount,omitempty"  protobuf:"bytes,9,opt,name=createServiceAccount"`
```
jtnord: possibly `NativeUser` or something along those lines.

pmuir: > I _think_ I see what you mean: its a flag that is driving logic, which isn't quite right for CRDs. The bit I'm missing is: aren't many bits of CRDs state... that we then use to drive logic? e.g. "If this is a type _X_ User then do _this_, but if its a type _Y_ User then do _that_". This is why I was asking about the naming, would it make more sense if instead modelled as a `userType` field or something?

Yes something like that makes a lot more sense to me.

pkg/apis/jenkins.io/v1/types_users.go
pmuir: Agreed, that seems more consistent with the way these APIs work.

pkg/apis/jenkins.io/v1/types_users.go
markawm: 👍 Great, thanks, I'll change the representation to something that indicates the type of User that was created.

pkg/apis/jenkins.io/v1/types_users.go
markawm: @ccojocar - just to address your questions: metadata shouldn't be inconsistent; we will (well, could) have a mix of users with & without a ServiceAccount, but only because they are of different types, some being 'local' K8S users (no SA) and some are external users (with SA). The logic and state should be consistent, whereby the controller will always ensure SA exists for those users that need it. There's no expectation that an admin or anyone else would ever change the state of this field: it is set at user creation and represents the type of user (SA required or not), and _that never changes_.

pkg/apis/jenkins.io/v1/types_users.go
```diff
@@ -11,6 +11,13 @@ import (
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +k8s:openapi-gen=true
```
ccojocar: nit: You can define a type for this.

```Go
type UserType string 

const (
    UserTypeLocal UserType = "User"
)
```

markawm: Good pointer, thanks.
I tried this out, but it didn't really seem to help much as it is ultimately is compared against the Kind in the Subject, which is itself a String. I found the _string_ constants were already defined there, though, so switched to using those.

markawm: @ccojocar, are you happy with this now (i.e. kept as `string`, but using the client's already defined constants).

ccojocar: It's fine.

PR #2821, opened 2019-01-23
fix: avoid failing create project if no GitService

jstrachan:
but we can find the git provider kind from the git secrets


pmuir:
/hold
Just testing batch PRs

PR #2824, opened 2019-01-23
fix: allow `jx get build logs jobName` to be used with prow

jstrachan:
which does not need the build number to be included


pmuir:
BAH, sorry!

PR #2826, opened 2019-01-23
fix(install): wait for vault to be initialized and unsealed before retrieving the auth token

ccojocar:
#### Submitter checklist

fixes #


ccojocar:
Thanks. I updated the code to use that retry.

PR #2862, opened 2019-01-25
fix: do not clean certs on cb addon install

amuniz:
#### Submitter checklist

#### Description
Cleaning certificates and issuers is not needed when installing this addon.
it was generating a full round of new certificate requests which sums up for the rate limit of let's encrypt.
So this is reducing the number of requested certificates when installing the addon.

This is based on a conversation with @ccojocar, he told me there is no need to recreate all certificates because a separate certificate request would be done for the addon automatically. I've tested this locally and it works fine (the certificate is issued correctly).

### DIFF
--- a/pkg/jx/cmd/create_addon_cloudbees.go
+++ b/pkg/jx/cmd/create_addon_cloudbees.go
@@ -161,18 +161,11 @@ To register to get your username/password to to: %s
 			return errors.Wrap(err, "ensuring cert-manager is installed")
 		}
 
-		ingressConfig.TLS = true
-		ingressConfig.Issuer = kube.CertmanagerIssuerProd
-		err = kube.CleanCertmanagerResources(client, o.Namespace, ingressConfig)
-		if err != nil {
-			return errors.Wrap(err, "creating cert-manager issuer")
-		}
-
 		values := []string{
 			"sso.create=true",
 			"sso.oidcIssuerUrl=" + dexURL,
 			"sso.domain=" + domain,
-			"sso.certIssuerName=" + ingressConfig.Issuer}
+			"sso.certIssuerName=" + kube.CertmanagerIssuerProd}
 
 		if len(o.SetValues) > 0 {
 			o.SetValues = o.SetValues + "," + strings.Join(values, ",")


pkg/jx/cmd/create_addon_cloudbees.go
```diff
@@ -161,18 +161,11 @@ To register to get your username/password to to: %s
 			return errors.Wrap(err, "ensuring cert-manager is installed")
 		}
 
-		ingressConfig.TLS = true
-		ingressConfig.Issuer = kube.CertmanagerIssuerProd
-		err = kube.CleanCertmanagerResources(client, o.Namespace, ingressConfig)
```
ccojocar: I think, the addon installation will fail, if a different namespace than `jx` is used. 

This doesn't clean up the certificates but the issuer and recreates it again according to the ingress configuration. This ensure that the addon always has an up to date production issuer, which can be used by the `sso-operator`.

amuniz: [It is removing certificates too](https://github.com/jenkins-x/jx/blob/cff7d93a9b86dc3dbeac696f5e9eee9f347194bd/pkg/kube/crds.go#L567-L568). So it forces a new round of certificates for all services in the namespace, thus certificate rate limit is hit quickly.

Does it make sense to install the addon in a different namespace? Should we force `jx` for this addon? It is a cluster-wide service, all teams would share it, so I guess it makes sense to have it always in the `jx` namespace.

Do you have some other possible solution in mind? Having all certificates regenerated is causing a lot of pain.

ccojocar: >t is removing certificates too. So it forces a new round of certificates for all services in the  >namespace, thus certificate rate limit is hit quickly.

These are the issuer CA certificates not the services certificates.

>Does it make sense to install the addon in a different namespace? Should we force jx for this >addon? It is a cluster-wide service, all teams would share it, so I guess it makes sense to >have it always in the jx namespace.

I wound't couple the installation to `jx` namespace which is the default dev namespace. In a multi-team setup, there are multiple dev namespaces which could potentially have a dedicated cb addon installed. 

>Do you have some other possible solution in mind? Having all certificates regenerated is >causing a lot of pain.

I think the issuer is recreated because for instance the email might change. It can also be updated and only created if no issuer exists in that namespace, but this will have the same effect.

amuniz:
Superseded by #2893

PR #2893, opened 2019-01-29
fix(issuer): do not regenerate issuers if it is not needed

amuniz:
#### Submitter checklist

#### Description
Do not regenerate certificate issuers if it is not needed (it's already the issuer required by ingress config and email matches).

This is an alternative to https://github.com/jenkins-x/jx/pull/2862

### DIFF
--- a/pkg/kube/pki/cert-manager.go
+++ b/pkg/kube/pki/cert-manager.go
@@ -2,6 +2,8 @@ package pki
 
 import (
 	"github.com/jenkins-x/jx/pkg/kube"
+	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/util"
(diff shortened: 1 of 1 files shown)


pkg/kube/pki/cert-manager.go
```diff
@@ -109,9 +111,14 @@ func CreateCertManagerResources(certclient certclient.Interface, targetNamespace
 		return nil
 	}
 
+	// do not recreate the issuer if it is already there and correctly configured
+	if alreadyConfigured(certclient, targetNamespace, ic) {
+		return nil
+	}
+
 	err := CleanCertManagerResources(certclient, targetNamespace, ic)
```
ccojocar: I think we don't need to clean up if does not exist.

ccojocar: Actually, we need when the update fails :)

PR #2897, opened 2019-01-29
feat(certs): wait for certificate issuer on create addon cb

### LINKED ISSUE #2876: Watch/Wait for Certs to be issued by cert-manager when installing cloudbees addon
### Summary 

`jx upgrade ingress` has the option to wait for certs to be issued when using TLS (`--wait-for-certs`). However `jx create addon cloudbees` also needs a certificate and it's not waiting for it to be issued.

Run `jx create addon cloudbees` and then `jx cloudbees`.

### Expected behavior

TLS is correctly configured in new services.

### Actual behavior

TLS certificate is not yet configured.

amuniz:
#### Submitter checklist

#### Description

`jx create addon cloudbees --sso` now waits for the TLS configuration to be ready (certificate issued for the service).

Output:
```
$ jx create addon cloudbees --sso
Configuring single sign-on...
? Domain: my-domain.com
? Dex URL: https://dex.sso.my-domain.com
Looking for cert-manager deployment in namespace cert-manager...
Updating Helm repository...
Helm repository update done.
Waiting for cert: tls-core...
Ready Cert: tls-core
Addon installed successfully.

  jx cloudbees Open the app in a browser

``` 

fixes #2876

### DIFF
--- a/pkg/jx/cmd/create_addon_cloudbees.go
+++ b/pkg/jx/cmd/create_addon_cloudbees.go
@@ -254,7 +254,7 @@ To register to get your username/password to to: %s
 		}
 	}
 
-	log.Infof("Addon installed successfully.\n Run `jx cloudbees` to open the app in a browser\n")
+	log.Infof("Addon installed successfully.\n\n  %s Open the app in a browser\n\n", util.ColorInfo("jx cloudbees"))
 
 	return nil
 }
--- a/pkg/jx/cmd/create_addon_cloudbees.go
+++ b/pkg/jx/cmd/create_addon_cloudbees.go
@@ -204,7 +204,7 @@ To register to get your username/password to to: %s
 		if err != nil {
 			return errors.Wrap(err, "creating the cert-manager client")
 		}
-		err = pki.WaitCertificateExists(certMngrClient, certName, o.Namespace, 3*time.Minute)
+		err = pki.WaitCertificateIssuedReady(certMngrClient, certName, o.Namespace, 3*time.Minute)
 		if err != nil {
 			return err // this is already wrapped by the previous call
 		}
--- a/pkg/jx/cmd/create_addon_cloudbees.go
+++ b/pkg/jx/cmd/create_addon_cloudbees.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"time"
 
 	survey "gopkg.in/AlecAivazis/survey.v1"
 
@@ -178,7 +179,7 @@ To register to get your username/password to to: %s
 			"sso.create=true",
 			"sso.oidcIssuerUrl=" + dexURL,
 			"sso.domain=" + domain,
-			"sso.certIssuerName=" + ingressConfig.Issuer}
+			"sso.certIssuerName=" + pki.CertManagerIssuerProd}
 
 		if len(o.SetValues) > 0 {
 			o.SetValues = o.SetValues + "," + strings.Join(values, ",")
@@ -193,6 +194,21 @@ To register to get your username/password to to: %s
 	err = o.CreateAddon("cb")
 	if err != nil {
 		return err
+	}
+
+	if o.Sso {
+		// wait for cert to be issued
+		certName := pki.CertSecretPrefix + "core"
+		log.Infof("Waiting for cert: %s...\n", util.ColorInfo(certName))
+		certMngrClient, err := o.CreateCertManagerClient()
+		if err != nil {
+			return errors.Wrap(err, "creating the cert-manager client")
+		}
+		err = pki.WaitCertificateExists(certMngrClient, certName, o.Namespace, 3*time.Minute)
+		if err != nil {
+			return err // this is already wrapped by the previous call
+		}
+		log.Infof("Ready Cert: %s\n", util.ColorInfo(certName))
 	}
 
 	if o.Basic {
@@ -237,6 +253,8 @@ To register to get your username/password to to: %s
 			return err
 		}
 	}
+
+	log.Infof("Addon installed successfully.\n Run `jx cloudbees` to open the app in a browser\n")
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


pkg/jx/cmd/create_addon_cloudbees.go
```diff
@@ -195,6 +196,21 @@ To register to get your username/password to to: %s
...
 
+	if o.Sso {
+		// wait for cert to be issued
+		certName := pki.CertSecretPrefix + "core"
+		log.Infof("Waiting for cert: %s...\n", util.ColorInfo(certName))
+		certMngrClient, err := o.CreateCertManagerClient()
+		if err != nil {
+			return errors.Wrap(err, "creating the cert-manager client")
+		}
+		err = pki.WaitCertificateExists(certMngrClient, certName, o.Namespace, 3*time.Minute)
```
ccojocar: What's the reason to no use ` WaitCertificateIssuedReady`? `WaitCertificateExists` only checks if the `Certificate` resource is created by the cert-manager, which it always happens regardless if the certificate was successfully issued by Let's Encrypt or not. 

You really need to wait until the certificate is ready not only that exists. It happens quite often that the certificate exists but it is in a failed state. The effective certificate is stored in a separate secret only when the Cert is ready.

pkg/jx/cmd/create_addon_cloudbees.go
```diff
@@ -195,6 +196,21 @@ To register to get your username/password to to: %s
...
+		if err != nil {
+			return err // this is already wrapped by the previous call
```
ccojocar: Maybe provide a mindful message that the cert is not ready within the given timeout.

pkg/jx/cmd/create_addon_cloudbees.go
amuniz: Using `WaitCertificateIssuedReady` was my first thought, however I checked it does not work because [`isReady`](https://github.com/jenkins-x/jx/blob/048fe20a87aebad6ffbd13112c1e701cdf6daf48/pkg/kube/pki/certificate.go#L41-L44) is returning always false (not sure why), even after the right certificate is issued. Then I realized it is used from test code only so I figured it was not really intended for production use.

It's ok for me to use it though, do you have any idea of why it doesn't work/how to fix it?

amuniz: The problem is that the cert returned by `client.CertmanagerV1alpha1().Certificates(ns).Get(name, metav1.GetOptions{})` comes with  `Status.Conditions` as `nil`, so `cert.HasCondition` always returns `false`.

It works for tests because they are _artificially_ [setting the condition](https://github.com/jenkins-x/jx/blob/42d03fe574dd8ac2e8d001eb05084ab047621bbe/pkg/kube/pki/certificate_test.go#L23-L26).

This is because the cert has an empty `Status` (not sure why):

```
$ kubectl describe cert tls-core
Name:         tls-core
Namespace:    jx
Labels:       <none>
Annotations:  <none>
API Version:  certmanager.k8s.io/v1alpha1
Kind:         Certificate
Metadata:
  Cluster Name:
  Creation Timestamp:  2019-01-30T10:18:52Z
  Generation:          1
  Owner References:
    API Version:           extensions/v1beta1
    Block Owner Deletion:  true
    Controller:            true
    Kind:                  Ingress
    Name:                  core
    UID:                   6deff0f0-2478-11e9-b307-42010a840fe9
  Resource Version:        190105
  Self Link:               /apis/certmanager.k8s.io/v1alpha1/namespaces/jx/certificates/tls-core
  UID:                     70ed9c88-2478-11e9-b307-42010a840fe9
Spec:
  Acme:
    Config:
      Domains:
        core.jx.dev.amunizmartin.com
      Http 01:
        Ingress:
        Ingress Class:  nginx
  Dns Names:
    core.jx.dev.amunizmartin.com
  Issuer Ref:
    Kind:       Issuer
    Name:       letsencrypt-prod
  Secret Name:  tls-core
Status:
Events:  <none>
```

ccojocar: The status is being updated by cert-manager it takes a while. Normally, if is empty means that the certificate is not ready. Do you have a `tls-core` secret? Do you see any errors in the cert-manager logs?

pkg/jx/cmd/create_addon_cloudbees.go
amuniz: It's already printing `Waiting for cert: tls-core...
error: waiting for certificate "tls-core" to be ready in namespace "jx".: timed out waiting for the condition`

pkg/jx/cmd/create_addon_cloudbees.go
amuniz: I'm looking to this, I think it is a bug in cert-manager.

amuniz: Yes. Just verified. When the secret is already there then the cert status is not correctly set. If I remove the secret, it is recreated and the status is set.

amuniz: I think it could be fixed by this commit https://github.com/jetstack/cert-manager/commit/dc97dde2ef3c1b8695cdf5ecb35d17c60ac19207.

ccojocar: Yeah, we need to delete the secret as well. You can use this function `https://github.com/jenkins-x/jx/blob/44b45ef2c3d0290f54bb60d5136ef4c26e90d725/pkg/kube/pki/certificate.go#L62`

I implemented a while ago a feature in cert-manager which deletes the secret when a cert is deleted, but they didn't want to activate it by default. https://github.com/jetstack/cert-manager/pull/819

Probably, we should activate this option when the cert-manager is installed.

amuniz: Confirmed, it's fixed on `cert-manager` 0.6.0. Just switched to `WaitCertificateIssuedReady`.

PR #2901, opened 2019-01-29
Use chyle upstream

fiunchinho:
#### Submitter checklist

#### Description
We are currently using a fork of the chyle library to generate the changelog information. We don't have custom code in our fork AFAIK, so it'd be better to go back to upstream instead of rely on our own outdated fork.


fiunchinho:
I guess we need a /ok_to_test or /test

PR #2918, opened 2019-01-30
fix: fail early if oke options can't be fetched

carlossg:
and don't hide errors

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/jx/cmd/create_cluster_oke.go
+++ b/pkg/jx/cmd/create_cluster_oke.go
@@ -143,7 +143,8 @@ func (o *CreateClusterOKEOptions) createClusterOKE() error {
 	//we assume user has prepared the oci config file under ~/.oci/
 	imagesArray, kubeVersionsArray, shapesArray, latestKubeVersion, err := oke.GetOptionValues()
 	if err != nil {
-		fmt.Println("error")
+		fmt.Println("Error getting OKE option values")
+		return err
 	}
 
 	endpoint := o.Flags.Endpoint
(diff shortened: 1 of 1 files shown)


carlossg:
@hhuiliu I'm trying the oke create cluster and it's failing executing this oci command, which is also failing in the command line

```
$ oci ce node-pool-options get --node-pool-option-id all
RequestException: ('Connection broken: IncompleteRead(0 bytes read, 109 more expected)', IncompleteRead(0 bytes read, 109 more expected))
```

carlossg:
Also raised in https://github.com/oracle/oci-cli/issues/119

carlossg:
I got the command to work after adding the OKE policy but the error is very confusing

```
Your tenancy must also have the following required policy statement defined in the root compartment of your tenancy here by a user with administrative privileges:

allow service OKE to manage all-resources in tenancy
```

pkg/jx/cmd/create_cluster_oke.go
```diff
@@ -143,7 +143,8 @@ func (o *CreateClusterOKEOptions) createClusterOKE() error {
 	//we assume user has prepared the oci config file under ~/.oci/
 	imagesArray, kubeVersionsArray, shapesArray, latestKubeVersion, err := oke.GetOptionValues()
 	if err != nil {
-		fmt.Println("error")
+		fmt.Println("Error getting OKE option values, have you created the OKE policy? https://docs.cloud.oracle.com/iaas/Content/ContEng/Concepts/contengpolicyconfig.htm")
+		return err
```
pmuir: This is not the right way to log error messages. use `errors.Wrapf`

PR #2956, opened 2019-02-01
Binaries/dependencies code refactoring

hekonsek:
I've refactored existing code related to binaries/dependencies management, in particular I've packaged it nicely into dedicated functions. This is some code cleaning before I add:
- dynamic version recognition from installed packages
- moving packages versions definitions from hard-coded representation into dedicated file stored in Cloud Environments

### DIFF
--- a/pkg/binaries/binaries.go
+++ b/pkg/binaries/binaries.go
@@ -4,7 +4,11 @@ import (
 	"fmt"
 	"github.com/jenkins-x/jx/pkg/log"
 	"github.com/jenkins-x/jx/pkg/util"
+	"gopkg.in/yaml.v2"
+	"io/ioutil"
+	"os"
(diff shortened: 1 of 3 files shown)


pkg/binaries/binaries.go
```diff
@@ -40,28 +44,90 @@ type VersionExtractor interface {
...
+			return nil, err
+		}
+		return binariesVersions, nil
+	}
+	return map[string]string{}, nil
 }
 
+func RememberInstalledPackage(packageName string, version string) error {
+	versions, err := LoadInstalledPackages()
+	if err!= nil {
```
ccojocar: nit: space after `err`

hekonsek: Fixed. Thanks!

PR #2963, opened 2019-02-01
add a note preferring patch to update

jtnord:
#### Submitter checklist

#### Description

Add a note to the API docs that we should be using `PATCH` over `UPDATE` where possible.

fixes #


docs/apidocs/static_includes/_overview.html
```diff
@@ -48,6 +48,7 @@ <H4 id="resource-operations-update">Update</H4>
 Note: Replacing a resource object may not result immediately in changes being propagated to downstream objects.  For instance
 replacing a <CODE>ConfigMap</CODE> or <CODE>Secret</CODE> resource will not result in all <I>Pod</I>s seeing the changes unless the <I>Pod</I>s are
 restarted out of band.</P></LI>
+Note: <STRONG>Where possible using <CODE>Patch</CODE> is preferable over <CODE>UPDATE</CODE> as if new fields have been added to the resource object that is not known about by the running code these will be lost.
```
pmuir: Shouldn't this be inside the `<li>` with a `<br>` before it?

PR #3011, opened 2019-02-06
chore(bdd): don't use shorthands in tests

agentgonzo:
The -n flag can't be specified in create terraform so use the longhand version

Also expand environment variables in the arguments list


jstrachan:
@agentgonzo any chance you could rebase this PR please? we've had a flaky test thats now disabled

PR #3021, opened 2019-02-07
fix(sso): use the locked version instead of hard-coding for cert-manager, sso-operator, dex and vault-operator

ccojocar:
Also slightly refactored the DefaultChartURL to reduce duplication.

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/helm/constants.go
+++ b/pkg/helm/constants.go
@@ -1,6 +1,0 @@
-package helm
-
-const (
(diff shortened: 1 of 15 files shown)


agentgonzo:
Can you explain how this works differently with the empty strings rather than explicit version numbers?

jstrachan:
@agentgonzo using https://jenkins-x.io/architecture/version-stream/

Issue #2410, opened 2018-12-04
prow quickstart release builds get stuck, auto promoting to staging 


rawlingsj:
Even though the promotion is successful the release pipeline for the quickstart never completes.  The last entry in the logs say the promote can't get commit statuses:
```
Merge commit has not yet any statuses on repo project-demo/environment-jx-staging merge sha abc123
```
I can see the PR on the `environment-jx-staging` repo get's merged successfully so not too sure what's up yet.

rawlingsj:
As a quick workaround we could add the `--no-wait` flag to promote in the buildpack https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes/blob/master/packs/maven/pipeline.yaml#L32 when using prow ?  @jstrachan what do you think?

rawlingsj:
Rather than adding the workaround in the build backs which would affect non prow pipelines too, let's add it to jx, log an explanation and let's keep this issue open to investigate and fix the real issue.

dinfer:
I meet the same problem. and since the pr is merged, how can I get the fix?
the version of jenkins-x-platform is still `0.0.3402` for 15 days now. when will the jenkins-x-versions get updated? should I manually change the local version repo to force the platform get upgrade?

GingerBuffalo:
I think I'm hitting something similar to this issue. I used jx import on an existing project, and using static jenkins. I'm using gitlab enterprise edition, hosted within the jenkins x cluster. Everything appears to set up correctly, including webhooks. The webhooks are not showing any error. But jx promote is hanging on the pr sent to staging. The pr in staging is created just fine, but it's not auto approved. I've tried manually approving, but it appears jx promote can't actually see the pr (hence this popular warning message "WARNING: Failed to query the Pull Request last commit status for..."). 

As the webhooks appear to work just fine, I'm trying to see if maybe the issue is to do with how merge requests are configured to work on the gitlab server. If anyone here has any ideas, I'm all ears.

valdestron:
/remove-lifecycle stale

 Any news about this issue ?

dinfer:
/remove-lifecycle stale

It seems no one care about this issue.

Ubermensch01:
/reopen
I'm still seeing this issue in the pipeline, can anyone shed some light on it?

PR #3034, opened 2019-02-08
bug: update activities to complete on PR merge.

markawm:
#### Submitter checklist

#### Description

This is basically copying the workaround for issue #2410 from `PromoteOptions.waitForGitOpsPullRequest()` into
the workflow controller so that PipelineActivities get updated when the promotion PR merges. Normally, this is
done via `jx promote` during the build, but if that fails/times out then this ensures the workflow controller
tidies up the status to reflect that the PR is merged & promotion is complete.

fixes #

### DIFF
--- a/pkg/jx/cmd/controller_workflow.go
+++ b/pkg/jx/cmd/controller_workflow.go
@@ -8,7 +8,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/client/clientset/versioned"
 	typev1 "github.com/jenkins-x/jx/pkg/client/clientset/versioned/typed/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/gits"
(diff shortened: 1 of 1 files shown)


markawm:
I've duplicated some of the code from `jx promote` into the workflow controller, where I see _largely_ the same logic, but tbh I'm not sure why we have the same (or very very similar) code in both places: both:
 - Can these be merged? Is it just a refactoring that hasn't happened yet?
 - why are both the build job (via `jx promote`) and the workflow controller taking on what looks to me like essentially the same role in promotion?

jstrachan:
@markawm refactoring to share code would be good. The workflow controller came second; the longer term idea is to move more of the logic for promotion to a single controller and have less happen in the pipelines themselves; though we've yet to totally go 100% in on the workflow controller right now. We may be able to use bits of the workflow controller triggered via knative build pipeline at some point too to simplify the code + logic

pkg/jx/cmd/controller_workflow.go
```diff
@@ -507,6 +519,9 @@ func (o *ControllerWorkflowOptions) pollGitStatusforPipeline(activity *v1.Pipeli
 			log.Warnf("Failed to get PR number: %s", err)
 			return
 		}
+		if prNumber == 42 {
+			fmt.Printf("!!!Stop\n")
+		}
```
amuniz: wat?

markawm: D'oh - this was from when I was testing out the fix. But it got merged before I had a chance to remove it! :-(
See https://github.com/jenkins-x/jx/pull/3053

PR #3040, opened 2019-02-08
fix: lets bring back the `--view` option for `jx step creete task`

jstrachan:
and lets add `buildPack: none` for now to disable inheriting a build pack


dwnusbaum:
/lgtm

Is updating the mocking stuff something that we should be doing for PRs that change options for commands, or is it unrelated?

jstrachan:
@dwnusbaum I think we were hoping to automate the proxy generation on PRs - but yeah, its a bit tricky/troublesome at the moment

PR #3042, opened 2019-02-08
feat(kpipelines): Add loop-over-values to repeat step

abayer:
#### Submitter checklist

#### Description

Adds new syntax for a step to define a list of possible values for a
variable and run a set of one more steps with the current value of the
variable added to the environment, once for each value. Does allow
nesting of loops as well.

cc @dwnusbaum, @joseblas, @kshultzCB, @rawlingsj 

n/a


pkg/kpipelines/syntax/pipeline.go
```diff
@@ -674,49 +730,16 @@ func stageToTask(s Stage, pipelineIdentifier string, buildIdentifier string, nam
...
-								Env:     env,
-								Image:   stepImage,
-								Command: []string{step.Command},
-								Args:    step.Arguments,
-							})*/
-			} else {
-				return nil, errors.New("syntactic sugar steps not yet supported")
+			t.Spec.Steps = append(t.Spec.Steps, actualSteps...)
+			for k, v := range stepVolumes {
+				volumes[k] = v
```
dwnusbaum: Just noticed that we append volumes looping over a map on Line 746, so that will be nondeterministic, but we don't have any tests for volumes IIRC so not a problem yet.

abayer: d'oh!

abayer:
@rawlingsj I see that and raise you a...

/meow box

abayer:
...dangit.

/meow boxes

pkg/kpipelines/syntax/pipeline.go
dwnusbaum: And fwiw I introduced the bug in abayer/jx#13

abayer: heh, well, I'll clean that up in a followup.

rawlingsj:
> I see that and raise you a...

@abayer well played :)

abayer:
What can I say - I like cats in boxes.

PR #3043, opened 2019-02-08
fix: Make capital letters lowercase when creating resource names

dwnusbaum:
#### Submitter checklist

#### Description

Instead of converting uppercase characters into `-`, it seems better to convert them into their lowercase equivalent. I don't have any specific justification for the change, but in future Pipeline-related work we anticipate using this function for user-specified names, which may include capital letters, and the change seems harmless enough. Happy to hold off on it for now and open an issue instead if desired.


dwnusbaum:
D'oh this already happens because of strings.toLower... closing

Issue #2849, opened 2019-01-24
jx delete application fails to delete an app unless --org=<org> is passed


mikecirioli:
### Summary 
`jx delete application` fails to delete an app unless you explicitly specify the GH organization via the `--org=<org>` command line parameter.  This command should be using the `sourcerepositories` CRD to automatically supply this information so that you don't have to supply it on the command line.

### Steps to reproduce the behavior

1. Create a new quickstart app  (ex. golang-http app called `foobar`)
2. verify that you can see it via `jx get apps`
3. try to delete it using `jx delete application foobar`

### Expected behavior
The app will be deleted

### Actual behavior
App is not deleted
```
>$ jx delete application foobar3
Removing application foobar3 from environment Staging
? user name to submit the Pull Request mikecirioli
Created Pull Request: https://github.com/jx-dev-org-mikec/environment-jx-dev-mikec-jx-dev-mikec-staging/pull/13

Waiting for pull request https://github.com/jx-dev-org-mikec/environment-jx-dev-mikec-jx-dev-mikec-staging/pull/13 to merge
Failed to query the Pull Request last commit status for https://github.com/jx-dev-org-mikec/environment-jx-dev-mikec-jx-dev-mikec-staging/pull/13 ref 69abe75975757b610c6029e56f71eebd29abfa77 Could not find a status for repository jx-dev-org-mikec/environment-jx-dev-mikec-jx-dev-mikec-staging with ref 69abe75975757b610c6029e56f71eebd29abfa77
Request https://github.com/jx-dev-org-mikec/environment-jx-dev-mikec-jx-dev-mikec-staging/pull/13 is merged!
Removing application foobar3 from environment Production
No changes made to the GitOps Environment source code. Code must be up to date!
error: deleting application: deleting prow config for foobar3: removing repo /foobar3 from branch protection: Invalid path /foobar3 could not determine organisation and repository name
```
### Jx version

The output of `jx version` is:

```
NAME               VERSION
jx                 1.3.798
jenkins x platform 0.0.3306
Kubernetes cluster v1.10.11-gke.1
kubectl            v1.12.2
helm client        v2.11.0+g2e55dbe
helm server        v2.12.2+g7d2b0c7
git                git version 2.16.0
Operating System   Mac OS X 10.14.2 build 18C54
```

### Jenkins type

- [ ] Classic Jenkins
- [x] Serverless Jenkins

mikecirioli:
closed, this has been addressed by #3045

PR #3045, opened 2019-02-08
chore: fix delete application so that it tries to intelligently figure what org the associated repo belongs to

mikecirioli:
fix delete application so that it tries to intelligently figure what org the associated repo belongs to so that the `sourcerepository` crd object it is associated with can be deleted

#### Submitter checklist

#### Description
When deleting an application, `jx` should remove the `sourcerepository` metadata associated with it.  To do this, it needs to know the organization name (because the `sourcerepository` objects are named `<org>-<reponame>`)

I am not completely happy with this approach, and am creating this PR to solicit some feedback and ideas on how this could be better solved (even if it means revisiting the way we name the object metadata in the `sourcerepository` crd)

It might be worth adding a `--force` option to `jx delete` that could be used to clean up projects that are left in an inconsistent state.  For example, in testing this i was able to delete the prow config for my app, but not the CRD object.  Subsequent re-runs of the same `jx delete application` command abort prior to cleaning up the CRD with:
`error: deleting application: deleting prow config for golang1: removing repo jx-dev-org-mikec/golang1 from branch protection: repo golang1 not found in org jx-dev-org-mikec`

fixes #
#2849

### DIFF
--- a/pkg/jx/cmd/delete_application.go
+++ b/pkg/jx/cmd/delete_application.go
@@ -122,12 +122,6 @@ func (o *DeleteApplicationOptions) Run() error {
 		deletedApplications, err = o.deleteJenkinsApplication()
 	}
 
-	for _, deletedApplication := range deletedApplications {
-		err := repoService.DeleteSourceRepository(deletedApplication)
-		if err != nil {
-			log.Warnf("Unable to find application metadata for %s to remove", deletedApplication)
-		}
-	}
 	if err != nil {
 		return errors.Wrapf(err, "deleting application")
 	}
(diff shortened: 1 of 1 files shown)


pkg/jx/cmd/delete_application.go
```diff
@@ -155,20 +149,40 @@ func (o *DeleteApplicationOptions) deleteProwApplication(repoService kube.Source
...
-				return deletedApplications, fmt.Errorf("could not get org for %s. use --org", util.ColorInfo(applicationName))
+				return deletedApplications, fmt.Errorf("error in sourcerepository service %s", err.Error())
+			}
+
+			srObjects := []v1.SourceRepository{}
+			for sr := range srList.Items {
+				if srList.Items[sr].Spec.Repo == applicationName {
+					srObjects = append(srObjects, srList.Items[sr])
+					if len(srObjects) > 1 {
+						return deletedApplications, fmt.Errorf("application %s exists in multiple orgs, use --org to specify the app to delete", util.ColorInfo(applicationName))
```
agentgonzo: At this point, you could present the user with a filterable list of the multiple orgs and ask them which one to use, rather than returning an error.

mikecirioli: yeah, i was thinking that would make sense.  Seems fairly easy to do as well.

PR #3064, opened 2019-02-11
fix(kpipelines): Override workspace from podtemplate, fix args generation

abayer:
#### Submitter checklist

#### Description

We were using the working dir from the podTemplate, and we weren't mashing together the command and args specified in the YAML into something suitable for `/bin/sh -c`. More work needed on setting working dir, but this unblocks things for the moment.

cc @dwnusbaum 

n/a

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -96,7 +96,7 @@ require (
 	github.com/kevinburke/ssh_config v0.0.0-20180317175531-9fc7bb800b55 // indirect
 	github.com/keybase/go-crypto v0.0.0-20181127160227-255a5089e85a // indirect
 	github.com/knative/build v0.3.0
-	github.com/knative/build-pipeline v0.0.0-20190203191138-7b5938be3a54
+	github.com/knative/build-pipeline v0.0.0-20190207174440-709bc8516d6d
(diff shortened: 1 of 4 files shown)


abayer:
The `/bin/sh -c` is what's used as the base command for pod template containers, so I'm deriving from that.

go.mod
```diff
@@ -96,7 +96,7 @@ require (
 	github.com/kevinburke/ssh_config v0.0.0-20180317175531-9fc7bb800b55 // indirect
 	github.com/keybase/go-crypto v0.0.0-20181127160227-255a5089e85a // indirect
 	github.com/knative/build v0.3.0
-	github.com/knative/build-pipeline v0.0.0-20190203191138-7b5938be3a54
+	github.com/knative/build-pipeline v0.0.0-20190207174440-709bc8516d6d
```
pmuir: Just wanted to confirm you definitely meant to update this dependency and it wasn't done magically by go (as the changeset doesn't seem to require it?)

abayer: I may have just updated it, but it gets us in sync with what theoretically we need to end up having at runtime, so can't hurt. Also I'll have another PR in the next day or two bumping it further once https://github.com/knative/build-pipeline/pull/496 is merged anyway. =)

PR #3066, opened 2019-02-12
Extract fact crd

fiunchinho:
#### Submitter checklist

#### Description
Fact is a property of PipelineActivity. We want to make Fact a first class citizen of the API, creating its own CRD. The idea is partly to keep the RBAC simpler (as many apps are going to want to write Facts), partly to reduce contention on PipelineActivity, partly to avoid PipelineActivity getting bloated.

Apps need to be updated to use the new Fact CRD.


pkg/apis/jenkins.io/v1/types_fact.go
```diff
@@ -0,0 +1,99 @@
...
+	Status FactStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// FactSpec is the specification of a Fact
+type FactSpec struct {
+	Name             string        `json:"name" protobuf:"bytes,1,opt,name=name"`
+	ID               int           `json:"id" protobuf:"bytes,2,opt,name=id"`
+	FactType         string        `json:"factType" protobuf:"bytes,3,opt,name=factType"`
+	Measurements     []Measurement `json:"measurements" protobuf:"bytes,4,opt,name=measurements"`
+	Statements       []Statement   `json:"statements" protobuf:"bytes,5,opt,name=statements"`
```
jglick: Lack of `,omitempty` here caused issues for existing records after #2931. We should ensure that all optional fields are treated that way by the schema.

pkg/apis/jenkins.io/v1/types_fact.go
```diff
@@ -0,0 +1,99 @@
...
+
+// Measurement contains the value measured on this fact
+type Measurement struct {
+	Name             string   `json:"name" protobuf:"bytes,1,opt,name=name"`
+	MeasurementType  string   `json:"measurementType" protobuf:"bytes,2,opt,name=measurementType"`
+	MeasurementValue int      `json:"measurementValue" protobuf:"bytes,3,opt,name=measurementValue"`
+	Tags             []string `json:"tags,omitempty" protobuf:"bytes,4,opt,name=tags"`
+}
+
+// Statement is the Fact statement
```
jglick: Not very useful documentation. What is a “statement”? How does it differ from a “measurement”? When would you use it? When extracting the new CRD we should take the opportunity to review all the existing fields and think carefully about what they are needed for and why, and record that documentation.

pkg/apis/jenkins.io/v1/types_fact.go
```diff
@@ -0,0 +1,106 @@
...
+	CodeCoverageCountTypeInstructions = "Instructions"
+	CodeCoverageCountTypeBranches     = "Branches"
+	CodeCoverageCountTypeComplexity   = "Complexity"
+	CodeCoverageCountTypeLines        = "Lines"
+	CodeCoverageCountTypeMethods      = "Methods"
+	CodeCoverageCountTypeClasses      = "Classes"
+)
+
+const (
+	MeasurementPercent = "percent"
```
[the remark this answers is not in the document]
jglick: In particular, is this supposed to be a `MeasurementType`? If so, that field should link back here to the options.

pkg/apis/jenkins.io/v1/types_fact.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +genclient:noStatus
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +k8s:openapi-gen=true
+
+// Fact represents observed facts
```
jglick: Discuss what this means!

pkg/apis/jenkins.io/v1/types_fact.go
```diff
@@ -0,0 +1,99 @@
...
+type Fact struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
```
jglick: Should there be resource links to the owning `PipelineActivity` for GC purposes?

pkg/apis/jenkins.io/v1/types_fact.go
```diff
@@ -0,0 +1,99 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	Spec   FactSpec   `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
+	Status FactStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// FactSpec is the specification of a Fact
+type FactSpec struct {
+	Name             string        `json:"name" protobuf:"bytes,1,opt,name=name"`
+	ID               int           `json:"id" protobuf:"bytes,2,opt,name=id"`
```
jglick: I have never been clear on what either of these are actually for.

pkg/apis/jenkins.io/v1/types_fact.go
```diff
@@ -0,0 +1,99 @@
...
+	FactType         string        `json:"factType" protobuf:"bytes,3,opt,name=factType"`
+	Measurements     []Measurement `json:"measurements" protobuf:"bytes,4,opt,name=measurements"`
+	Statements       []Statement   `json:"statements" protobuf:"bytes,5,opt,name=statements"`
+	Original         Original      `json:"original,omitempty" protobuf:"bytes,6,opt,name=original"`
+	Tags             []string      `json:"tags,omitempty" protobuf:"bytes,7,opt,name=tags"`
+	SubjectReference string        `json:"subject" protobuf:"bytes,8,opt,name=source"`
+}
+
+// FactStatus is the status for an Environment resource
```
jglick: Huh?

pkg/apis/jenkins.io/v1/types_fact.go
```diff
@@ -0,0 +1,99 @@
...
+type Original struct {
+	MimeType string   `json:"mimetype,omitempty" protobuf:"bytes,1,opt,name=mimetype"`
+	URL      string   `json:"url,omitempty" protobuf:"bytes,1,opt,name=url"`
+	Tags     []string `json:"tags,omitempty" protobuf:"bytes,8,opt,name=tags"`
+}
+
+// Recommended measurements for static program analysis
+const (
+	StaticProgramAnalysisTotalClasses   = "TotalClasses"
+	StaticProgramAnalysisTotalBugs      = "TotalBugs"
```
jglick: I am uncomfortable with domain-specific constants like this being defined in `jx`. Perhaps they are broader than `jx-app-jacoco` but there is some intermediate level of specificity—a library for coverage-related apps.

In any case we need to carefully consider which sorts of facts are supposed to be consolidated between apps. If I have a project which collects both JaCoCo and output of some other tool, are they each supposed to register their own facts? Or merge them somehow? This is perhaps a stretch to imagine, but it is not at all unlikely in the case of linters etc.

pkg/apis/jenkins.io/v1/types_fact.go
jglick: @pmuir pointed me to [these docs](https://jenkins-x.io/contribute/apidocs/#openapi) suggesting

```golang
// +optional
```

fiunchinho: I can't find the difference between `optional` and `omitempty`. Shouldn't `omitempty` be enough?

pkg/apis/jenkins.io/v1/types_fact.go
fiunchinho: Agree. @pmuir @snoopdave can you provide a clear definition of the different fields present in the Fact CRD? I've read the Core3 data model, but I'm still not sure how to define them.

fiunchinho: I added the definitions found on the Core3 Data model, but I think they can be improved.

pkg/apis/jenkins.io/v1/types_fact.go
fiunchinho: I added some docs on how to link the Facts to a PipelineActivity using labels.

pkg/apis/jenkins.io/v1/types_fact.go
pmuir: `omitempty` has an impact on the marshaled JSON, `optional` only has an impact on the schema and docs.

pkg/apis/jenkins.io/v1/types_fact.go
pmuir: One for @snoopdave

pkg/apis/jenkins.io/v1/types_fact.go
pmuir: > I am uncomfortable with domain-specific constants like this being defined in `jx`. Perhaps they are broader than `jx-app-jacoco` but there is some intermediate level of specificity—a library for coverage-related apps.

Agreed

pkg/apis/jenkins.io/v1/types_fact.go
```diff
@@ -0,0 +1,113 @@
...
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	Spec   FactSpec   `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
+	Status FactStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// FactSpec is the specification of a Fact
+type FactSpec struct {
+	Name             string        `json:"name" protobuf:"bytes,1,opt,name=name"`
```
pmuir: There should be a `ResourceReference` (see https://github.com/jenkins-x/jx/blob/master/pkg/apis/jenkins.io/v1/register.go#L76) that points to e.g. the `PipelineActivity` - this is the standard format used e.g. by Kubernetes ownership references. You should also consider adding an ownershop

fiunchinho: Would it work making `SubjectReference` field be a ResourceReference instead of string?

pkg/apis/jenkins.io/v1/types_fact.go
fiunchinho: is the jx-app-jacoco app (or any other) currently using these constants? Should we remove it for now and think where to place them?

pkg/apis/jenkins.io/v1/types_fact.go
pmuir: yes that makes sense - what does `Subject` mean though?

fiunchinho: The thing that the Facts refer to. I can rename it to ResourceReference if you guys think that makes more sense.

pkg/apis/jenkins.io/v1/types_fact.go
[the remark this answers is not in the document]
snoopdave: yes, there should be an enumeration of measurement types defined, at least: count, percent, boolean

pkg/apis/jenkins.io/v1/types_fact.go
snoopdave: from the original design a Measurement is different from a Statement. A a measurement is a percentage or a count; something measured. A Statement represents a decision that was made, for example that a promition was approved or denied.

pkg/apis/jenkins.io/v1/types_fact.go
snoopdave: The naming is not great. In the original design, we talked about being able to declare the types of facts the the system accepts. So, you might have one type of resource that defines all of the FactTypes allowed by the system, and their types. Then you would have a separate kind of resource that represents the actual facts created by a pipeline.  

In the current implementation, Fact and Fact type seem are merged together into one thing. I'm not sure that is what we want.  

/cc @fiunchinho

fiunchinho: It's https://github.com/cloudbees/jx-apps-manifest the type of facts that the system accepts?

pkg/apis/jenkins.io/v1/types_fact.go
fiunchinho: I'll update the doc in the CRD with the following comments:

```
// Measurement is a percentage or a count, something measured that the system will capture within a fact
```

and

```
// Statement represents a decision that was made, for example that a promotion was approved or denied
```

Btw @snoopdave , what about `Original`? Should we change it's structure? @jglick was saying that current implementation is not really compatible with having several reports.

pkg/apis/jenkins.io/v1/types_fact.go
snoopdave: yes, I forgot about that work.

pkg/apis/jenkins.io/v1/types_fact.go
pmuir: I think `jx-app-jacoco` uses them.

pkg/apis/jenkins.io/v1/types_fact.go
pmuir: No it kinda makes sense to me :-)

fiunchinho:
We've decided to split this in two:
- First extract the CRD as it's
- Improve the CRD applying our current knowledge


I believe the work for extraction is ready. Anything else that you folks think is required before merging?

pkg/apis/jenkins.io/v1/types_fact.go
pmuir: > We've decided to split this in two:
> 
> * First extract the CRD as it's
> * Improve the CRD applying our current knowledge
> 
> I believe the work for extraction is ready. Anything else that you folks think is required before merging?

Yes, the `ResourceReference` shouldn't be a string ;-)

pmuir:
The `ResourceReference` should be a resource reference, not a string.

PR #3070, opened 2019-02-12
[wip] feat: use existing values when upgrading apps

pmuir:
#### Submitter checklist

#### Description
Also adds support for batch mode for installing apps

fixes #


hferentschik:
Would be nice to have an issue as well which might give a bit more context. Hard to know otherwise what this is about.

pkg/extensions/apps_values.go
```diff
@@ -0,0 +1,164 @@
...
+data:
+  {{ .key }}: {{ .value }}
+kind: Secret
+metadata:
+  name: {{ .name }} 
+type: Opaque
+{{- end }}
+`
+
+func StashValues(values []byte, name string, jxClient versioned.Interface, ns string) error {
```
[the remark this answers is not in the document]
hferentschik: What's our take on godoc?

pkg/extensions/apps_values.go
```diff
@@ -0,0 +1,167 @@
+package extensions
```
hferentschik: package "extension"? Is this a leftover from the fact that 'app' were once extensions?

pkg/jx/cmd/add_app_test.go
```diff
@@ -238,6 +238,134 @@ func TestAddAppWithSecrets(t *testing.T) {
...
+	err := o.Run()
+	assert.NoError(t, err)
+	err = console.Close()
+	<-donec
+	assert.NoError(t, err)
+	t.Logf(expect.StripTrailingEmptyLines(console.CurrentState()))
+
+	// Validate that the secret reference is generated and the secret is in the chart
+	// chart, _, _, _, _, _, _, _, _, valueFiles, _, _, _ :=
+	testOptions.MockHelmer.VerifyWasCalledOnce().
```
hferentschik: So this test is testing the actual CLI command up to the Helm abstraction which is in this case mocked, right?

pkg/jx/cmd/common_environments.go
```diff
@@ -324,7 +324,7 @@ func (o *CommonOptions) CreateAddRequirementFn(chartName string, alias string, v
 				err = util.CopyFile(valuesFiles[0], rootValuesFileName)
 				if err != nil {
 					return errors.Wrapf(err, "cannot copy values."+
-						"yaml to chartName directory %s", appDir)
+						"yaml to %s directory %s", chartName, appDir)
```
hferentschik: Why do you use string concatenation '+' and string interpolation here? I'd remove the '+'

pkg/surveyutils/jsonschema.go
```diff
@@ -279,15 +299,16 @@ func (o *JSONSchemaOptions) recurse(name string, prefixes []string, requiredFiel
...
+			ignoreMissingValues, in, out, outErr)
 		if err != nil {
 			return err
 		}
 	case "number":
 		validators := additionalValidators
 		validators = append(validators, FloatValidator())
-		err := o.handleBasicProperty(name, prefixes, numberValidator(required, validators, t), t, output, in, out,
-			outErr)
+		err := o.handleBasicProperty(name, prefixes, numberValidator(required, validators, t), t, output, existingValues, askExisting,
```
hferentschik: that's a heck of a parameter list. It seems you also do this at several places. Would it not make sense to introduce a parameter struct?

pkg/surveyutils/test_data/acceptExisting.test.schema.json
```diff
@@ -0,0 +1,14 @@
...
+  "description": "test values.yaml",
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string",
+      "title": "What is your name?",
+      "description": "Enter your name"
+    }
+  }
+}
```
hferentschik: newlines missing

pkg/util/types.go
```diff
@@ -68,6 +68,21 @@ func AsSliceOfStrings(unk interface{}) ([]string, error) {
 	return result, nil
 }
 
+//AsMapOfStringsIntefaces attempts to convert unk to a map[string]interface{}
+func AsMapOfStringsIntefaces(unk interface{}) (map[string]interface{}, error) {
+	v := reflect.ValueOf(unk)
+	v = reflect.Indirect(v)
+
+	if v.Kind() != reflect.Map {
+		return make(map[string]interface{}), fmt.Errorf("cannot convert %v (%v) to map[string]interface{}", v.Type(), v)
```
hferentschik: not returning nil? If one forgets to check the error and continues there is no difference to unk being an empty map of string to interface.

pkg/jx/cmd/add_app_test.go
pmuir: Yes.

pkg/extensions/apps_values.go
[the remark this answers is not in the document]
pmuir: I need to add that still

pkg/jx/cmd/common_environments.go
pmuir: That's probably the stupid string wrapper junk in Goland ;-)

PR #3106, opened 2019-02-14
feat(create task) lets checkout using a revision

rawlingsj:


abayer:
Ooooo, excellent change.

/lgtm

PR #3120, opened 2019-02-15
feat: Add PipelineStructure as CRD/API/etc

abayer:
#### Submitter checklist

#### Description
Also add Depth and Parent to PipelineStructureStage so that we can build out runtime mapping from pods to stages. There will be followups to this. =)

cc @dwnusbaum 

n/a


pkg/apis/jenkins.io/v1/types_pipelinestructure.go
```diff
@@ -0,0 +1,665 @@
...
+
+// MangleToRfc1035Label - Task/PipelineStructureStep names need to be RFC 1035/1123 compliant DNS labels, so we mangle
+// them to make them compliant. Results should match the following regex and be
+// no more than 63 characters long:
+//     [a-z]([-a-z0-9]*[a-z0-9])?
+// cf. https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+// body is assumed to have at least one ASCII letter.
+// suffix is assumed to be alphanumeric and non-empty.
+// TODO: Combine with kube.ToValidName (that function needs to handle lengths)
+func MangleToRfc1035Label(body string, suffix string) string {
```
dwnusbaum: nit: Now that this code is in a different package, I think we can import `github.com/jenkins-x/jx/pkg/kube` without cycles and replace uses of this method with `kube.ToValidNameTruncated(foo, 63-len(suffix)-1)+"-"+suffix`.

pkg/apis/jenkins.io/v1/types_pipelinestructure.go
```diff
@@ -0,0 +1,665 @@
...
+	// +optional
+	Steps []PipelineStructureStep `json:"steps,omitempty" protobuf:"bytes,5,opt,name=steps"`
+	// +optional
+	Stages []*PipelineStructureStage `json:"stages,omitempty" protobuf:"bytes,6,opt,name=stages"`
+	// +optional
+	Parallel []*PipelineStructureStage `json:"parallel,omitempty" protobuf:"bytes,7,opt,name=parallel"`
+	// +optional
+	Post []PipelineStructurePost `json:"post,omitempty" protobuf:"bytes,8,opt,name=post"`
+	// +optional
+	Parent *string `json:"parent,omitempty" protobuf:"bytes,9,opt,name=parent"`
```
dwnusbaum: nit : I would drop this and depth from this struct and handle all of the graph-related stuff with a separate `pipelinestructure_graph_traversal` file or similar for clear separation of concerns. We'd need pointers to the actual `PipelineStructureStages` in this struct itself to avoid having to construct a separate graph anyway, so it seems cleanest for this struct to map very closely to the YAML format itself. We can always cache the graph with its related PipelineStructure in whatever controller cares about these objects so we don't have to keep recreating it if performance is a concern. These are the kind of data structures I think we'd need for graph-related stuff:

```
Node {
    Stage           *PipelineStructureStage
    Depth           int8
    Children        []*Node
    Parent          *Node
    PreviousSibling *Node
    NextSibling     *Node
}

Graph {
    FirstStage *Node // maybe not?
    Nodes      map[*PipelineStructureStage]*Node
}
```

pkg/kpipelines/pipelines.go
```diff
@@ -214,3 +222,554 @@ func PipelineResourceName(gitInfo *gits.GitRepository, branch string, context st
 	resourceName := kube.ToValidNameTruncated(dirtyName, 35)
 	return resourceName
 }
+
+var randReader = rand.Reader
+
+func scopedEnv(newEnv []v1.PipelineStructureEnvVar, parentEnv []corev1.EnvVar) []corev1.EnvVar {
```
dwnusbaum: Any particular reason to move this stuff from its previous location? I don't feel too strongly about it, but I liked that `syntax/pipeline.go` had nothing to do with client-go related code or updating CRDs in the cluster or anything, whereas now this file now has both the pure translation logic and non-pure CRD creation/update logic (although they don't interact). `step_create_task.go` also has some translation logic as well, so we don't really have a clear separation between the files in general. Not really something to fix now but would be good to be thinking about where we can draw boundaries to keep the code easy to maintain.

pkg/apis/jenkins.io/v1/types_pipelinestructure.go
abayer: Yeah, actually, we don't need `Depth` or `Parent`, but I also don't think we need anything special for graph traversal - given a stage name and a `PipelineStructure`, we can easily traverse it to find the parent/etc. I don't know if we'll ever actually have a need for full-blown graph traversal along the lines of what we have for the `FlowGraph` in Jenkins Pipeline - here we just need a way to look at a `Task` (well, technically to look at a `Pod` to get its `TaskRun` to get its `Task` and `PipelineRun` and `Pipeline`, but I'm gonna shortcut here), get its corresponding stage name, and find where it is in the `PipelineStructure` so that we can populate `PipelineActivity` correctly.

pkg/kpipelines/pipelines.go
abayer: Separation of concerns - `types_pipelinestructure.go` is just the `PipelineStructure` CRD, its validation, and methods on its types, while `pipelines.go` has all the translation. It felt silly to have a separate package/file for just the translation when we already had this package/file here we could move it into. And we will be rewriting the non-pure-translation stuff soon to go build pack->`PipelineStructure`->`Pipeline` et al, so splitting that up just for now doesn't make a lot of sense to me.

pkg/apis/jenkins.io/v1/types_pipelinestructure.go
abayer: Will do.

abayer: Amusingly, we're only using the suffix bit of this in tests. =)

dwnusbaum: Yeah, I'm still thinking about how exactly we want to deal with this, especially given that Tektōn will truncate our names and add its own suffix in some cases...

abayer:
/assign pmuir 

@pmuir, I had to rework `kube.FixSchema` to resolve against the root document to handle recursive references. Feel free to shoot this down. =)

abayer:
Well, bugger, I'm getting
```
 Failed to retrieve team settings: failed to setup the dev environment for namespace 'jx': the server could not find the requested resource (post environments.jenkins.io) - falling back to default settings...
Using helmBinary helm with feature flag: none
Context "gke_abayer-jx-experiment_us-west1-b_abayer-test-cluster" modified.
error creating cluster configuring the team settings in the dev environment: updating the team setttings in the dev environment: failed to setup the dev environment for namespace 'jx': the server could not find the requested resource (post environments.jenkins.io)
error: configuring the team settings in the dev environment: updating the team setttings in the dev environment: failed to setup the dev environment for namespace 'jx': the server could not find the requested resource (post environments.jenkins.io)
```
spinning up a cluster now, so something's wrong beyond what I could test in `pkg/kube/crds_test.go`. Sigh. I'll come back to this over the weekend.

abayer:
Closed in favor of #3129

PR #3129, opened 2019-02-17
feat(kpipelines): Add PipelineStructure CRD

abayer:
#### Submitter checklist

#### Description

This will allow us to record the syntax-based structure of a pipeline,
including nested and parallel stages, so that we can then map from a
running Pod to the stage (and possible parent stages) it belongs
to. This can then be used to populate PipelineActivity properly, and
to get logs correctly.

Note that this mapping can't actually be implemented until we move to
a current version of Build Pipeline in the chart we install - once
that's done, I'll follow this up with a change to actually perform
that mapping.

Also added helper methods for creating ParsedPipeline and
PipelineStructure in tests. Existing tests are only partially
converted currently - that'll be updated shortly.

(also, this replaces my failed attempt in this direction over in #3120)

As mentioned above, this is just laying the groundwork for future functionality.

n/a


dwnusbaum:
(Approval is assuming we figure out the failing tests before merging, not sure if there is a link that I am not seeing or if we have to build locally to figure out what tests are failing)

jstrachan:
@abayer we had a flaky test we've just disabled - wonder if you could rebase + push again?

abayer:
@jstrachan Most recent push was rebased and is current now. Just waiting for the previous test run to finish so the new one can start. =)

abayer:
/assign @rawlingsj 
'cos hey, why not. =)

abayer:
/retest
These tests, they do like to hang.

abayer:
Ha, actually, I shoulda rebased on `master` after #3142 anyway.

PR #3153, opened 2019-02-19
fix(kipelines): PipelineStructure Name should be unique for each build

abayer:
#### Submitter checklist

#### Description

I was using the Pipeline name, which is all we have at generation
time, but is...not so good when it comes to actually creating it in
the cluster, since then the PipelineStructure name is non-unique. So
update it to use the PipelineRun name before we create it.

n/a


abayer:
/test this

...I'm pretty sure this is flaky/unrelated to the change, but not 100% convinced yet.

Issue #1016, opened 2018-06-12
Refactor the cmd package by grouping the related commands in sub-packages


ccojocar:
It would be nice to organise the related commands in sub-packages, and maybe move out some general purpose packages such as log.

garethjevans:
I was also wondering if we could make the commands a little more testable.  The runCommand* methods are a little hard to mock out

garethjevans:
Something like https://npf.io/2015/06/testing-exec-command/

jstrachan:
that sounds good!

rawlingsj:
@ccojocar with your recent refactoring is this something that's more achievable?

/kind enhancement 
/priority important-longterm

ccojocar:
@rawlingsj I think we have done a fair amount of refactoring moving out the common code in separate packages. I would say that now it would be possible to start splitting the commands in separate packages.

pmuir:
Why did you close this @ccojocar it's still a big mess.

ccojocar:
We can keep it open, but we can also follow up with small issue.

ccojocar:
A few refactorings go merged. They are required to start splitting the clients factory and common in separate packages from cmd.

ccojocar:
Next step is to extract the clients factory in a separate package, and use it only as a private field in the `CommonOptions`. The goal is to avoid using it directly in the `cmd` package.

ccojocar:
Next step at this stage is to move the `CommonOptions` in a separate package. Finally, we can group the commands in separate packages after this is done.

hferentschik:
Heads up, I am working on a pull request to restructure the commands into subpackages. I'll open a follow-up issue I guess.

PR #3178, opened 2019-02-20
refactor: extract the cmd clients factory in a separate package and set up the CommonOptions before creating the commands

ccojocar:
#### Submitter checklist

#### Description

- Extract the cmd clients factory in a separate package and set up the CommonOptions before creating the commands
- Make the common flags persistent for all commands
- Remove some short flags names which were conflicting with the short name of batch flag
- Update the description of common flags
- Move the PullSecrets option from common commands into the specific commands (create env and devpod) 

part of #1016

fixes #


pkg/jx/cmd/create_env_test.go
```diff
@@ -133,15 +133,15 @@ func TestCreateEnvRun(t *testing.T) {
 	helmValuesConfig.ExposeController.Config.HTTP = "false"
 	helmValuesConfig.ExposeController.Config.TLSAcme = "false"
 
+	commonOpts := cmd.NewCommonOptionsWithFactory(factory)
+	commonOpts.In = console.In
+	commonOpts.Out = console.Out
+	commonOpts.Err = console.Err
```
pmuir: I wonder if we should do a helper func for this as it's quiet a common pattern in tests - we can add that later!

ccojocar: Agree. Let's add it as a follow up.

PR #3187, opened 2019-02-21
refactor CLI for tekton

jstrachan:


pkg/prow/prow.go
```diff
@@ -24,9 +24,9 @@ import (
 const (
 	Hook = "hook"
 
-	KnativeBuildAgent         = "knative-build"
-	KnativeBuildPipelineAgent = "knative-pipeline-run"
-	KubernetesAgent           = "kubernetes"
+	KnativeBuildAgent = "knative-build"
+	TektonAgent       = "takton"
```
abayer: ...`takton`? =)

jstrachan: fixed thanks!

PR #3230, opened 2019-02-26
think env var need to reference the param from the pipelinerun not the input params

rawlingsj:


abayer:
No, this isn't right - the `${params.whatever}` reference is used when you're referencing a parameter defined in the Pipeline CRD *within* the Pipeline CRD - but the env vars here are going to be used in the Task CRD, where you would need to refer to it as `${input.params.whatever}`. I know, confusing.

rawlingsj:
ok - I was getting
```
error: failed to generate Task for build pack pipeline YAML: /Users/jr/go/src/github.com/jenkins-x/jx/jenkins-x.yml: failed to apply generated Pipeline: failed to create/update the task jenkins-x-jx-tekton1-ci in namespace jx: failed to get PipelineResource jenkins-x-jx-tekton1-ci with tasks.tekton.dev "jenkins-x-jx-tekton1-ci" not found after failing to create a new one: Internal error occurred: admission webhook "webhook.tekton.dev" denied the request: mutation failed: non-existent variable in "${inputs.params.version}" for step env[VERSION]: taskspec.steps.env[VERSION]
````
without this

abayer:
I could be wrong - I'm in a customer meeting so can't check the code right now. =) If this works, then go for it.

abayer:
@rawlingsj Can you post/link to the `task-whatever.yaml` you get when you run `jx step create task --output=. --no-apply`?

rawlingsj:
@abayer think @jstrachan has a fix on the way which you should see shortly.  Thanks for the spot!

abayer:
cool cool cool. =)

PR #3231, opened 2019-02-26
fix for Task.spec.inputs.params when using NGP

jstrachan:


pkg/jx/cmd/clients/factory.go
```diff
@@ -382,13 +383,45 @@ func (f *factory) ResetSecretsLocation() {
...
 }
 
+func (f *factory) getVaultName(namespace string) (string, error) {
+	name := ""
+	context, err := cluster.Context(f.kubeConfig)
+	if err != nil {
+		return name, err
+	}
+	if context == nil {
+		kubeClient, _, err := f.CreateKubeClient()
```
ccojocar: Is it the client created successfully if the context is nil?

pkg/jx/cmd/clients/factory.go
```diff
@@ -382,13 +383,45 @@ func (f *factory) ResetSecretsLocation() {
...
+		if err != nil {
+			return name, err
+		}
+		cm, err := kube.GetConfigMap(kubeClient, namespace, kube.ConfigMapNameJXInstallConfig)
```
ccojocar: There is a helper function for this `kube.ReadInstallValues`.

pkg/jx/cmd/clients/factory.go
```diff
@@ -382,13 +383,45 @@ func (f *factory) ResetSecretsLocation() {
...
+		if err != nil {
+			return name, errors.Wrapf(err, "no ConfigMap %s in namespace %s", kube.ConfigMapNameJXInstallConfig, namespace)
+		}
+		if cm.Data == nil {
+			return name, fmt.Errorf("no data for ConfigMap %s in namespace %s", kube.ConfigMapNameJXInstallConfig, namespace)
+		}
+		name = cm.Data["clusterName"]
```
ccojocar: This const is defined in `kube.ClusterName`.

jstrachan: lemme use that

pkg/jx/cmd/clients/factory.go
jstrachan: if there's no config then we can still create a k8s client yeah: https://github.com/jenkins-x/jx/blob/976b6e8d676ed1c6bfda05db85b3727876769c8a/pkg/jx/cmd/clients/factory.go#L617

PR #3235, opened 2019-02-27
feat: Add apps metadata during gitops install

daveconde:
#### Submitter checklist

#### Description
This PR enhances app metadata to include app specific metadata when an app is added to a gitops environment. If an existing App resource exists in an app it is updated, otherwise a new App resource is added to the gitops environment repo along with the app. 

App templates that are in the environment repo are then applied to the app chart during `jx step env apply`


pkg/apps/gitops.go
```diff
@@ -30,7 +30,7 @@ func (o *GitOpsOptions) AddApp(app string, dir string, version string, repositor
 	options := environments.EnvironmentPullRequestOptions{
 		ConfigGitFn: o.ConfigureGitFn,
 		Gitter:      o.Gitter,
-		ModifyChartFn: environments.CreateAddRequirementFn(app, alias, version,
+		ModifyChartFn: environments.CreateAddRequirementFn(o.Helmer, app, alias, version,
```
pmuir: I try to put things like Helmer at the end of the arg list - so that we pass 'variables' first, then interfaces.

pkg/environments/gitops.go
```diff
@@ -371,7 +374,7 @@ func CreateUpgradeRequirementsFn(all bool, chartName string, alias string, versi
 // valuesFiles (an array of paths to values.yaml files to add). The chartDir is the unpacked chart being added,
 // which is used to add extra metadata about the chart (e.g. the charts readme, the release.yaml, the git repo url and
 // the release notes) - if this points to a non-existant directory it will be ignored.
-func CreateAddRequirementFn(chartName string, alias string, version string, repo string,
+func CreateAddRequirementFn(helmer helm.Helmer, chartName string, alias string, version string, repo string,
```
pmuir: Likewise, try to put helmer at the end.

pkg/environments/gitops.go
```diff
@@ -601,3 +608,95 @@ func CreateNestedRequirementDir(dir string, requirementName string, requirementD
...
+func UpdateAppResource(helmer helm.Helmer, fetchedChartDir string, outputDir string, name string, repository string) error {
+	outputTemplateDir := filepath.Join(outputDir, "templates")
+	templatesDirExists, err := util.DirExists(outputTemplateDir)
+	if err != nil {
+		return err
+	}
+	if !templatesDirExists {
+		os.Mkdir(outputTemplateDir, os.ModePerm)
+	}
+	helmClient, foundHelmCLI := helmer.(*helm.HelmCLI)
```
pmuir: This seems really wrong as Helmer is supposed to hide these details. I would suggest modifying `Helmer` to add the function you need.

pkg/jx/cmd/add_app_test.go
```diff
@@ -79,7 +79,7 @@ func TestAddAppForGitOps(t *testing.T) {
 	// Validate the branch name
 	envDir, err := o.CommonOptions.EnvironmentsDir()
 	assert.NoError(t, err)
-	devEnvDir := filepath.Join(envDir, testOptions.OrgName, testOptions.DevEnvRepoInfo.Name)
```
pmuir: We should probably add a test for templates...

PR #3240, opened 2019-02-27
fix: Don't keep trying to get pods for completed PipelineRuns

abayer:
#### Submitter checklist

#### Description

When a PipelineRun has finished and its pods have been GC'd, we should
just ignore it for the purposes of `PipelineActivity` and `jx get
build logs`.

n/a


abayer:
Grr, my pre commit hook to format didn’t work!

PR #3245, opened 2019-02-28
tektonify jx CI and add next gen pipeline :boom:

rawlingsj:
/hold


abayer:
We may want to wait for https://github.com/knative/build-pipeline/pull/565 before we do this for realzies - since it makes long task names result in infinite numbers of `TaskRun`s getting created. =)

rawlingsj:
Sounds good to me. Let’s wait for that.

PR #3257, opened 2019-03-01
feat: Add PipelineConfig.Env for global environment variables

abayer:
#### Submitter checklist

#### Description

Works in both from-yaml and from-build-pack paths of step_create_task

Yeah, needs tests, but that'll come when we rewrite `step_create_task.go` to unify the paths anyway. =)

n/a


abayer:
grr, legit test failures - fixing.

PR #3258, opened 2019-03-01
fixes for tekton build packs + BDD tests

jstrachan:


pkg/jx/cmd/get_build_logs.go
```diff
@@ -240,22 +240,21 @@ func (o *GetBuildLogsOptions) getProwBuildLog(kubeClient kubernetes.Interface, t
...
-	}
-
 	args := o.Args
 	pickedPipeline := false
 	if len(args) == 0 {
 		if o.BatchMode {
 			return util.MissingArgument("pipeline")
 		}
+		var err error
+		if tektonEnabled {
```
abayer: Good catch!

PR #3274, opened 2019-03-04
fix: support forking on environment repos if user different to owner

jstrachan:


ccojocar:
@jstrachan Integration tests seems to be failing.

PR #3277, opened 2019-03-05
Support creating vault on AWS

msvticket:
#### Submitter checklist

#### Description

Adding support for vault on AWS with DynamoDB as storage.

This is untested work in progress that is submitted to get feedback.

### DIFF
--- a/pkg/jx/cmd/create_vault.go
+++ b/pkg/jx/cmd/create_vault.go
@@ -2,6 +2,9 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/jenkins-x/jx/pkg/cloud"
+	"github.com/jenkins-x/jx/pkg/kube/serviceaccount"
+	"k8s.io/client-go/kubernetes"
 	"time"
 
 	"github.com/banzaicloud/bank-vaults/operator/pkg/client/clientset/versioned"
@@ -11,7 +14,9 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/jenkins-x/jx/pkg/cloud/gke"
+	"github.com/jenkins-x/jx/pkg/cloud/amazon"
 	gkevault "github.com/jenkins-x/jx/pkg/cloud/gke/vault"
+	awsvault "github.com/jenkins-x/jx/pkg/cloud/amazon/vault"
 	"github.com/jenkins-x/jx/pkg/jx/cmd/templates"
 	"github.com/jenkins-x/jx/pkg/kube"
 	kubevault "github.com/jenkins-x/jx/pkg/kube/vault"
@@ -22,13 +27,14 @@ import (
 )
 
 const (
-	gkeKubeProvider  = "gke"
 	exposedVaultPort = "8200"
 )
 
 var (
 	createVaultLong = templates.LongDesc(`
 		Creates a Vault using the vault-operator
+
+        The necessary flags depends on the provider of the kubernetes cluster. 
 `)
 
 	createVaultExample = templates.Examples(`
@@ -44,12 +50,18 @@ var (
 type CreateVaultOptions struct {
 	CreateOptions
 
-	GKEProjectID      string
-	GKEZone           string
+	GKECreateVaultOptions
+	kubevault.AWSConfig
 	Namespace         string
 	SecretsPathPrefix string
 
 	IngressConfig kube.IngressConfig
+}
+
+// GKECreateVaultOptions the options for vault on GKE
+type GKECreateVaultOptions struct {
+	GKEProjectID      string
+	GKEZone           string
 }
 
 // NewCmdCreateVault  creates a command object for the "create" command
@@ -74,12 +86,29 @@ func NewCmdCreateVault(commonOpts *CommonOptions) *cobra.Command {
 		},
 	}
 
+	// GKE flags
 	cmd.Flags().StringVarP(&options.GKEProjectID, "gke-project-id", "", "", "Google Project ID to use for Vault backend")
 	cmd.Flags().StringVarP(&options.GKEZone, "gke-zone", "", "", "The zone (e.g. us-central1-a) where Vault will store the encrypted data")
+
+	awsCreateVaultOptions(cmd, &options.AWSConfig)
+
 	cmd.Flags().StringVarP(&options.Namespace, "namespace", "n", "", "Namespace where the Vault is created")
 	cmd.Flags().StringVarP(&options.SecretsPathPrefix, "secrets-path-prefix", "p", vault.DefaultSecretsPathPrefix, "Path prefix for secrets used for access control config")
 
 	return cmd
+}
+
+func awsCreateVaultOptions(cmd *cobra.Command, options *kubevault.AWSConfig) {
+	// AWS flags
+	cmd.Flags().StringVarP(&options.DynamoDBRegion, "aws-dynamodb-region", "", "", "The region to use for storing values in AWS DynamoDB")
+	cmd.Flags().StringVarP(&options.DynamoDBTable, "aws-dynamodb-table", "", "vault-data", "The table in AWS DynamoDB to use for storing values")
+	cmd.Flags().StringVarP(&options.KMSRegion, "aws-kms-region", "", "", "The region of the AWS KMS key to encrypt values")
+	cmd.Flags().StringVarP(&options.KMSKeyID, "aws-kms-key-id", "", "", "The ID or ARN of the AWS KMS key to encrypt values")
+	cmd.Flags().StringVarP(&options.S3Bucket, "aws-s3-bucket", "", "", "The name of the AWS S3 bucket to store values in")
+	cmd.Flags().StringVarP(&options.S3Prefix, "aws-s3-prefix", "", "vault-operator", "The prefix to use for storing values in AWS S3")
+	cmd.Flags().StringVarP(&options.S3Region, "aws-s3-region", "", "", "The region to use for storing values in AWS S3")
+	cmd.Flags().StringVarP(&options.AccessKeyID, "aws-access-key-id", "", "", "Access key id of service account to be used by vault")
+	cmd.Flags().StringVarP(&options.SecretAccessKey, "aws-secret-access-key", "", "", "Secret access key of service account to be used by vault")
 }
 
 // Run implements the command
@@ -100,14 +129,10 @@ func (o *CreateVaultOptions) Run() error {
 		return errors.Wrap(err, "retrieving the team settings")
 	}
 
-	if teamSettings.KubeProvider != gkeKubeProvider {
-		return errors.Wrapf(err, "this command only supports the '%s' kubernetes provider", gkeKubeProvider)
+	if teamSettings.KubeProvider != cloud.GKE && teamSettings.KubeProvider != cloud.AWS && teamSettings.KubeProvider != cloud.EKS {
+		return errors.Wrapf(err, "this command only supports the '%s' kubernetes provider", cloud.GKE)
 	}
 
-	return o.createVaultGKE(vaultName)
-}
-
-func (o *CreateVaultOptions) createVaultGKE(vaultName string) error {
 	kubeClient, team, err := o.KubeClientAndNamespace()
 	if err != nil {
 		return errors.Wrap(err, "creating kubernetes client")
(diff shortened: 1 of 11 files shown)


pkg/jx/cmd/create_vault.go
```diff
@@ -127,23 +145,53 @@ func (o *CreateVaultOptions) createVaultGKE(vaultName string) error {
...
 	}
 
-	err = gke.Login("", true)
+	kubeClient, _, err := o.KubeClientAndNamespace()
+	if err != nil {
+		return err
+	}
+	clusterName, err := cluster.ShortName(o.Kube())
+	if err != nil {
+		return err
```
ccojocar: Please wrap this error

pkg/jx/cmd/create_vault.go
```diff
@@ -127,23 +145,53 @@ func (o *CreateVaultOptions) createVaultGKE(vaultName string) error {
...
+	}
+	log.Infof("Created service account %s for Vault authentication\n", util.ColorInfo(vaultAuthServiceAccount))
+	if kubeProvider == cloud.GKE {
+		err = o.createVaultGKE(vaultOperatorClient, vaultName, kubeClient, clusterName, vaultAuthServiceAccount)
+	}
+	if kubeProvider == cloud.AWS || kubeProvider == cloud.EKS {
+		err = o.createVaultAWS(vaultOperatorClient, vaultName, kubeClient, clusterName, vaultAuthServiceAccount)
+	}
+	if err != nil {
+		return err
```
ccojocar: wrap error

pkg/jx/cmd/create_vault.go
```diff
@@ -232,20 +269,45 @@ func (o *CreateVaultOptions) createVault(vaultOperatorClient versioned.Interface
...
+	if o.S3Bucket == "" {
+		return fmt.Errorf("Missing S3 bucket flag")
+	}
+	if o.KMSKeyID == "" {
+		return fmt.Errorf("Missing AWS KMS key id flag")
+	}
+	if o.DynamoDBRegion == "" || o.KMSRegion == "" || o.S3Region == "" {
+		defaultRegion, err := amazon.ResolveRegionWithoutOptions()
+		if err != nil {
+			return err
```
ccojocar: wrap err

pkg/kube/vault/vault.go
```diff
@@ -57,6 +60,25 @@ type GCSConfig struct {
 	HaEnabled string `json:"ha_enabled"`
 }
 
+// AWSConfig keeps the vault configuration for AWS
+type AWSConfig struct {
```
ccojocar: Please define the tags fro these fields.

pkg/kube/vault/vault.go
```diff
@@ -103,6 +125,7 @@ type Telemetry struct {
 // Storage configuration for Vault storage
 type Storage struct {
 	GCS GCSConfig `json:"gcs"`
+	DynamoDB DynamoDBConfig `json:"dynamodb"`
```
ccojocar: I think you need to add `omitempty` to use these configs independently.

pkg/kube/vault/vault.go
```diff
@@ -121,19 +144,90 @@ func SystemVaultNameForCluster(clusterName string) string {
...
-// CreateVault creates a new vault backed by GCP KMS and storage
-func CreateVault(kubeClient kubernetes.Interface, vaultOperatorClient versioned.Interface, name string, ns string,
+// CreateGKEVault creates a new vault backed by GCP KMS and storage
+func CreateGKEVault(kubeClient kubernetes.Interface, vaultOperatorClient versioned.Interface, name string, ns string,
 	gcpServiceAccountSecretName string, gcpConfig *GCPConfig, authServiceAccount string,
 	authServiceAccountNamespace string, secretsPathPrefix string) error {
 
+	vault, err := InitializeVault(kubeClient, name, ns, authServiceAccount, authServiceAccountNamespace, secretsPathPrefix)
+	if err != nil {
+		return err
```
ccojocar: wrap err

pkg/kube/vault/vault.go
```diff
@@ -121,19 +144,90 @@ func SystemVaultNameForCluster(clusterName string) string {
...
+	return err
+}
+
+// CreateAWSVault creates a new vault backed by AWS KMS and DynamoDB storage
+func CreateAWSVault(kubeClient kubernetes.Interface, vaultOperatorClient versioned.Interface, name string, ns string,
+	awsConfig *AWSConfig, authServiceAccount string, authServiceAccountNamespace string, secretsPathPrefix string) error {
+
+	vault, err := InitializeVault(kubeClient, name, ns, authServiceAccount, authServiceAccountNamespace, secretsPathPrefix)
+	if err != nil {
+		return err
```
ccojocar: wrap err

pkg/kube/vault/vault_test.go
```diff
@@ -43,7 +43,7 @@ func TestCreateVault(t *testing.T) {
 
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
-			err := vault.CreateVault(client, vaultclient, tc.name, tc.namespace, tc.gcpSecretName,
+			err := vault.CreateGKEVault(client, vaultclient, tc.name, tc.namespace, tc.gcpSecretName,
```
ccojocar: It would be nice to have a test also for AWS.

pkg/jx/cmd/create_vault.go
msvticket: I don't think it's needed since ShortName returns the error wrapped

pkg/kube/vault/vault.go
msvticket: What tags? The equivalent type for Google Cloud (`GCPConfig`) doesn't have any tags for the fields.

pkg/kube/vault/vault.go
msvticket: Yes, but I think I also need to make GCS and DynamoDB pointers:
https://www.sohamkamani.com/blog/golang/2018-07-19-golang-omitempty/#values-that-cannot-be-omitted

Now I remember: That was what I was supposed to fix this morning before creating the pull request ;)

pkg/jx/cmd/create_vault.go
msvticket: OK

pkg/jx/cmd/create_vault.go
msvticket: OK

pkg/kube/vault/vault.go
msvticket: Is it really needed when all errors returned from InitializeVault are wrapped?

pkg/kube/vault/vault.go
msvticket: Is it really needed when all errors returned from InitializeVault are wrapped?

pkg/kube/vault/vault_test.go
msvticket: I'll see what I can do. I'll first push the above fixes.

msvticket:
I have never used GKE, so unfortunately I doubt I'll be able to test that. I will certainly test on EKS and probably on cluster created on AWS with kops as well.

pkg/kube/vault/vault_test.go
msvticket: I have now pushed an equivalent test for AWS

msvticket:
I realised that giving aws-service-account-secret as an option doesn't work that well, since the namsespace to put the secret is created by `jx install`. I'm now thinking of adding the options aws_access_key_id and aws_secret_access_key and then let `jx` create the secret.

msvticket:
I have tested `jx install --no-tiller --gitops --provider=eks --vault ...`
Vault seem to get started. But in the step

> Applying generated chart

it failed. Didn't seem related to vault though.

```
error: error when retrieving current configuration of:
Resource: "/v1, Resource=configmaps", GroupVersionKind: "/v1, Kind=ConfigMap"
Name: "", Namespace: "jx"
Object: &{map["apiVersion":"v1" "data":map["urltemplate":"" "domain":"35.164.185.29.nip.io" "email":"" "exposer":"Ingress" "issuer":"" "tls":"false"] "kind":"ConfigMap" "metadata":map["annotations":map["jenkins.io/chart":"env" "kubectl.kubernetes.io/last-applied-configuration":""] "creationTimestamp":<nil> "labels":map["jenkins.io/chart-release":"jenkins-x" "jenkins.io/version":"0.0.1"] "namespace":"jx"]]}
from server for: "/var/folders/5d/j7hj422s4j1c7g9zl9f913mh0000gn/T/helm-template-workdir-827396897/jenkins-x/output/env/templates/ingress-config-configmap.yaml": resource name may not be empty'
```

msvticket:
Aha,  the error seemed to be due to an existing jenkins-x-dev-environment directory. Retrying after removing that and the jx namespace worked.

There are a couple of problems left, but they seem unrelated to my changes: 

- The staging environment is registered twice and production not at all.
- The build of the environment fails with: `error: fetching secrets files from vault: retrieving the system Vault: cannot find cluster name as no ConfigMap jx-install-config in namespace : failed to get configmap jx-install-config in namespace , an empty namespace may not be set when a resource name is provided`

msvticket:
I should test `create vault` as well

msvticket:
Fixed `jx step helm apply` in staging environment with the last commit.

I have now tried 
```
jx create addon vault-operator
jx create vault ...
```
on a kops cluster on aws that wasn't setup with gitops. So that seems to work as well.

Ah, `jx delete vault` doesn't work. Fixing...

ccojocar:
@msvticket The commits history looks strange. Please could you rebase these changes on top of master and squash your commits?

I will like to have another review before merge it. Thanks

msvticket:
Yeah, I something went wrong with some rebase... How about now?

I realised I haven't tested `jx create cluster eks -no-tiller --gitops --vault ...` or `jx create cluster aws -no-tiller --gitops --vault ...` with eks or aws or as provider. I can't see that it would work now, but probably would with just adding support for handling the aws vault command line flags. Should I add that in this PR?

pkg/jx/cmd/create_vault.go
```diff
@@ -74,14 +86,31 @@ func NewCmdCreateVault(commonOpts *CommonOptions) *cobra.Command {
 		},
 	}
 
+	// GKE flags
 	cmd.Flags().StringVarP(&options.GKEProjectID, "gke-project-id", "", "", "Google Project ID to use for Vault backend")
 	cmd.Flags().StringVarP(&options.GKEZone, "gke-zone", "", "", "The zone (e.g. us-central1-a) where Vault will store the encrypted data")
+
+	awsCreateVaultOptions(cmd, &options.AWSConfig)
```
ccojocar: I am wondering is not cleaner to have two distinct commands:
`jx create vault aws`
`jx create vault gcp`

pkg/jx/cmd/create_vault.go
```diff
@@ -100,14 +129,10 @@ func (o *CreateVaultOptions) Run() error {
 		return errors.Wrap(err, "retrieving the team settings")
 	}
 
-	if teamSettings.KubeProvider != gkeKubeProvider {
-		return errors.Wrapf(err, "this command only supports the '%s' kubernetes provider", gkeKubeProvider)
+	if teamSettings.KubeProvider != cloud.GKE && teamSettings.KubeProvider != cloud.AWS && teamSettings.KubeProvider != cloud.EKS {
+		return errors.Wrapf(err, "this command only supports the '%s' kubernetes provider", cloud.GKE)
```
ccojocar: This logs message needs to be updated to print also `AWS/EKS`.

pkg/jx/cmd/create_vault.go
```diff
@@ -127,23 +152,53 @@ func (o *CreateVaultOptions) createVaultGKE(vaultName string) error {
 		return errors.Wrap(err, "creating vault operator client")
 	}
 
-	return o.createVault(vaultOperatorClient, vaultName)
+	return o.createVault(vaultOperatorClient, vaultName, teamSettings.KubeProvider)
 }
 
 // DoCreateVault creates a vault in the existing namespace.
```
ccojocar: nit: replace `DoCreateVault with `createVault` in the comment

pkg/jx/cmd/create_vault.go
```diff
@@ -286,3 +365,18 @@ func (o *CreateVaultOptions) exposeVault(vaultService string) error {
...
+func CreateAuthServiceAccount(client kubernetes.Interface, vaultName, namespace, clusterName string) (string, error) {
+	serviceAccountName := AuthServiceAccountName(vaultName)
+	_, err := serviceaccount.CreateServiceAccount(client, namespace, serviceAccountName)
+	if err != nil {
+		return "", errors.Wrap(err, "creating vault auth service account")
+	}
+	return serviceAccountName, nil
+}
+
+// AuthServiceAccountName creates a service account name for a given vault and cluster name
```
ccojocar: What is the reason to move this function here? I think that it is also used by other commands like `delete vault`.

ccojocar: Also I think the name was truncated to avoid exceeding the allowed resource name length.

pkg/jx/cmd/install.go
```diff
@@ -274,6 +275,7 @@ func NewCmdInstall(commonOpts *CommonOptions) *cobra.Command {
 	cmd.Flags().StringVarP(&options.Flags.Provider, "provider", "", "", "Cloud service providing the Kubernetes cluster.  Supported providers: "+KubernetesProviderOptions())
 
 	cmd.AddCommand(NewCmdInstallDependencies(commonOpts))
+	awsCreateVaultOptions(cmd, &options.AWSConfig)
```
ccojocar: These adds quite a bunch of options to install command which are only used for AWS/EKS. Maybe we should move these out into something more specific `create cluster eks` command, or is some configuration which is created during  when the cluster is created.

pkg/kube/vault/vault_test.go
```diff
@@ -10,23 +11,34 @@ import (
...
-func TestCreateVault(t *testing.T) {
+type VaultTestCase struct {
+	name               string
+	namespace          string
+	err                bool
+	authServiceAccount string
+	secretsPathPrefix  string
+}
+
+type vaultGKETestCase struct {
```
ccojocar: Normally this is defined inside of the test.

pkg/kube/vault/vault_test.go
```diff
@@ -35,34 +47,77 @@ func TestCreateVault(t *testing.T) {
...
-			rb, err := client.RbacV1().ClusterRoleBindings().Get(tc.name, metav1.GetOptions{})
-			assert.NoError(t, err, "should retrieve vault cluster role binding without error")
-			assert.NotNil(t, rb, "created vault cluster role binding should not be nil")
+
+			validateVault(err, vaultclient, tc.VaultTestCase, t, client)
 		})
 	}
 }
+
+type vaultAWSTestCase struct {
```
ccojocar: I would move this one inside of the test.

ccojocar:
@jstrachan This change is adding 9 additional options to the install command which are only required for AWS/EKS. I think we should move them in something more specific. What's your opinion on this?

pkg/jx/cmd/install.go
msvticket: `jx create cluster eks` and `jx create cluster aws` should have these options as well, but I want to use terraform to create my cloud resources, so I will never use these commands. I suppose one could split up `jx install` into several commands as well instead of having the option `--provider`. But I'll wait for @jstrachan to pitch in before starting with such a change.

pkg/kube/vault/vault_test.go
msvticket: OK

pkg/kube/vault/vault_test.go
msvticket: I'll remove it altogether instead.

pkg/jx/cmd/create_vault.go
msvticket: It was in a GKE specific place before while it will probably be used for all cloud providers. Also it is only used in create_vault. Maybe it should be used in `delete vault`, but it isn't now. So I'll let it be for now.

pkg/jx/cmd/create_vault.go
msvticket: I started implementing that but when realising that one then should check that teamSettings.KubeProvider matches the command made I got cautious. It underlines that splitting into 2 commands goes against the principle that jx shouldn't require the user to give information already known. It might still be the best way, but I'll wait for @jstrachan to pitch in.

jstrachan:
@ccojocar the AWS stuff is in a separate struct so sounds good to me for now - we can always refactor more later?

ccojocar:
@jstrachan It's fine form my side. I think we need some validation on GKE with this PR, to verify that the existing functionality still works.

polothy:
Are there docs on how to use this for AWS?  Also wondering if the kms/dynamo/s3 resources could be shared between teams or each time would need their own 3.

msvticket:
I don't know enough about the inner workings of vault to know if the same resources can be reused for different teams. But in practice I don't think it is possible to have several GitOps operated teams yet due to assumptions in the code that the dev namespace is called jx. That shouldn't be that hard to change though.

Some documentation (requiring knowledge about AWS and Terraform though) should soon be up. Until then you can check my pull request jenkins-x/jx-docs#1332 .

mjosephd:
@msvticket, do you know when `jx create cluster eks -no-tiller --gitops --vault` will be available, is there a separate PR that I could follow perhaps?

amoussawi:
I see that creating a vault requires an S3bucket, but I don't see it being used anywhere when creating a vault?

PR #3286, opened 2019-03-06
fix for create env on prow and add filtering to jx get build log on tekton

jstrachan:


pkg/jx/cmd/get_build_logs.go
```diff
@@ -502,7 +502,7 @@ func (o *GetBuildLogsOptions) loadPipelines(kubeClient kubernetes.Interface, tek
 			log.Warnf("Error creating PipelineRunInfo for PipelineRun %s: %s\n", pr.Name, err)
 			return names, defaultName, buildMap, pipelineMap, err
 		}
-		if pri != nil {
+		if pri != nil && o.BuildFilter.BuildMatches(pri.ToBuildPodInfo()) {
```
abayer: I think we'll eventually want to change `BuildFilter` to be a `PipelineRunInfoFilter`, but it does seem easier to just keep it compatible for now.

jstrachan: yeah - or we could have a filter interface which we support on both `BuildRunInfo` and `PipeilneRunInfo` that both take the filter parmeters and do their thing

PR #3289, opened 2019-03-06
fix(tekton): Improve error handling when PipelineConfig is nil

dwnusbaum:
#### Submitter checklist

#### Description

Previously, running `jx step create task --pack=none` with a `jenkins-x.yml` that did not have a `pipelineConfig` section would fail with a nil pointer error. This PR changes that to a controlled failure with an error message.

/assign @abayer 

N/A

### DIFF
--- a/pkg/jx/cmd/step_create_task_test.go
+++ b/pkg/jx/cmd/step_create_task_test.go
@@ -48,11 +48,12 @@ func TestGenerateTektonCRDs(t *testing.T) {
 	}
 
 	cases := []struct {
-		name         string
-		language     string
-		repoName     string
-		organization string
-		branch       string
+		name           string
+		language       string
+		repoName       string
+		organization   string
+		branch         string
+		expectingError bool
 	}{
 		{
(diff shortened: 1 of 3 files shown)


pkg/jx/cmd/step_create_task_test.go
```diff
@@ -48,11 +48,12 @@ func TestGenerateTektonCRDs(t *testing.T) {
...
-		language     string
-		repoName     string
-		organization string
-		branch       string
+		name           string
+		language       string
+		repoName       string
+		organization   string
+		branch         string
+		expectingError bool
```
dwnusbaum: The expected error message includes a machine-dependent file path. We could create error types that take the path and then use a function to construct them in the test to make sure the test fails with the intended error, but this seemed good enough for now. Happy to change this up to match a specific error if desired.

abayer: Nah, that's fine for me at least for the time being. If we ever care more specifically in the future, we can do that.

PR #3295, opened 2019-03-07
fix: lets not fail a `jx get build log` if we find a pod thats invalid

jstrachan:
lets just ignore bad pods for now (logging warnings in verbose mode) as it could be the pod wasn't useful anyway, there could be other pods.

also we want the `jx get build logs --wait owner/repo/branch` command to wait and return a valid result when the pod starts up and is valid as we use this in our BDD tests


abayer:
argh, not sure if I ever really did anything with `--wait` - time to test that out more.

PR #3296, opened 2019-03-07
fix: Add cleanup for GKE addresses

### LINKED ISSUE #3285: Extend the `jx gc gke` command to remove unused items from GKE
### Summary 

During the process of creating and destroying clusters by user/automated tests, it leaves many unused items, which build up over time and breach our quota. We already have a mechanism to generate a script which clears down some items, but it would be useful if that were extended to also cover:
- IP addresses
- Instance groups

cagiti:
#### Description
This change adds cleanup commands for more items not removed when clusters are removed from GKE

fixes #3285

### DIFF
--- a/pkg/jx/cmd/gc_gke.go
+++ b/pkg/jx/cmd/gc_gke.go
@@ -52,6 +52,11 @@ type Rule struct {
 	TargetTags []string `json:"targetTags"`
 }
 
+type address struct {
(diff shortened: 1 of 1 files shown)


cagiti:
@rawlingsj this seems to be a lot healthier since we moved it over to build on tekton

rawlingsj:
awesome!
/lgtm

PR #3336, opened 2019-03-13
 tektonify jx CI and add next gen pipeline

rawlingsj:


pkg/jx/cmd/step_create_task.go
```diff
@@ -274,6 +274,8 @@ func (o *StepCreateTaskOptions) Run() error {
 		if err != nil {
 			return errors.Wrapf(err, "failed to apply Tekton CRDS")
 		}
+		// only include labels on PipelineRuns because they're unique, Task and Pipeline are static resources so we'd overwrite existing labels if applied to them too
```
abayer: Yeah, we have an open task to come up with a better way of managing pipelines and tasks in that regard. There are all kinds of ways things could go weird with changes happening during a run, etc, but we also don’t want to go over the top with unique pipelines/tasks for each run.

rawlingsj:
forgive me

PR #3349, opened 2019-03-14
Added os.Exit(1) to get app <name> when no apps found

imeredith:
jx now os.Exit(1) if a specific app is requested and it does not exist.

```
➜  jx get app test
No Apps found
➜  echo $?
1
```


imeredith:
Is https://github.com/jenkins-x/jx/blob/master/pkg/jx/cmd/get_apps_test.go#L191 causing an issue because of the os.Exit 

I'm having issues trying to run the tests locally. The following command seems to find no tests to run.

```
export TEST=TestGetAppNotFound
make test1
```

rawlingsj:
while we move releases to new infra
/hold

imeredith:
@rawlingsj What is the status of the hold. Im unclear if there have actual problems

pkg/jx/cmd/get_apps.go
```diff
@@ -90,6 +91,9 @@ func (o *GetAppsOptions) Run() error {
 
 	if len(apps.Items) == 0 {
 		fmt.Fprint(o.Out, "No Apps found\n")
+		if(len(o.Args) > 0) {
+			os.Exit(1)
+		}
```
pmuir: Why do we not just return an error here?

hferentschik: I don't think it is a good idea either to exit from here. There really should be a single exit point.

hferentschik: I am trying to figure out what the aim here is really. So the list of returned apps is empty and now you want to add an additional check for the length of the command line arguments. So this will check whether you just ran `jx get apps` or `jx get apps foo`. Why do you want to behave differently for the latter case? You believe that the exit code of `jx get apps foo` should be non 0 is the app is not installed? 

I guess that is reasonable, but would you in this case not also have different messages?
For the case of `jx get apps` one displays "No apps found" whereas for the case of `jx get app foo` one displays "App foo is not installed" or "App foo cannot be found".

Also what happens in the case of `jx get app foo bar`? What is the expected outcome? Do we have a test for this?

imeredith: So the main goal of this pr is the exit code if no apps are found if you explicitly request a specific one. As you point out, the command supports listing more than one.

Bdd-jx tests need to detect if an app is installed, and currently calls kubectl directly to look for the app, and if not found kubectl returns code 1. The idea of this PR was to use the jx command directly.

I'm wondering if there should just be a specific get app command vs get apps.

jtnord:
/test integration
just want to test a trigger

pkg/jx/cmd/get_apps.go
hferentschik: > So the main goal of this pr is the exit code if no apps are found if you explicitly request a specific one. As you point out, the command supports listing more than one.

> Bdd-jx tests need to detect if an app is installed, and currently calls kubectl directly to look for the app, and if not found kubectl returns code 1. The idea of this PR was to use the jx command directly.

Got you. I guess the change will do what you intended. Personally, I'd have a different more customized message for the case where get apps is called with a specific list of apps, but none is installed. But fair enough.

IMO it would have been awesome if you created an issue first outlining the above. This would have provided the required context and make any kind of archaeological activity further down the lines easier. I realise though that this is not (yet) common practice in this repo ;-)

I would recommend so to at least adjust the commit message to match the conventional commit style - https://www.conventionalcommits.org/en/v1.0.0-beta.3/. I don't think it is enforced (yet), but most seem to stick to it. It also aids the automatic changelog generation, which tries to detect and categorize commits with conventional commit message style.

> I'm wondering if there should just be a specific get app command vs get apps.

I don't think this is a good idea, mainly because this does not align with the rest of the Kube tooling around, most importantly `kubecetl` where the plural versions of a noun usually is an alias to the singular form.

dgozalo:
/hold 
something went wrong with the merge

pmuir:
Fixed by #3535

PR #3393, opened 2019-03-21
feat: add utility to generate a JSON patch.

markawm:
#### Submitter checklist

#### Description
Added a utility to generate a JSON patch. Returns the JSON patch bytes ready to be applied, e.g. using the generated clientset.

fixes #

### DIFF
--- a/pkg/kube/patch.go
+++ b/pkg/kube/patch.go
@@ -2,6 +2,9 @@ package kube
 
 import (
 	"encoding/json"
+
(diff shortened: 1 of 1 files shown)


markawm:
I have a use in a downstream project... when I searched the code to see if this was already there, I found a few uses of strategic patch, but nothing to do a JSON patch. tbh I'm not too sure of the differences - I read some details about the way lists are handled, but I'm not clear on the pros & cons. In the meantime having the _option_ to do either seemed useful...

PR #3423, opened 2019-03-26
fix: lets default to recreate the vault bucket

jstrachan:
which makes installing vault more reliable


pkg/cloud/gke/vault/vault_backend.go
```diff
@@ -80,21 +84,35 @@ func CreateGCPServiceAccount(kubeClient kubernetes.Interface, vaultName, namespa
...
+	exists, err := gke.BucketExists(projectID, bucketName)
 	if err != nil {
 		return "", errors.Wrap(err, "checking if Vault GCS bucket exists")
 	}
 	if exists {
-		return bucketName, nil
+		if !recreate {
+			return bucketName, nil
+		}
+		if batchMode {
```
ccojocar: I think, the condition should be `!bachMode` when prompting.

jstrachan: thanks - doh - fixed!

jstrachan:
@ccojocar fixed that boolean issue you spotted btw ;)

PR #3425, opened 2019-03-26
fix: don’t try to overwrite user id

pmuir:
#### Submitter checklist

fixes #

### DIFF
--- a/pkg/users/git.go
+++ b/pkg/users/git.go
@@ -3,6 +3,10 @@ package users
 import (
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/errors"
+
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"gopkg.in/src-d/go-git.v4/plumbing/object"
 
 	jenkinsv1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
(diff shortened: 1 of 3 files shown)


pkg/users/git.go
```diff
@@ -71,6 +73,14 @@ func (r *GitUserResolver) Resolve(user *gits.GitUser) (*jenkinsv1.User, error) {
 		}
 		new := r.GitUserToUser(gitUser)
 		id = gitUser.Login
+		// Check if the user id is available, if not append "-<n>" where <n> is some integer
+		for i := 0; true; i++ {
+			_, err := r.JXClient.JenkinsV1().Users(r.Namespace).Get(id, v1.GetOptions{})
+			if err == nil {
```
davidcurrie: Shouldn't this be negated? We break out if we couldn't retrieve the user...

Better still is to use `errors.IsNotFound`:
https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L457

pkg/users/git_test.go
```diff
@@ -202,6 +202,72 @@ func TestFindUserByFromGitProviderWithNoEmail(t *testing.T) {
...
+	}()
+	assert.NoError(t, err)
+	gitUser := gits.GitUser{
+		Login: gitUser1.Login,
+		Email: gitUser1.Email,
+	}
+	_, err = resolver.Resolve(&gitUser)
+	assert.NoError(t, err)
+	// Validate that we have both users
+	// TODO
```
davidcurrie: TODO ?!

pmuir: Ooops ;-)

pkg/users/git.go
pmuir: Thanks, good point

PR #3426, opened 2019-03-26
expose the defaultRole option in the Cloudbees addon

jtnord:
the CloudBees addon can create users with a default role assigned.

This is useful for some projects like Jenkins-X who want all users to be
able to sign on and see the build result of pull-requests (and even
smaller companies that only have one team where everyone should have
committer type access.

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/jx/cmd/create_addon_cloudbees.go
+++ b/pkg/jx/cmd/create_addon_cloudbees.go
@@ -45,9 +45,10 @@ var (
 // CreateAddonCloudBeesOptions the options for the create spring command
 type CreateAddonCloudBeesOptions struct {
 	CreateAddonOptions
-	Sso      bool
-	Basic    bool
-	Password string
+	Sso         bool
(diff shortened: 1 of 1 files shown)


pkg/jx/cmd/create_addon_cloudbees.go
```diff
@@ -75,6 +76,7 @@ func NewCmdCreateAddonCloudBees(commonOpts *CommonOptions) *cobra.Command {
 	}
 
 	cmd.Flags().BoolVarP(&options.Sso, "sso", "", false, "Enable single sign-on")
+	cmd.Flags().StringVarP(&options.DefaultRole, "default-role", "", "", "The default role to apply to new users. Defaults to no role and applies to the admin namespace only")
```
markawm: (nit) - IMHO its nicer to put these flags into constants so that they rename more safely if anyone does so in the future.

PR #3435, opened 2019-03-26
fix(tekton) make pipelines and tasks unique per-build

dwnusbaum:
#### Submitter checklist

- [ ] Change is covered by existing or new tests. (I'm not sure how much work it would be to add functional tests here for concurrent builds, but it seems like we'd want BDD tests for this kind of thing. The change is partially covered by existing tests, but we are not testing the relevant edge cases (multiple concurrent builds))

#### Description

Otherwise, updating a Pipeline while a build is in progress causes the running Pipeline to adjust the rest of its execution to match the new state, which means the build would execute partly using the old version and partly using the new version, which is a little confusing in most cases but could be a serious correctness issue depending on the nature of the changes to the Pipeline.

[Here](https://gist.github.com/dwnusbaum/8e7222372c789f3e863138b1b2efee71) is an example Gist showing the build logs for a Pipeline where one of the Tasks was updated while the Pipeline was still running. The logs show that the Pipeline ran the original version of the first Task, but the updated version of the second Task.

Uniqueness is handled by adding build numbers to Pipelines and Tasks. Build numbers are tracked per-branch using annotations on SourceRepository resources. _There is no migration code that looks at PipelineActivity_, so anyone currently running pipelines via Tekton will see the build numbers reset to 0. If this seems like a serious problem to reviewers I can try to add some kind of migration code. 

Note that right now we have a lot of pipeline-related resources without owner references that are presumably not getting GC'd in any timely manner, and this PR doesn't fix that, and it probably makes it a little worse because now each build will have its own versions of Pipelines and Tasks (see [here](https://drive.google.com/file/d/1wDd7uz-Bty67s93cS0Sp54NUjwIvRaoF/view?usp=sharing) (click through to draw.io to see the full diagram) for diagrams of the state of resources before and after this patch, and a diagram of what I think would make more sense for the future).

/assign @abayer

N/A

### DIFF
--- a/pkg/jx/cmd/step_create_task.go
+++ b/pkg/jx/cmd/step_create_task.go
@@ -36,14 +36,14 @@ const (
 
 var (
 	createTaskLong = templates.LongDesc(`
-		Creates a Knative Pipeline Run for a project
+		Creates a Tekton Pipeline Run for a project
 `)
 
 	createTaskExample = templates.Examples(`
-		# create a Knative Pipeline Run and render to the console
+		# create a Tekton Pipeline Run and render to the console
 		jx step create task
 
-		# create a Knative Pipeline Task
+		# create a Tekton Pipeline Task
 		jx step create task -o mytask.yaml
 
 		# view the steps that would be created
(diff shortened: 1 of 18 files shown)


abayer:
/lgtm
@rawlingsj - are we good to remove the hold?

dwnusbaum:
Just rebased and pushed b9bdc0e to handle the merge conflicts after #3469 and the would-be test failures after #3421.

abayer:
@rawlingsj ping again re: the hold. =)

PR #3406, opened 2019-03-22
fix(addon) upgrades ignored previous values

jtnord:
#### Submitter checklist

- [x] Change is code complete and matches issue description.
- [x] Change is covered by existing or new tests.

#### Description

when upgrading the addon any previous values for customosation where
lost.  Thic change tells helm to merge existing values with ones that
where used previously when upgrading.

#### Special notes for the reviewer(s)

Please think about any other usage that this may break, I hope I have confined it to just addons.

#### Which issue this PR fixes

fixes #3399

Issue #3399, opened 2019-03-21
jx upgrade addon ignores any options used when the addon was installed


jtnord:
### Summary 

If you install an addon that uses some options in the helm chard (e.g. `jx create addon cloudbees`) then when you upgrade the addon with `jx upgrade` the options get reset to the defailts.

### Steps to reproduce the behaviour

install the cloudbees addon with sso enabled.
wait
upgrade the cloudbees addon

### Expected behavior

the cloudbees addon is installed and configured with SSO

### Actual behavior

the cloudbees addon is installed but is not configured to use SSO

### Jx version

The output of `jx version` is:

```
NAME               VERSION
jx                 [32m1.3.980[0m
jenkins x platform [32m0.0.3368[0m
Kubernetes cluster [32mv1.12.5-gke.5[0m
kubectl            [32mv1.13.0[0m
helm client        [32mv2.11.0+g2e55dbe[0m
helm server        [32mv2.11.0+g2e55dbe[0m
git                [32mgit version 2.16.2.windows.1[0m
Operating System   [32mWindows 10 Pro 1809 build 17763[0m
```

### Jenkins type

- [ ] Classic Jenkins
- [x] Serverless Jenkins

jtnord:
SO I have reverted this and and upgraded the jx cli locally so I am running with the revert and I can no longer reproduce the original issue.

helm versions are the same in client and server, only difference is that I had upgraded the platform but that should have not be involved in running this command.

PR #3447, opened 2019-03-27
partial revert of #3406 / #3399

jtnord:
this did not only re-use the values passed by set but also the previous
default values of the chart (and did not even merge with the default
values of the new version of the chart).

#3406 caused a serious regression where it was not the options passed to --set that where re-used but also the previous default values from when the chart was installed.

when this includes the version of the software to install things get a little wonky and you end up upgrading the chart but it keeps the old software.  also any new flags in the chart end up being missing so you can not upgrade the chart.  For now just revert it whilst a better fix is thought about.

#### Submitter checklist

fixes #

@amuniz @markawm @pmuir

### DIFF
--- a/pkg/jx/cmd/upgrade_addons.go
+++ b/pkg/jx/cmd/upgrade_addons.go
@@ -157,7 +157,7 @@ func (o *UpgradeAddonsOptions) Run() error {
 					return errors.Wrap(err, "backing up the prow config")
 				}
 			}
-			err = o.Helm().UpgradeChart(chart, k, ns, "", false, -1, false, false, values, valueFiles, "", "", "", true)
+			err = o.Helm().UpgradeChart(chart, k, ns, "", false, -1, false, false, values, valueFiles, "", "", "", false)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


amuniz:
FTR we were hitting something very near to https://github.com/helm/helm/issues/3957

jtnord:
> change looks fine to me, but if its "danger will robinson" on the reuseValues flag, then I'd suggest at least a warning on the function doc. to be very careful. I doubt most other people would know what to say without this background, and could pick the wrong value.

I wanted to revert the whole change, but I couldn't do that cleanly as there where merge conflict and want something quick before there where too many breakages.
the flag does do what it says on the tin - I guess I read it wrong to begin with as I am sure I understood something different at the time,

jtnord:
>  FTR we were hitting something close to helm/helm#3957

yup - that is the newly introduced bug.
before that (and what the original fix we are reverting here tried to fix) we where hitting helm/helm/issues/3702

Either way - this sucks...

jtnord:
/test all
/reset
/test this

one of these must work...

rawlingsj:
lets merge this as part of a batch
/hold

PR #3448, opened 2019-03-27
Fuller revert of 3406

jtnord:
this did not only re-use the values passed by set but also the previous
default values of the chart (and did not even merge with the default
values of the new version of the chart).

#3406 caused a serious regression where it was not the options passed to --set that where re-used but also the previous default values from when the chart was installed.

when this includes the version of the software to install things get a little wonky and you end up upgrading the chart but it keeps the old software. also any new flags in the chart end up being missing so you can not upgrade the chart. For now just revert it whilst a better fix is thought about.

this subsumes #3447 but I expect to have missed a file and this not to pass CI so the other is there to prevent breakages asap.

#### Submitter checklist

fixes #


jtnord:
/test integration
lets get them logs!

jtnord:
/test integration
take 2

PR #3459, opened 2019-03-28
Don't bind mount docker.sock if we use Kaniko or Tekton

ndeloof:
#### Submitter checklist

#### Description

Access to Docker.sock introduce a major security issue, letting build / PR access arbitrary resources on the node. This is required to run `docker build` for classic builds, but once  Kaniko or Tekton is selected during installation, we don't need build pods to bind-mount docker.sock anymore.

This assumes Tekton always uses Kaniko. Have been told on slack but can't confirm by myself without better understanding of tekton.

### DIFF
--- a/pkg/config/helm_values.go
+++ b/pkg/config/helm_values.go
@@ -23,6 +23,11 @@ type ExposeController struct {
 type JenkinsValuesConfig struct {
 	Servers JenkinsServersValuesConfig `json:"Servers,omitempty"`
 	Enabled *bool                      `json:"enabled,omitempty"`
+	Agent   JenkinsAgentValuesConfig   `json:"Agent,omitempty"`
+}
+
+type JenkinsAgentValuesConfig struct {
+	DockerHostPath string  `json:"DockerHostPath,omitempty"`
 }
 
 type ProwValuesConfig struct {
--- a/pkg/jx/cmd/install.go
+++ b/pkg/jx/cmd/install.go
@@ -1078,6 +1078,12 @@ func (options *InstallOptions) configureHelmValues(namespace string) error {
 		enableControllerBuild := true
 		helmConfig.ControllerBuild.Enabled = &enableControllerBuild
 	}
+
+	if options.Flags.Kaniko || options.Flags.Tekton {
+		// As we use Kaniko to build docker images, we don't need docker.sock bind mount in build containers
+		helmConfig.Jenkins.Agent.DockerHostPath = ""
+	}
+
 	return nil
 }
 


pkg/config/helm_values.go
```diff
@@ -23,6 +23,11 @@ type ExposeController struct {
 type JenkinsValuesConfig struct {
 	Servers JenkinsServersValuesConfig `json:"Servers,omitempty"`
 	Enabled *bool                      `json:"enabled,omitempty"`
+	Agent   JenkinsAgentValuesConfig   `json:"Agent,omitempty"`
+}
+
+type JenkinsAgentValuesConfig struct {
```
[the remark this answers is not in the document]
ndeloof: as the whole file is written this way (as many others) I would assume this is not a blocker
That being said, I need to admit some comment to explain this `helm_value.go` file models https://github.com/jenkins-x/jenkins-x-platform/blob/master/values.yaml would help contributors.

ndeloof:
"link" detail point to this pull-request. Where can I check the logs to understand what's going wrong here ?

pkg/jx/cmd/install.go
```diff
@@ -1078,6 +1078,12 @@ func (options *InstallOptions) configureHelmValues(namespace string) error {
 		enableControllerBuild := true
 		helmConfig.ControllerBuild.Enabled = &enableControllerBuild
 	}
+
+	if options.Flags.Kaniko || options.Flags.Tekton {
+		// As we use Kaniko to build docker images, we don't need docker.sock bind mount in build containers
+		helmConfig.Jenkins.Agent.DockerHostPath = ""
```
ccojocar: In case of `Tekton`, Jenkins installation should be disabled.

ndeloof: Where are the tekton pod templates defined then ?

ccojocar: There isn't any template. Typically, the volumes are mounted in the pipeline/tasks CRDs definitions. For Jenkins X, they are generated from jenkins-x.yaml file + buildpack file, which contains the high level pipeline definition.

ndeloof: When I tried Tekton I got Task set with a bind mount for docker.sock. Not sure where this comes from.

ccojocar: Yeah, that should be fixed now. I saw this as well, but in the latest version from yesterday, it doesn't seem to be the case.

ndeloof: tested just now (1.3.1075) and still
```
Mounts:
      /builder/home from home (rw)
      /etc/podinfo from podinfo (ro)
      /home/jenkins from workspace-volume (rw)
      /home/jenkins/.docker from volume-0 (rw)
      /tools from tools (rw)
      /var/run/docker.sock from docker-daemon (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from tekton-bot-token-5thvv (ro)
      /workspace from workspace (rw)
```

ccojocar: @warrenbailey @garethjevans Any idea why we are still mounting the docker socket? I saw in the prod cluster that it wasn't mounted.

PR #3467, opened 2019-03-28
Issue #3314 Use PATCH over UPDATE when updating CRs

### LINKED ISSUE #3314: Use PATCH over UPDATE when updating CRs
Instead of using the `Update` method for the various generated clientset interfaces, we should use the `Patch` approach, specifying a JSON path over updating the whole CR.
This approach reduces the amount of data sent over the wire and it should reduce the number of conflicts when trying to update.

hferentschik:
#### Submitter checklist

#### Description

* Moved code generation into its own cmd
* Introduce for each CRD type a PatchUpdate method which applies the update via a JSON patch

fixes #3314


Makefile
```diff
@@ -32,13 +33,8 @@ GITHUB_ACCESS_TOKEN := $(shell cat /builder/home/git-token 2> /dev/null)
...
 CGO_ENABLED = 0
 
-all: build
-full: check
-check: build test
-
-version:
 ifeq (,$(wildcard pkg/version/VERSION))
-TAG := $(shell git fetch --all -q 2>/dev/null && git describe --abbrev=0 --tags 2>/dev/null)
+TAG := $(shell git describe --abbrev=0 --tags 2>/dev/null)
```
hferentschik: FYI, I am removing the `git fetch` here. It takes a lot of time and I don't think is needed. For sure not during development. 

If it is an issue on CI or during release I prefer to address this somewhere else.

hferentschik: If we want to do the setting of `VERSION` conditionally within a `version` target, we need to use `eval`. The way the Makefile was written, `VERSION` was always evaluated. Just because the `ifeq` appeared under `version:` does not mean it is only evaluated when `version` is called.

Makefile
```diff
@@ -14,16 +14,17 @@
...
 NAME := jx
 GO := GO111MODULE=on go
 GO_NOMOD :=GO111MODULE=off go
 REV := $(shell git rev-parse --short HEAD 2> /dev/null || echo 'unknown')
-#ROOT_PACKAGE := $(shell $(GO) list .)
 ROOT_PACKAGE := github.com/jenkins-x/jx
 GO_VERSION := $(shell $(GO) version | sed -e 's/^[^0-9.]*\([0-9.]*\).*/\1/')
-PKGS := $(shell go list ./... | grep -v generated)
-GO_DEPENDENCIES := cmd/*/*.go cmd/*/*/*.go pkg/*/*.go pkg/*/*/*.go pkg/*//*/*/*.go
+GO_DEPENDENCIES := $(call rwildcard,pkg/,*.go) $(call rwildcard,cmd/jx/,*.go)
```
hferentschik: The existing `GO_DEPENDENCIES` were not quite right. It was not covering all potential Go files. This should be better.

cmd/codegen/codegen.go
```diff
@@ -0,0 +1,13 @@
+package main
+
+import (
+	"github.com/jenkins-x/jx/cmd/codegen/app"
+	"os"
+)
+
+func main() {
```
hferentschik: So codegen is its own binary now. There is still a dependency to methods under `pgk`, however minimal. TBH, I am tempted to break the link completely and have a true standalone tool. There is not much benefit in sharing here. On the contrary.

pkg/client/clientset/versioned/typed/jenkins.io/v1/app_expansion.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// AppExpansion expands the default CRUD interface for App.
+type AppExpansion interface {
+	PatchUpdate(app *v1.App) (result *v1.App, err error)
+}
+
+// PatchUpdate takes the representation of an app and updates using Patch generating a JSON patch to do so.
+// Returns the server's representation of the app, and an error, if there is any.
+func (c *apps) PatchUpdate(app *v1.App) (*v1.App, error) {
```
hferentschik: @pmuir, So this is the essence of using JSON patch. Each CRD type gets its `PatchUpdate`. I am leveraging the go-client cache here, so I don't think we need the explicit deep clone.

Makefile
pmuir: makes sense.

pkg/client/clientset/versioned/typed/jenkins.io/v1/app_expansion.go
pmuir: where is the generator for this?

PR #3473, opened 2019-03-29
fix: use GitOps with jx upgrade platform

jstrachan:


pkg/jx/cmd/upgrade_platform.go
```diff
@@ -431,3 +443,18 @@ func (o *UpgradePlatformOptions) repairAdminSecrets(fileName string) error {
 
 	return nil
 }
+
+func (o *UpgradePlatformOptions) upgradePlatformViaGitOps(devEnv *v1.Environment, targetVersion string, versionsDir string, configStore configio.ConfigStore) error {
+	opts := &UpgradeAppsOptions{}
+	opts.CommonOptions = o.CommonOptions
+	opts.ReleaseName = "jenkins-x"
```
ccojocar: I would use a constant for this string e.g. `JenkinsXPlatformReleaseName`.

pkg/jx/cmd/upgrade_platform.go
```diff
@@ -145,20 +145,32 @@ func (o *UpgradePlatformOptions) Run() error {
...
 
+	isGitOps, devEnv := o.GetDevEnv()
+	if isGitOps {
+		if devEnv == nil {
+			return fmt.Errorf("no Dev environment found")
+		}
+		if devEnv.Spec.Source.URL == "" {
+			return fmt.Errorf("Dev environment does not have source URL")
+		}
+		return o.upgradePlatformViaGitOps(devEnv, targetVersion, versionsDir, configStore)
```
ccojocar: This does not check the current version before raising the PR. I think, the update should be skipped if the `targetVersion` == `Current version`.  The logic which verifies this is below.

Issue #3490, opened 2019-04-01
git fetch executed every time make is called


hferentschik:
### Summary 

At the moment when calling any build target in the Makefile is executed, a `git fetch --all` is executed. This adds several seconds to the build time and is not needed during development.

The reason it exists might be that during a release build the checkout does not contain tags. 

The Makefile contains a _version_ target which on the first inspection tries to resolve the issue, by only calling `git fetch`  when the `version` target is executed. However, this is not correct.
In the code below, the if statement is always executed. It does not matter that it appears under `version:`. 

```
version:
ifeq (,$(wildcard pkg/version/VERSION))
TAG := $(shell git fetch --all -q 2>/dev/null && git describe --abbrev=0 --tags 2>/dev/null)
ON_EXACT_TAG := $(shell git name-rev --name-only --tags --no-undefined HEAD 2>/dev/null | sed -n 's/^\([^^~]\{1,\}\)\(\^0\)\{0,1\}$$/\1/p')
VERSION := $(shell [ -z "$(ON_EXACT_TAG)" ] && echo "$(TAG)-dev+$(REV)" | sed 's/^v//' || echo "$(TAG)" | sed 's/^v//' )
else
VERSION := $(shell cat pkg/version/VERSION)
endif
```

### Steps to reproduce the behavior

Echo the VERSION variable in a target which does not depend on _version_.

### Expected behavior

`git fetch` should only be executed in case of a release. It might even be best to make sure that the checkout on the builder server already gets the tags and it is not the resposibilty of the make to do so.

hferentschik:
merged via pull request #3492

PR #3492, opened 2019-04-01
refactor: avoid git fetch on each build

hferentschik:
- per default VERSION will be set to a dev version
- removed if statement checking for VERSION file
- introduced target set-release-version to make sure the latest tag is used to
  determine next version (using git fetch)
- extracted Docker related targets into Makefile.docker

fixes issue #3490


Makefile
```diff
@@ -14,16 +14,17 @@
 # limitations under the License.
 #
 
+# Make does not offer a recursive wildcard function, so here's one:
+rwildcard=$(wildcard $1$2) $(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2))
```
hferentschik: @jstrachan - I double checked that his works for macOS. I am using 'GNU Make 3.81'. If it does not work for you, I think we can safely remove it. There is really no big benefit here. We can rely on Go figuring things out.

Makefile
```diff
@@ -32,18 +33,9 @@ GITHUB_ACCESS_TOKEN := $(shell cat /builder/home/git-token 2> /dev/null)
 PEGOMOCK_PACKAGE := github.com/petergtz/pegomock/
 CGO_ENABLED = 0
 
-all: build
-full: check
-check: build test
+TAG := $(shell git describe --abbrev=0 --tags 2>/dev/null)
+VERSION := $(shell echo "$(TAG)-dev+$(REV)" | sed 's/^v//')
```
hferentschik: @jstrachan per default we create a dev version without `git fetch`

Makefile
```diff
@@ -159,30 +141,40 @@ inttestbin:
...
 
-release: check
+set-release-version:
+    # ensure we have the latest tag
+	@git fetch --all -q 2>/dev/null
+	$(eval VERSION = $(shell git describe --abbrev=0 --tags 2>/dev/null | sed 's/^v//' ))
+	@echo Release version: $(VERSION)
+
+.PHONY: release
+release: set-release-version check
```
hferentschik: @jstrachan - for the release we depend on a new `set-release-version` which calls `git fetch` and dynamically updates VERSION using eval.

jstrachan:
@hferentschik btw when we are creating a new release there is the `$VERSION` env var we could use to simplify the makefile?

hferentschik:
> btw when we are creating a new release there is the $VERSION env var we could use to simplify the makefile?

Ahh, even better. In this case, we could do something like create a dev VERSION if there is none set, otherwise use the VERSION as passed through the environment. Makes much more sense.

jstrachan:
@hferentschik yeah - then folk could set a dev `$VERSION` env var once on startup then we never have  to run any git commands on a build which will further speed things up a bit

Makefile
```diff
@@ -32,18 +33,9 @@ GITHUB_ACCESS_TOKEN := $(shell cat /builder/home/git-token 2> /dev/null)
 PEGOMOCK_PACKAGE := github.com/petergtz/pegomock/
 CGO_ENABLED = 0
 
-all: build
-full: check
-check: build test
+# set dev version unless VERSION is explicitly set via environment
+VERSION ?= $(shell echo "$$(git describe --abbrev=0 --tags 2>/dev/null)-dev+$(REV)" | sed 's/^v//')
```
hferentschik: @jstrachan, only set VERSION if not explicitly specified via the environment. For the case VERSION is not set use `git describe`.

hferentschik:
@jstrachan 

> yeah - then folk could set a dev $VERSION env var once on startup then we never have to run any git commands on a build which will further speed things up a bit

+1 I updated the Makefile. If VERSION is set in the environment use it, otherwise `git describe` is used. Without the fetch it is not really a problem. Also, if someone is bothered by it, he can always set now an explicit VERSION.

hferentschik:
@jstrachan on a different note, could `Jenkinsfile` and myvalues.yaml.template` not be removed now?

jstrachan:
@hferentschik we can zap `Jenksinfile` now yeah - though we still use `myvalues.yaml.template` to run the BDD tests

PR #3495, opened 2019-04-01
fix: build number generation for tekton

jstrachan:


pkg/jx/cmd/step_create_task.go
```diff
@@ -169,6 +169,9 @@ func (o *StepCreateTaskOptions) Run() error {
 	}
 	o.devNamespace = ns
 
+	if o.Verbose {
```
ccojocar: @jstrachan I think, we should configure the `logrus` log-level to `info` in one place before starting the command execution when `Verbose` flag is set. This will have the benefit that we won't need to guard the logs with `if o.Verbose {}`.

PR #3496, opened 2019-04-01
feature: added a way to show the status of the app and provided funti…

### LINKED ISSUE #3270: Wrong options for 'jx get apps'
### Summary 

The new `jx get apps` commands has unexepcted/wrong options.

```
$ jx get apps -h
Display installed Apps

Aliases:
Apps, app, apps
Examples:
  # List all apps
  jx get apps

  # Display details about the app called cheese
  jx get app cheese
Options:
  -o, --output='': The output format such as 'yaml'
Usage:
  jx get Apps [flags] [options]
```

There is an -o/output option. This is unexpected and also not working since even if specified the default table is displayed.

### Expected behavior

I would expect to have other options to filter and environment/namespace which aligns with the other `jx get` commands.

On a slightly different note, should there be an option to show the status of the App? @pmuir once mentioned that `get apps` should replace `helm status`, but afaict `get apps` does not contain any status information. 

Last but not least, I don't think that `apps` should be aliased to `Apps` with capital 'A'.

### Actual behavior

See above

### Jx version

jx --version
1.3.948

### Jenkins type

- [ ] Classic Jenkins
- [x] Serverless Jenkins

dgozalo:
#### Submitter checklist

#### Description

- Adds a "Status" column to the `jx get app(s)`

- Makes it possible to format the output of the table to `json` or `yaml` with `-o` or `--output`

- Added the `--status` command to get the detailed Helm chart resources information

- The `-o` or `--output` flags should also affect the output of `--status` providing the output in the indicated format.

BDD tests still pending. Depending on open PRs.

Adding newly generated mocks.

fixes #3270


pkg/helm/helm_cli.go
```diff
@@ -464,6 +464,14 @@ func (h *HelmCLI) StatusRelease(ns string, releaseName string) error {
 	return h.runHelm("status", releaseName)
 }
 
+// StatusReleaseWithOutput returns the output of the helm status command for a given release
+func (h *HelmCLI) StatusReleaseWithOutput(ns string, releaseName string, outputFormat string) (string, error) {
```
hferentschik: +1

pkg/helm/helm_cli_test.go
```diff
@@ -219,6 +219,28 @@ func TestStatusRelease(t *testing.T) {
 	verifyArgs(t, helm, runner, expectedArgs...)
 }
 
+func TestStatusReleaseWithOutputNoFormat(t *testing.T) {
+	expectedArgs := []string{"status", releaseName}
```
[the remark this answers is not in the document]
dgozalo: My reasoning was that, when testing calls to the Helm CLI (or any CLI for that matter) we should always test that the parameters passed in that specific function won't change, and since there's an ìf`in that function, it's always good to test that fork as well. 

I do reckon that there's another extremely similar test for another function, but I wanted to guard myself on this one.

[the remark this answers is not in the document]
hferentschik: I see. The test itself is definitely inline with the others in this file. Happy to keep it. I literally tried to understand your reasoning here, since I feel skeptical about these tests.

dgozalo:
> Looking basically ok. Maybe if we could reduce the commit to only the changes required it would be easier to take in.

Should be fixed now. `make fmt` removed most of the unwanted changes.

pmuir:
Looks like an error with the git service in the BDD tests

pkg/jx/cmd/get_apps.go
```diff
@@ -117,7 +189,21 @@ func (o *GetAppsOptions) generateTable(apps *v1.AppList, kubeClient kubernetes.I
 func (o *GetAppsOptions) generateTableHeaders(apps *v1.AppList) table.Table {
 	t := o.createTable()
 	t.Out = o.CommonOptions.Out
-	titles := []string{"Name", "Version", "Description", "Chart Repository"}
+	titles := []string{"Name", "Version", "Chart Repository", "Status", "Description"}
```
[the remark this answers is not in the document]
hferentschik: I am still missing the namespace though. I installed the JaCoCo App into two different namespaces and the result of get apps is:
```
$jx get apps
Name                  Version   Chart Repository                Status   Description
jx-app-jacoco         1.0.0-dev http://127.0.0.1:8879           DEPLOYED A Helm chart for the JenkinsX JaCoCo App
jx-app-test-lifecycle 0.0.2     http://chartmuseum.jenkins-x.io DEPLOYED A Helm chart for Kubernetes
jx-app-jacoco         1.0.0-dev http://127.0.0.1:8879           DEPLOYED A Helm chart for the JenkinsX JaCoCo App
```

Now I do a:
```
$jx get apps --status jx-app-jacoco
Provide only one app for status check
```
This is confusing, since it is not obvious to me how to differentiate the two JaCoCo Apps. Effectively, I need for example:
```
$ jx get apps --status jx-app-jacoco -n jx
```

@pmuir it is a valid usecase to install an App into multiple namespaces, right?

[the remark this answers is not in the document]
dgozalo: Ah, I thought we were going to do this in a different issue. 

I'll change it tomorrow if we need to show the ns.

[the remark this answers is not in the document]
hferentschik: Not sure whether one needs to enforce it. It gets cumbersome for the default case where everything is in the default namespace. 

Displaying the namespace will to start with make things transparent and give the user a pointer on to use the namespace flag. I think this is not such a big mental jump.

Apart from this, one could think about more specific messages. Instead of having one generic message if multiple apps match, one could for the case of same app in multiple namespaces respond with: "The specified app is installed in multiple namespaces. Use the namespace flag to further qualify the app" (something like this).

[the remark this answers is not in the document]
dgozalo: Finally, as we talked, the info shown in `--status` will always be the resources of the app when it was first installed, so there's no point in filtering this by namespace. 

Also added a better message when multiple different applications are passed to `--status`.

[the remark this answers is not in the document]
hferentschik: Not filtering, but displaying right? I still can have two different apps installed into different namespaces in which case it would be nice to know where to look.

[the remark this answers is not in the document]
dgozalo: The namespace will still appear in the table when executing `jx get apps` and could be filtered.

This only applies when you add `--status`, as there isn't really much of a point to filter / show the namespace, as  atm it's showing the same for all installed apps.

pkg/jx/cmd/get_apps.go
```diff
@@ -93,21 +129,58 @@ func (o *GetAppsOptions) Run() error {
...
+	if o.Output != "" {
+		appsResult := o.generateTableFormatted(apps)
+		return o.renderResult(appsResult, o.Output)
+	}
+	table := o.generateTable(apps, kubeClient)
 	table.Render()
 	return nil
 }
 
+func (o GetAppsOptions) generateAppStatusOutput(app *v1.App) error {
```
ccojocar: I guess, you missed a pointer on the receiver.

pkg/jx/cmd/get_apps.go
```diff
@@ -93,21 +129,58 @@ func (o *GetAppsOptions) Run() error {
...
+	output, _ := o.Helm().StatusReleaseWithOutput(o.Namespace, app.Labels[helm.LabelAppName], "json")
+	return o.printHelmResourcesWithFormat(output)
+}
+
+func (o GetAppsOptions) generateTableFormatted(apps *v1.AppList) appsResult {
```
ccojocar: ditto

pkg/jx/cmd/get_apps.go
```diff
@@ -93,21 +129,58 @@ func (o *GetAppsOptions) Run() error {
...
+		appsResult := o.generateTableFormatted(apps)
+		return o.renderResult(appsResult, o.Output)
+	}
+	table := o.generateTable(apps, kubeClient)
 	table.Render()
 	return nil
 }
 
+func (o GetAppsOptions) generateAppStatusOutput(app *v1.App) error {
+	output, _ := o.Helm().StatusReleaseWithOutput(o.Namespace, app.Labels[helm.LabelAppName], "json")
```
ccojocar: I wouldn't ignore the error here. Maybe print a warning message.

pkg/jx/cmd/get_apps.go
```diff
@@ -93,21 +129,58 @@ func (o *GetAppsOptions) Run() error {
...
+	return o.printHelmResourcesWithFormat(output)
+}
+
+func (o GetAppsOptions) generateTableFormatted(apps *v1.AppList) appsResult {
+	releasesMap, _ := o.Helm().StatusReleases(o.Namespace)
```
ccojocar: error is ignored

pkg/jx/cmd/get_apps.go
```diff
@@ -93,21 +129,58 @@ func (o *GetAppsOptions) Run() error {
...
+	results := appsResult{}
+	for _, app := range apps.Items {
+		name := app.Labels[helm.LabelAppName]
```
ccojocar: It might be that `Labels` is nil.

pkg/jx/cmd/get_apps.go
```diff
@@ -93,21 +129,58 @@ func (o *GetAppsOptions) Run() error {
...
+		if name != "" {
+			results.AppOutput = append(results.AppOutput, appOutput{
+				Name:            name,
+				Version:         app.Labels[helm.LabelAppVersion],
+				Description:     app.Annotations[helm.AnnotationAppDescription],
```
ccojocar: Also here `Annotations` might be nil.

pkg/jx/cmd/get_apps.go
```diff
@@ -93,21 +129,58 @@ func (o *GetAppsOptions) Run() error {
...
+				Status:          releasesMap[name].Status,
+			})
+		}
+	}
+	return results
+}
+
 func (o *GetAppsOptions) generateTable(apps *v1.AppList, kubeClient kubernetes.Interface) table.Table {
 	table := o.generateTableHeaders(apps)
+	releasesMap, _ := o.Helm().StatusReleases(o.Namespace)
```
ccojocar: error is ignored?

pkg/jx/cmd/get_apps.go
```diff
@@ -117,7 +190,31 @@ func (o *GetAppsOptions) generateTable(apps *v1.AppList, kubeClient kubernetes.I
...
+	if o.Output == "" {
+		fmt.Fprintln(o.Out, h.helmInfoStatus.Resources)
+		return nil
+	}
+	return o.renderResult(h, o.Output)
+
+}
+
+func checkDifferentAppsInAppsSlice(apps *v1.AppList) bool {
+	x, a := apps.Items[0], apps.Items[1:]
```
ccojocar: If `Items` has size 0, this will panic.

dgozalo: Following the normal flow of execution, the apps.items slice getting to this not exported method should never be empty.

pkg/jx/cmd/get_apps.go
hferentschik: works as well, the conversion is implicit. A pointer receiver might be better of course and consistent with the rest. Good catch.

pkg/jx/cmd/test_data/get_apps/get_app_status.json
```diff
@@ -0,0 +1,7 @@
...
+    }
+  }
+}
```
hferentschik: If you want to make GutHub happy you should add a newline

pkg/jx/cmd/test_data/get_apps/get_apps_table_json.json
```diff
@@ -0,0 +1,11 @@
...
+  "items": [
+    {
+      "appName": "<NAME_PLACEHOLDER>",
+      "version": "0.0.1",
+      "description": "My test chart description",
+      "chartRepository": "http://chartmuseum.jenkins-x.io",
+      "status": "DEPLOYED"
+    }
+  ]
+}
```
hferentschik: same as above

pkg/jx/cmd/get_apps.go
ccojocar: That's right but is good to have a consistent style. e.g. the linter will flag this inconsistency if the pointer are mixed with non-pointer receivers.

Issue #3505, opened 2019-04-02
PatchUpdate methods are swallow error


hferentschik:
### Summary 

The introduced `PatchUpdate` method for the CRDs is not properly propagating all errors.
For example:

```golang
func (c *pipelineActivities) PatchUpdate(pipelineActivity *v1.PipelineActivity) (*v1.PipelineActivity, error) {
	resourceName := pipelineActivity.ObjectMeta.Name

	// force retrieval from cache
	options := metav1.GetOptions{ResourceVersion: "0"}
	orig, err := c.Get(resourceName, options)
	if err != nil {
		return nil, err
	}

	patch, err := util.CreatePatch(orig, pipelineActivity)
	patched, err := c.Patch(resourceName, types.JSONPatchType, patch)

	return patched, nil
}
```

The errors from `CreatePatch` and `Patch` need to be propagated

hferentschik:
Merged via pull request #3507

PR #3507, opened 2019-04-02
fix: propagate error from PatchUpdate

hferentschik:
fixes issue #3505

#### Submitter checklist

#### Description

Propagates error from PatchUpdate for custom CRDs


hferentschik:
@pmuir  now with tests added :-)

PR #3512, opened 2019-04-02
fix: Use /workspace/workspace everywhere for non-build-packs

abayer:
#### Submitter checklist

#### Description

This needs to get rationalized eventually, but for now, since the
unified generation path change, we can end up with a mismatch between
input target path and working directories, so in those scenarios,
force usage of the old /workspace/workspace.

I'd like to just switch the default to fall back on `/workspace/source`, but I'm afraid of breaking anything that's using `/workspace/workspace` currently, hence pinging in Warren to see if we can get away with making that change (i.e., is anything using `/workspace/workspace` in prod currently?).

/assign @jstrachan 
/assign @rawlingsj 
/assign @warrenbailey 
/assign @ccojocar 

n/a


jstrachan:
@abayer not a biggie but we've tried to use `/workspace/source` to avoid the confusing looking `/workspace/workspace` btw https://github.com/jenkins-x/jx/blob/9b52d70b493d36d4e7fe2f45ef1f917716fadd1c/pkg/jx/cmd/step_create_task.go#L152

abayer:
@jstrachan - Yeah, I pinged @warrenbailey to see whether I'll break anything if I just change it to `/workspace/source` everywhere - at the moment, this PR is just an attempt to unbreak some scenarios without breaking any others, but if we're comfortable enough switching to `/workspace/source` everywhere, I can hack that in ASAP.

warrenbailey:
Jenkins-x-serverless and Jenkins-x-builders both have /workspace/workspace hard coded, though we can change it pretty quick once this PR is merged. 

There's potentially a few others too, but we can fix forward once they start breaking. It's pretty easy fix

abayer:
Hrm, integration tests aren't restarting...

PR #3515, opened 2019-04-02
Add PR author and title to PipelineActivity

### LINKED ISSUE #3514: Add PR author and title to PipelineActivity
### Summary 

Add PR title and author to PipelineActivity (or commit author and last commit message for branch builds).

### Jx version

```
COPY OUTPUT HERE
```

### Jenkins type

- [ ] Classic Jenkins
- [ ] Serverless Jenkins

amuniz:
#### Submitter checklist

#### Description
This is adding two new fields to `PipelineActivity`: `author` and `pullTitle`

It requires an HTTP request to GH to retrieve the information, once it is set no more requests will be done for a given activity.

fixes #3514


amuniz:
I don't see any obvious error in the logs other than:
```
error: Failed to create git server with kind ghe at url https://github.beescloud.com in team bdd-ghe-jx-pr-3515-1: context canceled
{"component":"entrypoint","error":"wrapped process failed: exit status 1","level":"error","msg":"Error executing test process","time":"2019-04-02T15:20:03Z"}
```
Any idea?

amuniz:
And `integration` fail with a lot of:

```
# github.com/jenkins-x/jx/pkg/util/mocks
pkg/util/mocks/commander.go:17:34: undefined: pegomock.Option
FAIL	github.com/jenkins-x/jx/pkg/cloud/aks [build failed]
```
Which seems to be a mock thing. Is it possible to run integration tests locally?

markawm:
🤔 ISTR seeing that error before when I had mismatched versions of pegomock: I had a mock that had been generated with a later version of the pegomock command line tool than the pegomock library I was compiling against, so it referenced a new type that then didn't exist.

markawm:
Also, this may help:
```
go test
...
	-c
	    Compile the test binary to pkg.test but do not run it
	    (where pkg is the last element of the package's import path).
	    The file name can be changed with the -o flag.
```

So you can use that to _just_ compile the integration test without having to run it (since it modifies cluster 'stuff'...)

amuniz:
> I had a mock that had been generated with a later version of the pegomock command line tool than the pegomock library I was compiling against

Interesting. I think that's not possible now as `pegomock` is installed as part of `make generate` using the exact version defined in `go.mod`

amuniz:
```
$> GOPATH=~/go GO111MODULE=on go test -c -tags=integration ./...
cannot use -c flag with multiple packages
```
:(

amuniz:
FTR there were 2 different issues:

1. Infra instability
2. Wrong mocks generated by a the wrong pegomock version (pegomock/master was being used instead of the specific revision determined by the `Makefile`). This was fixed in `Makefile` by using `$(GO)` instead of just `go` when computing the pegomock sha to checkout.

amuniz:
Just added a new test covering the feature

pkg/jx/cmd/controller_build.go
```diff
@@ -321,6 +328,116 @@ func (o *ControllerBuildOptions) createPromoteStepActivityKey(buildName string,
...
+// completeBuildSourceInfo sets the PR author and PR title from GitHub in the given PA
+// If the PA is a branch build it then sets the commit author and last commit message
+func (o *ControllerBuildOptions) completeBuildSourceInfo(activity *v1.PipelineActivity) {
+
+	log.Infof("[BuildInfo] Completing build info for PipelineActivity=%s\n", activity.Name)
+
+	gitInfo, err := gits.ParseGitURL(activity.Spec.GitURL)
+	if err != nil {
+		log.Warnf("Cannot parse git repo URL [%s]. Error: %s\n", activity.Spec.GitURL, err)
+		return
```
ccojocar: any reason to shallow the error? I would propagate it to the caller and let the caller decide how to handle it.

pkg/jx/cmd/controller_build.go
```diff
@@ -321,6 +328,116 @@ func (o *ControllerBuildOptions) createPromoteStepActivityKey(buildName string,
...
+	}
+	if activity.Spec.Author != "" {
+		// info already set, save some GH requests
+		return
+	}
+
+	secrets, err := o.LoadPipelineSecrets(kube.ValueKindGit, "github")
+	if err != nil {
+		log.Warnf("Cannot load pipeline secrets. Error: %s\n", err)
```
ccojocar: ditto

pkg/jx/cmd/controller_build.go
```diff
@@ -321,6 +328,116 @@ func (o *ControllerBuildOptions) createPromoteStepActivityKey(buildName string,
 	}
 }
 
+// completeBuildSourceInfo sets the PR author and PR title from GitHub in the given PA
+// If the PA is a branch build it then sets the commit author and last commit message
+func (o *ControllerBuildOptions) completeBuildSourceInfo(activity *v1.PipelineActivity) {
+
+	log.Infof("[BuildInfo] Completing build info for PipelineActivity=%s\n", activity.Name)
```
ccojocar: nit: I would consider adding a filed fortmatter instead of adding the context in each log:
```
logrusutil.NewDefaultFieldsFormatter(nil, logrus.Fields{"component": "BuildInfo"})

// NewDefaultFieldsFormatter returns a DefaultFieldsFormatter,
// if wrappedFormatter is nil &logrus.JSONFormatter{} will be used instead
func NewDefaultFieldsFormatter(
	wrappedFormatter logrus.Formatter, defaultFields logrus.Fields,
) *DefaultFieldsFormatter {
	res := &DefaultFieldsFormatter{
		WrappedFormatter: wrappedFormatter,
		DefaultFields:    defaultFields,
	}
	if res.WrappedFormatter == nil {
		res.WrappedFormatter = &logrus.JSONFormatter{}
	}
	return res
}
```

pkg/jx/cmd/controller_build.go
```diff
@@ -321,6 +328,116 @@ func (o *ControllerBuildOptions) createPromoteStepActivityKey(buildName string,
...
+			log.Warnf("Cannot read last branch commit from Github. Error: %s\n", e)
+			return
+		}
+		if len(gitCommits) > 0 {
+			if gitCommits[0] != nil && gitCommits[0].Author != nil {
+				activity.Spec.Author = gitCommits[0].Author.Login
+				activity.Spec.LastCommitMessage = gitCommits[0].Message
+			}
+		}
+		log.Infof("[BuildInfo] PipelineActicity set with author=%s and last message=%s\n", activity.Spec.Author, activity.Spec.LastCommitMessage)
```
ccojocar: I think, this can get verbose if the last commit message is large.

pkg/jx/cmd/controller_build.go
```diff
@@ -321,6 +328,116 @@ func (o *ControllerBuildOptions) createPromoteStepActivityKey(buildName string,
...
+}
+
+func (o *ControllerBuildOptions) getGithubProvider(secrets *corev1.SecretList, gitInfo *gits.GitRepository) (gits.GitProvider, error) {
+	// this internal provider is only used during tests
+	if o.gitHubProvider != nil {
+		return o.gitHubProvider, nil
+	}
+
+	// production code always goes this way
+	if secrets != nil {
```
ccojocar: This is specific only to GitHub. 

I would consider using instead of this block of code: `CommonOptions.getPipelineGitAuth` + `gits.CreateProvider` which abstracts away the provider type.

pkg/jx/cmd/controller_build.go
amuniz: Computing values of this new fields should be a _best effort_ action, I don't want to block any other process because of this (a temporary GH issue, or a rate limit hit should not prevent the local operation of updating an activity). That's why I just logged errors and moved on.

pkg/jx/cmd/controller_build.go
amuniz: This is not using `logrus`, migrating the the whole project to use it should be another PR.

pkg/jx/cmd/controller_build.go
amuniz: I intentionally implemented this for `github.com` only as it is the only fully supported git provider. Also, I don't think it makes sense make the secrets part generic and not the rest (info retrieval is using the GH API client). Making this really generic would require a lot more work.

ccojocar: It makes sense, but please use the helper methods instead of implementing this on your own. It is less maintains and the risk to break things is lower. Thanks

pkg/jx/cmd/controller_build.go
ccojocar: I can imaging that this method can return an error, which gets logged from where it is called, without blocking anything.

This style of logging a message and return nothing is not very common in Go.

pkg/jx/cmd/controller_build.go
amuniz: I see, you want to abstract the secrets retrieval so any change on secrets storage is not impacting this. Got it. Fixing.

amuniz:
@ccojocar your suggestion is applied

pkg/jx/cmd/controller_build.go
amuniz: @ccojocar changed. Care to re-review please?

PR #3516, opened 2019-04-02
fix(userrole) removing multiple roles was not correct

jtnord:
#### Submitter checklist

#### Description

When removing multiple roles only one of the roles would actually be
removed.  turned out the diff implementation was not correct so fixed
that and added some unit tests.

fixes #


pkg/kube/roles.go
```diff
@@ -135,6 +137,7 @@ func UpdateUserRoles(kubeClient kubernetes.Interface, jxClient versioned.Interfa
 					if err != nil {
 						return errors.Wrapf(err, "Failed to remove User %s kind %s as a Subject of EnvironmentRoleBinding %s: %s", userName, userKind, name, err)
 					}
+					break
```
jtnord: no point looping when we have found the match.

pkg/util/strings.go
```diff
@@ -176,7 +176,7 @@ func DiffSlices(oldSlice []string, newSlice []string) ([]string, []string) {
 
 	for _, name := range oldSlice {
 		if StringArrayIndex(newSlice, name) < 0 {
-			toDelete = append(toInsert, name)
+			toDelete = append(toDelete, name)
```
jtnord: the fix :-o

PR #3523, opened 2019-04-03
feature: hiearchical config for prow.

pmuir:
Note that this is still under development, but want to get this part merged so that keeping up with trunk is easier

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/pipelinescheduler/config.go
+++ b/pkg/pipelinescheduler/config.go
@@ -1,0 +1,395 @@
+package pipelinescheduler
+
+import (
+	"time"
+)
+
+// SchedulerLeaf is a wrapper around Scheduler that can also store the org and repo that the scheduler is defined for
+type SchedulerLeaf struct {
+	*Scheduler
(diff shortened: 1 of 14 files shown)


pkg/pipelinescheduler/config.go
```diff
@@ -0,0 +1,388 @@
...
+// configurations in the parent scheduler
+type Presubmits struct {
+	// Items are the Presubmit configurtations
+	Items []*Presubmit
+	// Replace the existing items
+	Replace bool
+}
+
+// Presubmit defines a job configuration for pull requests
+type Presubmit struct {
```
ccojocar: Is this copied from `prow/jobs.go`? It will be challenging to keep it in sync. I am wondering if we can use directly the prow package here.

pmuir: It's similar but not the same. We can't reuse as (a) the prow config is a mess and (b) the structure is different.

PR #3547, opened 2019-04-04
fix(helm): helm template splitting on empty sections

polothy:
#### Submitter checklist

#### Description

When using Helm Template (tillerless) feature, it can produce invalid YAML files when sections of the YAML file being split apart contain only newlines or comments.

Example of error:

```
[unable to decode "/tmp/helm-template-workdir-992343983/jx/output/cert-manager/charts/cert-manager/charts/webhook/templates/part0-rbac.yaml": Object 'Kind' is missing in '{"metadata":{"annotations":{"jenkins.io/chart":"cert-manager"},"labels":{"jenkins.io/chart-release":"jx","jenkins.io/version":"2"}}}']
```

If you looked inside of something like `part0-rbac.yaml` you would see just JX annotations/labels but no valid k8s object.

The offending line: https://github.com/jetstack/cert-manager/blob/9757b2aed1144b8d57d74848675ecdcc069942d6/deploy/charts/cert-manager/webhook/templates/rbac.yaml#L2

None.

None.


polothy:
Thanks @jstrachan !

PR #3559, opened 2019-04-05
WIP Adding tests to the feature: hiearchical config for prow

dgozalo:
#### Submitter checklist

#### Description

This is just a PR that completes the tests from the feature developed by @pmuir and adds some minor refactoring.

This is Work in progress.

Not so sure about the inclusion of the `spew` dependency, but it's the only one that I've found that is capable of dereferencing a struct of pointers, otherwise it would look like a bunch of memory addresses instead of the data of the struct.


rawlingsj:
/hold

PR #3584, opened 2019-04-09
Stop copying workspace as output for final tasks

joseblas:
#### Submitter checklist

#### Description
We're currently specifying an output resource for every Task, but we don't need that for a "final" Task - i.e., any Task that is not specified as runAfter or from on any other Task. We should change that to not have PipelineTask outputs for those cases, and to have a null outputs on the Task spec itself. This would also mean that the final Task in a parallel branch shouldn't have outputs either, since we don't merge together workspaces from parallel tasks.

fixes #


dwnusbaum:
/ok-to-test

@joseblas I think you need to rebase the PR onto master so it doesn't look like you made 25 commits.

pkg/tekton/syntax/pipeline.go
```diff
@@ -1114,23 +1116,54 @@ func (j *ParsedPipeline) GenerateCRDs(pipelineIdentifier string, buildIdentifier
...
 		tasks = append(tasks, linearTasks...)
-		p.Spec.Tasks = append(p.Spec.Tasks, createPipelineTasks(stage, p.Spec.Resources[0].Name)...)
+		p.Spec.Tasks = append(p.Spec.Tasks, pipelineTasks...)
 		structure.Stages = append(structure.Stages, stage.getAllAsPipelineStructureStages()...)
 	}
 
 	return p, tasks, structure, nil
 }
 
+func wipeOutput(stage *transformedStage, taskName string, index int, tasksLen int, isLastStage bool) bool {
```
abayer: I'd probably name this something like `shouldRemoveWorkspaceOutput`.

pkg/tekton/syntax/pipeline.go
```diff
@@ -1114,23 +1116,54 @@ func (j *ParsedPipeline) GenerateCRDs(pipelineIdentifier string, buildIdentifier
...
 		previousStage = stage
 
+		pipelineTasks := createPipelineTasks(stage, p.Spec.Resources[0].Name)
+
 		linearTasks := stage.getLinearTasks()
-		for _, lt := range linearTasks {
+
+		for index, lt := range linearTasks {
+			wipeOutput := wipeOutput(stage, lt.Name, index, len(linearTasks), isLastStage)
+			if wipeOutput {
```
abayer: Could just be `if wipeOutput(...) {`

joseblas: I'd rather go for the variable, is more clear. But I'll change it.

joseblas:
@abayer @dwnusbaum I've updated the branch with the latest changes from master (some conflicts)

rawlingsj:
/hold

abayer:
Thanks, @joseblas!

PR #3585, opened 2019-04-09
Support batching from prow and pipelinerunner fixes from Andrew B

rawlingsj:


rawlingsj:
@jstrachan since my last rebase I need to fix some tests - am on it ATM should push soon.

PR #3589, opened 2019-04-09
fix(gke): GetCurrentProject can be multi-line causing error(s)

deanesmith:
#### Description
`gcloud config get-value project` returns a multi-line value which includes the current configuration and then the project ID.  GetCurrentProject() is not handling this.

None reported yet


rawlingsj:
/hold

PR #3612, opened 2019-04-11
feat: replace vault: references with secrets

pmuir:
#### Submitter checklist

fixes #

### DIFF
--- a/pkg/helm/helm_helpers.go
+++ b/pkg/helm/helm_helpers.go
@@ -488,6 +488,10 @@ type InstallChartOptions struct {
 	Username    string
 	Password    string
 	VersionsDir string
+	InstallOnly bool
+	NoForce     bool
+	Wait        bool
+	UpgradeOnly bool
 }
 
 // InstallFromChartOptions uses the helmer and kubeClient interfaces to install the chart from the options,
@@ -519,6 +523,11 @@ func InstallFromChartOptions(options InstallChartOptions, helmer Helmer, kubeCli
 	if err != nil {
 		return errors.WithStack(err)
 	}
+	cleanup, err := DecorateWithSecrets(&options, vaultClient)
+	defer cleanup()
+	if err != nil {
+		return errors.WithStack(err)
+	}
 	if options.Ns != "" {
 		annotations := map[string]string{"jenkins-x.io/created-by": "Jenkins X"}
 		kube.EnsureNamespaceCreated(kubeClient, options.Ns, nil, annotations)
@@ -528,7 +537,13 @@ func InstallFromChartOptions(options InstallChartOptions, helmer Helmer, kubeCli
 		return errors.Wrap(err, "failed to convert the timeout to an int")
 	}
 	helmer.SetCWD(options.Dir)
-	return helmer.UpgradeChart(chart, options.ReleaseName, options.Ns, options.Version, true, timeout, true, false, options.SetValues, options.ValueFiles, options.Repository, options.Username, options.Password)
+	if options.InstallOnly {
+		return helmer.InstallChart(chart, options.ReleaseName, options.Ns, options.Version, timeout,
+			options.SetValues, options.ValueFiles, options.Repository, options.Username, options.Password)
+	}
+	return helmer.UpgradeChart(chart, options.ReleaseName, options.Ns, options.Version, !options.UpgradeOnly, timeout,
+		!options.NoForce, options.Wait, options.SetValues, options.ValueFiles, options.Repository,
+		options.Username, options.Password)
 }
 
 // HelmRepoCredentials is a map of repositories to HelmRepoCredential that stores all the helm repo credentials for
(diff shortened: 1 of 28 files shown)


pkg/vault/helpers.go
```diff
@@ -94,3 +98,46 @@ func ReadBasicAuth(client Client, path string) (*config.BasicAuth, error) {
...
 	}, nil
 }
+
+// ToURI constructs a vault: URI for the given path and key
+func ToURI(path string, key string) string {
+	return fmt.Sprintf("vault:%s:%s", path, key)
+}
+
+// ReplaceURIs will replace any vault: URIs in a string, using the vault client
+func ReplaceURIs(s string, client Client) (string, error) {
```
ccojocar: I would be nice to have some unit tests for this function.

pkg/vault/helpers.go
```diff
@@ -94,3 +98,46 @@ func ReadBasicAuth(client Client, path string) (*config.BasicAuth, error) {
...
+	var err error
+	answer := vaultURIRegex.ReplaceAllStringFunc(s, func(found string) string {
+		// Stop once we have an error
+		if err == nil {
+			pathAndKey := strings.TrimSuffix(strings.TrimPrefix(strings.TrimPrefix(found, "vault:"), "\""), "\"")
```
ccojocar: You can use only:
```Go
strings.Trim(strings.TrimPrefix(found, "vault:"), "\"")
```
It will trim both suffix and prefix.

pkg/vault/helpers.go
```diff
@@ -94,3 +98,46 @@ func ReadBasicAuth(client Client, path string) (*config.BasicAuth, error) {
...
+			parts := strings.Split(pathAndKey, ":")
+			if len(parts) != 2 {
+				err = errors.Errorf("cannot parse %s as path:key", pathAndKey)
+				return ""
+			}
+			secret, err1 := client.Read(parts[0])
+			if err1 != nil {
+				err = err1
+				return ""
+			} else {
```
ccojocar: You don't need this `else`. It make the code more readable.

pkg/vault/helpers.go
```diff
@@ -94,3 +98,46 @@ func ReadBasicAuth(client Client, path string) (*config.BasicAuth, error) {
...
+				if v, ok := secret[parts[1]]; !ok {
+					err = errors.Errorf("unable to find %s in secret at %s", parts[1], parts[0])
+					return ""
+				} else {
```
ccojocar: Also you don't need this `else`.

pkg/vault/helpers.go
```diff
@@ -94,3 +98,46 @@ func ReadBasicAuth(client Client, path string) (*config.BasicAuth, error) {
...
+		if err == nil {
+			pathAndKey := strings.TrimSuffix(strings.TrimPrefix(strings.TrimPrefix(found, "vault:"), "\""), "\"")
+			parts := strings.Split(pathAndKey, ":")
+			if len(parts) != 2 {
+				err = errors.Errorf("cannot parse %s as path:key", pathAndKey)
+				return ""
+			}
+			secret, err1 := client.Read(parts[0])
+			if err1 != nil {
+				err = err1
```
ccojocar: I would wrap this error with more context.

pkg/vault/helpers.go
```diff
@@ -94,3 +98,46 @@ func ReadBasicAuth(client Client, path string) (*config.BasicAuth, error) {
...
+				return ""
+			} else {
+				if v, ok := secret[parts[1]]; !ok {
+					err = errors.Errorf("unable to find %s in secret at %s", parts[1], parts[0])
+					return ""
+				} else {
+					result, err1 := util.AsString(v)
+					if err1 != nil {
+						err = err1
```
ccojocar: I would wrap the error with some context.

pkg/vault/helpers.go
```diff
@@ -94,3 +98,46 @@ func ReadBasicAuth(client Client, path string) (*config.BasicAuth, error) {
...
+						return ""
+					}
+					return result
+				}
+			}
+		}
+		return found
+	})
+	if err != nil {
+		return "", err
```
ccojocar: I would wrap this error with some context.

pkg/helm/helm_helpers.go
```diff
@@ -541,6 +556,45 @@ type HelmRepoCredential struct {
...
+func DecorateWithSecrets(options *InstallChartOptions, vaultClient vault.Client) (func(), error) {
+	cleanup := func() {
+	}
+	if vaultClient != nil {
+		newValuesFiles := make([]string, 0)
+		cleanup = func() {
+			for _, f := range newValuesFiles {
+				err := util.DeleteFile(f)
+				if err != nil {
+					log.Errorf("Deleting temp file %s\n", f)
```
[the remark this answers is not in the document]
pmuir: because you are going to defer the cleanup() method almost certainly, so you just push the work on to the caller. Neither approach is great.

pkg/vault/helpers.go
pmuir: I did do a test but for the helm func, I'll copy it in here.

pkg/vault/helpers.go
pmuir: Hound thought that too, but I do as I need the value of `v` in the else. Bug in the linter ;-)

PR #3616, opened 2019-04-11
fix: lets return the error if a git command fails

jstrachan:


rawlingsj:
/hold

PR #3618, opened 2019-04-11
switch logging from custom Jenkins X package to logrus to standardise…

rawlingsj:


rawlingsj:
> One thing we might consider is to create our own internal "GetLogger" helper

Yeah nice idea, that way we can also include timestamps etc.  I'll see if I can do that in a parallel PR now but I've not got long left today and want to get this merged as it's blocking other PRs to avoid conflicts.

rawlingsj:
lets see if there's a better way to do this, we could change the Jenkins X log to use logrus and change the logging template there.
/hold

rawlingsj:
I think based @hferentschik comment and a question about switching between JSON and text formatted logs whether or not in-cluster from @garethjevans earlier, we could do better here.  If I don't get it done in time I'll abandon this PR and unhold the other PRs on the repo.

rawlingsj:
closed in preference to https://github.com/jenkins-x/jx/pull/3624 thanks for the feedback @hferentschik :)

PR #3675, opened 2019-04-18
fix: jenkins-x.yml agent.container was getting cleared for buildpacks

abayer:
#### Submitter checklist

#### Description

That was wrong. Don't do that.

/assign @jstrachan 
/assign @dwnusbaum 

n/a


abayer:
/hold

Putting on hold because I want @jstrachan to make sure the test case fits his issue.

abayer:
/hold cancel

So @jstrachan's issue was actually that his YAML had `agent.container` in the wrong place in `jenkins-x.yml` - which is a different issue (we don't do strict unmarshalling so bogus keys just are silently ignored). But I added an additional test anyway. =)

PR #3737, opened 2019-04-24
feat: install Istio using the version passed in the cmd line

carlossg:
Allows passing `--version x.x.x` instead of using always latest

#### Submitter checklist

#### Description

Currently istio is always installed using `latest`. This patch honors the already existing options `--version` and `-v` of `jx create addon istio`

This does not use the `jenkins-x-versions` repo unfortunately as it would require some refactoring, given the istio release is downloaded before the versions repo is cloned, suggestions are welcome.

fixes #


carlossg:
build is fixed now

Issue #3752, opened 2019-04-25
creating a new ng gke cluster fails on windows


jtnord:
### Summary 

Creating a new cluster on gke fails on windows.

### Steps to reproduce the behavior

on windows run `jx create cluster gke --cluster-name=plat-shared-rbac --ng=true`

### Expected behavior

it works

### Actual behavior

it eventually bombs out

```
Wrote chart values.yaml C:\workarea\source\github\teilo-org\jenkins-x-dev-environment\env\values.yaml generated from directory tree
generated helm C:\workarea\source\github\teilo-org\jenkins-x-dev-environment\env\values.yaml

controllerbuild:
  "":
    enabled: true
controllerworkflow:
  "":
    enabled: false
jenkins:
  "":
    enabled: false
jenkins-x-platform:
  PipelineSecrets:
    DockerConfig: |-
      {
          "credHelpers": {
              "gcr.io": "gcr",
              "us.gcr.io": "gcr",
              "eu.gcr.io": "gcr",
              "asia.gcr.io": "gcr",
              "staging-k8s.gcr.io": "gcr"
          }
      }
  postinstalljob:
    enabled: "true"


Using values files: C:\workarea\source\github\teilo-org\jenkins-x-dev-environment\env\values.yaml, C:\workarea\source\github\teilo-org\jenkins-x-dev-environment\env\secrets.yaml, C:\workarea\source\github\teilo-org\jenkins-x-dev-environment\env\templates\jx-install-config-secret.yaml
Applying chart overrides
Copying chart override C:\workarea\source\github\teilo-org\jenkins-x-dev-environment\env\templates\dev-env.yaml
WARNING: Error copying template C:\workarea\source\github\teilo-org\jenkins-x-dev-environment\env\templates\dev-env.yaml to C:\workarea\source\github\teilo-org\jenkins-x-dev-environment\env\charts\env\templates\dev-env.yaml
Copying chart override C:\workarea\source\github\teilo-org\jenkins-x-dev-environment\env\templates\ingress-config-configmap.yaml
WARNING: Error copying template C:\workarea\source\github\teilo-org\jenkins-x-dev-environment\env\templates\ingress-config-configmap.yaml to C:\workarea\source\github\teilo-org\jenkins-x-dev-environment\env\charts\env\templates\ingress-config-configmap.yaml
Copying chart override C:\workarea\source\github\teilo-org\jenkins-x-dev-environment\env\templates\jx-install-config-configmap.yaml
[...]

PR #3753, opened 2019-04-25
create cluster fixes for windows (and others)

jtnord:
#### Submitter checklist

#### Description

I suspect that the parent folder does not exist for some reason, however if the last file to copy works then the error is swallowed and this leaves the user with no idea what is wrong.

if a template file can not be copied then the result will not be as
expected and there is little point in continuing

see #3752


pkg/jx/cmd/step_helm_apply.go
```diff
@@ -242,36 +242,47 @@ func (o *StepHelmApplyOptions) Run() error {
 func (o *StepHelmApplyOptions) applyTemplateOverrides(chartName string) error {
 	log.Infof("Applying chart overrides\n")
 	templateOverrides, err := filepath.Glob(chartName + "/../*/templates/*.yaml")
```
jstrachan: wonder if that line should be:
```
	templateOverrides, err := filepath.Glob(filepath.Join(chartName, "..", "*", "templates", "*.yaml"))
```

jtnord: possibly, not sure.
in this case it works as it goes into the loop and fails to copy the files so it did match stuff

jtnord: docs for glob says it calls [`filepath.Match`](https://golang.org/pkg/path/filepath/#Match) which has some interesting behaviour on windows.  Also if the chartName contains special characters like `-` then they should be escaped (I think the chartname may contain a dash.

pkg/jx/cmd/step_helm_apply.go
```diff
@@ -242,36 +242,47 @@ func (o *StepHelmApplyOptions) Run() error {
...
+			depChartName := writeTemplateParts[len(writeTemplateParts)-3]
+			templateName := writeTemplateParts[len(writeTemplateParts)-1]
+			depChartDir := filepath.Join(depChartsDir, depChartName)
+			// If the chart directory does not exist explode the tgz
+			exists, err := util.DirExists(depChartDir)
+			if err != nil {
+				return errors.Wrapf(err, "Checking directory %s", depChartDir)
+			}
+			if !exists {
+				chartArchives, _ := filepath.Glob(filepath.Join(depChartsDir, depChartName+"*.tgz"))
```
jtnord: this is the existing logic - but if it enters here it will not unpack the archive as it was expecting to do as the Glob will not match anything as the directory does not exist??!!!

pkg/jx/cmd/step_helm_apply.go
jtnord: `-` only needs escaping in a choice `[...]`

pkg/jx/cmd/step_helm_apply.go
jtnord: now I could just change the logic here - but the comment on [line 259](https://github.com/jenkins-x/jx/pull/3753/files#diff-ee8c207b832edc0919065b6959cbfdcbR259) implied that this should be unpacking something if the directory does not exist - so in that case the question becomes where is this thing that should be unpacked?

@jstrachan any ideas here?

pkg/version/version_data.go
```diff
@@ -57,6 +57,11 @@ type StableVersion struct {
 // LoadStableVersion loads the stable version data from the version configuration directory returning an empty object if there is
 // no specific stable version configuration available
 func LoadStableVersion(wrkDir string, kind VersionKind, name string) (*StableVersion, error) {
+	if filepath.IsAbs(name) {
```
jtnord: this should be the thing that allows us to install correctly - however it is impossible to test at the moment (on either  OS-X or windows) as the vault installation is failing.

hferentschik:
What's the state of this pull request?

ccojocar:
@jtnord Can we remove the hold on this PR? Is anything left to be addressed?

jtnord:
I have never been able to test this as cluster creation was still failing even after my changes due to further issues.

the fix fixed at least an observed issue - but at the same time I am not convinced that this was not just putting a sticky plaster over a previous issue and there was something more nefarious going on.

I am no longer working on this - so feel free to take over.

ccojocar:
@jtnord Thanks for update. I am not sure who can continue working. How much work is still left? Is it just validating that it work with Windows 10?

jtnord:
@ccojocar so there a a couple of things I think need to happen.

1) the git repo for cluster gitops (not environment) was being created in the CWD - this is different to the ones for creating environment repos - so this could be the source of the original issue.   I don't know what the expectations exactly where around where this should be created, but certainly this fix should be looked at carefully by someone that knows that to ensure it is not just masking an earlier problem.

2) yes - I would say that cluster creation needs to be attempted / tested on windows 10 (both with and without this fix; as it is now a bit old and I am guessing things in this area may have changed somewhat)

hferentschik:
What shall we do with this PR now? It seems it is out of date. I suggest closing it and creating a new one once the code is rebased and ready to be merged. @jtnord wdyt?

jtnord:
to be clear I am no longer working on this, thus I have stake in the way forward.

hferentschik:
I am going to close this PR.

PR #3790, opened 2019-04-29
Fix #3008: use rev-list if git describe doesn't find a commit hash

### LINKED ISSUE #3008: jx step changelog does not find commit hash if tag already exists for imported projects
### Summary 
Issue in https://github.com/jenkins-x/jx/issues/2934 is not fixed.  jx step changelog works if all the tags were created by jx.  If there are existing tags referring to commits in the master branch, this will fail.
### Expected behavior
```
++ cat ../../VERSION
+ jx step changelog --version v5.3.58
Using batch mode as inside a pipeline
Generating change log from git ref 6809c62fe648241816733687309ef1e06e76a712 => c2e1b808b30a350734a1e32b68f87fc36a66607e
Unable to find user: jenkins-x.googlegroups.com -- users.jenkins.io "jenkins-x.googlegroups.com" not found
Unable to find user: jenkins-x.googlegroups.com -- users.jenkins.io "jenkins-x.googlegroups.com" not found
```
### Actual behavior
```
++ cat ../../VERSION
+ jx step changelog --version v5.3.57
Using batch mode as inside a pipeline
Generating change log from git ref 5.3.55 => 14372096fc7930810876ec69281fc39c7b3b43cd
error: object not found

script returned exit code 1
```

### Jx version
1.3.831

```
COPY OUTPUT HERE
```

### Jenkins type

- [x] Classic Jenkins
- [ ] Serverless Jenkins

hangchan:
#### Submitter checklist

fixes # 3008

### DIFF
--- a/pkg/gits/git_cli.go
+++ b/pkg/gits/git_cli.go
@@ -660,7 +660,12 @@ func (g *GitCLI) GetPreviousGitTagSHA(dir string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("failed to find previous tag for project in %s : %s", dir, err)
 	}
+
+	if len(previousTag) != 40 {
+		return g.gitCmdWithOutput(dir, "rev-list", "--tags", "--skip=2", "--max-count=1")	
+	}
 	return previousTag, err
+
 }
 
 // GetRevisionBeforeDate returns the revision before the given date


pkg/gits/git_cli.go
```diff
@@ -660,7 +660,12 @@ func (g *GitCLI) GetPreviousGitTagSHA(dir string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("failed to find previous tag for project in %s : %s", dir, err)
 	}
+
+	if len(previousTag) != 40 {
```
hferentschik: I am confused. In which cases would _previousTag_ ever be a sha. IMO the whole function is broken.
AFAIK, `git describe --abbrev=0 --debug --tags --always` will always only display tag names, no shas. From the [docs](https://git-scm.com/docs/git-describe):

> The command finds the most recent tag that is reachable from a commit. If the tag points to the commit, then only the tag is shown.

Only if it is not a tag or the `--long` option is specified you get a human-readable object name of the form "v0.0.1-1-g23fece4" where the part after the 'g' is the 7 char abbreviation of the sha.

So either one needs to use the `--long` options and extract the abbreviated sha using a regexp or similar or one needs to always determine the actual sha for `previousTag` given that it is a tag name. Something equivalent to:

```
$ git rev-list < previousTag>
```

Also for the case where there is no previous tag I would think this function throws an error. Is this what we want?

ccojocar:
@hangchan @hferentschik Should we close this PR?

hferentschik:
+1

hferentschik:
Closing this pull request in favour of the discussion on issue #3866 and pull request #4027.

PR #3793, opened 2019-04-29
fix: adding local app fails to build dependencies

carlossg:
helm dependency build is ran in the wrong dir

#### Submitter checklist

fixes #


pkg/helm/helm_helpers.go
```diff
@@ -449,13 +449,25 @@ func CombineValueFilesToFile(outFile string, inputFiles []string, chartName stri
 	return nil
 }
 
+// IsLocal returns whether this chart is being installed from the local filesystem or not
+func IsLocal(chart string) bool {
+	b := strings.HasPrefix(chart, "/") || strings.HasPrefix(chart, ".") || strings.Count(chart, "/") > 1
```
hferentschik: Will this work on Windows? Maybe a combination of filepath.Abs and util.FileExists would be better. If the file exist it is local, otherwise not.

carlossg: that was there before, I haven't changed it
the `FileExists` check is in the following lines

hferentschik: Ahh, I see. You extracted it from InspectChart. Fair enough. And yes, it does the util.FileExists. I still argue that this line is not quite right and will fail on Windows.

I'd replace this with:

```go
absolutePath, err := filepath.Abs(chart)
if err != nil {
   // maybe a log message here as well
   return false
}

exists, err := util.FileExists(absolutePath)
if err != nil {
	return false
}
return exists
```

Don't you think this is better?

carlossg: I don't know if it makes any difference unless you'd want to log the absolute path, if I'm passing a relative path `FileExists` would return correctly.

hferentschik: So what this code doing atm? It is supposed to be some (Unix) shortcut to say whether the string is a local path, right? It if starts with '/' it is most likely an absolute path the same with '.'. But why would it also return `true` for `strings.Count(chart, "/") > 1`? This condition will be `true` for pretty much any URL as well, right?

carlossg: A remote chart would be `something/something` so I'm guessing the intent was that 1 `/` is ok, more than one is a path.
I don't think you can pass urls to install charts.

pmuir: Carlos is right, it's an attempt to detect when it's not a chart name (e.g. `cheddar`) or a repo/name (e.g. `cheese/cheddar`). Making the path absolute is somewhat irrelevant here as, if the user supplies the path, it has to be correct for the chart to be located - Go will correct resolve the path without us fiddling with it. To improve Windows support we might want to detect if backslashes are used or if the path starts with a drive name. This patch is worth adding regardless.

hferentschik: Absolute path or not, point being is that this method will return true for URLs as well. This code is not a proper check whether a specified string is a local path/file. Also it hard codes the path separator, making it only work on Unix like systems. In my book this method is buggy it should be fixed.

PR #3819, opened 2019-05-01
Update README.md to add codescene reporting

michaelneale:
hey @ccojocar @rawlingsj found this near tool called "codescene" which gives all sorts of analysis around code health, hostspots and things that have a lot of churn or need refactoring. 

Thought it is worth a try (at the moment it looks at a few repos)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 # JX 
 
+JX is a command line tool for installing and using [Jenkins X](http://jenkins-x.io/)
+
+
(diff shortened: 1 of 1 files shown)


michaelneale:
what it looks like: 

Screen Shot 2019-05-01 at 4 03 49 pm

ccojocar:
It looks interesting! I like the hotsports. They seems to be relevant :). Does it integrate with the repositories as a GitHub application?

michaelneale:
@ccojocar I think it might be an oauth app - not sure (can't seem to see where it went as an install app)

michaelneale:
yes it is an oauth app (I approved it, I can see it there now)

PR #3841, opened 2019-05-02
chore: Initial work on unification of build pack and jenkins-x.yml syntax

abayer:
#### Submitter checklist

#### Description

Here specifically, we're getting rid of `PipelineAgent` and
`PipelineStep`, so that you can use the `jenkins-x.yml` syntax for
agents and steps within a build pack or for overriding/adding to a
build pack in `jenkins-x.yml`. The existing syntax still works in
build pack context, but can't be used when defining the whole pipeline
- you'll get validation errors telling you what you should do
instead.

More is likely to come - but since this is functional as is with no
backwards compatibility concerns that I can find, hey, might as well
toss it up.

cc @sharepointoscar 
/assign @jstrachan 
/assign @pmuir 
/assign @rawlingsj 
/assign @dwnusbaum 

n/a

### DIFF
--- a/pkg/tekton/syntax/pipeline_test.go
+++ b/pkg/tekton/syntax/pipeline_test.go
@@ -1737,7 +1737,7 @@ func PipelineOptionsRetry(count int8) PipelineOptionsOp {
 // PipelineEnvVar add an environment variable, with specified name and value, to the pipeline.
 func PipelineEnvVar(name, value string) PipelineOp {
 	return func(parsed *syntax.ParsedPipeline) {
-		parsed.Environment = append(parsed.Environment, syntax.EnvVar{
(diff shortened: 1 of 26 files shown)


pkg/tekton/syntax/pipeline_test.go
```diff
@@ -1363,13 +1363,65 @@ func TestFailedValidation(t *testing.T) {
...
+			name: "agent_container_field",
+			expectedError: (&apis.FieldError{
+				Message: "the container field is deprecated - please use image instead",
+				Paths:   []string{"container"},
+			}).ViaField("agent"),
+		},
 	}
 
 	for _, tt := range tests {
-		if tt.name != "multiple_stages" {
```
dwnusbaum: Good catch!

Issue #3846, opened 2019-05-03
Add --create option to namespace command which will create non existent


hferentschik:
### Summary 

Allow for creating namespace when using `jx ns <name>`.

### Steps to reproduce the behavior

There is no such option yet.

### Expected behavior

```
$ jx ns --create foo
```

Selects the namespace, potentially creating it if it does not exist.

hferentschik:
Merged via pull request #3847

PR #3847, opened 2019-05-03
feat: adding ability to create namespace via 'jx ns --create <name>

hferentschik:
fixes issue #3846

#### Submitter checklist

#### Description

Adds a `--create` option to `jx ns <name>` and adds some tests around the namespace command.

### DIFF
--- a/pkg/jx/cmd/namespace.go
+++ b/pkg/jx/cmd/namespace.go
@@ -2,14 +2,15 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/pkg/errors"
 
 	"github.com/jenkins-x/jx/pkg/kube"
-	"github.com/pkg/errors"
 
 	"github.com/spf13/cobra"
 
 	"github.com/jenkins-x/jx/pkg/jx/cmd/opts"
 	"github.com/jenkins-x/jx/pkg/jx/cmd/templates"
+	api_errors "k8s.io/apimachinery/pkg/api/errors"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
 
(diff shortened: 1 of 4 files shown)


ccojocar:
I am wondering why not using the `kubectl create ns <name> instead of the --create option. Are there any benefits in duplicating it in jx CLI?

hferentschik:
> I am wondering why not using the kubectl create ns <name> instead of the --create option. Are there any benefits in duplicating it in jx CLI?

Would the "what for" argument not apply for the namespace command in general? I am a bit concerned in general about the various non-core helper commands in `jx`. I'd argue the whole command should maybe not be there. This might also reduce potential confusion for users between concepts of pure Kubernetes namespaces and a jx environment. 

That said, I see how `jx ns` is a bit more convenient to use compared to plain kubectl and I added the flag for completeness and because I missed ` kubectl create ns <name>`. I thought for creating
a namespace I need to use the `-f` option and provide a namespace spec. Compared to this the additional flag on `jx` was much easier. 

I feel the flag does not hurt and I did a bit of code cleanup as well and added some tests.

ccojocar:
It sounds good to me! I just wanted to raise awareness when some functionality gets duplicated form kubectl or from some other popular CLI. It is fine If there is a good reason, otherwise all these features will require some maintenance and bug fixing.

hferentschik:
> I just wanted to raise awareness when some functionality gets duplicated form kubectl or from some other popular CLI. It is fine If there is a good reason, otherwise all these features will require some maintenance and bug fixing.

Completely agree. I was wondering whether for the sake of clarity it would make sense to move this pure helper commands (for example `jx ns`) into a CLI extension app. Something like `jx-kubectl-helpers`. This would reduce the scope of `jx` and increases the clarity imo. AFAIU the capabilities to extend the CLI via Apps is there. This might be a use case for it.  That's a different discussion though. Just putting out the idea.

PR #3848, opened 2019-05-03
fix: lets add a FakeFactory for easier testing

jstrachan:
without requiring mocks

### DIFF
--- a/pkg/jx/cmd/clients/fake/fake_factory.go
+++ b/pkg/jx/cmd/clients/fake/fake_factory.go
@@ -40,10 +40,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	metricsclient "k8s.io/metrics/pkg/client/clientset/versioned"
-
-	// this is so that we load the auth plugins so we can connect to, say, GCP
-
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
 type fakefactory struct {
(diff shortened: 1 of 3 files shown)


pkg/jx/cmd/clients/fake/fake_factory.go
```diff
@@ -0,0 +1,405 @@
+package fake
```
ccojocar: Any reason not to use https://github.com/jenkins-x/jx/blob/master/pkg/jx/cmd/clients/mocks/factory.go? 

It is generated from the interface.

jstrachan: I wanted to use fakes rather than mocks

ccojocar: I think we need to be careful with the fake to not instantiate any real clients in the unit tests. Also it will require manually maintenance every time the factory interface gets modified. 

I would add some docs at the top when to use fake vs mock (e.g. for which tests case it is suited).

jstrachan: yeah - its gonna take some maintenance as we add methods to the factory interface - though the build will fail so we'll catch it.

good point on the docs though; I'm a huge fan of fake versus mocks for k8s/CRD related stuff - so much easier to write, less brittle and whatnot

ccojocar: I think for CRDs/k8s resources fake is great because of the k8s fake clients. I believe we can also instantiate the mock with fake clients only for CRDs resources. Probably we will need some boilerplate for this to make it easy for tests to use it.

PR #3850, opened 2019-05-03
feat: Allow overriding/removing steps from build pack by name

abayer:
#### Submitter checklist

#### Description

See [pkg/jx/cmd/test_data/step_create_task/override-step/jenkins-x.yml](pkg/jx/cmd/test_data/step_create_task/override-steps/jenkins-x.yml)
for example syntax.

This allows replacing named steps from build packs by pipeline, stage
(aka lifecycle - I opted to call it stage since that's more consistent
with jenkins-x.yml syntax and eventually I'll make the lifecycles into
stages, albeit ones that share a task/pod), and step name. If no
pipeline is specified, the stage/lifecycle and step name are replaced
in all pipelines, with similar logic for stage/lifecycle.

Steps can be replaced with a single step, multiple steps, or no step
at all to remove them.

/assign @jstrachan 
/assign @rawlingsj 
/assign @dwnusbaum 
cc @sharepointoscar 

n/a


abayer:
fwiw, this and #3841 overlap a fair amount so when one of them gets merged, I'll have to clean up the other. Won't be hard - just changing struct names around.

abayer:
oops - this may actually break if it's merged at the same time as #3841 - if that happens, I'll get a PR up fixing it immediately.

PR #3877, opened 2019-05-07
 Enable gcs buckets as default storage for Jenkins X when using GCP

### LINKED ISSUE #3815: Enable gcs buckets as default storage for Jenkins X when using GCP
### Summary 

When you install Jenkins X with GKE, you should get long term long term storage of logs, using GCS, out of the box.

Things we need to do:

0) Make a question to ask if you want this
1) During the install create the bucket, and set any permissions needed (probably none)
2) During the install, enable storage of logs in gcs using team settings (right now it's

```
jx edit storage -c default  --bucket-url gs://<bucket name>
jx edit storage -c logs  --bucket-url gs://<bucket name>
```

3) Write a bdd test to ensure that this is actually happening!

dgozalo:
…the Google Cloud Storage permissions are set

feature: added a way to provide the name of the long term storage bucket and fall back to a random generated one if not provided

feature: added long term storage when installing Jx, also makes sure the Google Cloud Storage permissions are set

#### Submitter checklist

#### Description
Added a way to enable Long Term Storage for a Jenkins X installation on GKE by providing the flag `--long-term-storage`. 

It also adds the posibility to provide the target bucket, it will do the following checks to it:

- Checks if it exists within the current GCP project
- If it does, just use that one
- If it doesn't, try to create it, if it fails generate a random named one

Note: This feature needs the scope `"https://www.googleapis.com/auth/devstorage.full_control"` to be enabled in the cluster, which is added if the answer to the question `Would you like to access Google Cloud Storage / Google Container Registry?` was positive.

fixes #3815

### DIFF
--- a/pkg/jx/cmd/install.go
+++ b/pkg/jx/cmd/install.go
@@ -3,6 +3,8 @@ package cmd
 import (
 	"encoding/base64"
 	"fmt"
+	gkeStorage "github.com/jenkins-x/jx/pkg/cloud/gke/storage"
+	"github.com/jenkins-x/jx/pkg/kube/cluster"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -23,10 +25,10 @@ import (
 	kubevault "github.com/jenkins-x/jx/pkg/kube/vault"
 	"github.com/jenkins-x/jx/pkg/vault"
 
-	jenkinsio "github.com/jenkins-x/jx/pkg/apis/jenkins.io"
+	"github.com/jenkins-x/jx/pkg/apis/jenkins.io"
 
 	"github.com/jenkins-x/jx/pkg/addon"
-	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	"github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/auth"
 	"github.com/jenkins-x/jx/pkg/cloud/aks"
 	"github.com/jenkins-x/jx/pkg/cloud/amazon"
@@ -117,6 +119,8 @@ type InstallFlags struct {
 	NoGitOpsVault               bool
 	NextGeneration              bool
 	StaticJenkins               bool
+	LongTermStorage             bool
+	LongTermStorageBucketName   string
 }
 
 // Secrets struct for secrets
@@ -218,6 +222,7 @@ This repository contains the source code for the Jenkins X Development Environme
   }
 }
 `
+	longTermStorageFlagName = "long-term-storage"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/jx/cmd/install.go
```diff
@@ -117,6 +119,8 @@ type InstallFlags struct {
 	NoGitOpsVault               bool
 	NextGeneration              bool
 	StaticJenkins               bool
+	LongTermStorage             bool
```
hferentschik: Are these options needed? I thought the intent was to  per default store into a GCS bucker in case the provider is GKE?

pkg/jx/cmd/install.go
```diff
@@ -356,6 +360,8 @@ func (options *InstallOptions) addInstallFlags(cmd *cobra.Command, includesInit
 	cmd.Flags().BoolVarP(&flags.Kaniko, "kaniko", "", false, "Use Kaniko for building docker images")
 	cmd.Flags().BoolVarP(&flags.NextGeneration, "ng", "", false, "Use the Next Generation Jenkins X features like Prow, Tekton, No Tiller, Vault, Dev GitOps")
 	cmd.Flags().BoolVarP(&flags.StaticJenkins, "static-jenkins", "", false, "Install a static Jenkins master to use as the pipeline engine. Note this functionality is deprecated in favour of running serverless Tekton builds")
+	cmd.Flags().BoolVarP(&flags.LongTermStorage, "long-term-storage", "", false, "Enable the Long Term Storage option to save logs and other assets into a GCS bucket")
```
hferentschik: The way it is worded now, it is a GKE option only? In the case of Vault we use something like:

> Sets up a Hashicorp Vault for storing secrets during installation (supported only for GKE)

Maybe we should align how we present this in the help string.

pkg/jx/cmd/install.go
dgozalo: Default answer will be yes, but can always give the option to disable.

pkg/jx/cmd/install.go
```diff
@@ -2263,6 +2274,168 @@ func (options *InstallOptions) configureBuildPackMode() error {
...
+		}
+
+		isWriteEnabled, err := gke.IsGCSWriteRoleEnabled(options.installValues[kube.ClusterName],
+			options.installValues[kube.Zone])
+		if err != nil {
+			return errors.Wrap(err, "checking if we the storage write scope is enabled")
+		}
+
+		if !isWriteEnabled {
+			log.Warn("Write access to Google Cloud Storage is not enabled, skipping long term storage")
```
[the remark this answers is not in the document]
dgozalo: Probably, if they can't use long term storage but they didn't see this error when installing, it could be misleading as they will think the logs will be in the bucket. 

So we will just fail fast when this happens.

dgozalo:
> basically looks good! may be worth adding a bdd test too?

Yes, was the next step after fixing all the comments.

PR #3878, opened 2019-05-07
Fix the lazily created ecr docker registry organisation

amoussawi:
#### Submitter checklist

#### Description
This PR fixes this issue:
- When importing a project into an aws cluster, the wrong organisation is passed to the LazyCreateRegistry function, where the the git org is passed instead of the docker registry org.

fixes #

### DIFF
--- a/pkg/jx/cmd/import.go
+++ b/pkg/jx/cmd/import.go
@@ -989,7 +989,7 @@ func (options *ImportOptions) ensureDockerRepositoryExists() error {
 		dockerRegistry := cm.Data["docker.registry"]
 		if dockerRegistry != "" {
 			if strings.HasSuffix(dockerRegistry, ".amazonaws.com") && strings.Index(dockerRegistry, ".ecr.") > 0 {
-				return amazon.LazyCreateRegistry(kubeClient, ns, region, dockerRegistry, orgName, appName)
+				return amazon.LazyCreateRegistry(kubeClient, ns, region, dockerRegistry, options.getDockerRegistryOrg(), appName)
 			}
 		}
 	}


pkg/jx/cmd/import.go
```diff
@@ -989,7 +989,7 @@ func (options *ImportOptions) ensureDockerRepositoryExists() error {
 		dockerRegistry := cm.Data["docker.registry"]
 		if dockerRegistry != "" {
 			if strings.HasSuffix(dockerRegistry, ".amazonaws.com") && strings.Index(dockerRegistry, ".ecr.") > 0 {
-				return amazon.LazyCreateRegistry(kubeClient, ns, region, dockerRegistry, orgName, appName)
+				return amazon.LazyCreateRegistry(kubeClient, ns, region, dockerRegistry, options.getDockerRegistryOrg(), appName)
```
hferentschik: Good catch, _DockerRegistryOrg_ is really not taken into account at all here. I think, however, that just using `options.getDockerRegistryOrg()` is not quite right either. The help for this option says:

> The name of the docker registry organisation to use. If not specified then the Git provider organisation will be used

So it not set, the git provider org should be used. So really it probably should be more like this in the beginning: 

```go
var orgName string
if options.getDockerRegistryOrg() != "" {
  orgName := options.getDockerRegistryOrg()
}
else {
  orgName := options.getOrganisationOrCurrentUser()
}
```

hferentschik: On second thought, we might even have to go a bit further. AFAIK the org/namespace is optional on ECR, so one might be explicitly set to the empty string via `DockerRegistryOrg`. There is a function [IsFlagExplicitlySet](https://github.com/jenkins-x/jx/blob/75a3e8bd1150077d832cbc0fc66bc6ea66db0c8c/pkg/jx/cmd/opts/common.go#L1087) in `CommonOptions` which can be used to check whether a given flag was explicitly set. This function should probably be used in the above if statement to determine whether `DockerRegistryOrg` should be used or the Git org (via `options.getOrganisationOrCurrentUser()`). WDYT?

hferentschik:
@amoussawi, thanks for your pull request.  Much appreciated. I added some comments. Let me know what you think.

pkg/jx/cmd/import.go
amoussawi: Replying to your first comment, if you check the the `getDockerRegistryOrg` function, it already does that.

amoussawi: Regarding your second comment, yes org is optional in ECR, and allowing empty org should be fine, but the problem is that everywhere in jenkins x project it's assumed that there's a docker registry org, so that requires modifications in a lot of places such as build packs.

PR #3879, opened 2019-05-07
Skip ECR docker repository creation if a project doesn't need a docker image/file

amoussawi:
#### Submitter checklist

#### Description
This PR fixes this issue:
- When importing a project, if there's no Dockerfile and the project doesn't require one, the code still goes and creates an ecr repository for the project, this PR skips that step in that case.

fixes #

### DIFF
--- a/pkg/jx/cmd/import.go
+++ b/pkg/jx/cmd/import.go
@@ -900,9 +900,16 @@ func (options *ImportOptions) doImport() error {
 		jenkinsfile = defaultJenkinsfileName
 	}
 
-	err = options.ensureDockerRepositoryExists()
+	dockerfileExists, err := util.FileExists("Dockerfile")
 	if err != nil {
 		return err
+	}
+
+	if dockerfileExists {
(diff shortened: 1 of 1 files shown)


pkg/jx/cmd/import.go
```diff
@@ -900,11 +900,18 @@ func (options *ImportOptions) doImport() error {
 		jenkinsfile = defaultJenkinsfileName
 	}
 
-	err = options.ensureDockerRepositoryExists()
+	dockerfileExists, err := util.FileExists("Dockerfile")
```
hferentschik: ooi, are you, in this case, using the import with the `--no-draft=true` option? Otherwise you get a `Dockerfile`, right?

Say you add a `Dockerfile` later, you would expect that the repo is then manually created?

amoussawi: Yes, I'm using `--no-draft=true`.
If I added a Dockerfile later, then yes I would manually create the repo.

PR #3891, opened 2019-05-08
feat: 'jx step create task' should create a PipelineActivity early

joseblas:
#### Submitter checklist

#### Description

it creates a PipelineActivity within *jx step create task* cmd

fixes #

### DIFF
--- a/pkg/jx/cmd/step_create_task.go
+++ b/pkg/jx/cmd/step_create_task.go
@@ -407,7 +407,13 @@ func (o *StepCreateTaskOptions) Run() error {
 		log.Infof("created tekton CRDs for %s\n", run.Name)
 	}
 
-	// output results for invokers of this command like the pipelinerunner
+	activityKey := o.GeneratePipelineActivity(ns)
+
+	if o.Verbose {
+		log.Infof(" PipelineActivity for %s created successfully", pipeline.Name)
+	}
+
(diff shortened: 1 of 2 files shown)


abayer:
Can you add `PipelineActivity` expected results to `step_create_task_test.go` like the existing pipeline, structure, tasks, etc?

abayer:
/assign @jstrachan 
James, does this look like what you were aiming for?

pkg/jx/cmd/step_create_task.go
```diff
@@ -428,9 +437,48 @@ func (o *StepCreateTaskOptions) Run() error {
...
+		},
+	}
+
+	if key != nil {
+		jx, _, err := o.JXClientAndDevNamespace()
+		if err != nil {
+			log.Error("error retrieving jxclient")
+			return err
+		}
+		_, _, err = key.GetOrCreate(jx, ns)
```
abayer: So rather than actually applying the CRD here, we should return it from here, and `GenerateTektonCRDs` should return it with the other CRDs. It should then pass it to `applyPipeline` for the actual applying, same as we do with the other CRDs. That keeps the generation of the CRDs structs separate from creating them in k8s, which makes for cleaner testing, lets the `—no-apply` flag still work, etc.

PR #3908, opened 2019-05-10
fix: 'jx step create task --no-apply' gets an error: "Failed to outpu…

joseblas:
fix: 'jx step create task --no-apply' gets an error: "Failed to output Tekton CRDs: mkdir : no such file or directory"

fix: 'jx step create task --view' gets a panic error

#### Submitter checklist

#### Description
After testing locally https://github.com/jenkins-x/jx/pull/3891 I've noticed that --no-apply fails because no output was provided, so I've added a default value and a message to change that.

Besides, when --view there was a panic error,, so after CRDs creation the function is returned to avoid the problem. The tasks are already printed.

### DIFF
--- a/pkg/jx/cmd/step_create_task.go
+++ b/pkg/jx/cmd/step_create_task.go
@@ -145,7 +145,7 @@ func NewCmdStepCreateTask(commonOpts *opts.CommonOptions) *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&options.Dir, "dir", "d", "", "The directory to query to find the projects .git directory")
-	cmd.Flags().StringVarP(&options.OutDir, "output", "o", "", "The directory to write the output to as YAML")
+	cmd.Flags().StringVarP(&options.OutDir, "output", "o", "out", "The directory to write the output to as YAML. Defaults to 'out'")
 	cmd.Flags().StringVarP(&options.Branch, "branch", "", "", "The git branch to trigger the build in. Defaults to the current local branch name")
 	cmd.Flags().StringVarP(&options.Revision, "revision", "", "", "The git revision to checkout, can be a branch name or git sha")
 	cmd.Flags().StringVarP(&options.PipelineKind, "kind", "k", "release", "The kind of pipeline to create such as: "+strings.Join(jenkinsfile.PipelineKinds, ", "))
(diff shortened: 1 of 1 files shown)


pkg/jx/cmd/step_create_task.go
```diff
@@ -316,6 +316,11 @@ func (o *StepCreateTaskOptions) Run() error {
 
 	if o.NoApply {
 		o.BuildNumber = "1"
+		if o.OutDir == "" {
+			log.Infof("No output directory provided, using %s\n", util.ColorInfo("out"))
```
hferentschik: OOI, why not make `out` the default where the options are defined - https://github.com/jenkins-x/jx/blob/98a094d98c282fe4d42e533b0abda3b078538632/pkg/jx/cmd/step_create_task.go#L148? 

Or maybe even require it to be set and abort if not. 

Just trying to understand the reasoning for this option. ATM, I would think the first option would be the most consistent one.

joseblas:
@hferentschik, I've applied those changes. Could you take a look?

pkg/jx/cmd/step_create_task.go
```diff
@@ -145,7 +145,7 @@ func NewCmdStepCreateTask(commonOpts *opts.CommonOptions) *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&options.Dir, "dir", "d", "", "The directory to query to find the projects .git directory")
-	cmd.Flags().StringVarP(&options.OutDir, "output", "o", "", "The directory to write the output to as YAML")
+	cmd.Flags().StringVarP(&options.OutDir, "output", "o", "out", "The directory to write the output to as YAML. Defaults to 'out'")
```
hferentschik: 👍

pkg/jx/cmd/step_create_task.go
```diff
@@ -403,6 +403,11 @@ func (o *StepCreateTaskOptions) Run() error {
 	if err != nil {
 		return errors.Wrap(err, "failed to generate Tekton CRD")
 	}
+
+	if o.ViewSteps {
```
hferentschik: Would it not make more sense to remove

```golang
	if o.ViewSteps {
		return nil, nil, nil, nil, nil, o.viewSteps(tasks...)
	}
```

from within [GenerateTektonCRDs ](https://github.com/jenkins-x/jx/blob/master/pkg/jx/cmd/step_create_task.go#L608) and move it here:

```golang
if o.ViewSteps {
   o.viewSteps(tasks...)
   return nil
}
```

Right now the code is confusing to me, since just returning _nil_ keeps me wondering where the "viewing" part is. One needs to search for all usages of `ViewSteps` to understand.

joseblas: I guess this was created like that just to save some computing only.

hferentschik: > I guess this was created like that just to save some computing only.

Probably. Feels a bit of premature optimisation.

joseblas: indeed @hferentschik 
I've just uploaded those changes

PR #3909, opened 2019-05-10
chore: improve codegen integration and add dep tidy rule

hferentschik:


Makefile
```diff
@@ -78,6 +78,11 @@ get-test-deps: ## Install test dependencies
 	$(GO_NOMOD) get github.com/axw/gocov/gocov
 	$(GO_NOMOD) get -u gopkg.in/matm/v1/gocov-html
 
+tidy-deps: ## Cleans up dependencies
```
hferentschik: @ccojocar, I added this rule to do the dep cleanup. It makes sure the tooling dependencies which might get removed as part of the `mod tidy` get added as well. WDYT?

Makefile.codegen
```diff
@@ -15,15 +20,21 @@ test-codegen: ## Test the code geneator
...
 # Anything generated by this target should be checked in
-generate: generate-clean build-codegen generate-mocks generate-openapi generate-client ## Generate the Go code (crds, mocks, openapi, client)
+generate: generate-clean build-codegen install-generate-deps generate-mocks generate-openapi generate-client ## Generate the Go code (crds, mocks, openapi, client)
 	@$(MAKE) fmt
 	@echo "Generation complete"
 
-PEGO_MOCK := $(GOPATH)/bin/pegomock
-$(PEGO_MOCK):
-	$(GO) get github.com/petergtz/pegomock/...@v2.2.0
+install-generate-deps:
```
hferentschik: I mainly added this for convenieve in conjunction with `tidy-deps` which would otherwise require to run the full code generation which does install the tooling as well in case it is missing.

Makefile.codegen
```diff
@@ -57,8 +68,12 @@ generate-clean: clean ## Delete code generation tools
 	@rm -f $(GOPATH)/bin/gen-apidocs
 	@rm -f $(GOPATH)/bin/pegomock
 
+stash:
+	# Making sure repo has no outstanding changes
```
hferentschik: This will make sure any pre-existing modifications gets stashed prior to verify-generation-complete

Makefile
ccojocar: lgtm

PR #3927, opened 2019-05-12
Various fixes and improvements around syntax unification

### LINKED ISSUE #3925: jx create step is broken
### Summary 

@abayer  The commit 98a094d98c282fe4d42e533b0abda3b078538632 broke the functionality of `jx create step`. Now it adds an empty `agent` (unnecessary clutter) and `loop` with `null` as values for `steps` and `values` inside. Such a loop fails to run and it does not pass the validations.

On top of that, `command` (it was `sh` before) does not execute at all so even if the `loop` section is removed manually after executinh `jx create step`, it does not do anything.

If the fix is not quick, can we revert the change to `create_step.go` while waiting for reorganization to finish?

abayer:
#### Submitter checklist

#### Description

* Gets rid of superfluous `agent` and `loop` in output, adds tests to verify that we're generating what's expected. 
* Add support for `loop` in build packs and lifecycle step addition.
* Add support for `containerOptions` on `pipelineConfig` in build pack or `jenkins-x.yml`.
* Add test for `pipelineConfig.env` and lifecycle step addition.

/assign @jstrachan 
cc @vfarcic  - I'm not seeing the issue with `command` not executing, so if you could test this by hand and let me know how it looks, I'd appreciate it.

fixes #3925

### DIFF
--- a/pkg/jenkinsfile/pipeline.go
+++ b/pkg/jenkinsfile/pipeline.go
@@ -149,7 +149,7 @@ func (p *PipelineOverride) MatchesStage(name string) bool {
 // PipelineConfig defines the pipeline configuration
 type PipelineConfig struct {
 	Extends     *PipelineExtends `json:"extends,omitempty"`
-	Agent       syntax.Agent     `json:"agent,omitempty"`
+	Agent       *syntax.Agent    `json:"agent,omitempty"`
 	Env         []corev1.EnvVar  `json:"env,omitempty"`
 	Environment string           `json:"environment,omitempty"`
 	Pipelines   Pipelines        `json:"pipelines,omitempty"`
@@ -518,8 +518,7 @@ func LoadPipelineConfigAndMaybeValidate(fileName string, resolver ImportFileReso
 	pipelines.RemoveWhenStatements(jenkinsfileRunner)
 	if clearContainer {
 		// lets force any agent for prow / jenkinsfile runner
-		config.Agent.Label = ""
-		config.Agent.Container = ""
+		config.Agent = &syntax.Agent{}
 	}
 	if config.Extends == nil || config.Extends.File == "" {
 		config.defaultContainerAndDir()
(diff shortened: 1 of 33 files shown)


abayer:
Jenkinsfile creation is busted, will be back to fix that in a bit.

vfarcic:
I created the following pipeline to test the `loop` instruction.

```yaml
buildPack: go
pipelineConfig:
  agent: {}
  pipelines:
    pullRequest:
      build:
        preSteps:
        - sh: make unittest
        - loop:
            variable: ARCH
            values:
            - arch-1
            - arch-2
            - arch-3
            steps:
            - command: echo
              args:
              - this
              - is
              - a
              - loop
              - with
              - ${ARCH}
```

When I check the logs of a build it does not show any trace of the command in the `loop`. It's as if it's not executed at all.

vfarcic:
Here's another example of something that I missunderstood or that doesn't work.

```yaml
buildPack: go
pipelineConfig:
  agent: {}
  pipelines:
    pullRequest:
    env:
    - name: NAMESPACE
      value: "12345"
      promote:
        steps:
        - sh: env
```

The logs output of the `env` command does not show `NAMESPACE`. If I misunderstood the syntax, it should throw an error. If, on the other hand, I did get it right, the `env` instruction is ignored.

vfarcic:
When syntax is incorrect (e.g., when `jx step syntax validate pipeline`) there is no error in the logs of the build with `jenkins-x.yml` with the incorrect syntax. As an example, I moved `env` from the previous comment inside `pullRequest` and there was no indication in the build logs that there's anything wrong. Would it make sense to add `jx step syntax validate pipeline` as the first step of pipeline execution?

abayer:
re logging of validation errors - that's on our plate to fix up. The problem is that the generation of the CRDs from the `jenkins-x.yml` is happening in the `pipelinerunner` controller, but the specific error that results doesn't actually make it anywhere, since the build never actually can start. The plan is to actually change to `pipelinerunner` kicking off a pipeline run that does nothing but generate the CRDs and launch the CRDs as a second, chained pipeline run, so that we'll actually have build logs/results for cases where `jenkins-x.yml` is invalid.

re: `loop` and the `env` thing - could you open issues for those? This PR is just aiming at fixing `jx create step`. The `env` thing definitely should work, so if it's not, there's a bug, while `loop` theoretically should work but I wouldn't be at all surprised if using it via buildpack syntax doesn't work yet.

abayer:
I take that back on `loop` - I've now added proper support of `loop` from build packs or `preSteps` etc syntax to this PR.

abayer:
Just added `containerOptions` support in `pipelineConfig`.
/assign @rawlingsj

abayer:
@vfarcic - added https://github.com/jenkins-x/jx/pull/3927/files#diff-77585df8009c6d9a6d6af9d711a9a270 (with relevant change in the equivalent `tasks.yml` as well) to verify the env thing works right.

vfarcic:
Was this PR released? I see that it was merged but the current release (2.0.118) is still from two days ago.

rawlingsj:
We had some infra issues earlier and missed a couple of automated builds.  I've just triggered a release now so should be available in ~30 mins.

PR #3932, opened 2019-05-13
fix: 'jx step create task --no-apply' shouldn't bump version

joseblas:
Signed-off-by: Jose Blas Camacho Taboada <joseblas@gmail.com>

#### Submitter checklist

#### Description
noApply should not bump the current version

fixes #

### DIFF
--- a/pkg/jx/cmd/step_create_task.go
+++ b/pkg/jx/cmd/step_create_task.go
@@ -76,6 +76,7 @@ type StepCreateTaskOptions struct {
 	CustomLabels      []string
 	CustomEnvs        []string
 	NoApply           bool
+	DryRun            bool
 	Trigger           string
 	TargetPath        string
 	SourceName        string
@@ -156,6 +157,7 @@ func NewCmdStepCreateTask(commonOpts *opts.CommonOptions) *cobra.Command {
 	cmd.Flags().StringVarP(&options.CloneGitURL, "clone-git-url", "", "", "Specify the git URL to clone to a temporary directory to get the source code")
 	cmd.Flags().StringVarP(&options.PullRequestNumber, "pr-number", "", "", "If a Pull Request this is it's number")
 	cmd.Flags().BoolVarP(&options.NoApply, "no-apply", "", false, "Disables creating the Pipeline resources in the kubernetes cluster and just outputs the generated Task to the console or output file")
+	cmd.Flags().BoolVarP(&options.DryRun, "dry-run", "", false, "Disables creating the Pipeline resources in the kubernetes cluster and just outputs the generated Task to the console or output file, without side effects")
 	cmd.Flags().BoolVarP(&options.ViewSteps, "view", "", false, "Just view the steps that would be created")
 
 	options.AddCommonFlags(cmd)
@@ -315,7 +317,7 @@ func (o *StepCreateTaskOptions) Run() error {
 		}
 	}
 
-	if o.NoApply {
+	if o.NoApply || o.DryRun {
 		o.BuildNumber = "1"
 	} else {
 		jxClient, _, err := o.JXClient()
@@ -427,12 +429,14 @@ func (o *StepCreateTaskOptions) Run() error {
 	o.Results.PipelineRun = run
 	o.Results.Structure = structure
 
-	if o.NoApply {
+	if o.NoApply || o.DryRun {
+		log.Infof("Writing output ")
 		err := o.writeOutput(o.OutDir, pipeline, tasks, run, resources, structure, activityKey)
 		if err != nil {
 			return errors.Wrapf(err, "Failed to output Tekton CRDs")
 		}
 	} else {
+		log.Infof("Applying changes ")
 		err := o.applyPipeline(pipeline, tasks, resources, structure, run, o.GitInfo, o.Branch, activityKey)
 		if err != nil {
 			return errors.Wrapf(err, "failed to apply Tekton CRDs")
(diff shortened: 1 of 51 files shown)


pkg/jx/cmd/test_data/step_create_task/from_yaml/pipeline.yml
```diff
@@ -5,30 +5,40 @@ metadata:
   name: abayer-js-test-repo-really-long-1
   namespace: jx
 spec:
-  params: []
+  params:
```
hferentschik: @joseblas, @abayer - ooi, what is your experience in terms of maintenance of these type of tests. It is not quite labour intensive to adjust these tests as changes are made to the code? Just wondering.

pkg/jx/cmd/test_data/step_create_task/from_yaml/pipeline.yml
```diff
@@ -5,30 +5,40 @@ metadata:
...
-        name: workspace
-        resource: abayer-js-test-repo-really-long
-    runAfter:
-      - build-a-really-long-stage-name-please-but-not-too-long-thanks
-    taskRef:
-      name: abayer-js-test-repo-really-long-second-1
-status: {}
+      taskRef:
+        name: abayer-js-test-repo-really-long-second-1
+status: {}
```
hferentschik: For what's its worth, there is a newline missing. Same for some more files in this commit.

pkg/jx/cmd/step_create_task.go
```diff
@@ -1431,13 +1431,53 @@ func (o *StepCreateTaskOptions) viewSteps(tasks ...*pipelineapi.Task) error {
...
+
 func (o *StepCreateTaskOptions) setVersionOnReleasePipelines(pipelineConfig *jenkinsfile.PipelineConfig) error {
 	if o.NoReleasePrepare || o.ViewSteps {
 		return nil
 	}
 	version := ""
 
-	if o.PipelineKind == jenkinsfile.PipelineKindRelease {
+	if o.NoApply {
+		version, err := getVersionFromFile(o.Dir)
```
hferentschik: Where would the version file come from in this case? Is the idea that one can manually create the file and then it is picked up when your run `step create task` with the `--no-apply` option?

hferentschik: What is the intended semantics of  `--no-apply`. From the docs, I would assume that it does all the "normal" things, but then just does not apply the generated resources. But it seems we are not also behaving differently while creating the CRDs?

hferentschik:
@joseblas you wan to run `make fmt`. There is a Go format issue in `step_create_task.go`

hferentschik:
Seems something in the integration tests `make integration` fails as well:

```
--- FAIL: TestGenerateTektonCRDs (0.20s)
    --- FAIL: TestGenerateTektonCRDs/maven_build_pack (0.10s)
        step_create_task_test.go:252: Generated Tasks did not match expected: 
            {*v1alpha1.TaskList}.Items[0].Spec.Steps[9].Name:
                -: "promote-step13"
                +: "promote-step10"
            {*v1alpha1.TaskList}.Items[0].Spec.Steps[9].Env[?->18]:
                -: <non-existent>
                +: v1.EnvVar{Name: "VERSION", Value: "${inputs.params.version}"}
            {*v1alpha1.TaskList}.Items[0].Spec.Steps[9].Env[?->19]:
                -: <non-existent>
                +: v1.EnvVar{Name: "PREVIEW_VERSION", Value: "${inputs.params.version}"}
```

pkg/jx/cmd/step_create_task.go
joseblas: The problem to be fixed is when using *noApply* a version bump happens. That happens in this function _setVersionOnReleasePipelines_. 
In order to fix it, a new option has been added (if noApply) that will not bump the version. It will read the VERSION file that will be created by jx in the initial release. If there is no VERSION file, a default one will be added alongside a warning in the logs. 
The version bump is generated by:

`
Steps: []*syntax.Step{
					{
						Command: "jx step next-version --use-git-tag-only --tag",
						Name:    "next-version",
						Comment: "tags git with the next version",
					},
				},
`
 Having said that, the behaviour is exactly the same for creating the CRDs and the only difference is this function

pkg/jx/cmd/test_data/step_create_task/from_yaml/pipeline.yml
joseblas: Indeed, but I think in this case is worth keeping.

hferentschik: Ok

pkg/jx/cmd/step_create_task.go
hferentschik: > The problem to be fixed is when using noApply a version bump happens. That happens in this function setVersionOnReleasePipelines.

Why is that an issue? Should the version bump not only depend on the pipeline kind?

> In order to fix it, a new option has been added (if noApply) that will not bump the version.

I am still trying to understand where this becomes an issue. Is this a testing issue or is this an issue in an actual production workflow?

> It will read the VERSION file that will be created by jx in the initial release.

But we never commit the VERSION file, right?

joseblas: the problem is when you run* jx step create task --no-apply* because the version in bumped, and that shouldn't happen. No Apply should change anything.

hferentschik: I am wondering whether we are not trying to do drive to distinct things with on option. `--no-apply` should just be about whether the resources are applied to the cluster or not. No circumventing any logic. Then there could be for example a `--dry-run` with the semantics of no modifications to the source/repo. With this option enabled certain side-steps are taken to avoid modifications.

joseblas: I guess this is noapply is trying to achieve, a dry run.

abayer:
/lgtm cancel

@hferentschik and I talked about this this morning, and then I talked with @joseblas later - it feels like the right thing to do would be to add a new `--dryrun` option that basically does the same thing as `--no-apply` but without any side effects, so that the output of `--no-apply` could still be used with `kubectl apply -f ...` to apply the CRDs to the cluster in the same form they'd be in if `--no-apply` hadn't been specified.

joseblas:
> /lgtm cancel
> 
> @hferentschik and I talked about this this morning, and then I talked with @joseblas later - it feels like the right thing to do would be to add a new `--dryrun` option that basically does the same thing as `--no-apply` but without any side effects, so that the output of `--no-apply` could still be used with `kubectl apply -f ...` to apply the CRDs to the cluster in the same form they'd be in if `--no-apply` hadn't been specified.

@abayer @hferentschik, in the current impl when running --no-apply a new tag is created (and pushed). I guess the expected outcome for this is:
  -- no-apply: generates CRDs files and a new tag locally (doesn't push to server)
  -- dry run: generates CRDs only.

abayer:
@joseblas - assuming that `--no-apply` currently doesn't push the tag, then yeah. Basically just leave `--no-apply` behavior as is, and make `--dry-run` do the same thing as `--no-apply` without doing the tagging.

joseblas:
@abayer, I think it pushes the tag, that was my main concern about this ticket.

joseblas:
after an offline discussion with @abayer this is the expected outcome:
-- no-apply: no changes regarding this feature
-- dry run: generates CRDs only.

pkg/jx/cmd/step_create_task.go
```diff
@@ -155,6 +156,7 @@ func NewCmdStepCreateTask(commonOpts *opts.CommonOptions) *cobra.Command {
 	cmd.Flags().StringVarP(&options.CloneGitURL, "clone-git-url", "", "", "Specify the git URL to clone to a temporary directory to get the source code")
 	cmd.Flags().StringVarP(&options.PullRequestNumber, "pr-number", "", "", "If a Pull Request this is it's number")
 	cmd.Flags().BoolVarP(&options.NoApply, "no-apply", "", false, "Disables creating the Pipeline resources in the kubernetes cluster and just outputs the generated Task to the console or output file")
+	cmd.Flags().BoolVarP(&options.DryRun, "dryrun", "", false, "Disables creating the Pipeline resources in the kubernetes cluster and just outputs the generated Task to the console or output file, without side effects")
```
hferentschik: At other places in the code we use `dry-run`. I think we should align.

pkg/jx/cmd/step_tag.go
```diff
@@ -88,6 +89,8 @@ func NewCmdStepTag(commonOpts *opts.CommonOptions) *cobra.Command {
 	cmd.Flags().StringVarP(&options.Flags.Dir, "dir", "", "", "the directory which may contain a 'jenkins-x.yml'")
 	cmd.Flags().StringVarP(&options.Flags.ChartValueRepository, "charts-value-repository", "r", "", "the fully qualified image name without the version tag. e.g. 'dockerregistry/myorg/myapp'")
 
+	cmd.Flags().BoolVarP(&options.Flags.NoApply, "no-apply", "", false, "Do not push the tag to the server, this is used for example in dry runs")
```
hferentschik: +1

joseblas:
@hferentschik already changed the dry-run.

joseblas:
@hferentschik rebase needed as more tests were added to master.

joseblas:
@hferentschik tests are now passing after a rebase and the fix.

PR #3952, opened 2019-05-14
feat(tekton): Add a potentially multiline script field for steps

### LINKED ISSUE #3767: Allow the execution of multiple commands within a single pipeline step
### Summary 

Right now one cannot execute multiple commands in a single pipeline step. The best workaround is using '&&':

```yaml
- name: copy
  image: busybox
  command: mkdir
  args:
  - -p
  - /workspace/go/src/github.com/jenkins-x/exposecontroller
  - "&&"
  - cp
  - -R
  - "./"
  - /workspace/go/src/github.com/jenkins-x/exposecontroller
```

### Expected behavior

It would be nice to have a syntax which allows executing multiple commands, eg by using/adding a _script_ key.

abayer:
#### Submitter checklist

#### Description

This allows doing things like:
```
steps:
  - name: A script step
    script: |
      ls -la
      echo tada
      for d in 1 2 3 4; do
        echo "d is ${d}";
      done
  - name: A command step
    command: echo
    args: ['a normal step']
```

/assign @hferentschik 
/assign @jstrachan 
/assign @warrenbailey 
/assign @rawlingsj 
/assign @dwnusbaum 
cc @sharepointoscar @vfarcic 

fixes #3767

### DIFF
--- a/pkg/jx/cmd/test_data/step_create_task/script-step/jenkins-x.yml
+++ b/pkg/jx/cmd/test_data/step_create_task/script-step/jenkins-x.yml
@@ -1,0 +1,38 @@
+pipelineConfig:
+  env:
+    - name: FRUIT
+      value: BANANA
+  pipelines:
+    release:
+      pipeline:
+        env:
+          - name: GIT_AUTHOR_NAME
+            value: somebodyelse
+        options:
+          containerOptions:
+            resources:
+              requests:
+                cpu: 0.1
+                memory: 64Mi
+        agent:
+          image: nodejs
+        stages:
+          - name: Build-a-really-long-stage-name-please-but-not-too-long-thanks
+            steps:
+              - script: |
+                  echo hello world
+                  ls -la
+                env:
+                  - name: ANOTHER_VAR
+                    value: Another value
+          - name: Second
+            steps:
+              - command: echo
+                args:
+                  - hi ${FRUIT}
+            options:
+              containerOptions:
+                resources:
+                  limits:
+                    cpu: 0.4
+                    memory: 256Mi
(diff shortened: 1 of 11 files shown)


pkg/jx/cmd/test_data/step_create_task/script-step/jenkins-x.yml
```diff
@@ -0,0 +1,38 @@
...
+            resources:
+              requests:
+                cpu: 0.1
+                memory: 64Mi
+        agent:
+          image: nodejs
+        stages:
+          - name: Build-a-really-long-stage-name-please-but-not-too-long-thanks
+            steps:
+              - script: |
```
dwnusbaum: My understanding is that the shell used to execute the script defaults to `/bin/sh` but could be overridden by the pod template. I wonder if people would want to be able to specify a different shell here since it would be so easy to write scripts directly in jenkins-x.yml. You could still do that in `script` as-is, but it would be a bit awkward with quotes. 

That said, maybe forcing `/bin/sh` as a default is a good way to encourage people to put their scripts in source control somewhere if they are complicated enough to want a different shell.

abayer: Yeah, I'd say stick with forcing the shell - if someone really cares enough, they can hack the `script`, but it's not worth the potential pain of someone specifying an overriding shell that doesn't exist on the image, IMO.

jstrachan:
@abayer damn looks like it needs a rebase - but love this change!

vfarcic:
I like the idea behind this change, but I'm not sure that it's worth introducing yet another instruction. Can we use `command` for that as well (instead of adding `script`)?

abayer:
@vfarcic Wow, how did I not realize you can actually do this with `command` already...oy. Closing this. =)

vfarcic:
I wasn't aware that we can do that with `command`, but rather hinted that multi-line should still be done with it. Now I learned something new 👍

PR #3953, opened 2019-05-14
fix(cmd): application environment was always the last environment

polothy:
#### Submitter checklist

#### Description

The JX Environment listed in `ApplicationEnvironmentInfo` here:

https://github.com/jenkins-x/jx/blob/443b821c5093b71eabbd1eca0d7334b6f89429af/pkg/jx/cmd/get_applications.go#L47

Is always set to the last environment that was iterated over.  I think this is caused by variable shadowing.  The fix was to clone the environment struct.

Manually built and tested the fix.

None.


polothy:
Thanks!

PR #3977, opened 2019-05-16
Fix: isolated path for go in codegen

pmuir:
#### Submitter checklist

fixes #


cmd/codegen/app/generate_clientset.go
```diff
@@ -83,7 +83,7 @@ func NewGenerateClientSetCmd(genOpts GenerateOptions) *cobra.Command {
 	cobraCmd.Flags().StringVarP(&o.InputPackage, optionInputPackage, "i", "", "Input package, must specify")
 	cobraCmd.Flags().StringVarP(&o.OutputPackage, optionOutputPackage, "o", "", "Output package, must specify")
 	cobraCmd.Flags().StringVarP(&o.InputBase, optionInputBase, "", wd, "Input base, defaults working directory")
-
+	cobraCmd.Flags().BoolVarP(&o.NotIsolated, optionNotIsolated, "", false, "do not isolate from the users GOPATH")
```
hferentschik: Why is the option negative? I find "isolated" easier to process/understand.

cmd/codegen/app/generate_clientset.go
```diff
@@ -109,11 +109,18 @@ func (o *ClientSetGenerationOptions) Run() error {
 		return errors.Wrapf(err, "ensure GOPATH is set correctly")
 	}
 
-	err = generator.InstallCodeGenerators(o.GeneratorVersion)
+	gopath := util.GoPath()
+	if !o.NotIsolated {
```
hferentschik: I find this type of if statements hard to process. See above.

hferentschik: So per default, you don't want to install into the GOPATH?

cmd/codegen/app/generate_options.go
```diff
@@ -20,6 +21,7 @@ const (
 	optionOutputBase      = "output-base"
 	optionBoilerplateFile = "boilerplate-file"
 	optionModuleName      = "module-name"
+	optionNotIsolated     = "not-isolated"
```
hferentschik: 'isolated' or 'isolated-install'?

cmd/codegen/app/generate_docs.go
```diff
@@ -74,17 +76,24 @@ func run(o *GenerateDocsOptions) error {
 	}
 	util.AppLogger().Infof("generating docs to %s\n", o.OutputBase)
 
-	err = generator.InstallGenAPIDocs(o.GeneratorVersion)
+	gopath := util.GoPath()
```
hferentschik: Could this handling of the isolated GOPATH creation not be handled in something like `func GoPath(isolated bool)`?

cmd/codegen/util/go.go
```diff
@@ -168,3 +171,41 @@ func getModGraph(dir string) (string, error) {
...
+	}
+	h := os.Getenv("USERPROFILE") // windows
+	if h == "" {
+		h = "."
+	}
+	return h
+}
+
+func ConfigDir() (string, error) {
+	path := os.Getenv("JX_HOME")
```
hferentschik: So per default, you want to install this into JX_HOME? Personally, I am no fan of that. It is not what I would expect. What's the driving reason for this change?

There is btw a clean target in the codegen Makefile which deletes the binaries. That probably would need to be looked at as well.

cmd/codegen/generator/client_set.go
```diff
@@ -25,14 +26,14 @@ var (
 )
 
 // InstallCodeGenerators installs client-gen from the kubernetes-incubator/reference-docs repository.
-func InstallCodeGenerators(version string) error {
+func InstallCodeGenerators(version string, gopath string) error {
```
hferentschik: might be worth to rename these variables to something like `toolGoPath` to create a bit of a differentiation

cmd/codegen/util/go.go
hferentschik: FYI, my plan was to:

* make codegen a proper submodule with its own go.mod
* in the main build install and use a pinned version of codegen via `go get`
* adjust the make rules in the codgen Makefile to allow for development and use of the latest version

cmd/codegen/app/generate_clientset.go
pmuir: Should be on by default, so did the negative. TBH I'm not even sure this should be optional, I put this in mostly for you ;-)

cmd/codegen/app/generate_clientset.go
pmuir: exactly, it doesn't work at all. I'm not entirely sure how it works for you.

cmd/codegen/app/generate_options.go
pmuir: Don't understand the comment?

cmd/codegen/app/generate_docs.go
pmuir: yes, it could.

cmd/codegen/util/go.go
pmuir: > What's the driving reason for this change?

It doesn't work doing it in the gopath as you already have stuff cloned there onto branches - no way we should be reusing that.

pmuir: > There is btw a clean target in the codegen Makefile which deletes the binaries. That probably would need to be looked at as well.

Will do.

pmuir: > FYI, my plan was to:

> make codegen a proper submodule with its own go.mod
> in the main build install and use a pinned version of codegen via go get
> adjust the make rules in the codgen Makefile to allow for development and use of the latest version

That's orthogonal to this.

cmd/codegen/generator/client_set.go
pmuir: That would be confusing - what's a "Tool Go Path"? This is just the `GOPATH` to use.

cmd/codegen/util/go.go
pmuir: I think with your "expect" thing, you are thinking as the developer of the feature, not the end user. As the end user I would expect these tools to work in isolation, not by fiddling with my development environment.

cmd/codegen/app/generate_clientset.go
pmuir: will change it to global

Makefile.codegen
```diff
@@ -57,17 +57,6 @@ codegen-openapi: build-codegen
 	@echo "Generating OpenAPI structs for Kubernetes Clients"
 	./build/$(CODE_GEN_BIN_NAME) --generator-version $(OPENAPI_GEN_VERSION) openapi --output-package=pkg/client --input-package=github.com/jenkins-x/jx/pkg/apis --group-with-version=jenkins.io:v1
 
-generate-clean: clean ## Delete code generation tools
```
hferentschik: hmm, not so sure this is a good idea. This ensured you never had "stale" tools.

cmd/codegen/app/generate_clientset.go
```diff
@@ -83,7 +83,7 @@ func NewGenerateClientSetCmd(genOpts GenerateOptions) *cobra.Command {
 	cobraCmd.Flags().StringVarP(&o.InputPackage, optionInputPackage, "i", "", "Input package, must specify")
 	cobraCmd.Flags().StringVarP(&o.OutputPackage, optionOutputPackage, "o", "", "Output package, must specify")
 	cobraCmd.Flags().StringVarP(&o.InputBase, optionInputBase, "", wd, "Input base, defaults working directory")
-
+	cobraCmd.Flags().BoolVarP(&o.Global, global, "", false, "use the users GOPATH")
```
hferentschik: Why not make the default `true`?

cmd/codegen/app/generate_docs.go
hferentschik: but not worth doing?

cmd/codegen/util/go.go
```diff
@@ -64,39 +67,112 @@ func EnsureGoPath() error {
...
 			"GO111MODULE": modulesMode,
+			"GOPATH":      gopath,
 		},
 	}
 	out, err := goGetCmd.RunWithoutRetry()
 	if err != nil {
 		return errors.Wrapf(err, "error running %s, output %s", goGetCmd.String(), out)
 	}
-
+	if !goModules && version != "" {
```
hferentschik: Is this branch used?

cmd/codegen/app/generate_clientset.go
pmuir: Because it doesn't function properly except in some edge cases (you have none of the code already checked out in your `GOPATH`, or if you do, `master` is checked out).

cmd/codegen/util/go.go
pmuir: Yes, it's used if something that is being got doesn't support go modules, and a version is specified. For example line 498 of `open_api.go`.

Makefile.codegen
pmuir: ok - let me do that properly.

pmuir: https://github.com/jenkins-x/jx/pull/3977/files#diff-e1f8204ecc7085af67da9ba2d35cc266R85

cmd/codegen/app/generate_clientset.go
hferentschik: Would have been nice with an issue related to this providing more context.

cmd/codegen/util/go.go
hferentschik: > It doesn't work doing it in the gopath as you already have stuff cloned there onto branches - no way we should be reusing that.

Maybe. But then I'd choose a tmp directory which gets deleted afterwards and not ~/.jx

hferentschik: > That's orthogonal to this.

trueish

PR #3979, opened 2019-05-16
feat: Automate dns managed zone creation for tenant cluster

cagiti:
Adding the ability to for jx to create a managed-zone within CloudDNS

### DIFF
--- a/pkg/jx/cmd/create_cluster_gke.go
+++ b/pkg/jx/cmd/create_cluster_gke.go
@@ -1,20 +1,21 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
 	"strings"
 	"time"
 
 	"github.com/jenkins-x/jx/pkg/jx/cmd/helper"
+	"github.com/pkg/errors"
+
+	osUser "os/user"
 
 	"github.com/Pallinder/go-randomdata"
 	"github.com/jenkins-x/jx/pkg/cloud"
 	"github.com/jenkins-x/jx/pkg/features"
 	"github.com/jenkins-x/jx/pkg/kube"
+	"github.com/jenkins-x/jx/pkg/tenant"
 	"gopkg.in/AlecAivazis/survey.v1"
-
-	osUser "os/user"
 
 	"regexp"
 
@@ -55,6 +56,8 @@ type CreateClusterGKEFlags struct {
 	Scopes          []string
 	Preemptible     bool
 	EnhancedApis    bool
+	CloudBeesDomain string
+	CloudBeesAuth   string
 }
 
 const (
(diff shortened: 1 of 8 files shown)


pkg/jx/cmd/create_cluster_gke.go
```diff
@@ -550,3 +578,71 @@ func validateClusterName(clustername string) error {
...
+	// Checking whether dns api is enabled
+	err = gke.EnableAPIs(projectID, "dns")
+	if err != nil {
+		return "", errors.Wrap(err, "enabling the dns api")
+	}
+
+	// Create domain if it doesn't exist and return name servers list
+	nameServers, err := createTenantsSubDomainDNSZone(projectID, domain)
+	if err != nil {
+		return "", err
```
ccojocar: I would wrap this error

pkg/jx/cmd/create_cluster_gke.go
```diff
@@ -550,3 +578,71 @@ func validateClusterName(clustername string) error {
...
+	return nil
+}
+
+// createTenantsSubDomainDNSZone creates the tenants DNS zone if it doesn't exist
+// and returns the list of name servers for the given domain and project
+func createTenantsSubDomainDNSZone(projectID string, domain string) ([]string, error) {
+	var nameServers = []string{}
+	err := gke.CreateManagedZone(projectID, domain)
+	if err != nil {
+		return []string{}, err
```
ccojocar: I would add some context to the error in case it fails, it is more obvious where.

pkg/jx/cmd/create_cluster_gke.go
```diff
@@ -550,3 +578,71 @@ func validateClusterName(clustername string) error {
...
+	}
+	nameServers, err = gke.GetManagedZoneNameServers(projectID, domain)
+	if err != nil {
+		return []string{}, err
```
ccojocar: ditto

pkg/jx/cmd/create_cluster_gke.go
```diff
@@ -114,6 +123,8 @@ func NewCmdCreateClusterGKE(commonOpts *opts.CommonOptions) *cobra.Command {
 
 	options.addCreateClusterFlags(cmd)
 
+	cmd.Flags().StringVarP(&options.Flags.CloudBeesDomain, "cloudbees-domain", "", "", "When setting up a letter/tenant cluster, this creates a tenant cluster on the cloudbees domain which is retrieved via the required URL")
+	cmd.Flags().StringVarP(&options.Flags.CloudBeesAuth, "cloudbees-auth", "", "", "Auth used when setting up a letter/tenant cluster")
```
rawlingsj: let's address this comment in a followup PR but maybe we need to describe the format that the auth string needs to be provided as in the description.

cagiti: totally, I missed that

PR #3989, opened 2019-05-16
chore: dead code cleanup removing chromedp dependency and node package

### LINKED ISSUE #3987: Chrome on the host system should not be a dependency of JenkinsX
### Summary 
ChromeDP is included in a branch of code related to Git instantiation using the v1 variant of the ChromeDP API.
This means JX command line requires my machine to have Chrome installed. JX is generally for setting up lab environments on K8S clusters. This type of machine is more typically a server rather than consumer machine, where perhaps it makes less sense to install a consumer desktop web browser, if able to avoid it.

Open for discussion @wbrefvem @jstrachan 

Should JX be launching a browser? If so, lets put it in a container. If not, can remove this area of code for now

"v1"
v2 also does not have a "runner" package
https://github.com/jenkins-x/jx/blob/c9a4a473d3b4693fd0b48a8977c5f1ca8103bb11/pkg/jx/cmd/create_git_api_token.go#L11
Seems there is an overhaul of the ChromeDP API as mentioned in the Slack yesterday by Jeff Zhu:
https://github.com/chromedp/chromedp/commit/3d3bf22ccc321c3536c9c34fa5fefe7b575a6ba1

Source code inspection
https://github.com/jenkins-x/jx/blob/c9a4a473d3b4693fd0b48a8977c5f1ca8103bb11/pkg/jx/cmd/create_git_api_token.go#L10

### Expected behavior
Product installs all dependencies not mentioned in the manual, or does not require a compatible version of Chrome installed on the executing system
Is ChromeDP the right package to take as dependency here for firing up a browser or browser-equivalent? For something like obtaining a credential or creating the default token I would hope the related project will expose an API

Basically any solution other than taking a web browser as requirement will make sense.

### Actual behavior
Chrome is required by ChromeDP to function. This is not being encapsulated by running against a container with Chrome nor using an alternative dependency directly from GoLang to download and parse the Gitea site.

### Jx version
All since Feb 2018

N/A

### Jenkins type
N/A

### Kubernetes cluster
N/A

Ubuntu 16.04.6 LTS
N/A

xihaLong:
#### Description
Remove ChromeDP dependency from JX

#### Submitter checklist
- [ yes ] Change is code complete and matches issue description.
- [ nope ] Change is covered by existing or new tests.

/assign @jstrachan
/assign @wbrefvem 

fixes #3987


xihaLong:
@wbrefvem @jstrachan for consideration as approver and reviewer

xihaLong:
@rawlingsj @ccojocar @jglick also appropriate for review

xihaLong:
@wbrefvem This change affects git compatibility and may be relevant to plans in the area

pkg/jx/cmd/create_git_api_token.go
```diff
@@ -157,132 +149,8 @@ func (o *CreateGitTokenOptions) Run() error {
 
 // lets try use the users browser to find the API token
 func (o *CreateGitTokenOptions) tryFindAPITokenFromBrowser(tokenUrl string, userAuth *auth.UserAuth) error {
```
ccojocar: Why are you not removing completely this function if is dead code?

xihaLong: Can remove entirely

pkg/jx/cmd/create_git_api_token.go
```diff
@@ -112,12 +102,6 @@ func (o *CreateGitTokenOptions) Run() error {
 		userAuth.ApiToken = o.ApiToken
 	}
 
-	tokenUrl := gits.ProviderAccessTokenURL(server.Kind, server.URL, userAuth.Username)
-
-	if userAuth.IsInvalid() && o.Password != "" {
-		err = o.tryFindAPITokenFromBrowser(tokenUrl, userAuth)
```
jglick: If I understand correctly this is not “dead code” you are removing.

I know nothing about this usage of Chrome. I did remove the _other_ usage, for Jenkins, in #2651.

ccojocar:
Closing this PR since almost all tests are failing. You can provide a valid user name in the command line arguments to avoid starting Chrome.

xihaLong:
Thank you @ccojocar for moving this to a closed or held state while I needed to work out some other issues.

I cannot access the test results you linked above here. Is there a way I can view the summary or be granted permission to view the results?

"Internal Server Error
Failed to authenticate: github: user "xihaLong" not in required orgs or teams"

pkg/jx/cmd/create_git_api_token.go
xihaLong: Thank you @jglick for the feedback. I had to handle some internal issues.

Instead of referring to the whole change as dead code removal (in this case: code that should not or cannot work on anyone's machine and refers to a set of dependencies that do not exist generally), I may say this is in the same line of thought as #2651.

This creates a Chrome instance using a no-longer-existing-or-supported API set from ChromeDP, which was a problem for a couple users as mentioned in the Slack channel. It is impossible to launch unless you have several side effects from installing other things on the system and potentially manually install an outdated tool and an outdated version of Chrome.

Of course, all of this is mitigated by running the command by hand and entering the result in the command line arguments. Which means the only way known to run this code is to avoid the code blocks that launch a browser. By extension, those code blocks that as far as I can inspect should not work on anyone's machine... well, maybe they work in some form of systems integration test or the context of everyone at CloudBees somehow.

In general, it is not desirable to launch and scrape in the current manner. In line with the thinking of #2651 I will check what APIs may exist to replace the function and chat in the Slack to make sure I am on the same page as you/your team.

xihaLong: "no-longer-existing-or-supported" I am referring to ChromeDP here (link):

PR #3990, opened 2019-05-16
fix(cmd): GetJenkinsURL ignores the ns argument

polothy:
#### Submitter checklist

- [x] (**GUESSING IT IS**) Change is covered by existing or new tests.

#### Description

The `ns` argument to `GetJenkinsURL` gets immediately overwritten with the current namespace stored in kube config.  Very likely they are always the same, but in places like running JX in the cluster, they are not.

Just giving some more details as to how I ran into this problem: it was from doing a JX install from within the cluster.  On install, it builds the `CreateJenkinsUserOptions` command and specifies the Namespace it should run in in order to generate the Jenkins API token.  Then in `CreateJenkinsUserOptions.Run` it calls `CustomJenkinsURL`, passing in the namespace.  Once you make it to `GetJenkinsURL`, the namespace argument is ignored and it reads the namespace from the kube config or pod namespace.

I'm hoping the BDD tests cover this change.

None.


polothy:
Thanks!

PR #3995, opened 2019-05-16
fix: Get fork status from repo

### LINKED ISSUE #3994: jx create pullrequest ignores repo fork status
### Summary 

jx create pullrequest now calls the helper method `PushRepoAndCreatePullRequest` with the `fork` flag always set to true. This will fail if the repo being used for the pull request is NOT actually a fork with the message `422 Validation Failed [{Resource:PullRequest Field:head Code:invalid Message:}]`

tdcox:

#### Submitter checklist

#### Description
Uses the fork status from the repo details rather than the input parameter.

It's not clear why `fork` was being passed explicitly as an input so this change may break something in the GitOps functionality in gitops.go which is the only other consumer of this method. I have therefore just changed the source of the fork status but not removed this from the API. If all the tests still work for GitOps, then we need to clean this up by removing the `fork` parameter from the function.

fixes #3994

### DIFF
--- a/pkg/gits/helpers.go
+++ b/pkg/gits/helpers.go
@@ -245,7 +245,7 @@ func PushRepoAndCreatePullRequest(dir string, gitInfo *GitRepository, base strin
 	if username == "" {
 		return nil, fmt.Errorf("no git user name found")
 	}
-	if gitInfo.Organisation != username && fork {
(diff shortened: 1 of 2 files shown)


pmuir:
/lgtm

Yes, I tried to remove the fork parameter before, but I wasn't aware of this gitInfo - if all the bdd tests pass (which they do) then I think this is safe...

Issue #3342, opened 2019-03-14
Pipeline extensibility for Tekton


pmuir:
Currently there are some ways to extend pipeline e.g. you can put a pipeline definition in a build pack but there is no way to provide any intermediate configuration. This is also not consistent with the other ways we do extensibility which is via CRDs

Acceptance Criteria:

* Consistent with the rest of the JX extensibility
* Blocks pipeline execution until all extensions are applied.

hferentschik:
The current idea is to create a meta pipeline prior to the actual build pipeline. This meta pipeline gets also the source code injected and is then able to contribute to the pipeline by modifying the pipeline config (`jenkins-x.yml`) or any other file of the project (eg a `pom.xml`).

Once the meta pipeline is built and executed, the actual build pipeline is created and executed. 

Apps can "register" themselves as wanting to contribute/modify the pipeline. They do that via a label, eg:

```yaml
  labels:
    jenkins.io/app-name: jx-app-acme
    jenkins.io/app-type: pipeline-extension
``` 

When the meta pipeline is built, all Apps with app-type  pipeline-extension are discovered. Each of this apps needs to have a `PipelineExtension` defined in their App CRD:

```go
type PipelineExtension struct {
	// Name of the container specified as a DNS_LABEL.
	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
	// Docker image name.
	Image string `json:"image,omitempty" protobuf:"bytes,2,opt,name=image"`
	// Entrypoint array. Not executed within a shell.
	Command string `json:"command,omitempty" protobuf:"bytes,3,rep,name=command"`
	// Arguments to the entrypoint.
	Args []string `json:"args,omitempty" protobuf:"bytes,4,rep,name=args"`
}
```

`PipelineExtension` contains all the information needed to build a Tekton task to be executed as part of the meta pipeline. 

In a first version, there is no order in which the Apps get executed, but as a next step we could add a order/priority label as well.

Another idea which needs to be discussed further is to commit the changes each Apps makes to the config/source. This could either be done by the App itself or by the controller building the meta pipeline. TBD.

hferentschik:
Related pull requests:

- #4004 
- #4051

hferentschik:
[...]

PR #4004, opened 2019-05-17
refactor: extracting more functionality into pipelines.go 

hferentschik:
* Moving more code out of create_step_create
* Introducing a wrapper class for the Tekton CRDs

The motivation for the refactor is to have less code in the command class and the ability to share some of the code with the metapipleine stuff (#3342) I am working on.


hferentschik:
@joseblas ^^

abayer:
/assign @dwnusbaum 
Looks good so far!

hferentschik:
Rebased.

hferentschik:
@pmuir, I rebased. Given that it still has the 'approved' label should it not merge? Or does it need another lgtm label in which case what's the point of 'approve'?

PR #4010, opened 2019-05-17
Chore: run gnerate-docs as part of lint

pmuir:
#### Submitter checklist

- [ ] Change is code complete and matches issue description.
- [ ] Change is covered by existing or new tests.

#### Which issue this PR fixes

fixes #

PR #4011, opened 2019-05-17
fix: api doc generation

daveconde:
Fix issue where the api doc generation fails due to a map containing a complex type. While this seems to generate a valid json schema, the current implementation of apidocs-gen believes the type to be unused.


pmuir:
It may be worth merging https://github.com/jenkins-x/jx/pull/4010 into this as then we'll be checking the `make generate-docs` as part of the lint

PR #4020, opened 2019-05-20
Add BatchPipelineActivity information to batch builds and PR builds to get linked in the UI

dgozalo:
#### Submitter checklist

#### Description
When tide has multiple PRs ready to be merged in the merge pool, Prow will create a Batch job that will include references to each PR that comprises it. 

Until now, there was no way to link each PR to the batch build or vice versa in order to get information of the other parts to, for example, link to each other in the UI.

Now a batch build `PipelineActivity` will contain information of each PR that comprise it like the batch name (PR-#) and the build number that is included.

Each one of those PRs' `PipelineActivity` will contain information of the batch build number that is handling its code.

There is a special case for Prow that will attempt to merge the lowest PR of a batch build outside of the batch, then attempt to merge the batch. This will generate another build execution for that PR.

There's a reconciliation mechanism set that will look for its last build with the same `lastCommitSha` that also contains a batch build number, then it will update itself with that batch build number, update the batch build `PipelineActivity` to register itself as the last build for the given PR and then erase the batch build number from the old execution.

This way we can always keep updated information in the `PipelineActivities`. 

This can be merged independently of any PR in other repos within the same issue.


pkg/jx/cmd/controller_pipelinerunner.go
```diff
@@ -180,11 +180,12 @@ func (o *ControllerPipelineRunnerOptions) startPipelineRun(w http.ResponseWriter
 		return
 	}
 
-	// lets change this to support new pipelineresource type that handles batches
-	if len(pj.Refs.Pulls) > 0 {
+	// Only if there is one Pull in Refs, it's a PR build so we are going to pass it
+	if len(pj.Refs.Pulls) == 1 {
```
hferentschik: I'd would be tempted to create a function `isSinglePullRequest()`, even if it in the end just contains just one line. IMO such a method speaks more for itself than the current if statement + comment

hferentschik: So this is just to make sure we are passing the pull request number along, right?

pkg/jx/cmd/step/create/step_create_task.go
```diff
@@ -236,7 +237,7 @@ func (o *StepCreateTaskOptions) Run() error {
 				}
 			}()
 		}
-		err := o.mergePullRefs(o.cloneDir)
+		pr, err = o.mergePullRefs(o.cloneDir)
```
hferentschik: The thing which keeps confusing me here is that it seems that `PULL_REFS` is only passed via an env variable. Almost everything else we pass via option. This seems a bit arbitrary to me. I think we need a better plan here in general. 

One thing we could arleady do now is to have some validateOptions at the beginning of `Run` which determines what type of build we are dealing with, validates we have all parameters, parses the pull refs if needed.

pkg/kube/activity.go
```diff
@@ -200,6 +202,16 @@ func (k *PipelineActivityKey) GetOrCreate(jxClient versioned.Interface, ns strin
 		create = true
 		a = defaultActivity
 	}
+
+	numberOfRefs := len(k.PullRefs)
```
hferentschik: I'd make this a function of `PipelineActivityKey`

pkg/kube/activity.go
```diff
@@ -29,6 +30,7 @@ type PipelineActivityKey struct {
 	LastCommitMessage string
 	LastCommitURL     string
 	GitInfo           *gits.GitRepository
+	PullRefs          map[string]string
```
hferentschik: Would it make sense to have some `builtType` as well?

pkg/kube/activity.go
```diff
@@ -227,6 +247,145 @@ func (k *PipelineActivityKey) GetOrCreate(jxClient versioned.Interface, ns strin
...
+	}
+
+	currentBuildNumber, err := strconv.Atoi(k.Build)
+	if err != nil {
+		return errors.Wrapf(err, "error parsing the current build number for PipelineActivity %s", currentActivity.Name)
+	}
+
+	//It only makes sense to look for the build before this one
+	currentBuildNumber = currentBuildNumber - 1
+	for _, v := range activities.Items {
```
hferentschik: maybe this could be extracted into a method which describes what's going on here.

pkg/kube/activity_test.go
```diff
@@ -198,6 +199,252 @@ func TestCreateOrUpdateActivities(t *testing.T) {
 	//tests.Debugf("Has Promote %#v\n", promote)
 }
 
+func TestCreateOrUpdateActivityForBatchBuild(t *testing.T) {
```
hferentschik: 👍 nice with some tests

pkg/apis/jenkins.io/v1/zz_generated.deepcopy.go
```diff
@@ -172,6 +172,27 @@ func (in *Attachment) DeepCopy() *Attachment {
 	return out
 }
 
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
+func (in *BatchPipelineActivity) DeepCopyInto(out *BatchPipelineActivity) {
```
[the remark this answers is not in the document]
dgozalo: Worked fine. The only "issue" would be that the code generation binaries weren't in .jx/codegen/bin and I had to copy them manually.

pkg/jx/cmd/controller_pipelinerunner.go
dgozalo: The thing here is that, the `> 0` condition was being matched by both PullRequest anad Batch build types. This caused the batch build to incorrectly contain information about the first PullRequest it contained.

pkg/kube/activity.go
dgozalo: I think you can get the same result by checking the length of the `PullRefs`. I'm already creating a function to return the number of refs of a `PipelineActivityKey`. 

Adding a buildType property would just imply moving that len check to somewhere else just to initialize the variable.

Considering the length of PullRefs is being used elsewhere to do other things, I don't think it's necessary.

pkg/kube/activity.go
dgozalo: I think the resulting method would be messy due to all the parameters that it would need. I'd always to keep params lower than 3. 

I will however word it better so it's more understandable.

pkg/kube/activity.go
dgozalo: Agree, created two new funcs: isBatchBuild and isPRBuild. Should make the code much cleaner.

pkg/jx/cmd/step/create/step_create_task.go
dgozalo: They are actually passed as `o.CustomEnvs` and it's probably to keep the spec of Prow: https://github.com/oracle/kubernetes-test-infra/tree/master/prow (at Job Environment Variables).
This way we can always "check" we are respecting the spec.

pkg/apis/jenkins.io/v1/zz_generated.deepcopy.go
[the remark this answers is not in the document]
hferentschik: which target did you execute? There should be no need for a manual copy. Personally, I don't like that we put things into `.jx/codegen/bin`, but that's a different story.

pkg/kube/activity.go
hferentschik: ok

pkg/kube/activity.go
hferentschik: ok

> I think the resulting method would be messy due to all the parameters that it would need

which might point to a deeper lying issue. But anyways

pkg/jx/cmd/step/create/step_create_task.go
hferentschik: > They are actually passed as o.CustomEnvs and it's probably to keep the spec of Prow

Sure. But why do our controllers not also only act on env variables? Instead, there is this mixture of options and env variables which I find confusing.

hferentschik:
@dgozalo what is the state with this PR now?

pkg/apis/jenkins.io/v1/zz_generated.deepcopy.go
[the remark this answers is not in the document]
dgozalo: It was `make generate`

dgozalo:
> @dgozalo what is the state with this PR now?

Should be mergeable. I'll tackle the change to the garbage collection in another PR / issue.

pkg/kube/activity.go
```diff
@@ -200,6 +210,15 @@ func (k *PipelineActivityKey) GetOrCreate(jxClient versioned.Interface, ns strin
 		create = true
 		a = defaultActivity
 	}
+
+	if k.isBatchBuild() {
```
hferentschik: 👍

pkg/kube/activity.go
```diff
@@ -29,12 +30,21 @@ type PipelineActivityKey struct {
...
 	LastCommitURL     string
 	GitInfo           *gits.GitRepository
+	PullRefs          map[string]string
 }
 
 func (k *PipelineActivityKey) IsValid() bool {
 	return len(k.Name) > 0
 }
 
+func (k *PipelineActivityKey) isBatchBuild() bool {
```
hferentschik: 👍

PR #4034, opened 2019-05-21
chore: revert the clone methods

pmuir:
#### Submitter checklist

fixes #


hferentschik:
@pmuir, was the intention to get this merged or did you just want to show what we potentially could do?

pmuir:
Just to show potentially what we could do I think. I'm not 100% convinced this is safe. I'll try to do a local fix today.

hferentschik:
>  I'm not 100% convinced this is safe.

Why not? What are your concerns? It looks definitely much simpler ;-)

pmuir:
As we discussed this area is delicate and requires manual testing. Just because something is simple doesn't mean it's right.

PR #4035, opened 2019-05-21
Fix: There was an incompatibility problem between gcloud and go-cloud so we are going to create buckets with gsutil

dgozalo:
…, so we are going to create the buckets with gsutil

#### Submitter checklist

#### Description
A bug surfaced yesterday which didn't allow people with a newer version of gcloud to create buckets for long term storage.

After further investigation, an incompatibility between gcloud and `go-cloud`, the library we were using to abstract the cloud provider. 

The method to create buckets has changed to use gsutil instead.

fixes #

### DIFF
--- a/pkg/cloud/gke/storage/long_term_storage.go
+++ b/pkg/cloud/gke/storage/long_term_storage.go
@@ -5,40 +5,58 @@ import (
 	"github.com/jenkins-x/jx/pkg/cloud/gke"
 	"github.com/jenkins-x/jx/pkg/kube"
 	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/util"
 	"github.com/pkg/errors"
 	"github.com/satori/go.uuid"
 )
 
-func EnableLongTermStorage(installValues map[string]string, providedBucketName string,
-	createBucketFN func(bucketName string, bucketKind string) (string, error)) (string, error) {
-	var bucketName string
+// EnableLongTermStorage will take the cluster install values and a provided bucket name and use it / create a new one for gs
+func EnableLongTermStorage(installValues map[string]string, providedBucketName string) (string, error) {
 	if providedBucketName != "" {
-		exists, err := gke.BucketExists(installValues[kube.ProjectID], providedBucketName)
-		if err != nil {
-			return "", errors.Wrap(err, "checking if the provided bucket exists")
-		}
-		if exists {
-			bucketName = providedBucketName
-			return fmt.Sprintf("gs://%s", bucketName), nil
-		}
-
-		bucketURL, err := createBucketFN(providedBucketName, "gs")
-		if err == nil {
-			return bucketURL, nil
-		}
-		log.Warnf("Attempted to create the bucket %s in the project %s but failed, will now create a "+
-			"random bucket", providedBucketName, installValues[kube.ProjectID])
-	}
-
-	if providedBucketName == "" {
+		return ensureProvidedBucketExists(installValues, providedBucketName)
+	} else {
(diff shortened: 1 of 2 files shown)


pkg/cloud/gke/storage/long_term_storage.go
```diff
@@ -5,29 +5,30 @@ import (
...
 )
 
-func EnableLongTermStorage(installValues map[string]string, providedBucketName string,
-	createBucketFN func(bucketName string, bucketKind string) (string, error)) (string, error) {
+// EnableLongTermStorage will take the cluster install values and a provided bucket name and use it / create a new one for gs
+func EnableLongTermStorage(installValues map[string]string, providedBucketName string) (string, error) {
 	var bucketName string
 	if providedBucketName != "" {
-		exists, err := gke.BucketExists(installValues[kube.ProjectID], providedBucketName)
+		bucketName = providedBucketName
```
hferentschik: why do you need this assignment?

pkg/cloud/gke/storage/long_term_storage.go
```diff
@@ -5,29 +5,30 @@ import (
...
+		bucketURL, err := createBucket(bucketName, installValues)
 		if err == nil {
 			return bucketURL, nil
 		}
 		log.Warnf("Attempted to create the bucket %s in the project %s but failed, will now create a "+
-			"random bucket", providedBucketName, installValues[kube.ProjectID])
+			"random bucket", bucketName, installValues[kube.ProjectID])
 	}
 
 	if providedBucketName == "" {
```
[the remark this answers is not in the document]
hferentschik: I'd extract this:

```golang
uuid4, _ := uuid.NewV4()
	bucketName = fmt.Sprintf("%s-lts-%s", installValues[kube.ClusterName], uuid4.String())
	if len(bucketName) > 60 {
		bucketName = bucketName[:60]
	}
```

into a function `generateUniqueBucketName` or `generateRandomBucketName`

pkg/cloud/gke/storage/long_term_storage.go
```diff
@@ -40,5 +41,18 @@ func EnableLongTermStorage(installValues map[string]string, providedBucketName s
 		bucketName = bucketName[:60]
 	}
 
-	return createBucketFN(bucketName, "gs")
+	return createBucket(bucketName, installValues)
```
hferentschik: I find the flow in this method confusing. Partly because of the two if statements and partly because of the multiple returns.

hferentschik: I'd do something like:

```golang
func EnableLongTermStorage(installValues map[string]string, providedBucketName string) (string, error) {
	var bucketName string
	if providedBucketName != "" {
		bucketName = providedBucketName
		exists, err := gke.BucketExists(installValues[kube.ProjectID], bucketName)
		if err != nil {
			return "", errors.Wrap(err, "checking if the provided bucket exists")
		}
		if exists {
			return fmt.Sprintf("gs://%s", bucketName), nil
		}

		bucketURL, err := createBucket(bucketName, installValues)
		if err == nil {
			return bucketURL, nil
		}
		log.Warnf("Attempted to create the bucket %s in the project %s but failed, will now create a "+
			"random bucket", bucketName, installValues[kube.ProjectID])
         
                return createBucket(uniqueBucketName(installValues), installValues)
               
	} else 
		log.Info("No bucket name provided for long term storage, creating a new one")
                 return createBucket(uniqueBucketName(installValues), installValues)
	}
}
```

hferentschik: Maybe even put the block in the first if statement into a function!?

PR #4050, opened 2019-05-22
fix: remove the 'jx-' prefix from application name

ccojocar:
It seems that the applications are prefixed with `jx-` regardless of the namesapce
where they are installed. This can create confusion when for instance users are trying to
promote an application. They are often use the application with the jx prefix which
leads to an error, because the chart is published without the prefix.

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/kube/version.go
+++ b/pkg/kube/version.go
@@ -133,6 +133,12 @@ func GetAppName(name string, namespaces ...string) string {
 				}
 			}
 		}
+		// The applications seems to be prefixed with jx regardless of the namespace
+		// where they are deployed. Let's remove this prefix.
+		prefix := "jx-"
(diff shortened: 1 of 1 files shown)


pmuir:
I've never seen an application have `jx-` as a prefix...

ccojocar:
This is for `jx get applications` not apps :). You can see this if you create any quickstart and then execute `jx get applications`. The name is prefixed.

ccojocar:
@pmuir Is is possible to overwrite the status of `codecov/patch/defualt`? It seems to be failing because this PR does not increase the test coverage.

pmuir:
Yes, though right now it shouldn't enforce it. What did @abayer is correct

PR #4064, opened 2019-05-23
fix(tekton): Use a working and useful versions directory

abayer:
#### Submitter checklist

#### Description

We *were* using the typoed `cmmon_versions`, but once that typo was
fixed, we weren't using anything. Changing to just use the
now-existant `common_versions` wasn't particularly desirable, since it
messed with expected volumes etc, so instead let's use
`stable_versions`, which matches up with the pod templates we define
for the test.

/assign @hferentschik 
/assign @dwnusbaum 

n/a


pkg/jx/cmd/step_create_task_test.go
```diff
@@ -41,11 +41,11 @@ func TestGenerateTektonCRDs(t *testing.T) {
 	log.SetOutput(ioutil.Discard)
 	defer log.SetOutput(os.Stderr)
 
-	testVersionsDir := path.Join("test_data", "cmmon_versions")
```
hferentschik: So what happens here with the stuff under common_versions?

abayer:
Rebased!

`common_versions` is still being used by `version_test`, so I left it as is.

PR #4073, opened 2019-05-23
Add git repo URL to to jx get applications

wbrefvem:
#### Submitter checklist

#### Description
`jx get applications` now includes the application's git repo URL.


pkg/jx/cmd/get_applications.go
```diff
@@ -151,6 +154,19 @@ func (o *GetApplicationsOptions) generateTable(apps []string, envApps []EnvApps,
 	appEnvMap := map[string]map[string]*ApplicationEnvironmentInfo{}
 	for _, appName := range apps {
 		row := []string{appName}
+		jxClient, ns, err := o.JXClientAndDevNamespace()
```
ccojocar: I would create the client outside of the loop since a new client is not required for each iteration. Also I would return an error instead of just logging a warning. This looks like a fatal error.

pkg/jx/cmd/get_applications.go
```diff
@@ -151,6 +154,19 @@ func (o *GetApplicationsOptions) generateTable(apps []string, envApps []EnvApps,
...
+		if err != nil {
+			log.Warnf("Error getting jx client and dev namespace: %s", err)
+		}
+		sourceRepositories := jxClient.JenkinsV1().SourceRepositories(ns)
+		sourceRepository, err := GetSourceRepositoryForApplication(appName, sourceRepositories)
+		gitURL := "None Found"
+		if err != nil {
+			log.Warnf("Could not get Source Repository for application: %s", err)
```
ccojocar: Do we need to log here a warning or just go silent with no git URL? Maybe this will confuse the user.

pkg/jx/cmd/get_applications.go
```diff
@@ -243,6 +261,31 @@ func (o *GetApplicationsOptions) generateTable(apps []string, envApps []EnvApps,
...
+	}
+	return sourceRepository.Spec.Provider + "/" + sourceRepository.Spec.Org + "/" + sourceRepository.Spec.Repo + ".git"
+}
+
+// GetSourceRepositoryForApplication fetches a SourceRepository CRD for a particular application by trying to match its name to a repo.
+func GetSourceRepositoryForApplication(application string, sourceRepositories altv1.SourceRepositoryInterface) (*v1.SourceRepository, error) {
+	// We can't pull the SourceRepository by name because we don't know the org, so we iterate over all of them and match the application name to the repo.
+	repoList, err := sourceRepositories.List(metav1.ListOptions{})
+	if err != nil {
+		return nil, err
```
ccojocar: I would wrap this error with some context.

pkg/jx/cmd/get_applications.go
```diff
@@ -243,6 +261,31 @@ func (o *GetApplicationsOptions) generateTable(apps []string, envApps []EnvApps,
...
+	if sourceRepository.Spec.Provider == "" || sourceRepository.Spec.Org == "" || sourceRepository.Spec.Repo == "" {
+		return "None Found"
+	}
+	return sourceRepository.Spec.Provider + "/" + sourceRepository.Spec.Org + "/" + sourceRepository.Spec.Repo + ".git"
+}
+
+// GetSourceRepositoryForApplication fetches a SourceRepository CRD for a particular application by trying to match its name to a repo.
+func GetSourceRepositoryForApplication(application string, sourceRepositories altv1.SourceRepositoryInterface) (*v1.SourceRepository, error) {
+	// We can't pull the SourceRepository by name because we don't know the org, so we iterate over all of them and match the application name to the repo.
+	repoList, err := sourceRepositories.List(metav1.ListOptions{})
```
ccojocar: I am wondering how efficient is this for a large number of repositories (e.g. 50+).

pkg/jx/cmd/get_applications.go
wbrefvem: Agree on moving it out of the loop, I'll fix that. As far as the warning goes, I initially thought there might be a situation where the app name doesn't match the repo name, in which case I didn't want the command to fail. But actually there's no way for this to happen that I know of, so maybe you're right.

pkg/jx/cmd/get_applications.go
wbrefvem: That's a good question that I don't know the answer to, but I'm not sure there's another way to do this. You have to know the org + repo in order to build the SourceRepository name if you want to use Get for an individual SourceRepository. There's no way to get the org from the application, unless we were to change the naming of applications so that it's org + repo instead of just repo. Another solution is one unified CRD for apps and applications that references a SourceRepository. In my opinion this is the most robust solution, but obviously quite a bit more effort.

pkg/jx/cmd/get_applications.go
wbrefvem: That's a good question. I don't have strong feelings about this, it just felt right to add an error message here.

pkg/jx/cmd/get_applications.go
wbrefvem: Will do,

pkg/jx/cmd/get_applications.go
wbrefvem: Actually never mind, I was thinking about the git URL error message. This is definitely a fatal error.

pkg/jx/cmd/get_applications.go
ccojocar: +1 to unify the CRD for apps 

cc @pmuir

pmuir: Unifying apps and applications should be done but is a big job and completely orthogonal to this change - you still have to find the right crd for the name.

Anyway the solution to Cosmin's point is to use a label selector on the sourcerepository for already present `repository` label and then check the returned list which should be quite short.

You really need to get all the possible source repositories and handle the case that >1 exists.

wbrefvem: @pmuir That may reduce the size of the list returned in some cases, but I'm still not seeing a way to solve the ambiguity without reference to `owner` and `provider` as well, which is why I brought up putting them in a CRD, though it's a point well taken that unifying apps and applications is beside the point here. But adding a separate CRD would only add more work to that effort.

pmuir: > pmuir That may reduce the size of the list returned in some cases,

It will reduce it in every case except the one where every source repository you have has the same name but a different organisation. In the real world, that is every case ;-)

>  but I'm still not seeing a way to solve the ambiguity without reference to `owner` and `provider` as well, which is why I brought up putting them in a CRD, though it's a point well taken that unifying apps and applications is beside the point here. But adding a separate CRD would only add more work to that effort.

Yes. I see the point - I was confused as your solution suffers from the ambiguity ;-)

Probably a better way to handle this is via releases - look up the last release for the application (which you could do with a label and then sorting by version) and grab the git url

wbrefvem: @pmuir Okay, doing it by release makes sense, since the application name should match. I guess the same ambiguity is possible, for instance if you have a cluster with a bunch of forks of the same repo and each is a separate application. I'm not sure that's even possible, but it sounds like something that might happen in future, though to your point, this probably isn't a case to lose sleep over.

wbrefvem: The problem here is that releases, as far as I can tell, don't get labels, so a label selector is out. The application name is stored in the CRD but it's different from `Release.metadata.name`, so we'd need to do a field selector, except that `Release.spec.name` is not a supported field. So unless we can assume a version number (like 0.0.1 or something), which would allow us to select on `Release.metadata.name`, I'm officially stumped.

jstrachan:
@wbrefvem needs a rebase (looks like accept the new generated openapi thingy)

PR #4080, opened 2019-05-24
Chore: enable bdd coverage

pmuir:
#### Submitter checklist

fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,10 +73,10 @@ endif
 
 # Various codecov.io variables that are set from the CI envrionment if present, otherwise from locally computed values
 
-CODECOV_NAME := integration
+CODECOV_NAME ?= integration
 
 #ARGS is extra args added to the codecov uploader
-CODECOV_ARGS := "-n $(CODECOV_NAME)"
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -119,8 +121,15 @@ BUILDFLAGS += -ldflags "-X $(ROOT_PACKAGE)/pkg/features.FeatureFlagToken=$(FEATU
 RELEASE_ORG_REPO := cloudbees/cloudbees-jenkins-x-distro
 endif
 
+ifdef JUNIT_REPORT
+# REPORT_PIPE must go at end of line
+OUTFILE := $(shell mktemp)
+REPORT_PIPE := -test.v 2>&1 | tee $(OUTFILE) | go-junit-report > build/integration.junit.xml & tail -f $(OUTFILE)
```
ccojocar: Can we also keep the HTML report?

ccojocar: I'd like to open it in the browser. The Junit XML is hard to visualise without extra tools.

pmuir: yeah, i assumed no one would use that, i'll add it back in

pmuir: Actually that's the other PR I have open ;-)

Issue #4018, opened 2019-05-20
CLI upgrade fails


vfarcic:
### Summary 

Every once in a while, new releases are not available through `jx version` and `jx upgade cli`. Right now, for example, we cannot upgrade to `2.0.139`.

### Steps to reproduce the behavior

```bash
jx version
```

```
NAME               VERSION
jx                 2.0.137
Kubernetes cluster v1.10.11
kubectl            v1.14.0
git                git version 2.20.1 (Apple Git-117)
Operating System   Mac OS X 10.14.4 build 18E226
WARNING:
A new jx version is available: 2.0.139
? Would you like to upgrade to the new jx version? Yes
```

### Expected behavior

New version is installed.

### Actual behavior

```
ERROR: Error: Command failed  brew upgrade jx
error: exit status 1
```

### Jx version

```
NAME               VERSION
jx                 2.0.137
Kubernetes cluster v1.10.11
kubectl            v1.14.0
git                git version 2.20.1 (Apple Git-117)
Operating System   Mac OS X 10.14.4 build 18E226
```

### Jenkins type

Any

### Kubernetes cluster

Any

### Operating system / Environment

macOS

hferentschik:
I've seen this myself a couple of times. Needs some investigation.

garethjevans:
Are you using brew?

vfarcic:
In my case if fails every once in a while with brew

hferentschik:
Same here, also with brew.

udbscience:
It also downloads the wrong version, tried with v2.0.66 too.

jx upgrade cli --version="2.0.66"
Downloading https://github.com/jenkins-x/jx/releases/download/v2.0.311/jx-linux-amd64.tar.gz

PR #4086, opened 2019-05-27
fix(cmd): on macOS use 'brew info' to determine latest jx version

hferentschik:
fixes issue #4018

Not 100% sure whether the issue reported on #4018 is exactly the same, but this fix tries to make sure that if brew is used to upgrade jx, brew is also used to check the latest available version.

ATM, the latest available version is determined by the GitHub release which might not have properly propagated to Homebrew.

### DIFF
--- a/pkg/jx/cmd/upgrade_cli.go
+++ b/pkg/jx/cmd/upgrade_cli.go
@@ -1,7 +1,10 @@
 package cmd
 
 import (
+	"encoding/json"
+	"github.com/blang/semver"
 	"github.com/jenkins-x/jx/pkg/jx/cmd/helper"
+	"github.com/pkg/errors"
 	"runtime"
 
 	"github.com/jenkins-x/jx/pkg/jx/cmd/opts"
@@ -30,6 +33,15 @@ type UpgradeCLIOptions struct {
 	Version string
 }
 
+// BrewInfo contains some of the `brew info` data.
+type brewInfo struct {
+	Name     string
+	Outdated bool
+	Versions struct {
+		Stable string
+	}
+}
+
 // NewCmdUpgradeCLI defines the command
 func NewCmdUpgradeCLI(commonOpts *opts.CommonOptions) *cobra.Command {
 	options := &UpgradeCLIOptions{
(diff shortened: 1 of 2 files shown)


pkg/jx/cmd/upgrade_cli.go
```diff
@@ -83,3 +94,33 @@ func (o *UpgradeCLIOptions) Run() error {
...
 }
+
+func (o *UpgradeCLIOptions) latestAvailableJxVersion() (*semver.Version, error) {
+	var newVersion *semver.Version
+	if runtime.GOOS == "darwin" && !o.NoBrew {
+		brewInfo, err := o.GetCommandOutput("", "brew", "info", "--json", "jx")
+		if err != nil {
+			return nil, err
+		}
+		v, err := o.latestJxBrewVersion(brewInfo)
```
ccojocar: It seems that this error is not handled.

pkg/jx/cmd/upgrade_cli.go
```diff
@@ -83,3 +94,33 @@ func (o *UpgradeCLIOptions) Run() error {
...
+		newVersion, err = semver.New(v)
```
ccojocar: ditto

pkg/jx/cmd/upgrade_cli.go
```diff
@@ -83,3 +94,33 @@ func (o *UpgradeCLIOptions) Run() error {
 		return o.InstallJx(true, newVersion.String())
 	}
 }
+
+func (o *UpgradeCLIOptions) latestAvailableJxVersion() (*semver.Version, error) {
+	var newVersion *semver.Version
+	if runtime.GOOS == "darwin" && !o.NoBrew {
+		brewInfo, err := o.GetCommandOutput("", "brew", "info", "--json", "jx")
+		if err != nil {
+			return nil, err
```
ccojocar: I would wrap the errors with some message throughout the entire function.

hferentschik:
@ccojocar I updated the PR and rebased it. Also found another issue I tried to fix. The --version flag was actually never used.

PR #4089, opened 2019-05-28
(fix) Tidy Go modules

warrenbailey:
#### Submitter checklist

#### Description

Ran `go mod tidy` to remove some of the unused modules.

fixes #


go.mod
```diff
@@ -44,7 +44,6 @@ require (
 	github.com/gliderlabs/ssh v0.1.1 // indirect
 	github.com/go-ole/go-ole v1.2.1 // indirect
 	github.com/go-openapi/jsonreference v0.17.0
-	github.com/go-openapi/loads v0.19.0 // indirect
```
ccojocar: I think some of these are added by the code generation. They are indirect dependencies. Probably will break the lint which atm is checking for differences in these files.

cc @hferentschik

warrenbailey: Yeah I think this is why this fails, thanks :)

Issue #4104, opened 2019-05-28
`jx create cluster gke` failing on latest jx 


rawlingsj:
### Steps to reproduce the behavior
```
jx create cluster gke  --default-admin-password foo --zone europe-west1-c --cluster-name bar
```

### Actual behavior

```
ERROR: error creating cluster configuring and installing Prow: installing Prow: failed to install Tekton: after 2 attempts, last error: failed to run 'kubectl delete release --ignore-not-found -l jenkins.io/chart-release=tekton,jenkins.io/version!=0.0.38 --namespace jx' command in directory '', output: 'error: the server doesn't have a resource type "release"'
error: configuring and installing Prow: installing Prow: failed to install Tekton: after 2 attempts, last error: failed to run 'kubectl delete release --ignore-not-found -l jenkins.io/chart-release=tekton,jenkins.io/version!=0.0.38 --namespace jx' command in directory '', output: 'error: the server doesn't have a resource type "release"'
```

Unsure if related yet but I can see earlier in the console output a new error that I've not seen before when installing the ingress controller...
```
Applying generated chart stable/nginx-ingress YAML via kubectl in dir: /var/folders/n6/vfz_f9jn21g5l5n4p__rjdqc0000gn/T/helm-template-workdir-630724473/jxing/output
clusterrole.rbac.authorization.k8s.io/jxing-nginx-ingress unchanged
clusterrolebinding.rbac.authorization.k8s.io/jxing-nginx-ingress unchanged
configmap/jxing-nginx-ingress-controller unchanged
deployment.extensions/jxing-nginx-ingress-controller configured
deployment.extensions/jxing-nginx-ingress-default-backend configured
role.rbac.authorization.k8s.io/jxing-nginx-ingress unchanged
rolebinding.rbac.authorization.k8s.io/jxing-nginx-ingress unchanged
serviceaccount/jxing-nginx-ingress unchanged
Error from server (Invalid): error when applying patch:
{"spec":{"clusterIP":""}}
to:
Resource: "/v1, Resource=services", GroupVersionKind: "/v1, Kind=Service"
Name: "jxing-nginx-ingress-controller", Namespace: "kube-system"
[...]

PR #4107, opened 2019-05-29
Added extra logging

cagiti:
Added extra logging to help debug #4104


pkg/jx/cmd/opts/common.go
```diff
@@ -259,7 +262,9 @@ func (o *CommonOptions) JXClient() (versioned.Interface, string, error) {
 	if o.factory == nil {
 		return nil, "", errors.New("command factory is not initialized")
 	}
+
 	if o.jxClient == nil {
+		log.Infof("Creating new JX Client")
```
ccojocar: Shouldn't we use `Debug` for this? I can get extra verbose if you intent to use it permanently.

hferentschik:
@cagiti, is this still something we need after pr #4108 got merged? Should this be closed now?

PR #4109, opened 2019-05-29
feat(tekton): adding command to create and apply metapipeline

hferentschik:
This is work in progress for the meta pipeline functionality. The pull request adds a new command `step create pipeline` which generates Tekton CRDs to run a meta pipeline allowing Apps to modify the source prior to the execution of the actual build pipeline.

This is work in progress but could be merged, since it does not affect the current behaviour. The next step is to let the meta pipeline call a modified `step create task`.

To run:
```bash
jx --verbose --log-level debug step create pipeline --clone-git-url <repo-url> --branch master --kind release --no-apply
```

In order to have some example apps, one can install any existing App and then edit its CRD adding:

```yaml
metadata:
  labels:
    jenkins.io/app-type: pipeline-extension
...
spec:
  pipelineExtension:
    args:
    - -l
    - /workspace/source
    command: ls
    image: alpine:latest
    name: My App
```

Relates to issue #3342


pkg/tekton/pipelines.go
```diff
@@ -27,13 +27,17 @@ func GeneratePipelineActivity(buildNumber string, branch string, gitInfo *gits.G
 	name := gitInfo.Organisation + "-" + gitInfo.Name + "-" + branch + "-" + buildNumber
 	pipeline := gitInfo.Organisation + "/" + gitInfo.Name + "/" + branch
 	log.Infof("PipelineActivity for %s", name)
+	var toMerge map[string]string
```
hferentschik: @abayer heads up, @dgozalo's latest changes related to batch builds changed the way you can use `step create task`.

Without this change here you cannot just call `jx step crate task` anymore. You also need to specify `--clone-git-url <repo>` and `-e PULL_REFS=<valid-pull-ref>`.

This change will get the old behaviour back.

pkg/jx/cmd/step/create/step_create_meta_pipeline.go
```diff
@@ -0,0 +1,318 @@
...
+
+	err = o.handleResult(tektonClient, jxClient, tektonCRDs, buildNumber, o.Branch, o.PullRefs, ns, gitInfo)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (o *StepCreatePipelineOptions) validateCommandLineFlags() error {
```
hferentschik: @abayer, I'd really like to end up at a stage where it is clearly defined what parameters are needed. Also,. easy validation should occur early on.

pkg/tekton/metapipeline/metapipeline.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	// TODO - Add step which does call the 'step create task' functionality to build the actual build pipeline
+	return steps, nil
+}
+
+func buildEnvParams(params CRDCreationParameters) []corev1.EnvVar {
+	var envVars []corev1.EnvVar
+
+	envVars = append(envVars, corev1.EnvVar{
```
hferentschik: I might be missing some variables, but the ones here make obvious sense right now. If we need more and in particular if more is needed for creating the build pipeline needs to be seen.

pkg/tekton/metapipeline/metapipeline.go
```diff
@@ -0,0 +1,233 @@
...
+		return nil, errors.Wrap(err, "error  retrieving pipeline contributor apps")
+	}
+	return appsList.Items, nil
+}
+
+// BuildExtensionSteps builds a step (container) for each of the apps specified.
+func BuildExtensionSteps(extendingApps []jenkinsv1.App) ([]syntax.Step, error) {
+	var steps []syntax.Step
+
+	// TODO - Add step to create effective pipeline. This way apps will always have a pipeline config to load and work with,
```
hferentschik: @abayer, so here I'd like to create the step which generates the effective jenkins-x.yml. Maybe extract that into a new `step syntax command`

pkg/tekton/metapipeline/metapipeline.go
```diff
@@ -0,0 +1,233 @@
...
+			Image:     extension.Image,
+			Command:   extension.Command,
+			Arguments: extension.Args,
+		}
+
+		log.Debugf("App %s contributes with step %s", app.Name, util.PrettyPrint(step))
+		steps = append(steps, step)
+	}
+
+	// TODO - Add step which does call the 'step create task' functionality to build the actual build pipeline
```
hferentschik: @abayer, After the App steps, would come the step which calls the creation of the build pipeline CRDs. Effectively the current `step_create_task` minus the whole business of cloning and merging.

pkg/tekton/pipelines.go
abayer: Nice, good catch.

pkg/jx/cmd/step/create/step_create_meta_pipeline.go
abayer: Agreed.

pkg/tekton/metapipeline/metapipeline.go
abayer: Yeah, I'm gonna do that in a separate PR today or tomorrow.

hferentschik:
Closing this for now. I'll need to rebase and update the pull request.

PR #4116, opened 2019-05-29
feat: adding retries to Stages (tekton based solution)

joseblas:
#### Submitter checklist

#### Description
  Since Tekton already provides retries, one new param has to be passed to it.
 This PR just makes sure that the param coming from jenkins-x.yml goes into Tekton Pipeline CRD.

fixes #

### DIFF
--- a/pkg/tekton/syntax/pipeline.go
+++ b/pkg/tekton/syntax/pipeline.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/jenkins-x/jx/pkg/log"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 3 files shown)


pkg/jx/cmd/test_data/step_create_task/command-as-multiline-script/pipeline.yml
```diff
@@ -0,0 +1,35 @@
+apiVersion: tekton.dev/v1alpha1
```
abayer: This file should be removed - I'm guessing it's a lurking artifact from before the `cmd` package got reorganized. =)

pkg/tekton/syntax/pipeline.go
```diff
@@ -1534,6 +1533,13 @@ func (j *ParsedPipeline) GenerateCRDs(pipelineIdentifier string, buildIdentifier
 			return nil, nil, nil, err
 		}
 
+		o := stage.Stage.Options
+		if o.RootOptions != nil {
+			if o.Retry > 0 {
+				stage.Stage.Options.Retry = s.Options.Retry
+				log.Infof("Generating CRDs with %d retries", stage.Stage.Options.Retry)
```
abayer: Do we really want to be logging here? I tend to think we don't.

joseblas: Well, I think logging is good in general; is good for analysing possible bug or weird situations

pkg/jx/cmd/test_data/step_create_task/command-as-multiline-script/pipeline.yml
abayer: But it's not getting used in the tests - the one in `pkg/jx/cmd/step/create/test_data` is.

pkg/tekton/syntax/pipeline.go
abayer: Then a better logging message makes sense - like, say "setting retries to %d for stage %s".

PR #4117, opened 2019-05-29
test: increase parallelism of test target

hferentschik:
- allowing up to 4 test packages to be compiled and tested in parallel

with `p 1`
```sh
$ time (make test)
real	9m22.274s
user	8m5.280s
sys	1m51.307s
```

with `p 4`
```sh
$ time (make test)
real	3m17.259s
user	10m0.746s
sys	2m51.395s
```

@ccojocar, I think we talked about this once and you said there was a reason why we need `p1`. Is this still the case?


pmuir:
We needed `p1` because we had such a large package we used up all the ram. Now that we've refactored, we can increase the paralellism. How much ram is the test run peaking at with `p4`?

pmuir:
You really need to change this in the `slow-integration-tests` as well - those are the ones we use in CI

pmuir:
Closing as I changed the parallelism

Issue #4120, opened 2019-05-30
Only release the jx binary once versions has fully passed


pmuir:
Right now we release the jx binary on every PR. We should make these "pre-releases" and do the final release once versions has passed. This will need to update github, chocolatey, the website, brew dockerhub and dev-env

garethjevans:
This should help fix #4115

cagiti:
Approach:
- [ ] #4364 
- [ ] #4365 
- [ ] #4366

cagiti:
- [ ] Investigate upgrade cli so that users aren't able to install pre-release versions
- [ ] Merge in PR associated with #4365 

Assumptions, I can read the jx version within jenkins-x-versions repo from dependency-matrix file (matrix.md)

deanesmith:
After discussing with @pmuir, determined the urgency can be reduced in favor of other issues.

cagiti:
Completed 🎉

PR #4130, opened 2019-05-30
feat: Updated release process to use goreleaser

cagiti:
This contributes towards #4120

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -284,10 +284,9 @@ release: clean build test-slow-integration linux darwin win arm ## Release the b
 		cd ./build/arm; tar -zcvf ../../release/jx-linux-arm.tar.gz jx; \
 	fi
 
-	go get -u github.com/progrium/gh-release
-	gh-release checksums sha256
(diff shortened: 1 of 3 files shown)


jenkins-x.yml
```diff
@@ -102,6 +102,16 @@ pipelineConfig:
...
+              - name: tag-release
+                image: jenkinsxio/jx:2.0.128
+                command: git
+                args: ['tag', '-a', '${inputs.params.version}', '-m', '${inputs.params.version}']
+
+              - name: push-tag
+                image: jenkinsxio/jx:2.0.128
+                command: git
+                args: ['push', 'origin', '${inputs.params.version}']
+
```
rawlingsj: We shouldn't need to do the git tag here as this already happens in the pipelinerunner

Makefile
```diff
@@ -245,21 +245,15 @@ darwin: ## Build for OSX
 	chmod +x build/darwin/$(NAME)
 
 .PHONY: release
-release: clean build test-slow-integration linux darwin win arm ## Release the binary
```
pmuir: I would probably delete the platform specific targets now - the `Makefile` is pretty busy already and any simplification would be great

PR #4154, opened 2019-06-01
fix: pass helm values files to tekton chart installation

carlossg:

/hold this does not work because tekton is installed individually and not part of the platform.

It would need to have the tekton values in the top level

```yaml
pvc:
  size: 20Gi
```

https://github.com/jenkins-x/cloud-environments/pull/5


carlossg:
@jstrachan what do you think we should do to configure tekton with helm values?

PR #4158, opened 2019-06-02
jx get preview --current should print the url

heroic:
#### Submitter checklist

#### Description
`jx get preview --current` currently logs the url. We should print it so that it's easy for scripts to capture the output.


pmuir:
@heroic needs conflicts resolving

heroic:
> @heroic needs conflicts resolving

done!

PR #4220, opened 2019-06-10
feat: removed o.Debugf helper, using logger directly

### LINKED ISSUE #2916: Logging option should be global
### Summary 

There is no global option(s) for enabling additional/verbose logging.

```
$ jx options
No global options available
```

Some commands have a _verbose_ and _log-level_ flag, for example `jx create quickstart`. Others don't have it, for example `jx delete repo`

### Expected behavior

Ability to turn on verbose output and potentially configure its level should be a global option which one can enable independent of the issued command. 

### Actual behavior

See summary.

### Jx version

```
jx                 1.3.807
jenkins x platform 0.0.3320
Kubernetes cluster v1.10.11-gke.1
kubectl            v1.10.7
helm client        v2.10.0+g9ad53aa
helm server        v2.12.2+g7d2b0c7
git                git version 2.17.2 (Apple Git-113)
Operating System   Mac OS X 10.14.2 build 18C54
```

### Jenkins type

- [ ] Classic Jenkins
- [x] Serverless Jenkins

### Kubernetes cluster

GKE

macOS 10.14

garethjevans:
#### Submitter checklist

fixes #2916

### DIFF
--- a/pkg/jx/cmd/cmd.go
+++ b/pkg/jx/cmd/cmd.go
@@ -22,15 +22,13 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"strconv"
 	"strings"
 	"syscall"
 
 	"github.com/jenkins-x/jx/pkg/jx/cmd/add"
 	"github.com/jenkins-x/jx/pkg/jx/cmd/namespace"
 	"github.com/jenkins-x/jx/pkg/jx/cmd/promote"
-
-	"github.com/jenkins-x/jx/pkg/jx/cmd/helper"
-	"github.com/jenkins-x/jx/pkg/util"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -285,11 +283,21 @@ func fullPath(command *cobra.Command) string {
 }
 
 func setLoggingLevel(cmd *cobra.Command, args []string) {
-	level := cmd.Flag(opts.OptionLogLevel).Value.String()
-	err := log.SetLevel(level)
+	verbose, err := strconv.ParseBool(cmd.Flag(opts.OptionVerbose).Value.String())
 	if err != nil {
-		invalidOptionErr := util.InvalidOption(opts.OptionLogLevel, level, log.GetLevels())
-		helper.CheckErr(invalidOptionErr)
+		log.Logger().Errorf("Unable to determine log level")
+	}
+
+	if verbose {
+		err := log.SetLevel("debug")
+		if err != nil {
+			log.Logger().Errorf("Unable to set log level to debug")
+		}
+	} else {
+		err := log.SetLevel("info")
+		if err != nil {
+			log.Logger().Errorf("Unable to set log level to info")
+		}
 	}
 }
 
--- a/pkg/jx/cmd/create_mlquickstart.go
+++ b/pkg/jx/cmd/create_mlquickstart.go
@@ -120,12 +120,12 @@ func NewCmdCreateMLQuickstart(commonOpts *opts.CommonOptions) *cobra.Command {
 
 // Run implements the generic Create command
 func (o *CreateMLQuickstartOptions) Run() error {
-	o.Debugf("Running CreateMLQuickstart...\n")
+	log.Logger().Debugf("Running CreateMLQuickstart...\n")
 
 	interactive := true
 	if o.BatchMode {
 		interactive = false
-		o.Debugf("In batch mode.\n")
+		log.Logger().Debugf("In batch mode.\n")
 	}
 
 	authConfigSvc, err := o.CreateGitAuthConfigService()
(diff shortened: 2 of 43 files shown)


garethjevans:
Fixes #2916

pkg/jx/cmd/cmd.go
```diff
@@ -285,11 +283,21 @@ func fullPath(command *cobra.Command) string {
 }
 
 func setLoggingLevel(cmd *cobra.Command, args []string) {
-	level := cmd.Flag(opts.OptionLogLevel).Value.String()
-	err := log.SetLevel(level)
+	verbose, err := strconv.ParseBool(cmd.Flag(opts.OptionVerbose).Value.String())
```
hferentschik: What's the point really of `--verbose` at this stage. Is it not really enough with `--log-level`. All --verbose is doing is changing the level to debug.

If we are going down this route, I'd argue we should get rid of `--verbose` altogether.

hferentschik: What was the intention of --verbose in the first place? I guess it was adding additional info level messages to the user. Is this really useful? Not sure.

pkg/jx/cmd/create_mlquickstart.go
```diff
@@ -120,12 +120,12 @@ func NewCmdCreateMLQuickstart(commonOpts *opts.CommonOptions) *cobra.Command {
...
 // Run implements the generic Create command
 func (o *CreateMLQuickstartOptions) Run() error {
-	o.Debugf("Running CreateMLQuickstart...\n")
+	log.Logger().Debugf("Running CreateMLQuickstart...\n")
 
 	interactive := true
 	if o.BatchMode {
 		interactive = false
-		o.Debugf("In batch mode.\n")
+		log.Logger().Debugf("In batch mode.\n")
```
hferentschik: +1 for replacing o.Debug calls with a call to log.Logger()

pkg/jx/cmd/cmd.go
garethjevans: so, it looks like --log-level is only implemented in a handful of commands, where as --verbose is implemented nearly everywhere (its also all over our docs site and in all our demos). This seemed  like the change that would have the least impact to the users.

pkg/jx/cmd/create_mlquickstart.go
garethjevans: I also need to get ride of the \n at the end of every statement, but that can happen in another PR

garethjevans:
@hferentschik Tide should squash these commits automatically before merging, that's the way its configured on the jx repo.  If I squash too early I end up having to force push and I don't like doing that - feels dirty!

garethjevans:
Completely agree with your statement about the global options, I'm going to have a look at those in a later review, most seem to be related to the install (and not all commands)

pkg/jx/cmd/cmd.go
hferentschik: So are you suggesting to remove `--log-level` and just use `--verbose`? That would imply we only have info and debug level to work with, right? Seems a bit limiting. 

If we cannot get rid of `verbose` maybe we need to keep both where `--verbose` is effectively some sort of alias for `--log-level=debug`. Also `log-level` taking precedence over `verbose`.

Another alternative could be to just keep `verbose` as a CLI options, but allow to set any type of log level via an env variable, eg `JX_LOG_LEVEL`.

pkg/jx/cmd/create_mlquickstart.go
hferentschik: Well, you might as well do it now ;-)

hferentschik:
> Tide should squash these commits automatically before merging, that's the way its configured on the jx repo.

It does, but the commit messages get messy containing messages like "making Hound happy" since it still keeps the commit message from each commit. Also for each commit, you get a DCO sign-off line.

Another issue is that when multiple commits get squashed, the pull request title is used as a message for the squashed commit which often breaks the conventional commit message style.

Overall I don't think that squash is a good strategy for merging. I think rebase would be better. And no matter the method used, I think we as developers need to be more active in crafting commits the way we want them ending up in the repo.

Thought about bringing this up for discussion at the team level for quite some time. 

@jenkins-x/core ^^

> If I squash too early I end up having to force push and I don't like doing that - feels dirty!

I don't see an issue with force pushing. IMO this is the way to go.

hferentschik:
> Completely agree with your statement about the global options, I'm going to have a look at those in a later review, most seem to be related to the install (and not all commands)

Right, seem to be install related. Some things must have changed since I created issue #2916. At the time there were no global options shown in the help message. Now there seems to be too many :-)

I would expect to see `verbose`, `log-level`, `batch`.

pkg/jx/cmd/cmd.go
garethjevans: I like the idea of `Another alternative could be to just keep verbose as a CLI options, but allow to set any type of log level via an env variable, eg JX_LOG_LEVEL.` . I think that would work and have the least impact on existing users

cagiti:
I like the direction this change takes by simplifying how users gain more insight into what's happening. If users require more insight into what's happening then the approach of using an environment variable to toggle log level would work. Or perhaps a curl style flag where the number of v's determines the log level.

pkg/jx/cmd/cmd.go
hferentschik: I think there are also different type of users. For jx CLI users for which the differentiation between verbose [true|false] is probably enough. 

For developers, however, trace level logging might be useful (eg thinking about logging in a loop)

hferentschik:
@garethjevans how come this merged? It still had an outstanding "change request" from me. Does Prow not take this into account? If so, I would think this is a serious bug.

Issue #582, opened 2018-04-06
support Istio based user group promotion in Production


jstrachan:
it would be awesome to be able to support [istio](https://istio.io/) based rollouts of versions.

e.g. in Production it'd be awesome to rollout to different user groups:
```
Test -> Staging -> Production (internal developers -> early adopters -> rolling upgrade across everyone else)
```

Conceptually we could treat all of these things as kinds of `Environment`. e.g. so we could promote to `production-internal-developers` or `production-early adopters` etc

Then the current promotion model in Jenkins X could work well across Preview environments, namespace/cluster environments or graduated user group environments - whether thats done in app release CD pipelines, or when manually promoting things.

It would help also in the tooling (CLI, web, chatbot) to view the progress of a version through all the different kinds of environments. e.g. [jx get apps](http://jenkins-x.io/commands/jx_get_applications/) command would list what version is for which user group in Production.

## Implementation thoughts

We'd need to ensure that [istio](https://istio.io/) was installed in the Production environment; which we could check the `requirements.yaml` file and if the istio chart is not there, do a PR to automate it.

Then we'd need to apply the necessary Istio resources to implement which user group gets which version; which the [jx promote](http://jenkins-x.io/commands/jx_promote/) command could turn into the necessary Pull Requests on the environment git repositories.

I guess we need to start off with a sample somewhere of doing graduated rollouts of a version across user groups with Istio then see how we could generalise that for an app version. We may need to also modify the application's chart to add the necessary istio stuff?

jstrachan:
Agreed. Thanks for the links will noodle them some more.

So we separate releases from promotion already in Jenkins X - a ‘promote’ action can be triggered via a pipeline step, CLI or UI/chat bot or some micro service.

Today we have 2 kinds of promote action based on the metadata in the Environment CRD - either do a GitOps PR of a helm chart version change - or run `helm upgrade` directly if folks want to opt out of GitOps.

So I guess we need to add another promotion strategy - eg  have some metadata in an Environment CRD to define it as an Istio user group graduation group instead; so that when a promote is triggered on it for a version we do the Istio stuff instead of vanilla helm chart upgrading?

christian-posta:
so a "promote" sounds like it changes a version in an environment from v1 to v2.. and using vanilla k8s this would be some kind of rolling upgrade?

[...]

PR #4226, opened 2019-06-10
fix: too many open files installing istio addon

carlossg:
Each file inside the tarball is kept open while tarball is extracted

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/util/downloads.go
+++ b/pkg/util/downloads.go
@@ -252,23 +252,7 @@ func UnTargz(tarball, target string, onlyFiles []string) error {
 		}
 
 		path := filepath.Join(target, path.Base(header.Name))
-		info := header.FileInfo()
-		if info.IsDir() {
-			if err = os.MkdirAll(path, info.Mode()); err != nil {
-				return err
-			}
-			continue
-		}
-
-		file, err := os.OpenFile(path, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, info.Mode())
-		if err != nil {
-			return err
-		}
-		defer file.Close()
-		_, err = io.Copy(file, tarReader)
-		if err != nil {
-			return err
-		}
+		UnTarFile(header, path, tarReader)
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


pkg/util/downloads.go
```diff
@@ -252,23 +252,7 @@ func UnTargz(tarball, target string, onlyFiles []string) error {
...
-				return err
-			}
-			continue
-		}
-
-		file, err := os.OpenFile(path, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, info.Mode())
-		if err != nil {
-			return err
-		}
-		defer file.Close()
```
carlossg: this defer was not called until the function ended, so each file inside the for loop was open and not closed

hferentschik:
Do we have an issue for that? IMO it would be nice to capture these things in issues. I imagine that there was an error or panic occurring. For reference, it would be nice to have this information captured in an issue together with the fix.

Also, could one write a test verifying that this issue is fixed?

carlossg:
There's no panic, just the `too many open files` error
What's the point of having issues for PRs? just makes the information duplicated and harder to find and maintain

hferentschik:
> There's no panic, just the too many open files error

Which would have been useful context. Taking this "too many open files" as search string on the issues tab gets me https://github.com/jenkins-x/jx/issues/582 where I see the context out of which this whole thing arose. Note, that this search out of the box did not return this PR.

GitHub does in many regards abstract away the difference between pull requests and issues, but conceptually they are different things and even on a technical / UI level this difference is noticeable. 

> What's the point of having issues for PRs? just makes the information duplicated and harder to find and maintain

IMO the contrary is true. If we were to use an external issue tracker like JIRA, would you still say that creating an issue would not be necessary?

carlossg:
> If we were to use an external issue tracker like JIRA, would you still say that creating an issue would not be necessary?

in that case it would, because people would search in jira for issues. Here they can search in both issues and PRs at once

hferentschik:
> in that case it would, because people would search in jira for issues. Here they can search in both issues and PRs at once

You can, but it is not necessarily obvious

Issue #4273, opened 2019-06-13
jx upgrade client stuck at 2.0.219


tdcox:
### Summary 
```
➜ jx upgrade client
You are already on the latest version of jx 2.0.219
```
Now try brew:
```
➜ brew upgrade jx
Updating Homebrew...
--8<--
==> Upgrading 1 outdated package:
jenkins-x/jx/jx 2.0.219 -> 2.0.274
==> Upgrading jenkins-x/jx/jx
==> Downloading https://github.com/jenkins-x/jx/releases/download/v2.0.274/jx-darwin-amd64.tar.gz
```

Not sure if this was a problem with 219 or if it is an ongoing issue.

tdcox:
Hmmm. I'm getting some odd behaviour from `jx upgrade platform` too:
```
➜ jx upgrade platform
Using provider 'gke' from team settings
? A local Jenkins X versions repository already exists, pulling the latest: Yes
using stable version 2.0.330 from charts of jenkins-x/jenkins-x-platform from /Users/terry/.jx/jenkins-x-versions
Upgrading platform from version 4 to version 2.0.330
Read credentials for https://storage.googleapis.com/chartmuseum.jenkins-x.io from vault helm/repos
WARNING: Waiting for the fork of bootsbot/environment-d38-dev to appear...
Forked Git repository to https://github.com/bootsbot/environment-d38-dev
Fetched chart jenkins-x-platform to dir /var/folders/2n/n92yy0vd2pvfj1r8r3_ntp700000gn/T/jx-helm-fetch-jenkins-x-platform-922481433/jenkins-x-platform
WARNING: No changes made to the source code in /Users/terry/.jx/environments/Bootstrap-Jersey/environment-d38-dev. Code must be up to date!
```

Latest release of jenkins-x-platform in the repo seems to be 2.0.634

garethjevans:
This should be resolved in 2.0.277

Issue #4168, opened 2019-06-03
version upgrade bug


derrickburns:
### Summary 

```
$ jx version
NAME               VERSION
jx                 2.0.209
jenkins x platform 2.0.330
Kubernetes cluster v1.12.6-eks-d69f1b
kubectl            v1.14.1
helm client        Client: v2.14.0+g05811b8
git                git version 2.20.1
Operating System   Mac OS X 10.14.5 build 18F132
WARNING:
A new jx version is available: 2.0.212
? Would you like to upgrade to the new jx version? Yes

You are already on the latest version of jx 2.0.209
```

### Jx version

The output of `jx version` is:

```
COPY OUTPUT HERE
```

### Jenkins type

- [ ] Next Generation (Tekton + Prow)
- [ ] Classic Jenkins
- [ ] Serverless Jenkins (JenkinsFileRunner + Prow)

warrenbailey:
Hi @derrickburns, could do with a few more details on this one. Please fill out the following sections: summary, steps to reproduce, expected behaviour, actual behaviour, kubernetes cluster and jenkins type.

/priority awaiting-more-evidence
/kind bug

Wizofgoz:
I'm having the same problem on jx version 2.0.215
```
Wizofgoz:home mac$ jx version
NAME               VERSION
jx                 2.0.215
Kubernetes cluster v1.11.8-eks-7c34c0
kubectl            v1.14.0
helm client        v2.9.1+g20adb27
helm server        v2.9.1+g20adb27
git                git version 2.16.3
Operating System   Mac OS X 10.14.4 build 18E226
WARNING: 
A new jx version is available: 2.0.268
? Would you like to upgrade to the new jx version? Yes

You are already on the latest version of jx 2.0.215
```

Summary:
Attempts at upgrading cli fail claiming that you are running latest version of jx.

Steps to reproduce:
1. run `jx version`
2. confirm upgrade

Expected behavior:
Successfully upgrade to latest release

Actual behavior:
Upgrade fails and claims to already be running the latest release

Kubernetes cluster:
EKS

Jenkins Type:
Next Generation

hferentschik:
@Wizofgoz Could you re-run the command using the `--verbose` option to get some more trace.
Also, are you using homebrew and did you install jx via `brew`?

garethjevans:
This should be resolved in 2.0.277

Issue #4115, opened 2019-05-29
jx version does gets confused between the new and the old releasse


vfarcic:
### Summary 

When upgrading `jx` version, it offers a new release but instead upgrading the CLI it claims that I'm already running the latest.

### Steps to reproduce the behavior

```bash
jx version
```

```
jx version
NAME               VERSION
jx                 2.0.184
jenkins x platform 2.0.330
Kubernetes cluster v1.12.8
kubectl            v1.14.2
helm client        Client: v2.14.0+g05811b8
git                git version 2.20.1 (Apple Git-117)
Operating System   Mac OS X 10.14.5 build 18F203
WARNING:
A new jx version is available: 2.0.186? Would you like to upgrade to the new jx version? Yes

You are already on the latest version of jx 2.0.184
```

### Expected behavior

It should have upgraded to 2.0.186 as suggested.

### Actual behavior

It claims that 2.0.184 is the latest version even though it offered to upgrade to 2.0.186.

### Jx version

The output of `jx version` is:

```
NAME               VERSION
jx                 2.0.184
jenkins x platform 2.0.330
Kubernetes cluster v1.12.8
kubectl            v1.14.2
helm client        Client: v2.14.0+g05811b8
git                git version 2.20.1 (Apple Git-117)
Operating System   Mac OS X 10.14.5 build 18F203
```

### Jenkins type

Any

### Kubernetes cluster

Any

### Operating system / Environment

macOS

garethjevans:
Are you using brew to install jx?

vfarcic:
Yes.

garethjevans:
I think this is caused by the jx check looking directly at github, but the homebrew-jx release has not yet completed

garethjevans:
there is a window where this could happen.  We are moving the homebrew-jx update downstream of jenkins-x-versions (rather than jx) to improve this

vfarcic:
Can we make the release process wait until homebrew-jx update is finished? Otherwise, it is misleading at best.

vfarcic:
I don't think it's a "window". I just run the following.

```bash
NAME               VERSION
jx                 2.0.188
jenkins x platform 2.0.330
Kubernetes cluster v1.13.5-gke.10
kubectl            v1.14.2
helm client        Client: v2.14.0+g05811b8
git                git version 2.20.1 (Apple Git-117)
Operating System   Mac OS X 10.14.5 build 18F203
WARNING:
A new jx version is available: 2.0.192
[...]

Issue #3023, opened 2019-02-07
jx upgrade cli after a manual install (not via brew) fails


agentgonzo:
### Summary 
If you have previously installed jx via brew, then uninstalled it and the manually installed it via the download, a `jx upgrade cli` fails.

### Steps to reproduce the behavior
`brew install jx`
`brew uninstall jx`
Download and install jx manually
`jx upgrade cli`

### Expected behavior
the manually installed version of jx should be upgraded

### Actual behavior
it tries to install via brew, which fails
![image](https://user-images.githubusercontent.com/1652511/52411651-72dc8a00-2ad4-11e9-97da-3f7233d8a696.png)

### Jx version

The output of `jx version` is:

```
COPY OUTPUT HERE
![image](https://user-images.githubusercontent.com/1652511/52411660-76701100-2ad4-11e9-91b8-652dc46cd0f5.png)
```

### Jenkins type

- [ ] Classic Jenkins
- [ ] Serverless Jenkins

### Operating system / Environment
mac

msigler-cb:
First, remove jx using `brew brew uninstall jx`. Do `brew update` and `brew list` and not see jx. Also suggest `brew cleanup`. Then install per the guide: https://go.cloudbees.com/docs/cloudbees-core/kubernetes-cd-install-guide/install-base/ . This would include `curl -L https://github.com/jenkins-x/jx/releases/download/v1.3.620/jx-linux-amd64.tar.gz | tar xzv` and `sudo mv jx /usr/local/bin`. This should remove any preexisting association with brew, and install current version of jx without using brew. Any future upgrades should be using `jx upgrade cli --no-brew` which should prevent unintended upgrades with the rest of `brew update && brew upgrade`.  

This MacOS specific bug may also apply to any other client package manager.

msigler-cb:
I suspect the next time I do `jx version` and there is an upgrade available, that i will be prompted, and with `Y` i will get the brew error.

ccojocar:
Please could you try to run `brew upgrade jx` outside of jx? I suspect you will get something like this:

```
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/jx
Target /usr/local/bin/jx
already exists. You may want to remove it:
  rm '/usr/local/bin/jx'

To force the link and overwrite all conflicting files:
  brew link --overwrite jx

To list all files that would be deleted:
  brew link --overwrite --dry-run jx
```

[...]

PR #4275, opened 2019-06-13
fix: improved logic around upgrades and brew

garethjevans:
when using brew we now perform a `brew update` first.  Moved most of the logic around
version updates into UpgradeCLIOptions so it can be reused.


garethjevans:
Fixes #4273, #4168, #3023, #4115

cagiti:
Seems valid
/lgtm

PR #4297, opened 2019-06-17
Add user label to lts bucket

macox:

#### Submitter checklist

#### Description

Add a `created-by` label to long term storage buckets for easier identification


pkg/cloud/gke/storage/long_term_storage.go
```diff
@@ -55,6 +55,7 @@ func createBucket(bucketName string, installValues map[string]string) (string, e
 	log.Logger().Infof("The bucket %s does not exist so lets create it", infoBucketURL)
 	region := gke.GetRegionFromZone(installValues[kube.Zone])
 	err := gke.CreateBucket(installValues[kube.ProjectID], bucketName, region)
+	gke.AddBucketLabel(bucketName, gke.UserLabel())
```
garethjevans: On some of the linux distributions we have in cluster, the osUser.Current() returns nil, so we may need to handle the case where the label is blank

macox: @garethjevans So an empty wouldn't be sufficient? (if found && label != "" {....})

PR #4299, opened 2019-06-17
WIP: Remove platform update from updatebot config

cagiti:

Since we update jenkins-x-platform jx versions via jenkins-x-builders and update jenkins-x-builders jx versions via jx, it seems counter-intuitive to update jenkins-x-platform jx versions straight from jx.

### DIFF
--- a/.updatebot.yml
+++ b/.updatebot.yml
@@ -4,7 +4,6 @@ github:
     repositories:
     - name: homebrew-jx
     - name: jx-docs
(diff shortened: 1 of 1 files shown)


cagiti:
Since we update jenkins-x-platform jx versions via jenkins-x-builders and update jenkins-x-builders jx versions via jx, it seems counter-intuitive to update jenkins-x-platform jx versions straight from jx.

At the moment it's only true for different components.

PR #4302, opened 2019-06-17
fix: Added missing timestamps

### LINKED ISSUE #4289: jx create addon owasp-zap and jx create post preview job failing
### Summary 
`jx create addon owasp-zap` and `jx create post preview job` both fail with the same error:
```
error: Failed to update Development environment in namespace jx: Environment.jenkins.io "dev" is invalid: []: Invalid value: 
--8<--
environment map redacted
--8<--
spec.teamSettings.postPreviewJobs.spec.template.metadata.creationTimestamp in body must be of type string: "null"
spec.teamSettings.postPreviewJobs.metadata.creationTimestamp in body must be of type string: "null"
```
### Jx version

```
jx                 2.0.274
Kubernetes cluster v1.13.6-gke.5
kubectl            v1.14.2
git                git version 2.21.0
Operating System   Mac OS X 10.13.6 build 17G6030
```

### Jenkins type

- [x] Next Generation (Tekton + Prow)
- [ ] Classic Jenkins
- [ ] Serverless Jenkins (JenkinsFileRunner + Prow)

tdcox:

#### Submitter checklist

#### Description
Adds explict timestamps to spec.teamSettings.postPreviewJobs.metadata.creationTimestamp and
spec.teamSettings.postPreviewJobs.spec.template.metadata.creationTimestamp in order to pass validation.

fixes #4289

### DIFF
--- a/pkg/cmd/create/create_addon_owasp.go
+++ b/pkg/cmd/create/create_addon_owasp.go
@@ -102,11 +102,15 @@ func (o *CreateAddonOwaspOptions) Run() error {
 		}
 		settings.PostPreviewJobs = append(settings.PostPreviewJobs, batchv1.Job{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:   name,
-				Labels: labels,
+				Name:              name,
+				Labels:            labels,
+				CreationTimestamp: metav1.Now(),
 			},
 			Spec: batchv1.JobSpec{
 				Template: corev1.PodTemplateSpec{
+					ObjectMeta: metav1.ObjectMeta{
+						CreationTimestamp: metav1.Now(),
+					},
 					Spec: corev1.PodSpec{
 						Containers:    []corev1.Container{firstContainer},
 						RestartPolicy: corev1.RestartPolicyNever,
--- a/pkg/cmd/create/create_post_preview_job.go
+++ b/pkg/cmd/create/create_post_preview_job.go
@@ -112,11 +112,15 @@ func (o *CreatePostPreviewJobOptions) Run() error {
 		}
 		settings.PostPreviewJobs = append(settings.PostPreviewJobs, batchv1.Job{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:   name,
-				Labels: labels,
+				Name:              name,
+				Labels:            labels,
+				CreationTimestamp: metav1.Now(),
 			},
 			Spec: batchv1.JobSpec{
 				Template: corev1.PodTemplateSpec{
+					ObjectMeta: metav1.ObjectMeta{
+						CreationTimestamp: metav1.Now(),
+					},
 					Spec: corev1.PodSpec{
 						Containers:    []corev1.Container{firstContainer},
 						RestartPolicy: corev1.RestartPolicyNever,


pkg/cmd/create/create_addon_owasp.go
```diff
@@ -102,11 +102,15 @@ func (o *CreateAddonOwaspOptions) Run() error {
 		}
 		settings.PostPreviewJobs = append(settings.PostPreviewJobs, batchv1.Job{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:   name,
-				Labels: labels,
+				Name:              name,
+				Labels:            labels,
```
hferentschik: Is there a particular issue this tries to fix? How does it manifest itself?
AFAIK, one should not set the creation timestamp of the object meta. This is done on the server side. The docs seem to be quite clear here as well:

> CreationTimestamp is a timestamp representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations. Clients may not set this value.

hferentschik: Ahh, just saw the link to #4289

tdcox: Clients may not pass nil values to the API and thus the validation was failing.

hferentschik: Do you know where exactly the code is failing? Something is not quite right, but as said, I don't think one can/should just set the timestamp.

hferentschik: Does it maybe happen when the TeamSettings are creates/updated - https://github.com/jenkins-x/jx/blob/4e48c11af2416c014d777c6d01c9950018d934f9/pkg/apis/jenkins.io/v1/types_environment.go#L168. We keep an array of `batchv1.Job`. I can see how updating team settings would fail if the jobs themselves have not been created before. 

It seems PostPreviewJobs are more templates for the jobs. See also https://github.com/jenkins-x/jx/blob/ae0c16ca03873c53f8924d0ab1c68eab0574bde9/pkg/cmd/preview/preview.go#L652.

I think if it were me, I'd not store this information as a `batchv1.Job`. Seems we are only "supporting" image, name. , command and BackoffLimit. That's what I would store in a custom struct. @jstrachan WDYT?

At the very least, I'd add a comment.

hferentschik: An alternative could be to store the jobs are `JobSpecs` and do the lookup on the container name.

tdcox: I have no objection if you want to rework all the back-end functionality at some point, but this is broken in production and needs an urgent fix.

hferentschik: Bandages of this form are imo always a bad idea. I was hoping we could at least establish the root cause for the issue. So far my comments are just speculations.

hferentschik: Bandages of this form are imo always a bad idea. I was hoping we could at least establish the root cause for the issue. So far my comments are just speculations.

rawlingsj: As existing functionality is broken and this is a community contribution I suggest we accept the workaround PR and we raise an issue to track the real issue?

rawlingsj: Unless there is an obvious place we can help direct focus and help with a fix?

tdcox: It looks as though K8S started enforcing the prohibition against passing nil objects in fields like dates so the existing approach has stopped working. It does accept passing an explicit time stamp as valid for these objects, and that appears to be the recommended approach.

An alternative might be to modify the marshalling code to find all the edge cases involving timestamps and filter them out if the server is expected to set them but that could get messy fast.

hferentschik: > It looks as though K8S started enforcing the prohibition against passing nil objects in fields like dates so the existing approach has stopped working. It does accept passing an explicit time stamp as valid for these objects, and that appears to be the recommended approach.

Do you have a pointer to some of this? Funny that the timestamp would be enforced, but non other things like uid and resourceVersion. If the workaround is to add the timestamp, let's do that, but let' also add a comment preferably with a link to some information about this.

I'd also argue we should create a follow up issue to change what we store in the team settings. We really don't want to store actual `batchv1.Job` instances. We just need the job spec relevant data. The longer we wait with rectifying these type of things the harder it will be to change.

tdcox: I suspect that what is happening during all the marshalling and unmarshalling is that the absent timestamp field in the new request gets compared to the timestamp in the existing environment JSON and the diff produces creationTimestamp: nil rather than creationTimestamp: ""

This gets then gets rejected due to https://github.com/kubernetes/kubernetes/pull/74804

tdcox: It's moot anyway, since K8S just overrides it regardless.

PR #4307, opened 2019-06-17
fix(tekton): Pass the updated pipelineConfig, not the stale one

abayer:
#### Submitter checklist

#### Description

Without this, we use the original `pipelineConfig`, which may not have its pipelines populated.

/assign @daveconde 
/assign @hferentschik 

n/a


pkg/cmd/step/create/step_create_task.go
```diff
@@ -427,6 +425,7 @@ func (o *StepCreateTaskOptions) GenerateTektonCRDs(packsDir string, projectConfi
 	if projectConfig.NoReleasePrepare {
 		o.NoReleasePrepare = true
 	}
+	pipelineConfig := updatedProjectConfig.PipelineConfig
 	err = o.setVersionOnReleasePipelines(pipelineConfig)
```
hferentschik: So really, this is where the rub was, right? It passed the original /stale config to `setVersionOnReleasePipelines`.

hferentschik:
Even if I sound like a broken record now, I'd argue it would be worth creating an issue, describing how this bug manifested itself and how one can reproduce it.

pkg/cmd/step/create/step_create_task.go
abayer: Yup, exactly.

abayer:
@hferentschik Nope, you’re 100% right. I’ll do that.

PR #4308, opened 2019-06-18
feat: Create cluster GKE using yaml config

macox:

#### Submitter checklist

#### Description

Use [Viper](https://github.com/spf13/viper) in conjunction with the Cobra CLI to allow cluster creation config to be specified via yaml

### DIFF
--- a/pkg/cmd/create/create_cluster_gke.go
+++ b/pkg/cmd/create/create_cluster_gke.go
@@ -29,21 +29,21 @@ import (
 type CreateClusterGKEOptions struct {
 	CreateClusterOptions
 
-	Flags CreateClusterGKEFlags
+	Flags CreateClusterGKEFlags `mapstructure:"cluster"`
 }
 
 type CreateClusterGKEFlags struct {
 	AutoUpgrade     bool
-	ClusterName     string
+	ClusterName     string `mapstructure:"cluster-name"`
 	ClusterIpv4Cidr string
 	ClusterVersion  string
 	DiskSize        string
 	ImageType       string
-	MachineType     string
-	MinNumOfNodes   string
-	MaxNumOfNodes   string
+	MachineType     string `mapstructure:"machine-type"`
+	MinNumOfNodes   string `mapstructure:"min-num-nodes"`
+	MaxNumOfNodes   string `mapstructure:"max-num-nodes"`
 	Network         string
-	ProjectId       string
+	ProjectId       string `mapstructure:"project-id"`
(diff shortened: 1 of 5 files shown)


pkg/cmd/create/create_cluster_gke.go
```diff
@@ -162,6 +176,18 @@ func (o *CreateClusterGKEOptions) Run() error {
 }
 
 func (o *CreateClusterGKEOptions) createClusterGKE() error {
+	configFile := o.InstallOptions.Flags.ConfigFile
```
rawlingsj: Love this PR!  I wonder if we should move the config file flag and this section of code down into install options or something so it's reusable for all create cluster and install commands?

hferentschik: Probably all the way down to common options and the root command, so that any command can make use of this. I am also in the process of using Viper as part of binding environment variables to command options for `step create ` commands. So far I am using Viper just in the step commands directly similar to this code, but I was already wondering how this could be made more generic and usable for other commands.

pkg/cmd/create/create_cluster_gke.go
```diff
@@ -292,9 +322,11 @@ func (o *CreateClusterGKEOptions) createClusterGKE() error {
 			machineType = defaultMachineType
 			log.Logger().Infof(util.QuestionAnswer("Defaulting to machine type", machineType))
 		}
+	} else {
+		log.Logger().Infof(util.QuestionAnswer("Configured to machine type", machineType))
 	}
 
-	minNumOfNodes := o.Flags.MinNumOfNodes
+	minNumOfNodes := viper.GetString("min-num-nodes")
```
ccojocar: I would suggest that we parse the config in a central place. It is possible to define all fields in a struct and the use the `viper.Unmarshal` instead of fetching each value one by one through the code. Here is an example https://gist.github.com/chazcheadle/45bf85b793dea2b71bd05ebaa3c28644. 

This has the advantage that we can validate the fields early on and provide sensitive defaults from the very beginning.

pkg/cmd/create/create_cluster_gke.go
```diff
@@ -168,12 +194,14 @@ func (o *CreateClusterGKEOptions) createClusterGKE() error {
 		}
 	}
 
-	projectId := o.Flags.ProjectId
+	projectId := o.Flags.ProjectID
```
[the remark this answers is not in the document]
macox: No a change in this pr

PR #4320, opened 2019-06-18
feat(tekton): adding command to create and apply metapipeline

### LINKED ISSUE #3342: Pipeline extensibility for Tekton

hferentschik:
fixes #3342


hferentschik:
@abayer, had another go at `step_create_task` to make it work with the meta pipeline. Trying to keep both code paths alive until we have everything working end to end. Right now,  I am checking whether an effective project config exists and if so, just create the CRDs. If not, I old logic kicks in.

A couple of things I still want to look at but provided the old code path is not broken, we should merge asap so that we can try to sort out the last bits.

abayer:
Awesome! Great idea on keeping the old code path functional so that we don't have to nail down the `PipelineActivity`/`jx get build logs`/prow aspects of the metapipeline at the same time. Will give it a more thorough read tomorrow, but my initial impressions are overwhelming positive. =)

hferentschik:
@abayer you can see where things are going by running something like this:

```bash
$ ./build/jx --verbose step create pipeline --source-url https://github.com/hf-bee/test-go-project.git --branch master --kind release --no-apply --default-image foo/bar:1.0.0 --pull-refs master:abc --context foo
```

this should create the Tekton CRDs for the meta pipeline into the _out_ directory.

I started to integrate/use Viper to bind command line flags to environment variables. In my world, the various step commands should when executed in the pipeline receive their input from env variables.

abayer:
@hferentschik Very, very nice. I’ll try to look into the Prow side of things tomorrow.

hferentschik:
@abayer, remember the input params in the effective project config? You said this would be a cosmetic issue. How do I get them replaced? When I run the meta pipeline Tekton pipeline in the cluster, the CRD generation step fails with: 

```
: non-existent variable in "--destination=gcr.io/apps-dev-229310/test-go-project:${inputs.params.version}" for step arg[4]: taskspec.steps.arg[4]
```

Trying to figure out which piece of the puzzle I am missing atm

abayer:
@hferentschik Can you show me the `Task` CRD that's getting created?

hferentschik:
> Can you show me the Task CRD that's getting created?

No worries, I figured it out. Works now.

pkg/cmd/step/create/step_create_meta_pipeline.go
```diff
@@ -0,0 +1,325 @@
...
+
+	return tektonClient, jxClient, kubeClient, ns, nil
+}
+
+func (o *StepCreatePipelineOptions) buildLabels(gitInfo *gits.GitRepository) map[string]string {
+	// start with custom labels
+	labels, _ := util.ExtractKeyValuePairs(o.CustomLabels, "=")
+
+	// add some labels we always want to have based on the build information available
+	labels["owner"] = gitInfo.Organisation
```
abayer: Maybe we should propagate the `prowJobName` to both of the `PipelineRun`s at the very least? Or am I missing something and we already are?

hferentschik: sure, if we need that.

pkg/tekton/metapipeline/metapipeline.go
```diff
@@ -0,0 +1,278 @@
...
+
+func stepEffectivePipeline(params CRDCreationParameters) syntax.Step {
+	args := []string{"--output-dir", "."}
+	if params.Context != "" {
+		args = append(args, "--context", params.Context)
+	}
+
+	step := syntax.Step{
+		Name:      createEffectivePipelineStepName,
+		Comment:   "Pipeline step cresting the effective pipeline configuration",
```
abayer: Pretty sure you meant creating. =)

hferentschik: :-)

pkg/cmd/step/create/step_create_meta_pipeline.go
```diff
@@ -0,0 +1,325 @@
...
+
+	if o.Verbose {
+		log.Logger().Infof("creating meta pipeline CRDs for build %s of repository '%s/%s'", buildNumber, gitInfo.Organisation, gitInfo.Name)
+	}
+	extendingApps, err := metapipeline.GetExtendingApps(jxClient, ns)
+	if err != nil {
+		return err
+	}
+
+	pipelineName := tekton.PipelineResourceName(gitInfo, o.Branch, o.Context)
```
abayer: So this is an interesting thing - we probably do actually want the same `PipelineResource` name for both the metapipeline and the generated pipeline, but we need a different *name* for the `Pipeline` (and `PipelineRun`, `Task`, `PipelineStructure`). Otherwise we end up with dupes. So we should probably pass in both a `PipelineName` and a `MetaPipelineName`, both of which need to follow the `kube.ToValidTruncatedName` rules.

pkg/cmd/step/create/step_create_meta_pipeline.go
```diff
@@ -0,0 +1,325 @@
...
+		return err
+	}
+
+	podTemplates, err := o.getPodTemplates(kubeClient, ns, apps.AppPodTemplateName)
+	if err != nil {
+		return errors.Wrap(err, "unable to retrieve pod templates")
+	}
+
+	pipelineResourceName := tekton.PipelineResourceName(gitInfo, o.Branch, o.Context)
+	buildNumber, err := tekton.GenerateNextBuildNumber(tektonClient, jxClient, ns, gitInfo, o.Branch, retryDuration, pipelineResourceName)
```
abayer: Is this going to result in build number 1 for the metapipeline and build number 2 for the generated pipeline?

abayer: ...ideally they'd be the same number. Which I _think_ they will be. That'll let us use the `PipelineActivity.Pipeline` and `PipelineActivity.Build` to find both the metapipeline `PipelineRun` and the generated pipeline `PipelineRun` (based on labels on the `Pipeline`/`PipelineRun`/`Pod`/etc) for the purposes of updating the `PipelineActivity` and running `jx get build logs`. I think.

pkg/cmd/step/create/step_create_meta_pipeline.go
hferentschik: Right, this is something I'd like to discuss. I can show you how it works out atm.

abayer:
Athens is hosed, so builds are all failing.

pkg/tekton/pipelines.go
```diff
@@ -273,13 +294,17 @@ func CreateOrUpdatePipeline(tektonClient tektonclient.Interface, ns string, crea
 }
 
 // PipelineResourceName returns the pipeline resource name for the given git repository, branch and context
-func PipelineResourceName(gitInfo *gits.GitRepository, branch string, context string) string {
+func PipelineResourceName(gitInfo *gits.GitRepository, branch string, context string, pipelineType PipelineType) string {
```
hferentschik: @abayer hmm, I see quite a few names getting truncated, even before the addition of the pipeline type. For example, the names for the environment build. Is this not really an issue?

abayer: Correct. The CRD name is just cosmetic, really.

hferentschik: I see. I guess the object ID is used internally somehow. Still it gets a bit odd looking at things like `kubectl get pods` etc, everything where the name is visible.

This 64 char limit is really a pita imo.

PR #4323, opened 2019-06-19
fix: add a step to generate secrets.yaml from a template

jstrachan:
It is common for the secrets you need to inject into a variety of charts to share common values. So the new `jx step create install secrets` lets you define a template file `template-secrets.yaml` which you can safely check into git which is then used to create your `secrets.yaml` file using a more concise `secrets-parameters.yaml` file which you *do not* check into git. 

You can then use the `.gitignore` rule of `secrets*.yaml` to avoid ever checking in the `secrets-parameters.yaml` or the generated `secrets.yaml` files

e.g. you may inject your pipeline user name and token from your git provider into various charts like Tekton and Prow using the same values. Or you may want to use template functions to extract values for secrets from specific locations in the parameters file or from Vault locations etc.

So this templating mechanism lets us compose secret values from input parameters or arbitrary functions/locations in a very similar way to the use of go templates inside YAML files in helm charts themselves.

## Example

* here's the [template-secrets.yaml](https://github.com/jenkins-x/jx/blob/3997419076c02749421fbddff348022cd29ea2b1/pkg/cmd/step/create/test_data/step_create_install_secrets/template-secrets.yaml) that can be checked into git
* then here are the [secrets-parameters.yaml](https://github.com/jenkins-x/jx/blob/3997419076c02749421fbddff348022cd29ea2b1/pkg/cmd/step/create/test_data/step_create_install_secrets/secrets-parameters.yaml) which is either fetched from Vault directory or stored in some secret place (or on the local file system)
* running `jx step create install secrets` then generates this [secrets.yaml](https://github.com/jenkins-x/jx/blob/3997419076c02749421fbddff348022cd29ea2b1/pkg/cmd/step/create/test_data/step_create_install_secrets/expected-secrets.yaml)


pkg/cmd/step/create/step_create_install_secrets.go
```diff
@@ -0,0 +1,142 @@
...
+	"k8s.io/helm/pkg/chartutil"
+	"k8s.io/helm/pkg/engine"
+)
+
+const (
+	// SecretsTemplateFileName contains the template used to generate the secrets.yaml file which can be safely checked into git
+	SecretsTemplateFileName = "template-secrets.yaml"
+
+	// SecretsParametersFileName the secret parameter files which are never checked into git and may come from, say, Vault
+	SecretsParametersFileName = "secrets-parameters.yaml"
```
ccojocar: Can we use in this file the vault's URI based secrets lookup? The purpose of URI based secrets lookup was exactly to avoid committing secrets in git and have a dynamic mechanism to resolve thme. The secrets are resolved and interpolated before applying the charts. 

https://github.com/jenkins-x/environment-tekton-mole-dev/blob/783a8f6178c050c0b96c61016d1581bd0bc7a3a0/env/values.yaml#L92

I think the existing vault URI based lookup can be extended for k8s secrets and local secrets as follows: 

Vault:

``` 
vault:gitOps/jenkins-x/environment-tekton-mole-dev/feature-flag-api-key:token-passthrough
````

k8s secrets:
```
 k8s:<namespace>/<secret-name>:<secret-value>
```
Local:
```
```

```
 local:<yaml-file-path>/<secret-secret-path-in-yaml>:<secret-value>
```

File: pkg/cmd/step/create/step_create_install_secrets.go [hunk 2]
```diff
@@ -0,0 +1,142 @@
...
+)
+
+var (
+	createInstallSecretsLong = templates.LongDesc(`
+		Creates or updates the install/upgrade secrets.yaml file from a template you can safely check into git: 'template-secrets.yaml' and a local secret parameters file: 'secrets-parameters.yaml'
+
```
ccojocar: It would be great if we could be consistent with notation and with the secrets sources.

pkg/cmd/step/create/step_create_install_secrets.go
jstrachan: yeah - I did wonder about using a function in the template for this kinda thing - so `{{ .Vault.gitOps/jenkins-x/environment-tekton-mole-dev/feature-flag-api-key:token-passthrough }}` or something like that?

pmuir: Agree with @ccojocar.

* anything where there is a risk you might commit the actual secrets, we should avoid like the plauge
* the secrets plugin is odd and confusing, I think we should try to not use it at all

The URL approach is much better in my opinion and I think we can use the scheme Cosmin proposes to generalise it.

I'm not sure why we would need to use a function - is this when you are using non tillerless helm?

pmuir:
I'm 👎 on merging this PR. Instead we should extend the URLs and work out how to make them work with tillered helm

jstrachan:
@pmuir I initially thought of going all in on URLs - but then I hit things like maven `settings.xml` https://github.com/jenkins-x/jx/blob/3997419076c02749421fbddff348022cd29ea2b1/pkg/cmd/step/create/test_data/step_create_install_secrets/template-secrets.yaml#L14-L23 and docker `config.json` https://github.com/jenkins-x/jx/blob/3997419076c02749421fbddff348022cd29ea2b1/pkg/cmd/step/create/test_data/step_create_install_secrets/template-secrets.yaml#L14-L23 and even the secret for basic auth: https://github.com/jenkins-x/jx/blob/3997419076c02749421fbddff348022cd29ea2b1/pkg/cmd/step/create/test_data/step_create_install_secrets/template-secrets.yaml#L2 and figured templating was way more flexible as we are often including secret values (user/pwds) inside large amounts of text values in the generated `secrets.yaml`.

e.g. compare the input for a specific user/install needs to provide: [secrets-parameters.yaml](https://github.com/jenkins-x/jx/blob/3997419076c02749421fbddff348022cd29ea2b1/pkg/cmd/step/create/test_data/step_create_install_secrets/secrets-parameters.yaml) to the current [secrets.yaml](https://github.com/jenkins-x/jx/blob/3997419076c02749421fbddff348022cd29ea2b1/pkg/cmd/step/create/test_data/step_create_install_secrets/expected-secrets.yaml) thats generated by the generic template - so much simpler for the user to configure + the template hides the duplication inside Jenkins X / Apps and the composition of user/passwords into larger chunks of secret text like ingress passwords, maven `settings.xml` and docker `config.json` files.

The aim of the templating is to reduce the amount of user configuration required for secrets really - URLs don't help with templating

jstrachan:
@pmuir btw this is a stand alone step used for the alternative GitOps installer pipeline to bootstrap the secrets: https://github.com/jenkins-x/jx/issues/4326 - we don't need to necessarily include this into the general `jx step helm apply` logic - so folks can avoid using this approach if they don't want

jstrachan:
gonna try solve this in a different way...https://github.com/jenkins-x/jx/issues/4328

PR #4325, opened 2019-06-19
fix: make it easier to capture the preview URL from scripts/shell

### LINKED ISSUE #4324: jx get preview --current should only output the URL and not any other logging information
also use `fmt.Printf()` so that it can be easily captured in shells

jstrachan:
fixes #4324

### DIFF
--- a/pkg/cmd/get/get_preview.go
+++ b/pkg/cmd/get/get_preview.go
@@ -2,7 +2,6 @@ package get
 
 import (
 	"fmt"
-	"github.com/jenkins-x/jx/pkg/log"
 
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/get/get_preview.go
```diff
@@ -92,7 +91,8 @@ func (o *GetPreviewOptions) CurrentPreviewUrl() error {
 	}
 	for _, env := range envList.Items {
 		if env.Spec.Kind == v1.EnvironmentKindTypePreview && env.Name == name {
-			log.Logger().Infof("%s", env.Spec.PreviewGitSpec.ApplicationURL)
+			// lets log directly to stdout for easy capture of the URL from shell scripts
+			fmt.Println(env.Spec.PreviewGitSpec.ApplicationURL)
```
hferentschik: +1, I think we need to differentiate between logging and actual output of the command. The logging framework should not be used for the latter because it logs per default to stderr and because depending on the log level the expected output might not appear at all.

IMO, even better would be to let `CurrentPreviewUrl` return the URL as a string and handle the displaying in `Run`.

PR #4327, opened 2019-06-19
chore: minor rename of terraform service account

garethjevans:
brings the terraform service account name inline with the kaniko
and vault SAs


hferentschik:
commit is missing the sign-off :-(

PR #4330, opened 2019-06-19
fix: refactor out a sub-interface from Vault for VaultURL injection

jstrachan:
* so we can support local file system vault-like behaviour or real Vault from a small simple interface (which is a small subset of Vault client)
* same URL structure works for vault + local file system referencing


pmuir:
Looks good to me for the URL parts. @ccojocar you should look too though for a final lgtm :-)

pkg/vaulturl/vaulturl_client.go
```diff
@@ -0,0 +1,17 @@
+package vaulturl
```
ccojocar: I am wondering if we should name this `secreturl` instead. it is something more generic than vault. In my mind, vault is just one implementation of this interface.

pkg/vaulturl/localvault/client.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	return nil
+}
+
+// WriteObject writes a generic named object to the vault.
+// The secret _must_ be serializable to JSON.
+func (c *FileSystemClient) WriteObject(secretName string, secret interface{}) (map[string]interface{}, error) {
+	path := c.fileName(secretName)
+	dir, _ := filepath.Split(path)
+	err := os.MkdirAll(dir, util.DefaultWritePermissions)
```
ccojocar: I would use `c.Dir` instead of dir and remove the `filepath.Split`.

pkg/vaulturl/localvault/client.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to ensure that parent directory exists %s", dir)
+	}
+	err = helm.SaveFile(path, secret)
+	if err != nil {
+		return nil, err
```
ccojocar: Maybe wrap this error

pkg/vaulturl/helpers.go
```diff
@@ -0,0 +1,54 @@
...
+	"strings"
+
+	"github.com/jenkins-x/jx/pkg/util"
+	"github.com/pkg/errors"
+)
+
+var vaultURIRegex = regexp.MustCompile(`vault:[-_\w\/:]*`)
+
+// ReplaceURIs will replace any vault: URIs in a string, using the vault client
+func ReplaceURIs(s string, client Client) (string, error) {
```
ccojocar: I would rather make this a method part of `secreturl.Client`, this will allow us to extend it with other URI schemes (e.g. k8s secret).

pkg/vaulturl/helpers.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/jenkins-x/jx/pkg/util"
+	"github.com/pkg/errors"
+)
+
+var vaultURIRegex = regexp.MustCompile(`vault:[-_\w\/:]*`)
```
ccojocar: I would be explicit a use a URI scheme per client (e.g. vault:..., local:..., k8s:...) instead of hiding this. In this way, the user will be aware of the place where the secrets are stored.

pkg/vaulturl/vaulturl_client.go
jstrachan: good call!

jstrachan: then I guess the URLs need to be `secret:` too?

pkg/vaulturl/localvault/client.go
jstrachan: though the path can include folders which is why I did the Split. e.g. the test case uses a sub dir in the root `c.Dir`

pkg/vaulturl/localvault/client.go
jstrachan: I did consider - but the current message is pretty good https://github.com/jenkins-x/jx/blob/ccba49d49156e5c2b1d9594aa8fc339590132000/pkg/helm/helm_helpers.go#L318

pkg/vaulturl/helpers.go
jstrachan: the main reason I kept it here for now was to avoid touching `vault.Client` in case it interfered with your refactor ;).  Maybe we add a `ReplaceURIs` function to the `secreturl.Client` which for vault + localvault can reuse the same code + URL scheme - then do something different for k8s?

pkg/vaulturl/helpers.go
jstrachan: BTW one of the motivations of this was so that we don't have to search / replace all the values.yaml in a GitOps install if someone decides to switch from local file system secrets to vault - i.e. to use the same URI scheme in values.yaml

jstrachan: I'm hoping we can have a canonical git repo for installing Jenkins X with prow + tekton with + without vault without having to do a global replace of `vault:` <-> `local:`

pkg/vaulturl/helpers.go
jstrachan: how about we support `secret:` for either vault or local and `vault:` for only vault and `local:` for only local? then if folks want to be generic they use `secret:` URLs?

jstrachan: we could add other URLs for vault - e.g. to use a specific named vault too?

pkg/vaulturl/helpers.go
ccojocar: I would use `secret` instead. It will be difficult to use the same path for both vault and local, because vault is using a relative path to the root of the secrets backend which normally is mounted with a fix name `secret` for KV secret engine. It will be something like `secret/<path-in-vault>:secret-name` but the `secret` prefix is omitted from the URI, which is not totally correct if we want to support multiple vault secrets backends. Another backend will be mounted on a different prefix so the user won't be able to use that backend.

I think the user will be forced to store the secrets file to the same relative path which is similar with the relative path in vault.

Given this example:
https://github.com/jenkins-x/environment-tekton-mole-dev/blob/783a8f6178c050c0b96c61016d1581bd0bc7a3a0/env/values.yaml#L98

the user will have to store the secrets on a local path like this:
`<root-folder>/gitOps/jenkins-x/environment-tekton-mole-dev/jenkins-x-demo-sa:data`

pkg/vaulturl/helpers.go
ccojocar: 'secret:` sounds better to me in this case

pkg/vaulturl/localvault/client.go
ccojocar: alright

pkg/vaulturl/vaulturl_client.go
ccojocar: sgtm

jstrachan: ok refactoring to `secreturl` package but gonna stick with `vault:` and `local:` URLs

pkg/cmd/opts/common.go
```diff
@@ -1142,3 +1142,16 @@ func (o *CommonOptions) IsFlagExplicitlySet(flagName string) bool {
 	o.Cmd.Flags().Visit(explicitlySetFunc)
 	return explicit
 }
+
+// GetClusterName returns the current cluster name
+func (o *CommonOptions) GetClusterName() (string, error) {
```
ccojocar: We have already a helper function for this. I would be great to reuse it.
https://github.com/jenkins-x/jx/blob/442ddfc3176f50ef463faecb0152397dc861df38/pkg/kube/cluster/cluster.go#L14

pkg/helm/helm_helpers.go
```diff
@@ -574,7 +579,7 @@ type HelmRepoCredential struct {
 
 // DecorateWithSecrets will replace any vault: URIs with the secret from vault. Safe to call with a nil client (
 // no replacement will take place).
-func DecorateWithSecrets(options *InstallChartOptions, vaultClient vault.Client) (func(), error) {
+func DecorateWithSecrets(options *InstallChartOptions, vaultClient secreturl.Client) (func(), error) {
```
ccojocar: nit: I would rename the `vaultClient` => `secreturlClient` for readability reasons.

pkg/helm/helm_helpers.go
```diff
@@ -611,12 +619,41 @@ func DecorateWithSecrets(options *InstallChartOptions, vaultClient vault.Client)
 	return cleanup, nil
 }
 
+// LoadParameters loads the 'parameters.yaml' file if it exists in the current directory
+func LoadParameters(dir string, vaultClient secreturl.Client) (chartutil.Values, error) {
```
ccojocar: ditto: vaultClient => secreturlClient

pkg/helm/helm_helpers.go
```diff
@@ -611,12 +619,41 @@ func DecorateWithSecrets(options *InstallChartOptions, vaultClient vault.Client)
...
+	return chartutil.Values(m), err
+}
+
 // AddHelmRepoIfMissing will add the helm repo if there is no helm repo with that url present.
 // It will generate the repoName from the url (using the host name) if the repoName is empty.
 // The repo name may have a suffix added in order to prevent name collisions, and is returned for this reason.
 // The username and password will be stored in vault for the URL (if vault is enabled).
 func AddHelmRepoIfMissing(helmURL, repoName, username, password string, helmer Helmer,
-	vaultClient vault.Client, in terminal.FileReader,
+	vaultClient secreturl.Client, in terminal.FileReader,
```
ccojocar: ditto vaultClient => secreturlClient

pkg/helm/helm_helpers.go
```diff
@@ -663,7 +700,7 @@ func AddHelmRepoIfMissing(helmURL, repoName, username, password string, helmer H
 }
 
 // DecorateWithCredentials will, if vault is installed, store or replace the username or password
-func DecorateWithCredentials(repo string, username string, password string, vaultClient vault.Client, in terminal.FileReader,
+func DecorateWithCredentials(repo string, username string, password string, vaultClient secreturl.Client, in terminal.FileReader,
```
ccojocar: ditto

pkg/secreturl/client.go
```diff
@@ -0,0 +1,20 @@
+package secreturl
+
+// Client is a simple interface for Vault-like things such as `vault.Client` or a file system we can use to
+// access secret files and values in helm.
+//go:generate pegomock generate github.com/jenkins-x/jx/pkg/vault Client -o mocks/vault_client.go
```
ccojocar: name package name in the mock generator needs to be updated

github.com/jenkins-x/jx/pkg/vault => github.com/jenkins-x/jx/pkg/secreturl Client -o mocks/secreturl_client.go

pkg/cmd/opts/common.go
jstrachan: thanks! figure there must be but couldn't find it - fixed!

Issue #4334, opened 2019-06-19
step git merge only displays the help message


hferentschik:
### Summary 

Currently executing `jx step git merge` will only display the help message, no matter what arguments are passed or whether the `PULL_REFS` env variable is set.

This is due to a wrong setup of the `StepGitMergeOptions` struct.

### Steps to reproduce the behavior

Run:

`jx step git merge`

hferentschik:
fixed by pull request #4336

PR #4336, opened 2019-06-19
fix: making sure the right Run function is called for step git merge

hferentschik:
fixes issue #4334


hferentschik:
Once this is merged and a new jx builder image is created we need to update pipeline.go - https://github.com/jenkins-x/jx/blob/fc95231f13a77b149cd1c1db7d9930e71814ba78/pkg/tekton/syntax/pipeline.go#L31

Issue #4337, opened 2019-06-20
Updating default GitMergeImage for merge step in Tekton pipeline


hferentschik:
### Summary 

Currently, the image used for the merge step of the Tekton pipeline is gcr.io/jenkinsxio/builder-jx:0.1.477. This image contains a `jx` version with the big reported in issue #4334.
The image needs to be updated to a version containing an updated `jx` version

hferentschik:
Fixed by pull request #4338

PR #4338, opened 2019-06-20
fix: updating git merge image used in Tekton pipeline

hferentschik:
fixes issue #4337

### DIFF
--- a/pkg/tekton/syntax/pipeline.go
+++ b/pkg/tekton/syntax/pipeline.go
@@ -28,7 +28,7 @@ import (
 
 const (
 	// GitMergeImage is the default image name that is used in the git merge step of a pipeline
-	GitMergeImage = "gcr.io/jenkinsxio/builder-jx:0.1.477"
+	GitMergeImage = "gcr.io/jenkinsxio/builder-jx:0.1.527"
 
 	// WorkingDirRoot is the root directory for working directories.
 	WorkingDirRoot = "/workspace"
--- a/pkg/cmd/step/create/test_data/step_create_task/add-env-to-default-in-buildpack/tasks.yml
+++ b/pkg/cmd/step/create/test_data/step_create_task/add-env-to-default-in-buildpack/tasks.yml
@@ -29,7 +29,7 @@ items:
             - --verbose
           command:
             - jx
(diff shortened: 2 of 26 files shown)


pkg/tekton/syntax/pipeline.go
```diff
@@ -28,7 +28,7 @@ import (
 
 const (
 	// GitMergeImage is the default image name that is used in the git merge step of a pipeline
-	GitMergeImage = "gcr.io/jenkinsxio/builder-jx:0.1.477"
+	GitMergeImage = "gcr.io/jenkinsxio/builder-jx:0.1.527"
```
ccojocar: I would be great if we could move this out form the code into https://github.com/jenkins-x/jenkins-x-versions/tree/master/docker. It would make the update easier.

hferentschik: Do you have a pointer on how to do this? Not sure how to do it.

hferentschik: Wondering if we should merge this in the meantime, given that the 0.1.477 version is broken for merging.

ccojocar: https://github.com/jenkins-x/jx/blob/d5f7e09a56a7863ab5ae0c8cbb093f5b3c8797a6/pkg/cmd/step/create/step_create_task.go#L222

ccojocar: I think is fine to merge it. You can move it out in a follow up PR.

hferentschik: LOL, ahh that's what this is. I was wondering what this VersionResolver is, but never dived into it. That seems straight forward. There is already `docker/gcr.io/jenkinsxio/builder-jx.yml`. So really I just need to get this version and of course, ensure that version stream contains a recent enough version of builder-jx. 

For the latter, do i just need to create a pull request in the versions repo?

hferentschik:
@ccojocar created issue #4349 to follow up with. I'll use the version stream asap.

pkg/tekton/syntax/pipeline.go
ccojocar: yes, if you just introduce a new docker image in the version stream is no problem to merge the PR straight away.

PR #4348, opened 2019-06-20
feat: Updated release to use goreleaser

cagiti:
#### Description
This is part of a series of updates which intend on improving the way in which jx is released. The first part will replace progrium/gh-release for goreleaser https://goreleaser.com/

#4364

### DIFF
--- a/jenkins-x.yml
+++ b/jenkins-x.yml
@@ -107,7 +107,7 @@ pipelineConfig:
               - name: release
                 #image: docker.io/golang:1.11.5
                 # needs helm in the image for install_gitops_integration_test.go
-                image: gcr.io/jenkinsxio/builder-go:0.1.502
+                image: gcr.io/jenkinsxio/builder-go:0.1.533
                 command: make
                 args: ['release']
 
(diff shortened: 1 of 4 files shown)


jenkins-x.yml
```diff
@@ -107,7 +107,7 @@ pipelineConfig:
               - name: release
                 #image: docker.io/golang:1.11.5
                 # needs helm in the image for install_gitops_integration_test.go
-                image: gcr.io/jenkinsxio/builder-go:0.1.502
+                image: gcr.io/jenkinsxio/builder-go:0.1.533
```
jstrachan: btw wonder if it works without a version number; then it'll pick the latest from the version stream

cagiti: good call, It should then use the platform version right?

cagiti: I've removed the version if you're able to '/lgtm', thanks @jstrachan

jstrachan: hope it works now I suggested it ;) but yeah it should grab the version we maintain in the version stream: https://github.com/jenkins-x/jenkins-x-versions/blob/master/docker/gcr.io/jenkinsxio/builder-go.yml

PR #4350, opened 2019-06-20
fix: lets let `jx step create values` work with vault + local files

jstrachan:
also lets polish a little the directory structure + file names to be more concise
also lets allow the secret base path to be configured; when bootstrapping a new cluster we may not be able to find the name yet; so providing a consistent path to secrets helps avoid things moving during the bootstrap proces

### DIFF
--- a/pkg/cmd/step/create/step_create_values.go
+++ b/pkg/cmd/step/create/step_create_values.go
@@ -8,9 +8,6 @@ import (
 
 	"github.com/jenkins-x/jx/pkg/util"
 
-	"github.com/jenkins-x/jx/pkg/io/secrets"
-	"github.com/jenkins-x/jx/pkg/vault"
-
 	"github.com/pkg/errors"
 
 	"github.com/jenkins-x/jx/pkg/apps"
@@ -23,7 +20,7 @@ import (
 )
 
 var (
-	optionSecretsScheme = "secretsScheme"
+	optionSecretsScheme = "secrets-scheme"
 )
 
 var (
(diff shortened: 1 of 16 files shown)


pkg/cmd/step/create/step_create_values.go
```diff
@@ -23,7 +20,7 @@ import (
 )
 
 var (
-	optionSecretsScheme = "secretsScheme"
+	optionSecretsScheme = "secrets-scheme"
```
pmuir: ooh, nice catch!

pkg/surveyutils/jsonschema.go
```diff
@@ -771,12 +769,27 @@ func (o *JSONSchemaOptions) handleBasicProperty(name string, prefixes []string,
...
-			path := strings.Join([]string{o.VaultBasePath, strings.Join(prefixes, "-")}, "/")
+			// lets ignore the last prefix if its the same as the key
+			// to avoid redundant path
+			pathPrefixes := prefixes
+			lastIdx := len(prefixes) - 1
+			if prefixes[lastIdx] == dereferencedFormat {
+				pathPrefixes = prefixes[0:lastIdx]
+			}
+			path := strings.Join([]string{o.VaultBasePath, strings.Join(pathPrefixes, "-")}, "/")
 			secretReference := secreturl.ToURI(path, dereferencedFormat, o.VaultScheme)
```
pmuir: I think it's better to just use `prefixes[lastIdx]` as the key, rather than doing this dance of seeing if it's the same... Just so that we don't have to guess at what's happening with the paths

jstrachan: good call! another commit removes format as the key and uses the last path...

PR #4357, opened 2019-06-20
Created a step to handle JsonPatch syntax queries into Configmaps and a way to apply them into second level embedded resources

dgozalo:
…gMaps and literal json/yamls in the Data field

feature: changed the way the patch is applied

feature: added needed dependency

fix: removed unused file

feature: added a step to apply json patches to config maps and second level embedded items in config maps

#### Submitter checklist

#### Description
As a first step to be able to modify the configuration based on installed apps / addons, a step has to be created to aply JSON patches to ConfigMaps as some of them have second level embedded properties which are yaml / json themselves and Kubectl couldn't handle those.

You can specify the usual json patch types and follow the standard syntax.

Specifying the `first-level-property` flag would apply the patch into the value of the specified property only in JSONPatch format, if and only if the value is a valid yaml / json string.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,6 +36,7 @@ require (
 	github.com/dsnet/compress v0.0.0-20171208185109-cc9eb1d7ad76 // indirect
 	github.com/elazarl/goproxy v0.0.0-20181111060418-2ce16c963a8a // indirect
 	github.com/emirpasic/gods v1.9.0 // indirect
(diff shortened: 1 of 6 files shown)


dgozalo:
Mhmm, it's failing on `Step: Git Merge`, there's been a few other PRs failing there too lately.

PR #4359, opened 2019-06-20
chore: kubernetes-1.11.3 is not a module

pmuir:

#### Submitter checklist

fixes #


Makefile.codegen
```diff
@@ -26,13 +26,6 @@ generate: build-codegen install-generate-deps generate-mocks generate-openapi ge
 
 install-generate-deps:
 	$(GO) get github.com/petergtz/pegomock/...@$(PEGOMOCK_VERSION)
-	$(GO) get k8s.io/code-generator/cmd/client-gen@$(CLIENTSET_GENERATOR_VERSION)
```
hferentschik: Right, technically not needed here, but I had left it to be more explicit

cmd/codegen/app/generate_options.go
```diff
@@ -22,6 +23,7 @@ const (
 	optionBoilerplateFile = "boilerplate-file"
 	optionModuleName      = "module-name"
 	global                = "global"
+	optionVerbose         = "verbose"
```
hferentschik: 👍

cmd/codegen/generator/client_set.go
```diff
@@ -32,8 +32,8 @@ func InstallCodeGenerators(version string, gopath string) error {
 	}
 	for _, generator := range allGenerators {
 		path := fmt.Sprintf("%s/%s", basePath, generator)
-		util.AppLogger().Infof("installing %s in version %s via 'go get'", path, version)
-		err := util.GoGet(path, version, gopath, true, false)
+		util.AppLogger().Infof("installing %s version %s into %s", path, version, gopath)
+		err := util.GoGet(path, version, gopath, false, false)
```
hferentschik: Hmm, I don't think this is going to work, is it? With Go modules enabled, I don't think to specify a version works. I guess that is why the generated sources differ now.

cmd/codegen/util/commands.go
```diff
@@ -168,6 +168,8 @@ func (c *Command) RunWithoutRetry() (string, error) {
 	var r string
 	var e error
 
+	AppLogger().Debugf("Running %s %s %s", JoinMap(c.Env, " ", "="), c.Name, strings.Join(c.Args, " "))
```
hferentschik: Depending on what you are after you can print a map via `%+v`

hferentschik:
Seems there are some updates to generated files not checked in. Not sure whether the right generator versions are used.

cmd/codegen/util/commands.go
pmuir: Wanted to print the map as though it were environment variables.

pmuir:
yes - not sure quite what's going on there - i'll check

pmuir:
Ok, correct files are now checked in.

cmd/codegen/generator/client_set.go
pmuir: Right, with Go modules enabled choosing a version is a bit dodgy due to the way go implemented modues - it does all sorts of weird stuff including mucking with your `go.mod`. If you use `go get` without go modules, the code will do the `go get -d` (i.e. download only), pause, adjust the committish in use manually using git, and then run `go install` to complete the `go get`.

In essence `go get` is not really suitable for use in modules mode in the same way as it was before modules - this is quite well documented in the go issue tracker.

hferentschik: there is `gobin` which bypasses these issues and is probably the way to go for now until the go maintainers introduce better tooling.

hferentschik: >  If you use go get without go modules, the code will do the go get -d (i.e. download only), pause, adjust the committish in use manually using git, and then run go install to complete the go get.

So, you  are saying it does honor versions, just in an "awkward" way?

hferentschik:
needs a rebase

PR #4412, opened 2019-06-25
feat: Swapped out centos for alpine

cagiti:
Makes the jx image a little lighter :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
-FROM centos:7
+FROM alpine:3.10
 
-RUN yum install -y git
+RUN apk --no-cache --update add \
+    git \
+    bash
 
 ENTRYPOINT ["jx", "version"]
 


tdcox:
Is this going to have a knock-on effect to the other builder images?

Dockerfile
```diff
@@ -1,7 +1,9 @@
-FROM centos:7
+FROM alpine:3.10
```
hferentschik: +1, I love the idea of switching to Alpine. I am using Alpine myself when testing images and have a very similar Dockerfile. It makes building and pushing custom images more lightweight. For what it's worth I am using:

```
FROM alpine:3.10

RUN apk --update add ca-certificates git

COPY ./build/linux/jx /usr/bin/jx
```

I find the current `ENTRYPOINT` just gets into the way when using the image. I'd remove it.

One thing to keep in mind here is that we might want to try to align the used based images. @jstrachan and I had a discussion around this the other day and James was interested in somehow trying to align on a common base image. Also, he mentioned that centos was originally used due to some issues (DNS!?) with Alpine. For my part I have not seen any problems yet.

tdcox: Alpine has had horrendous issues relating to DNS resolution within Kubernetes that they never seemed to be able to resolve and generally blamed upon “dependencies outside their control”.

Alpine is also not compatible with certain physical infrastructure in the cloud, such as GPU cards, due to a lack of driver support. For this reason, the machine learning builders require Centos.

hferentschik: > Alpine has had horrendous issues relating to DNS resolution within Kubernetes that they never seemed to be able to resolve and generally blamed upon “dependencies outside their control”.

I guess you are referring to https://github.com/gliderlabs/docker-alpine/issues/255. I was not aware of this issue and have not noticed it before. 

I wonder though if it is really alpine specific. I wonder whether the issues we seeing at times to clone a git repo are not also related to DNS issues. But that's just a guess.

That said, I am not bound to alpine at all. I just would like to have a more leight weight default image.

> Alpine is also not compatible with certain physical infrastructure in the cloud, such as GPU cards, due to a lack of driver support. For this reason, the machine learning builders require Centos.

Sure, there are always special needs. But in the end you can always use custom images, right? 

But maybe this should be discussed on an issue instead.

hferentschik: I opened https://github.com/jenkins-x/jx/issues/4481

cagiti: > Is this going to have a knock-on effect to the other builder images?

It shouldn't, I don't believe this is used as a base image

hferentschik: How do we move forward with this PR?

hferentschik:
@cagiti shall we close this PR for now?

PR #4414, opened 2019-06-25
feat: Added auto create vault aws resource requirements

mjosephd:
Signed-off-by: Milinda Dias <mjosephd@gmail.com>

#### Submitter checklist

#### Description
This PR is used to create the AWS preresiquite resources automatically instead of relying on manual creation such as using a separate terraform script.

This would be the first OSS submission here.

fixes #

### DIFF
--- a/pkg/cloud/amazon/vault/vault_resources.go
+++ b/pkg/cloud/amazon/vault/vault_resources.go
@@ -1,0 +1,375 @@
+package vault
+
+import (
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/cloudformation"
+	"github.com/aws/aws-sdk-go/service/iam"
+	goformation "github.com/awslabs/goformation/cloudformation"
+	"github.com/awslabs/goformation/cloudformation/resources"
+	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/util"
+	"github.com/pkg/errors"
+	uuid "github.com/satori/go.uuid"
+)
+
+const stackNamePrefix = "jenkins-x-vault-stack"
+
+const (
+	readCapacityUnits  = 2
+	writeCapacityUnits = 2
+)
+
+// ResourceCreationOpts The input parameters to create a vault by default on aws
+type ResourceCreationOpts struct {
+	Region    string
+	Domain    string
+	Username  string
+	TableName string
+}
+
+// CreateVaultResources will automatically create the preresiquites for vault on aws
+func CreateVaultResources(vaultParams ResourceCreationOpts) (*string, *string, *string, *string, error) {
+	log.Logger().Infof("Creating vault presiquite resources with following values, %s, %s, %s, %s",
+		util.ColorInfo(vaultParams.Region),
+		util.ColorInfo(vaultParams.Domain),
+		util.ColorInfo(vaultParams.Username),
+		util.ColorInfo(vaultParams.TableName))
+
+	template := goformation.NewTemplate()
+
+	awsDynamoDbKey := "AWSDynamoDBTable"
+	dynamoDbTable := createDynamoDbTable(vaultParams.TableName)
+	template.Resources[awsDynamoDbKey] = &dynamoDbTable
+
+	awsS3BucketKey := "AWSS3Bucket"
+	s3Name, s3Bucket, err := createS3Bucket(vaultParams.Region, vaultParams.Domain)
+	if err != nil {
+		return nil, nil, nil, nil, errors.Wrapf(err, "Generating the vault cloudformation template failed")
+	}
+	template.Resources[awsS3BucketKey] = s3Bucket
+
+	awsIamUserKey := "AWSIAMUser"
+	iamUser := createIamUser(vaultParams.Username)
+	template.Resources[awsIamUserKey] = &iamUser
+
+	awsKmsKey := "AWSKMSKey"
+	kmsKey, err := createKmsKey([]string{awsIamUserKey})
+	if err != nil {
+		return nil, nil, nil, nil, errors.Wrapf(err, "Generating the vault cloudformation template failed")
+	}
+	template.Resources[awsKmsKey] = kmsKey
+
+	awsIamPolicy := "AWSIAMPolicy"
+	policy := createIamUserPolicy(vaultParams.Username, []string{awsDynamoDbKey, awsS3BucketKey, awsKmsKey, awsIamUserKey})
+	template.Resources[awsIamPolicy] = &policy
+
+	log.Logger().Infof("Generating the vault cloudformation template")
+
+	// and also the YAML AWS CloudFormation template
+	yaml, err := template.JSON()
+	if err != nil {
+		return nil, nil, nil, nil, errors.Wrapf(err, "Generating the vault cloudformation template failed")
+	}
+
+	log.Logger().Infof("Generated the vault cloudformation template successfully")
+
+	// had issues with json on resource, temporary workaround
+	yamlProcessed := string(yaml)
+	yamlProcessed = strings.Replace(yamlProcessed, "\"BUCKET_LIST_JSON\"", `{ "Fn::Sub" : "${`+awsS3BucketKey+`.Arn}" }`, -1)
+	yamlProcessed = strings.Replace(yamlProcessed, "\"BUCKET_JSON\"", `{ "Fn::Sub" : "${`+awsS3BucketKey+`.Arn}/*" }`, -1)
+	yamlProcessed = strings.Replace(yamlProcessed, "\"KMS_SUB_JSON\"", `{ "Fn::Sub" : "${`+awsKmsKey+`.Arn}" }`, -1)
+	yamlProcessed = strings.Replace(yamlProcessed, "\"DYNAMODB_JSON\"", `{ "Fn::Sub" : "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*" }`, -1)
+	yamlProcessed = strings.Replace(yamlProcessed, "\"KMS_USER_JSON\"", `[{ "Fn::Sub": "arn:aws:iam::${AWS::AccountId}:root"}, { "Fn::Sub": "arn:aws:iam::${AWS::AccountId}:user/`+vaultParams.Username+`"}]`, -1)
+
+	valueUUID, err := uuid.NewV4()
+	if err != nil {
+		return nil, nil, nil, nil, errors.Wrapf(err, "Generating UUID failed")
+	}
+
+	// Create dynamic stack name
+	stackName := stackNamePrefix + valueUUID.String()[:7]
+
+	runCloudformationTemplate(&yamlProcessed, &stackName)
+
+	kmsID, err := getKmsID(awsKmsKey, stackName)
+	if err != nil {
+		return nil, nil, nil, nil, errors.Wrapf(err, "Generating the vault cloudformation template failed")
+	}
+
+	accessKey, keySecret, err := createAccessKey(vaultParams.Region, vaultParams.Username)
+	if err != nil {
+		return nil, nil, nil, nil, errors.Wrapf(err, "Generating the vault cloudformation template failed")
+	}
+
+	return accessKey, keySecret, kmsID, s3Name, nil
+}
+
+func getKmsID(kmsKey string, stackName string) (*string, error) {
+	log.Logger().Infof("Retrieving the vault kms id")
+	sess := session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: session.SharedConfigEnable,
+	}))
+
+	svc := cloudformation.New(sess)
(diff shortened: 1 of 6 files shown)


mjosephd:
@ccojocar new PR mentioned in slack, apologies for the confusion.

mjosephd:
Resolved conflicts

mjosephd:
@jstrachan @ccojocar could you guys advice on how to proceed, thanks.

mjosephd:
@jstrachan @ccojocar could you guys advice on how to proceed or if there is anything else needed from my side, thanks.

daveconde:
@mjosephd could you please rebase and resolve conflicts and we'll try and get this running with our EKS BDD tests, once they are passing we can merge it.

mjosephd:
@daveconde done.

daveconde:
I have tried to run the code from the PR here https://github.com/jenkins-x/jenkins-x-versions/pull/352 I'm currently seeing an error in cloudformation: `AWSDynamoDBTable | CREATE_FAILED | vault-data already exists` is it possible to add checks to see if resources already exist?

mjosephd:
Sure. Though to clarify a few things, I add a random value suffix to the S3 bucket name and the CloudFormation name, though I haven't done for all resources. I'll go ahead and continue to add the suffixes to all the resources?

daveconde:
Thanks, I'll kick off the tests again once the other resources have the extra prefixes. Thanks for your contribution.

mjosephd:
Done

daveconde:
I am now getting an error due to requiring permissions for IAM:CreateUser. Would it be possible to be able to pass in an existing user with the correct permissions so these could be set up ahead of time by someone with the correct level of access?

dgozalo:
I've been looking into this and I can't understand the need for an User other than the current limitation on EKS that doesn't allow pods to assume roles, which would be the preferred approach.

If this is the case, we may want to look into other possibilities, as I don't think using an User, with a permanent set of credentials and authorizations is safe enough to use for Vault.

mjosephd:
@daveconde This was an error on CloudFormation? Let me retry and get back to you over the weekend.

Also could you share some details of the error, did it happen when you ran it once or on subsequent runs after the first one.

Also the PR is essentially based on the following
https://jenkins-x.io/getting-started/aws-terraform-install-gitops/

I decided not to change anything and stick to what was already there in terms of resources.

Which also creates a user IIRC from the code

```
# Create service account for vault. Should the policy
resource "aws_iam_user" "vault" {
  name = "vault_${var.region}"
}
```

hferentschik:
hi, just wondering what's the state on this PR?

BTW, Semantic Pull Request checks are failing since the commit messages are not according to the conventional commit [guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.4/).

ccojocar:
> @ccojocar what's your take on this? Given @dgozalo's security concerns is this something to move forward with?

I need to review the exact set of permissions but this is a good practice in security to have a dedicated user with a limit set of permissions and scope instead of a single user for entire infrastructure (all EKS cluster pods) with a large set of permissions (separation of concerns). This user should be limited only to the vault pods. Also the set of permissions is actually fixed and defined by vault requirements. In fact vault itself creates dynamically iam users for AWS secret engine (https://www.vaultproject.io/docs/secrets/aws/index.html#iam_user). 

I see the tension from operability point of view that the user provisioning the vault needs to have enough permissions to be able to create this dedicated IAM user and keep track of it for security audit and stuff. On the other hand, if we give access to KMS to all Pods in the cluster then means that any pod will be able to decrypt the vault secrets and in this way we bypass the vault's auth system.

pkg/cloud/amazon/vault/vault_resources.go
```diff
@@ -0,0 +1,374 @@
...
+	input := &iam.CreateAccessKeyInput{
+		UserName: aws.String(username),
+	}
+	result, err := svc.CreateAccessKey(input)
+
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "Unable to create access key")
+	}
+
+	log.Logger().Infof("Created the vault access key successfully")
```
ccojocar: log level info => debug

pkg/cloud/amazon/vault/vault_resources.go
```diff
@@ -0,0 +1,374 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "Unable to create vault preresiquite resources")
+	}
+
+	log.Logger().Infof("Ran the vault cloudformation template successfully")
+	return nil
+}
+
+/**
+We can create an access key using cloudformation, however this results the secret being logged.
```
ccojocar: when is the secret logged? When the CloudFormation template gets applied?

pkg/cmd/create/create_vault.go
```diff
@@ -326,34 +331,68 @@ func (o *CreateVaultOptions) createVaultGKE(vaultOperatorClient versioned.Interf
...
+	if o.AutoCreate == true {
 		defaultRegion, err := amazon.ResolveRegionWithoutOptions()
 		if err != nil {
 			return errors.Wrap(err, "finding default AWS region")
 		}
-		log.Logger().Infof("Region not specified, defaulting to %s", util.ColorInfo(defaultRegion))
-		if o.DynamoDBRegion == "" {
-			o.DynamoDBRegion = defaultRegion
+
+		o.DynamoDBRegion = defaultRegion
```
ccojocar: Shouldn't use first used the configured region and only when it is empty should detect the default?

pkg/cmd/create/create_vault.go
```diff
@@ -326,34 +331,68 @@ func (o *CreateVaultOptions) createVaultGKE(vaultOperatorClient versioned.Interf
...
+
+		accessKey, keySecret, kmsID, s3Name, tableName, err := awsvault.CreateVaultResources(awsvault.ResourceCreationOpts{
+			Region:    defaultRegion,
+			Domain:    domain,
+			Username:  username,
+			TableName: autoCreateTableName,
+		})
+
+		if err != nil {
+			return errors.Wrap(err, "An error occurred while creating the vault resources")
```
ccojocar: error messages are not capitalised

pkg/cmd/create/create_vault.go
```diff
@@ -326,34 +331,68 @@ func (o *CreateVaultOptions) createVaultGKE(vaultOperatorClient versioned.Interf
...
+
+		o.S3Bucket = *s3Name
+		o.KMSKeyID = *kmsID
+		o.AccessKeyID = *accessKey
+		o.SecretAccessKey = *keySecret
+		o.DynamoDBTable = *tableName
+
+	} else {
+		if o.S3Bucket == "" {
+			return fmt.Errorf("Missing S3 bucket flag")
```
ccojocar: error messages are not capitalised

pkg/cmd/create/create_vault.go
```diff
@@ -326,34 +331,68 @@ func (o *CreateVaultOptions) createVaultGKE(vaultOperatorClient versioned.Interf
...
+			Domain:    domain,
+			Username:  username,
+			TableName: autoCreateTableName,
+		})
+
+		if err != nil {
+			return errors.Wrap(err, "An error occurred while creating the vault resources")
+		}
+
+		o.S3Bucket = *s3Name
```
ccojocar: Is there any particular reason to use pointers here instead of plain strings?

pkg/cmd/create/create_vault.go
```diff
@@ -326,34 +331,68 @@ func (o *CreateVaultOptions) createVaultGKE(vaultOperatorClient versioned.Interf
...
+		o.AccessKeyID = *accessKey
+		o.SecretAccessKey = *keySecret
+		o.DynamoDBTable = *tableName
+
+	} else {
+		if o.S3Bucket == "" {
+			return fmt.Errorf("Missing S3 bucket flag")
+		}
+		if o.KMSKeyID == "" {
+			return fmt.Errorf("Missing AWS KMS key id flag")
```
ccojocar: ditto

pkg/cmd/create/create_vault.go
```diff
@@ -326,34 +331,68 @@ func (o *CreateVaultOptions) createVaultGKE(vaultOperatorClient versioned.Interf
...
 		}
-		if o.KMSRegion == "" {
-			o.KMSRegion = defaultRegion
+		if o.AccessKeyID == "" {
+			return fmt.Errorf("Missing AWS access key id flag")
```
ccojocar: ditto

pkg/cmd/create/create_vault.go
```diff
@@ -326,34 +331,68 @@ func (o *CreateVaultOptions) createVaultGKE(vaultOperatorClient versioned.Interf
...
 		}
-		if o.S3Region == "" {
-			o.S3Region = defaultRegion
+		if o.SecretAccessKey == "" {
+			return fmt.Errorf("Missing AWS secret access key flag")
```
ccojocar: ditto

mjosephd:
@ccojocar sure will get back to you, also in the meantime regarding the IAM user for vault question brought up by @hferentschik and @dgozalo are there any changes required with regards to that or would everything be fine as it is for **this** pr?

Also will update commits to match Semantic Pull Request.

pkg/cmd/create/create_vault.go
mjosephd: I initially tried using the return type `string` however that prevented me from returning `nil` as I got `cannot convert nil (untyped nil value) to string`. This is my first time at go so if there is a better/more standardish way at approaching this I can go ahead refactor.

pkg/cloud/amazon/vault/vault_resources.go
mjosephd: So basically if we use something like `Fn::GetAtt": [ "MyAccessKey", "SecretAccessKey" ]` it will get logged on the CloudFront logs IIRC. Apparently it was suggested not to use this approach to get secret key values (ref: https://forums.aws.amazon.com/thread.jspa?threadID=266209&tstart=0) thus using this method.

pkg/cloud/amazon/vault/vault_resources.go
```diff
@@ -0,0 +1,374 @@
...
+	policyDocument := resources.AWSIAMPolicy{
+		PolicyName: policyName,
+		Users:      []string{username},
+		PolicyDocument: PolicyRoot{
+			Version: "2012-10-17",
+			Statement: []PolicyDocument{
+				{
+					Sid:    "DynamoDB",
+					Effect: "Allow",
+					Action: []string{
```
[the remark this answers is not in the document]
mjosephd: I got this list from this section. https://jenkins-x.io/getting-started/aws-terraform-install-gitops/

pkg/cloud/amazon/vault/vault_resources.go
```diff
@@ -0,0 +1,374 @@
...
+	// and also the YAML AWS CloudFormation template
+	yaml, err := template.JSON()
+	if err != nil {
+		return nil, nil, nil, nil, nil, errors.Wrapf(err, "Generating the vault cloudformation template failed")
+	}
+
+	log.Logger().Infof("Generated the vault cloudformation template successfully")
+
+	// had issues with json on resource, temporary workaround
+	yamlProcessed := string(yaml)
```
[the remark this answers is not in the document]
mjosephd: I had issues mapping keys such as mapping "Fn::getAttr" to an attribute name then. I just started to have a look and it seems like I need to add `json:fieldname` to the struct. I'll refactor this and push an updated version.

mjosephd:
@ccojocar I pushed changes minus the template refactoring to have a look.

pkg/cloud/amazon/vault/vault_resources.go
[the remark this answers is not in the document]
ccojocar: Great! Thanks!

pkg/cloud/amazon/vault/vault_resources.go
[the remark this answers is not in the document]
ccojocar: Alright! I seems that they are all required according to the vault's documentation. https://www.vaultproject.io/docs/configuration/storage/dynamodb.html

pkg/cmd/create/create_vault.go
ccojocar: Usually, You can return an empty string `""` with an error instead of a pointer nil with an error. The error already signals that the returned strings are not valid.

pkg/cloud/amazon/vault/vault_resources.go
ccojocar: Thanks for clarification!

mjosephd:
@ccojocar done.

Note that with regards to mapping the Json to "Fn::Sub", there were issues. One of which it was attempted to be evaluated pre-AWS Cloud formation run (when the template is generated) but it did not evaluate correctly, there was no way that I could find to bypass Intrinsic processing either (reference : https://github.com/awslabs/goformation/issues/62). As a result I mapped the json to "Fn:Sub" (single colon) and did a string replace on the generated template string to make it double colon.

mjosephd:
@ccojocar, I noticed tekton has failed, I cannot check the results apparently. Is there anything required from my side? please let me know.

Issue #4362, opened 2019-06-20
metapipeline: PipelineActivity and jx get build logs need to treat generator and execution PipelineRuns as one pipeline


abayer:
`PipelineActivity` and `jx get build logs` currently assume every Tekton build has one and only one `PipelineRun` - but that's not going to be the case once the metapipeline (see #4320 and #4361) comes along. We're going to need to modify `PipelineRunInfo` and related code to treat both the CRD generator `PipelineRun` and the resulting actual execution `PipelineRun` as a single build, with proper ordering and results. The user shouldn't need to know that these are technically two separate "builds" behind the scenes - in fact, if the generator `PipelineRun` succeeds, we may not even want to bother exposing it in `PipelineActivity`? In any case, we need to figure this out. =)

cc @hferentschik @joseblas @dgozalo

deanesmith:
@abayer If this is not addressed in reasonable time, is the only risk here that in the event of a metapipeline execution only a portion of log data becomes available?

abayer:
Yeah, basically. That said, I don't think this will be *too* hard.

PR #4416, opened 2019-06-25
chore(tekton): Add build number to Tekton CRD labels

abayer:
#### Submitter checklist

#### Description

This will be helpful in the context of #4320 and #4362 for identifying all `PipelineRun`s etc for a particular `PipelineActivity`.

/assign @hferentschik 
/assign @dwnusbaum 

n/a

### DIFF
--- a/pkg/cmd/step/create/step_create_task.go
+++ b/pkg/cmd/step/create/step_create_task.go
@@ -610,13 +610,14 @@ func (o *StepCreateTaskOptions) createPipelineTaskParams() []pipelineapi.Param {
 func (o *StepCreateTaskOptions) setBuildValues() error {
 	labels := map[string]string{}
 	if o.GitInfo != nil {
-		labels["owner"] = o.GitInfo.Organisation
-		labels["repo"] = o.GitInfo.Name
+		labels[tekton.LabelOwner] = o.GitInfo.Organisation
+		labels[tekton.LabelRepo] = o.GitInfo.Name
 	}
-	labels["branch"] = o.Branch
+	labels[tekton.LabelBranch] = o.Branch
 	if o.Context != "" {
-		labels["context"] = o.Context
+		labels[tekton.LabelContext] = o.Context
 	}
+	labels[tekton.LabelBuild] = o.BuildNumber
 	return o.combineLabels(labels)
 }
 
--- a/pkg/cmd/step/create/test_data/step_create_task/add-env-to-default-in-buildpack/pipeline.yml
+++ b/pkg/cmd/step/create/test_data/step_create_task/add-env-to-default-in-buildpack/pipeline.yml
@@ -6,6 +6,7 @@ metadata:
   namespace: jx
   labels:
     branch: master
+    build: "1"
     owner: abayer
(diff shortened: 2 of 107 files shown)


pkg/cmd/step/create/step_create_task.go
```diff
@@ -610,13 +610,14 @@ func (o *StepCreateTaskOptions) createPipelineTaskParams() []pipelineapi.Param {
 func (o *StepCreateTaskOptions) setBuildValues() error {
 	labels := map[string]string{}
 	if o.GitInfo != nil {
-		labels["owner"] = o.GitInfo.Organisation
-		labels["repo"] = o.GitInfo.Name
+		labels[tekton.LabelOwner] = o.GitInfo.Organisation
```
dwnusbaum: This has no relation to the "owner" label [here](https://github.com/jenkins-x/jx/blob/e466371900aa308b36bc3c9dcb76876c976c42ba/pkg/apis/jenkins.io/v1/types_pipeline.go#L12), right? I tried to do a quick search to see if anything relevant was keying directly on "owner" but didn't find anything obvious.

pkg/cmd/step/create/step_create_task.go
```diff
@@ -610,13 +610,14 @@ func (o *StepCreateTaskOptions) createPipelineTaskParams() []pipelineapi.Param {
...
+		labels[tekton.LabelRepo] = o.GitInfo.Name
 	}
-	labels["branch"] = o.Branch
+	labels[tekton.LabelBranch] = o.Branch
```
dwnusbaum: Will this break anything in `activity.go`? I'm not sure if these labels make their way to `PipelineActivity` (if that is even what activity.go is about) or if the labels on `PipelineActivity` are set by something else. Right now that code uses `v1.LabelBranch` ("branch") in a few places ([example](https://github.com/jenkins-x/jx/blob/f4eaaee84a0b4c2a5497a50a203008879a3f1310/pkg/kube/activity.go#L318)). I think all of that code is unrelated to and unaffected by this change, but mentioning it just in case.

pkg/cmd/step/create/step_create_task.go
abayer: Correct. I'm not sure if anything's using the labels on `PipelineActivity`, but this is definitely independent of those.

pkg/cmd/step/create/step_create_task.go
abayer: Argh, actually, I think this might. Putting it on hold to dig into tomorrow.

abayer:
I've switched back to non-prefixed labels for now - there are too many considerations to deal with in terms of handling older builds, etc to make it worth going through everywhere that cares about the labels to make sure they're properly prefixed. We'll cross that bridge later.

pkg/cmd/step/create/step_create_task.go
jstrachan: our UI is using `owner` and `repo` labels :)

jstrachan: but yeah, all looks good

PR #4419, opened 2019-06-26
Corrected version number

cagiti:


cagiti:
can't believe I missed the major and minor version

Issue #4292, opened 2019-06-17
Refactor auth config to be self-contained


ccojocar:
### Summary 

Refactor auth-config to be self-contained to avoid the large number of issues due to overwriting of auth config by various commands.

Clearly define the location where the config is saved: local, vault, k8s secrets.

Separate the local user from pipeline user. The pipeline user should be stored separately either in vault or k8s secrets.

Cleanup the existing usage of oauth and isolate the configuration into the clients factory where the provider are created. 

### Jx version

The output of `jx version` is:

```
COPY OUTPUT HERE
```

### Jenkins type

- [ ] Next Generation (Tekton + Prow)
- [ ] Classic Jenkins
- [ ] Serverless Jenkins (JenkinsFileRunner + Prow)

Issue #3593, opened 2019-04-10
Split the personal auth token from pipeline auth token when storing secrets in Vault


ccojocar:
### Summary 

The user should be able to define a personal token besides of the pipeline token, which is stored locally on users machine (e.g. gitAuth.yaml), while the pipeline token is stored in vault.

### Jx version

The output of `jx version` is:

```
COPY OUTPUT HERE
```

### Jenkins type

- [ ] Classic Jenkins
- [ ] Serverless Jenkins

daveconde:
needs #5114

ccojocar:
Some work has been done to use the local token for some commands such as `import` and `create quickstart` but it requires addition work to handle this consistently across all commands running locally.

ccojocar:
related to #5463

Issue #3517, opened 2019-04-02
jx create terraform ... is not prompting for the user bot API Token


carlosrodlop:
### Summary 

During `jx create terraform ...` is not prompting for the user bot `API Token` thus the process is failing when it tries to create the terraform repos

```
Creating repository jx-support-beescloud/organisation-jx-support-beescloud
Creating Git repository jx-support-beescloud/organisation-jx-support-beescloud
error: Failed to create repository jx-support-beescloud/organisation-jx-support-beescloud due to: POST https://api.github.com/orgs/jx-support-beescloud/repos: 401 Bad credentials []
```

### Steps to reproduce the behavior

Full input command line as

```
jx create terraform --environment-git-owner jx-support-beescloud -o jx-support-beescloud --default-admin-password=mypassword --verbose
```

Console wizard

```
Your browser has been opened to visit:

    https://accounts.google.com/o/oauth2/auth?redirect_uri=http%3A%2F%2Flocalhost%3A8085%2F&prompt=select_account&response_type=code&client_id=32555940559.apps.googleusercontent.com&scope=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fuserinfo.email+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcloud-platform+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fappengine.admin+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcompute+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Faccounts.reauth&access_type=offline


Opening in existing browser session.
Checking Terraform Version...
Determined terraform version as 0.11.2
Terraform version appears to be valid
Git configured for user: carlosrodlop and email it.carlosrodlop
? How many clusters shall we create? 1
? Cluster 1 name: dev
? Cluster 1 provider: gke
? Would you like to install Jenkins X in cluster dev Yes
Using Git provider GitHub at https://github.com
? Do you wish to use jx-support-beescloud-bot as the Git user name? Yes


About to create repository organisation-jx-support-beescloud on server https://github.com with user jx-support-beescloud-bot
? Enter the new repository name:  organisation-jx-support-beescloud


Creating repository jx-support-beescloud/organisation-jx-support-beescloud
Creating Git repository jx-support-beescloud/organisation-jx-support-beescloud
[...]

PR #4425, opened 2019-06-26
[wip] refactor: clean up and refactor the auth config throughout the entire project

### LINKED ISSUE #4196: Pipeline bot token and users personal git token are getting mixed up on install
When installing Jenkins X the install process asks for a personal git username and token which is added to a local gitAuth.yaml file to be used later when creating quickstarts etc.  Then a pipeline bot token is asked for, this should be used by prow and tekton when cloning repos during builds and automated comments on issues / PRs etc.

If you have previously created a cluster or performed an install then a user will already have this gitAuth.yaml file with their personal git token as the first entry.  When a second install is made then jx automatically uses this existing git user as the pipeline user which results in a personal git token being used by the pipelines.

This is a significant issue which has caused lots of failed installs and confusion as it is non trivial to update the bot token once the install has happened.

@ccojocar has started work on pulling the pipeline bot token out from the gitAuth.yaml so that will only ever include a users token and the cluster will only ever have a pipeline bot username + token.

Raising this issue to track it.

/kind bug
/important critical-urgent
/area install
/assign ccojocar

ccojocar:
#### Submitter checklist

#### Description

The goal of this refactoring and cleanup is to:
- make the auth config more immutable and self-contained
- define a clear data model for the auth config
- keep the config loading in one place
- define services for config loading from various sources such as local file, k8s secrets, vault
- remove various local workarounds which merge the local config with the config from k8s secrets
- avoid duplicating the user auth config outside of the server auth config object
- clean up the git provider to only use the server auth config and rely on the current user set in that config
- split the local git auth config from the pipeline config (keep separate config services which are able to load this config from different sources).

This is still WIP and it does not entirely compile. 

There is still work to:
- Rework, cleanup and integrate the new auth config and git providers into the commands (it needs to be integrated in the existing `install` command until `boot` is ready)
- Make sure the various commands only updates the config over the services
- Rework and update the unit tests and integration tests
- Get the BDD tests passing 
- Validation of main configuration and commands:
   -  installation with Tekton + Prow and Jenkins classic 
   - With and without vault 
   - Test main commands execution locally and in the pipeline

fixes #4196 #4292 #2313 #3055 #3593 #3042 #981 #3517


ccojocar:
I am closing this PR because it is impossible to rebase at this time in order to continue with the work.

PR #4430, opened 2019-06-26
fix(tekton): Metapipeline should pass env/labels to generated pipeline

### LINKED ISSUE #4431: Metapipeline generation needs tests to verify expected arguments to step create task
The metapipeline's arguments to `step create task` aren't currently tested - we should be verifying that the right arguments are passed on to `step create task` and that they match what's passed into `step create metapipeline`. That means things like labels and environment variables.

abayer:
#### Submitter checklist

#### Description

We should be passing the custom labels/env vars/trigger/service account/source directory that we'll get from Prow (see https://github.com/jenkins-x/test-infra/blob/pipelinerunner/prow/cmd/pipeline/controller.go#L731-L739, though it's not actually passing anything but labels currently) on to the generated pipeline, not just use them on the metapipeline itself.

This also deals with propagating the `prowJobName` label, since that comes from Prow itself.

/assign @hferentschik 

Fixes #4431


abayer:
I also created #4431 to get tests for this, which I may or may not end up doing in this PR. =)

abayer:
Ok, now it does fix #4431. =)

abayer:
This also prepends `metapipeline-` in `tekton.PipelineResourceName` rather than appending it, so that it doesn't get truncated away.

pkg/cmd/step/create/step_create_meta_pipeline.go
```diff
@@ -141,7 +141,7 @@ func (o *StepCreatePipelineOptions) Run() error {
 		return errors.Wrap(err, "unable to retrieve pod templates")
 	}
 
-	pipelineName := tekton.PipelineResourceName(gitInfo, o.Branch, o.Context, tekton.MetaPipeline)
+	pipelineName := tekton.PipelineResourceNameFromGitInfo(gitInfo, o.Branch, o.Context, tekton.MetaPipeline)
```
hferentschik: +1,  I was not happy with this name either

pkg/tekton/metapipeline/metapipeline.go
```diff
@@ -44,8 +46,8 @@ type CRDCreationParameters struct {
 	PodTemplates   map[string]*corev1.Pod
 	Trigger        string
 	ServiceAccount string
-	Labels         map[string]string
-	EnvVars        []corev1.EnvVar
+	Labels         []string
+	EnvVars        []string
```
hferentschik: +1

pkg/tekton/metapipeline/metapipeline.go
```diff
@@ -164,9 +170,18 @@ func stepEffectivePipeline(params CRDCreationParameters) syntax.Step {
 func stepCreateTektonCRDs(params CRDCreationParameters) syntax.Step {
 	args := []string{"--clone-dir", filepath.Join(tektonBaseDir, params.SourceDir)}
 	args = append(args, "--build-number", params.BuildNumber)
+	args = append(args, "--trigger", params.Trigger)
```
hferentschik: +1, I guess now everything is explicitly passed what is needed for step_create_task. 

I still "somehow" would like to get away from this args building stuff, but not quite yet sure how. For now this makes  most sense.

pkg/tekton/pipelines.go
```diff
@@ -293,17 +293,20 @@ func CreateOrUpdatePipeline(tektonClient tektonclient.Interface, ns string, crea
...
+// PipelineResourceName returns the pipeline resource name for the given git org, repo name, branch and context
+func PipelineResourceName(organisation string, name string, branch string, context string, pipelineType PipelineType) string {
 	dirtyName := organisation + "-" + name + "-" + branch
 	if context != "" {
 		dirtyName += "-" + context
 	}
 
 	if pipelineType == MetaPipeline {
-		dirtyName += "-" + pipelineType.String()
+		dirtyName = pipelineType.String() + "-" + dirtyName
```
hferentschik: I was undecided on this one. The reason I in the end but it in the middle was, that if you do something like `kubectl get pipelines` you get the meta and execution pipeline "next" to each other due to the alphabetical order. Now you get the build and meta pipelines "grouped" together. 

The big drawback, of course, being that due to truncation it was hard to differentiate.

PR #4433, opened 2019-06-26
fix(tekton): Git org and repo should be replaced consistently

### LINKED ISSUE #4432: Mismatch in REPLACE_ME_APP_NAME during import and pipeline generation
### Summary 
There is an issue with **jx step syntax effective**. 

During **jx import** the folder name inside **charts** and **REPLACE_ME_APP_NAME** is replaced by **kube.ToValidName(strings.ToLower(repo_name))**. 

But during **jx step syntax effective** it's replaced by just raw repo_name causing inconsistencies

Create a repo with name a_b_c
Run **jx import --dry-run** and check charts folder app name
Run **jx step syntax effective** and check dir
There is a mismatch

### Expected behavior
Both names are same

### Actual behavior
Names are different

### Jx version

```
NAME               VERSION
jx                 2.0.329
Kubernetes cluster v1.11.8-eks-7c34c0
kubectl            v1.14.0
helm client        v2.11.0+g2e55dbe
helm server        v2.14.0+g05811b8
git                git version 2.13.1
Operating System   Mac OS X 10.14.4 build 18E227
```
### Jenkins type

- [x] Next Generation (Tekton + Prow)
- [ ] Classic Jenkins
- [ ] Serverless Jenkins (JenkinsFileRunner + Prow)

### Kubernetes cluster

What kind of Kubernetes cluster are you using & how did you create it?
EKS

abayer:
#### Submitter checklist

#### Description

`import` does the same replacements as `pkg/jenkinsfile/pipeline.go` does, but for Git org and repo, the replacement goes through `kube.ToValidName(strings.ToLower(...))`. We should be doing that in `pipeline.go` for build pack transformation as well.

And while I was here, I moved `pkg/kube/names.go` to `pkg/util/names.go` and renamed the functions to `ToValidKubeName` (and so on), since the `kube` package depends on way too many other packages as is (so I couldn't use it in the `jenkinsfile` package, for example).

/assign @dwnusbaum 
/assign @hferentschik 

fixes #4432

### DIFF
--- a/pkg/apps/install.go
+++ b/pkg/apps/install.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/jenkins-x/jx/pkg/kube/naming"
 	rbacv1 "k8s.io/api/rbac/v1"
 
 	corev1 "k8s.io/api/core/v1"
@@ -600,7 +601,7 @@ func toValidName(appName string, name string, id string) string {
 	if l > 20 {
 		l = 20
 	}
-	return kube.ToValidName(fmt.Sprintf("%s-%s", base[0:l], id))
+	return naming.ToValidName(fmt.Sprintf("%s-%s", base[0:l], id))
 }
 
 func (o *InstallOptions) getPrefixes() []string {
--- a/pkg/cmd/controller/controller_commitstatus.go
+++ b/pkg/cmd/controller/controller_commitstatus.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
+	"github.com/jenkins-x/jx/pkg/kube/naming"
 
 	"github.com/jenkins-x/jx/pkg/cmd/opts"
 	"github.com/jenkins-x/jx/pkg/prow/config"
@@ -271,7 +272,7 @@ func (o *ControllerCommitStatusOptions) onPod(pod *corev1.Pod, jxClient jenkinsv
 					buildNumber = buildId
 				}
 
-				pipelineActName := kube.ToValidName(fmt.Sprintf("%s-%s-%s-%s", org, repo, branch, buildNumber))
+				pipelineActName := naming.ToValidName(fmt.Sprintf("%s-%s-%s-%s", org, repo, branch, buildNumber))
 
 				// PLM TODO This is a bit of hack, we need a working build controller
 				// Try to add the lastCommitSha and gitUrl to the PipelineActivity
@@ -311,7 +312,7 @@ func (o *ControllerCommitStatusOptions) onPod(pod *corev1.Pod, jxClient jenkinsv
 
 						for _, ctx := range contexts {
 							if pullRequest != "" {
-								name := kube.ToValidName(fmt.Sprintf("%s-%s-%s-%s", org, repo, branch, ctx))
+								name := naming.ToValidName(fmt.Sprintf("%s-%s-%s-%s", org, repo, branch, ctx))
 
 								err = o.UpsertCommitStatusCheck(name, pipelineActName, sourceUrl, sha, pullRequest, ctx, pod.Status.Phase, jxClient, ns)
 								if err != nil {
--- a/pkg/cmd/create/create_chat_token.go
+++ b/pkg/cmd/create/create_chat_token.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
+	"github.com/jenkins-x/jx/pkg/kube/naming"
 
 	"github.com/jenkins-x/jx/pkg/auth"
 	"github.com/jenkins-x/jx/pkg/chats"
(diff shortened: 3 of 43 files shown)


pkg/util/names.go
```diff
@@ -1,4 +1,4 @@
-package kube
+package util
```
hferentschik: OOI, what is the motivation to move package? 

On that note, there is a k8s.go in the util package. It contains an `EncodeKubernetesName` which does not even seem to be used. What's about moving all the functions in this file into there (potentially removing `EncodeKubernetesName`) on the go?

names.go is very generic and when just looking at file names in a package. I somehow find `k8s.go` gives me a better idea on what to expect, leaving potentially the door open for none name related utility functions.

abayer: Circular dependency hell, basically. That said, I'll move it into `k8s.go`.

abayer:
Fair - will fix.

pkg/util/names.go
hferentschik: I see.

abayer:
@pmuir - done (and resolved a merge conflict when I noticed it)

hferentschik:
> Moving this out of the kube package and into util is not right as util is bloated enough.

Same as kube really. I think ideally it would stay in `kube` and we find another way to resolve the circular dependency.

> If you can't solve the circular dependencies then suggest you move it a new package like kubeutil nested under kube.

I'd argue `util` is in this case better package. Mainly also in the hope that more people come across it. My thinking is that people are more likely to "explore" `util` at some state in order to find useful functions than any other package. Also I struggle with `kube` vs `kubeutil`. Last but not least, there is k8s.go which, in the case of a `kubeutil` package, should be moved there.

> It would also be sensible to do this as two commits, one showing the package change and one showing the actual functional changes you are making

+1

hferentschik:
@abayer, ooi do you still have the compile error of the circular dependency?

abayer:
@hferentschik:

```
import cycle not allowed
package main
	imports github.com/jenkins-x/jx/cmd/jx/app
	imports github.com/jenkins-x/jx/pkg/cmd
	imports github.com/jenkins-x/jx/pkg/cmd/add
	imports github.com/jenkins-x/jx/pkg/apps
	imports github.com/jenkins-x/jx/pkg/environments
	imports github.com/jenkins-x/jx/pkg/gits
	imports github.com/jenkins-x/jx/pkg/auth
	imports github.com/jenkins-x/jx/pkg/vault
	imports github.com/jenkins-x/jx/pkg/config
	imports github.com/jenkins-x/jx/pkg/jenkinsfile
	imports github.com/jenkins-x/jx/pkg/kube
	imports github.com/jenkins-x/jx/pkg/jenkinsfile
```

abayer:
Specifically it's that `ModifyNamespace` in `kube/env.go` does stuff with `PipelineConfig`.

hferentschik:
> Specifically it's that ModifyNamespace in kube/env.go does stuff with PipelineConfig.

Maybe breaking this cycle would be better. Creating an _kube/env_ package and moving env.go in there might work. _env_cache.go_ could move as well in this case together with the relevant tests. That's what I would try, but I leave it up to you guys.

abayer:
@pmuir - your call - if you'd prefer I break the cycle, it looks like `env.go` (and related tests) could be pulled into their own package under `kube` without breaking anything.

abayer:
More specifically, `env.go`, `env_modify_namespace_integration_test.go`, `env_test.go`, `env_cache.go`, `envrolebindings.go`, and `envrolebindings_test.go` can all move - `envvar.go` can't, though.

pmuir:
I agree, the best bet is to break the cycle. @hferentschik understood re naming/packages, otoh there is sooo much in util, it's very hard to know what it does!

If we can use this as an opportunity to clean up `kube`, then I'm definitely for it - that's the next big problem after `cmd`.

abayer:
Alrightie, change incoming momentarily!

abayer:
...ok, not momentarily, because I think there are two more circular deps. Lemme unravel those.

abayer:
ha, and the things I need to move to break the circular dep themselves depend on stuff from various other things in `kube`...most notably the stuff in `names.go`, but a bunch of other files as well. @pmuir, @hferentschik - for now, I'm just going to move `names.go` into `pkg/kube/naming` and leave it at that, if that's ok. I'll open another issue for reorganizing `pkg/kube` as a whole.

pmuir:
Sounds good to me.

abayer:
Opened https://github.com/jenkins-x/jx/issues/4474

abayer:
whoops, missed `pipeline.go` in the move for some reason. Sigh. Fixed.

abayer:
...no, my IDE didn't actually fix things for me on the rename. Sigh. Fixing everything else now.

abayer:
Yay, it's good! Whenever you're ready, @pmuir or @hferentschik

hferentschik:
I like `naming` better than `kubeutils`. Thanks for opening  #4474

/approve

abayer:
Still looking for an lgtm... =)

abayer:
/hold

PR #4436, opened 2019-06-27
WIP: feat: make the loading of real clients fail in the unit tests

garethjevans:
This is a POC to see if we can enforce the failing of loading clients that
may cause side effects in a unit test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,7 @@ make-reports-dir:
 	mkdir -p $(REPORTS_DIR)
 
 test: make-reports-dir ## Run the unit tests
-	CGO_ENABLED=$(CGO_ENABLED) $(GOTEST) -count=1 $(COVERFLAGS) -failfast -short ./...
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -183,7 +183,7 @@ make-reports-dir:
 	mkdir -p $(REPORTS_DIR)
 
 test: make-reports-dir ## Run the unit tests
-	CGO_ENABLED=$(CGO_ENABLED) $(GOTEST) -count=1 $(COVERFLAGS) -failfast -short ./...
+	UNIT_TEST=true CGO_ENABLED=$(CGO_ENABLED) $(GOTEST) -count=1 $(COVERFLAGS) -failfast -short ./...
```
pmuir: This target isn't run during CI, you need to change this target https://github.com/jenkins-x/jx/blob/master/Makefile#L222

garethjevans:
This isn't going to work unless we can separate out the test contexts

PR #4438, opened 2019-06-27
feat: unit tests are run as part of a separate context

garethjevans:


pmuir:
This isn't needed. All go tests are run as part of `/test integration`

hferentschik:
If want to do this properly, we either need to create a new build tag 'unit' and mark all unit tests with them or we use something like !integration. We need to sets of tags and each test needs to have one. 

Shall we create an issue for this? I think it is worth it being able to properly differentiate between the different test types. 

For now I suggest to close this PR. @garethjevans, wdyt?

PR #4457, opened 2019-06-27
fix: lets default to jenkins x pipelines if there's a missing importmode

### LINKED ISSUE #4456: jx import should default to using `jenkins-x.yml` if using prow/tekton
it looks like things go wonky if folks lose the `dev Environments.spec.teamSetting` is missing `prowEngine: Tekton`

jstrachan:
if we lose the import mode for some reason on the teamSettings, lets default to jenkins x pipelines if we prow/tekton

fixes #4456


abayer:
/lgtm

What's caused this recent wave of the import mode vanishing in the first place, btw?

PR #4459, opened 2019-06-27
fix: Try uploading the release note with the release

cagiti:
Updating the release notes doesn't seem to work as expected, so the approach is to generate the release notes and then upload with the release artefacts.


cagiti:
All tests passed, it wouldn't merge automatically

PR #4466, opened 2019-06-27
fix(tekton): Activity and get build logs should support metapipeline

### LINKED ISSUE #4362: metapipeline: PipelineActivity and jx get build logs need to treat generator and execution PipelineRuns as one pipeline

abayer:
#### Submitter checklist

#### Description

So let's look for all pipelines that match an org/repo/branch/build number, which will give us both the metapipeline and the generated pipeline.

Tests are still needed, just want to see how this, y'know, *builds*.

/assign @hferentschik 

fixes #4362

### DIFF
--- a/pkg/cmd/controller/controller_build.go
+++ b/pkg/cmd/controller/controller_build.go
@@ -648,47 +648,53 @@ func (o *ControllerBuildOptions) updatePipelineActivityForRun(kubeClient kuberne
 	if allCompleted {
 		if failed {
 			spec.Status = v1.ActivityStatusTypeFailed
+		} else if pri.Type == tekton.MetaPipeline {
+			spec.Status = v1.ActivityStatusTypePending
 		} else {
 			spec.Status = v1.ActivityStatusTypeSucceeded
 		}
-		if !biggestFinishedAt.IsZero() {
-			spec.CompletedTimestamp = &biggestFinishedAt
-		}
 
-		// log that the build completed
-		logJobCompletedState(activity)
-
-		// lets ensure we overwrite any canonical jenkins build URL thats generated automatically
-		if spec.BuildLogsURL == "" {
-			podInterface := kubeClient.CoreV1().Pods(ns)
-
-			envName := kube.LabelValueDevEnvironment
-			devEnv := o.EnvironmentCache.Item(envName)
-			location := v1.StorageLocation{}
-			settings := &devEnv.Spec.TeamSettings
-			if devEnv == nil {
-				log.Logger().Warnf("No Environment %s found", envName)
-			} else {
-				location = settings.StorageLocationOrDefault(kube.ClassificationLogs)
+		if pri.Type != tekton.MetaPipeline {
+			if !biggestFinishedAt.IsZero() {
+				spec.CompletedTimestamp = &biggestFinishedAt
 			}
-			if location.IsEmpty() {
-				location.GitURL = activity.Spec.GitURL
-				if location.GitURL == "" {
-					log.Logger().Warnf("No GitURL on PipelineActivity %s", activity.Name)
+
+			// log that the build completed
+			logJobCompletedState(activity)
+
+			// TODO: This will need to be reworked for per-step logs, so leaving alone as part of metapipeline work
+			// lets ensure we overwrite any canonical jenkins build URL thats generated automatically
+			if spec.BuildLogsURL == "" {
+				podInterface := kubeClient.CoreV1().Pods(ns)
+
+				envName := kube.LabelValueDevEnvironment
+				devEnv := o.EnvironmentCache.Item(envName)
+				location := v1.StorageLocation{}
+				settings := &devEnv.Spec.TeamSettings
+				if devEnv == nil {
+					log.Logger().Warnf("No Environment %s found", envName)
+				} else {
+					location = settings.StorageLocationOrDefault(kube.ClassificationLogs)
 				}
-			}
+				if location.IsEmpty() {
+					location.GitURL = activity.Spec.GitURL
+					if location.GitURL == "" {
+						log.Logger().Warnf("No GitURL on PipelineActivity %s", activity.Name)
+					}
+				}
 
-			masker, err := kube.NewLogMasker(kubeClient, ns)
-			if err != nil {
-				log.Logger().Warnf("Failed to create LogMasker in namespace %s: %s", ns, err.Error())
-			}
+				masker, err := kube.NewLogMasker(kubeClient, ns)
+				if err != nil {
+					log.Logger().Warnf("Failed to create LogMasker in namespace %s: %s", ns, err.Error())
+				}
 
-			logURL, err := o.generateBuildLogURL(podInterface, ns, activity, pri.PipelineRun, pod, location, settings, o.InitGitCredentials, masker)
-			if err != nil {
-				log.Logger().Warnf("%s", err)
-			}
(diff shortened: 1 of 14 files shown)


abayer:
well, that Tekton failure is legit. =)

abayer:
yaaaay Tekton BDD passed. Good sign. =)

abayer:
/hold

Putting this back on hold - it's not working _quite_ right yet.

abayer:
/hold cancel

Ok, I think we're all good.

abayer:
Prow being picky about trailing spaces/newlines after commands is...annoying.

hferentschik:
Nice commit history 👍. I like :-) BTW, you know you don't have to sign-off your commits (adding your email) if you are GPG signing (option `-S`) your commits. Only works for members of jenkins-x. Just only found out recently myself.

pkg/cmd/controller/controller_build.go
```diff
@@ -648,47 +648,53 @@ func (o *ControllerBuildOptions) updatePipelineActivityForRun(kubeClient kuberne
 	if allCompleted {
 		if failed {
 			spec.Status = v1.ActivityStatusTypeFailed
+		} else if pri.Type == tekton.MetaPipeline {
```
hferentschik: I am not so familiar with this controller yet, but by looking at the code I am a bit confused by seeing something like `if allCompleted` and then  `ActivityStatusTypePending` if the type is tekton.MetaPipeline. Not quite sure how to do it differently atm though.

hferentschik: would it make more sense to do the switch between the type once at the beginning of the function and handle the case of meta pipeline or execution pipeline?

pkg/cmd/get/get_build_logs.go
```diff
@@ -233,7 +233,8 @@ func (o *GetBuildLogsOptions) getLastJenkinsBuild(name string, buildNumber int)
 	}
 }
 
-func (o *GetBuildLogsOptions) getProwBuildLog(kubeClient kubernetes.Interface, tektonClient tektonclient.Interface, jxClient versioned.Interface, ns string, tektonEnabled bool) error {
+// GetProwBuildLog prompts the user, if needed, to choose a pipeline, and then prints out that pipeline's logs.
+func (o *GetBuildLogsOptions) GetProwBuildLog(kubeClient kubernetes.Interface, tektonClient tektonclient.Interface, jxClient versioned.Interface, ns string, tektonEnabled bool) error {
```
hferentschik: So how does this change pan out in the end? `jx get build log` lists meta pipeline build as different items for which the logs can be retrieved?

hferentschik: How can I see how this looks like in the end?

pkg/cmd/get/get_build_logs.go
```diff
@@ -322,90 +322,98 @@ func (o *GetBuildLogsOptions) getProwBuildLog(kubeClient kubernetes.Interface, t
...
-			}
-			if stage.Pod == nil {
-				// The stage's pod hasn't been created yet, so let's wait a bit.
-				f := func() error {
-					selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: map[string]string{
-						pipeline.GroupName + pipeline.PipelineRunLabelKey: pri.PipelineRun,
-					}})
-					if err != nil {
-						return err
+		var pipelines []*tekton.PipelineRunInfo
```
hferentschik: If it were me I would for sure extract some functions here. This is ways too nested for my taste.

pkg/cmd/get/get_build_logs.go
```diff
@@ -505,27 +521,26 @@ func (o *GetBuildLogsOptions) loadBuilds(kubeClient kubernetes.Interface, ns str
...
 			defaultName = name
 		}
 	}
-	return names, defaultName, buildMap, pipelineMap, nil
+	return names, defaultName, buildMap, nil
 }
 
-func (o *GetBuildLogsOptions) loadPipelines(kubeClient kubernetes.Interface, tektonClient tektonclient.Interface, jxClient versioned.Interface, ns string) ([]string, string, map[string]builds.BaseBuildInfo, map[string]builds.BaseBuildInfo, error) {
+// LoadPipelines loads all available pipelines as PipelineRunInfos.
+func (o *GetBuildLogsOptions) LoadPipelines(kubeClient kubernetes.Interface, tektonClient tektonclient.Interface, jxClient versioned.Interface, ns string) ([]string, string, map[string][]builds.BaseBuildInfo, error) {
```
hferentschik: Why do you need to export this function? Is it just for testing? If so, then I think this is an anti-pattern. Then I rather put the test into the same package.

pkg/cmd/get/get_build_logs_test.go
```diff
@@ -0,0 +1,69 @@
+package get_test
```
hferentschik: I'd use package `get`. I think it is fully legit to test none exported functions.

pkg/cmd/controller/controller_build.go
abayer: `allCompleted` refers to the current executing `PipelineRun`, not the `PipelineActivity` as a whole. We want to treat the meta and execution `PipelineRun`s both as part of the same `PipelineActivity`, so if the meta pipeline didn't fail, we're expecting there to be an execution pipeline to come that we'll add to the `PipelineActivity`.

pkg/cmd/get/get_build_logs_test.go
abayer: Will do!

pkg/cmd/get/get_build_logs.go
abayer: `jx get build log` will treat the meta pipeline and execution pipeline as a single build, with the logs for all steps in both of them included in the proper order.

pkg/cmd/controller/controller_build.go
hferentschik: Sure. So it is more a "allStageCompleted"

abayer: Yeah, `allStagesCompleted` feels right. I'll rename it.

pkg/cmd/controller/controller_build.go
```diff
@@ -606,7 +606,7 @@ func (o *ControllerBuildOptions) updatePipelineActivityForRun(kubeClient kuberne
 	spec := &activity.Spec
 	var biggestFinishedAt metav1.Time
 
-	allCompleted := true
+	allStagesCompleted := true
```
hferentschik: :-)

pkg/cmd/get/get_build_logs_test.go
```diff
@@ -0,0 +1,68 @@
+package get
```
hferentschik: 👍

PR #4468, opened 2019-06-27
fix(gke): AddLabel should sanitize key and value, not =.

### LINKED ISSUE #4467: BDD cluster creation on GKE creates invalid `--labels` parameter for `create cluster`
`create.AddLabel` is currently sanitizing on the full label - i.e., `name=value`. That results in `name-value`, which isn't a valid label, obviously. We should be sanitizing `name` and `value`, but not the `=`.

abayer:
#### Submitter checklist

#### Description

We were sanitizing all of "name=value", resulting in a label of `name-value`. We should be sanitizing `name` and `value`, but not the full string.

/assign @hferentschik 
/assign @pmuir 

fixes #4467

### DIFF
--- a/pkg/cmd/create/create_cluster_gke.go
+++ b/pkg/cmd/create/create_cluster_gke.go
@@ -578,7 +578,7 @@ func AddLabel(labels string, name string, value string) string {
 		if labels != "" {
 			sep = ","
 		}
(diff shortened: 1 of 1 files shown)


abayer:
Closing in favor of #4462

PR #4301, opened 2019-06-17
chore(tekton): Add Tekton BDD test context to PRs

abayer:
This is waiting for @daveconde to nail down cluster cleanup before it's ready to merge.

/hold

Issue #4314, opened 2019-06-18
Switch step_create_task's interpret mode to step_syntax_effective optionally generating a shell script


abayer:
The new interpret mode for `step_create_task` allows for executing the steps in a pipeline locally - but the better way to do this would be to have `step_syntax_effective` optionally generate a shell script containing those steps, keeping `step_create_task` simpler.

hferentschik:
I think with `jx boot` there to stay, this becomes really important. How the interpret mode is currently implemented is not ideal. It is somehow mixed into the current code. There should be a clear separation of concerns.

PR #4479, opened 2019-06-27
fix(tekton): Clone before getting GitInfo in step create task

### LINKED ISSUE #4478: step create task tries to determine GitInfo from non existent directory
In the current version of _step_create_task_, the GitInfo struct is determined by calling `FindGitInfo(o.CloneDir)`. However, this call is made prior to `CloneDir` being set.

A better option is to get the GitInfo from the info from the passed source URL.

abayer:
#### Submitter checklist

#### Description

Metapipeline fun - if we don't clone first, we end up with no repo, which unsurprisingly is bad when we try to get the `GitInfo`. If the repo's already cloned (as will be the case when we switch over to `jx
step create pipeline`), we don't need to clone here, but for now...

I've tested this by hand after reproducing the original issue by hand, and it works, so yay. This sort of thing will get caught automatically once #4301 lands.

/assign @dwnusbaum 
/assign @hferentschik 

fixes #4478

### DIFF
--- a/pkg/cmd/step/create/step_create_task.go
+++ b/pkg/cmd/step/create/step_create_task.go
@@ -204,6 +204,33 @@ func (o *StepCreateTaskOptions) Run() error {
 		}
 	}
 
+	exists, err := o.effectiveProjectConfigExists()
+	if err != nil {
+		return err
+	}
+	if !exists {
+		// TODO this branch all things depending on it can be removed once the meta pipeline is working
+		// TODO keeping this to keep existing behavior until then (HF)
+		if o.CloneGitURL != "" {
+			o.CloneDir = o.cloneGitRepositoryToTempDir(o.CloneGitURL, o.Branch, o.PullRequestNumber, o.Revision)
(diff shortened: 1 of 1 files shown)


dwnusbaum:
Ah ok nvm, didn't read the description fully, looks like https://github.com/jenkins-x/jx/pull/4301 would test this behavior.

/lgtm

hferentschik:
Ahh, thanks. I was surprised the issue was already closed this morning. I had initially tried to make the clone URL obligatory and align more with step_create_metapipeline. That worked fine and it solved the use case of being called from the pipeline runner. However, I then realized that this probably won't work with the interpreted mode. 

I saw issue #4314, and I agree that it would be much better if for the interpreted mode we not reuse the step_create_task command. I'd rather have the boot code create an (effective) pipeline directly and then add some capability to the pipeline to execute locally or generate a script.

abayer:
Yeah, we should probably pivot to dealing with #4314 sooner than later.

PR #4485, opened 2019-06-28
fix: lets allow schedulers to be used when importing projects

### LINKED ISSUE #4483: lets add a scheduler mode on TeamSettings so that import uses Schedulers
if folks go all in on schedulers first to configure projects - rather than Prow `ConfigMaps: config+plugins` first - then we need a way for `jx import / jx create quickstart` to create the `SourceRepository` and regenerate the Prow `ConfigMaps` on import.

We should also support adding `--scheduler cheese` when importing if folks wanna use a particular flavour of scheduler

jstrachan:
rather than manually hacking the Prow ConfigMaps lets generate them from
    the Scheuler CRDs; using the specific scheduler on a SourceRepository or
    use the default one on your Team

    fixes #4483


abayer:
/hold

PR #4487, opened 2019-06-28
fix: Environment prefix defaulting to generated name

macox:

#### Submitter checklist

#### Description
Fix for environment prefix defaulting to a generated name, rather than using the cluster name


abayer:
/hold

Issue #4510, opened 2019-07-01
Semantic Pull Requests probot does not check all commits 


hferentschik:
The enabled probot does not verify each commit against conventional commit style, in fact in default mode it is even enough to have a pull request title in conventional commit format, but none of the commits itself.

IMO each commit should be checked.

hferentschik:
See config options of the bot: https://github.com/probot/semantic-pull-requests#configuration

hferentschik:
Fixed by PR #4511

PR #4511, opened 2019-07-01
fix: configuring semantic commit bot to check all commits

hferentschik:
fixes issue #4510


pmuir:
Ok, I removed the file I put in the wrong place, and changed it to commits and PR title as I had originally intended to set it.

PR #4513, opened 2019-07-01
chore: increase parallelism after refactor

pmuir:
#### Submitter checklist

fixes #


pmuir:
Still failing @4 so I think we can safely say we aren't there yet!

PR #4525, opened 2019-07-01
fix(tekton): Look for run by labels, not name

### LINKED ISSUE #4523: Next build number lookup uses PipelineRun name, which may be truncated
https://github.com/jenkins-x/jx/blob/eea23e76f73a39fa75d12788e0df3d259d25980e/pkg/tekton/pipelines.go#L151 - this works great if the `org-repo-branch-context` is short enough, but `PipelineResourceName` truncates that if it's more than 31 characters (thanks to limitations on Kubernetes resource name length). So a long enough base name results in non-unique `PipelineRun` etc names, which itself results in build number incrementing being shared across branches, etc. This is not ideal. We should look up `PipelineRun`s by labels instead.

abayer:
#### Submitter checklist

#### Description

`PipelineRun` names may be truncated to the point where they're not unique across branches and contexts, so we should be looking them up for determining the next build number by looking at their labels instead.

/assign @hferentschik 
/assign @jstrachan 
/assign @dwnusbaum 

fixes #4523

### DIFF
--- a/pkg/cmd/step/create/step_create_meta_pipeline.go
+++ b/pkg/cmd/step/create/step_create_meta_pipeline.go
@@ -142,7 +142,7 @@ func (o *StepCreatePipelineOptions) Run() error {
 	}
 
 	pipelineName := tekton.PipelineResourceNameFromGitInfo(gitInfo, o.Branch, o.Context, tekton.MetaPipeline)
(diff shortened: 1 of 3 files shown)


abayer:
@dgozalo - yup, in https://github.com/jenkins-x/jx/pull/4416

dgozalo:
> @dgozalo - yup, in #4416

Interesting, I'm petty sure I had the latest version before creating the cluster, will try again.

PR #4528, opened 2019-07-01
fix(tekton): Branch and revision need to be used in metapipeline

abayer:
#### Submitter checklist

#### Description

Specifically, we need to pass the revision to clone to the metapipeline itself or it'll just always checkout master. Which isn't ideal. And we need to pass the branch onward when generating the real pipeline or it also just ends up building master.

/assign @hferentschik 

n/a

### DIFF
--- a/pkg/tekton/metapipeline/metapipeline.go
+++ b/pkg/tekton/metapipeline/metapipeline.go
@@ -10,6 +10,7 @@ import (
 	"github.com/jenkins-x/jx/pkg/client/clientset/versioned"
 	"github.com/jenkins-x/jx/pkg/gits"
 	"github.com/jenkins-x/jx/pkg/kube"
+	"github.com/jenkins-x/jx/pkg/prow"
 	"github.com/jenkins-x/jx/pkg/tekton"
 	"github.com/jenkins-x/jx/pkg/tekton/syntax"
 	"github.com/jenkins-x/jx/pkg/util"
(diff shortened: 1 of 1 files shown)


pkg/tekton/metapipeline/metapipeline.go
```diff
@@ -72,7 +73,18 @@ func CreateMetaPipelineCRDs(params CRDCreationParameters) (*tekton.CRDWrapper, e
 		return nil, err
 	}
 
-	resources := []*pipelineapi.PipelineResource{tekton.GenerateSourceRepoResource(params.PipelineName, params.GitInfo, "")}
+	revision := params.Branch
+	if params.PullRef != "" {
```
hferentschik: I think this might be a bit too simple. Let me think about it for a bit. One thing we could do is to add a Revision option to `StepCreatePipelineOptions` and pass it through via `CRDCreationParameters`, but I'd like to spend a bit more time on this.

hferentschik: Unrelated, but I think here is another issue - https://github.com/jenkins-x/jx/blob/fe66ed51204667e37ed067f65a7ebc67f7c75bbb/pkg/cmd/step/create/step_create_task.go#L248

we need to move the parsing of the pull refs higher up so that it is initialised in case cloning is needed.

abayer:
/hold

Putting on hold to make sure this doesn't get inadvertently merged.

hferentschik:
@abayer, I think we can close this PR ow in favour of #4569

abayer:
Closing in favor of #4569.

PR #4529, opened 2019-07-01
chore: set default values for external-dns during jx boot

rawlingsj:
Signed-off-by: rawlingsj <rawlingsj80@gmail.com>


pkg/cmd/step/create/step_create_install_values.go
```diff
@@ -142,6 +142,38 @@ func (o *StepCreateInstallValuesOptions) defaultMissingValues(values map[string]
...
+		o.Provider = requirements.Provider
+		if o.Provider == "" {
+			log.Logger().Warnf("No provider configured\n")
+		}
+	}
+
+	subDomain := "." + ns + "."
+
+	// if using GKE with a custom domain let's default to using external-dns and TLS
+	if o.Provider == cloud.GKE && !strings.Contains(o.Domain, "nip.io") {
```
rawlingsj: @jstrachan this will default to using external-dns and certmanager if a domain is provided and on GKE.  I'm thinking we should probably have an override flag or better UX but I could do with a hand once it's merged and been used to see what works best.

PR #4546, opened 2019-07-02
fix(install): use YAML import mode for Tekton

polothy:
#### Submitter checklist

- [ ] Change is covered by existing or new tests. **(Don't know)**

#### Description

Was testing out a local [kind](https://kind.sigs.k8s.io) cluster to play with Tekton and friends. My install command was:

```
jx install --provider=kubernetes --no-default-environments --default-admin-password=admin --advanced-mode --on-premise --external-ip 'MY_IP' --gitops=false --prow=true --tekton=true --static-jenkins=false --vault=false
```

Then, when I tried to do a quickstart, it applied a normal `Jenkinsfile` instead of `jenkins-x.yaml`.  Turned out the import mode on the dev environment team settings was set to `Jenkinsfile` instead of `YAML`.

100% guessing on where to fix it. But did test the fix, team settings were correct after the change.

N/A

### DIFF
--- a/pkg/cmd/create/install.go
+++ b/pkg/cmd/create/install.go
@@ -1622,6 +1622,7 @@ func (options *InstallOptions) configureProwInTeamSettings() error {
 			settings.ProwEngine = v1.ProwEngineTypeKnativeBuild
 			if options.Flags.Tekton {
 				settings.ProwEngine = v1.ProwEngineTypeTekton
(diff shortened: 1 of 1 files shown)


abayer:
/hold
Putting on hold so #4552 can land first.

polothy:
Let me know if I need to rebase

PR #4561, opened 2019-07-03
feat: brew package update logic

cagiti:
Associated with #4596 
This change should allow the update of brew packaged components to be updated using their version and sha256.


pkg/cmd/step/create/pr/step_create_pr_brew.go
```diff
@@ -0,0 +1,93 @@
...
+	cmd.Flags().StringVarP(&options.Sha, "sha", "", "", "The sha of the brew archive to update")
+	return cmd
+}
+
+// Run implements this command
+func (o *StepCreatePullRequestBrewOptions) Run() error {
+	if err := o.ValidateOptions(); err != nil {
+		return errors.WithStack(err)
+	}
+	if o.Version == "" {
```
pmuir: We should probably do this in the `o.ValidateOptions()`

pkg/cmd/step/create/pr/step_create_pr_brew.go
```diff
@@ -0,0 +1,93 @@
...
+		return util.MissingOption("version")
+	}
+	if o.Sha == "" {
+		return util.MissingOption("sha")
+	}
+	if o.SrcGitURL == "" {
```
pmuir: We should probably do this in the `o.ValidateOptions()`

pkg/brew/brew.go
```diff
@@ -0,0 +1,116 @@
...
+	//									 $1			$2	$3		   $4
+	shaRegex = regexp.MustCompile(`^(.*sha256) (\")([0-9a-z]+)(\")$`)
+)
+
+//UpdateVersion scans the directory structure rooted in dir for files that match brewNameRegex and replaces any
+// lines starting with FROM <name>:, ENV <name> or ARG=<name> with the newVersion
+func UpdateVersion(dir string, newVersion string) ([]string, error) {
+	//shaPrefix := fmt.Sprintf("url %s:", name)
+	oldVersions := make(map[string]bool)
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
```
pmuir: It would have probably been easier to use a `filepath.Glob` here to find the files, as you are just searching by extension.

pkg/brew/brew.go
```diff
@@ -0,0 +1,116 @@
...
+	for oldVersion := range oldVersions {
+		answer = append(answer, oldVersion)
+	}
+	sort.Strings(answer)
+	return answer, nil
+}
+
+//UpdateSha scans the directory structure rooted in dir for files that match brewNameRegex and replaces any
+// lines starting with sha with the newVersion
+func UpdateSha(dir string, newSha string) ([]string, error) {
```
pmuir: It would probably be better to do both the version and the sha at the same time to avoid having to walk twice - but it's not a big deal.

pkg/brew/brew.go
```diff
@@ -0,0 +1,116 @@
...
+			if brewNameRegex.MatchString(filepath.Base(path)) {
+				bytes, err := ioutil.ReadFile(path)
+				if err != nil {
+					return errors.Wrapf(err, "reading %s", path)
+				}
+				brewFile := string(bytes)
+				answer := make([]string, 0)
+				for _, line := range strings.Split(brewFile, "\n") {
+					found := false
+					if shaRegex.MatchString(line) {
```
pmuir: It would be a more efficient to do submatches here as then you only need to execute the regex once. I recently added a utility to do a replace submatches with a function which is basically what you need I think https://github.com/jenkins-x/jx/blob/master/pkg/cmd/step/create/pr/step_create_pr_regex.go#L113

PR #4562, opened 2019-07-03
chore: minor logging improvements to git helper

garethjevans:
Removed verbose flag, using Debugf instead


abayer:
/hold
Putting on hold so #4552 can land first.

PR #4565, opened 2019-07-03
chore: Cut down on resource requests for BDD tests

### LINKED ISSUE #4574: codecov.io/bash call in covered jx binary sleeps 30 seconds after failed upload or query
When the covered `jx` binary executes, it runs https://codecov.io/bash to upload the coverage data. That bash script sleeps for 30 seconds if it fails to query Codecov, or if it fails to upload to Codecov. Digging into executing Tekton BDD tests, I'm seeing the bash call take >2 minutes when run within the pipeline created by the BDD test, with multiple `sleep 30`s showing up. I saw it with the upload, but if there are more than 3 failures for upload, or more than 3 failures for query, it just skips the upload or query, so this makes me think that the Tekton BDD tests, at the very least, are completely failing to upload.

abayer:
#### Submitter checklist

#### Description

3 CPUs and 3gb RAM was overkill for the request, IMO. Let's drop the CPU to 1 and the RAM to 2gb.

Adding the `tekton` context has resulted in a bunch of `PENDING` pods due to requests not being satisfied, hence this.

/assign @daveconde 
/assign @pmuir 

fixes #4574

### DIFF
--- a/jenkins-x-bdd.yml
+++ b/jenkins-x-bdd.yml
@@ -36,8 +36,8 @@ pipelineConfig:
                     cpu: 4
                     memory: 6144Mi
                   requests:
-                    cpu: 3
-                    memory: 3072Mi
+                    cpu: 1
+                    memory: 2048Mi
                 env:
                   - name: CODECOV_TOKEN
                     valueFrom:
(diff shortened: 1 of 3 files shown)


abayer:
/hold

Holding for #4552 to merge.

abayer:
We should get confirmation that this did the trick when the CI finishes - there should then be references to `#tektone2e` in the codecov report.

Issue #4551, opened 2019-07-03
Meta pipeline needs to merge all pull refs as first step in the pipeline


hferentschik:
Currently, the meta pipeline code does only rely on Tekton to check out the source. Tekton, however, only checks out a single revision. With Prow we can have multiple commits getting merged into a branch (batch build). 

To support this we need to use the `jx` merge step to merge all shas specified in a pull ref.

hferentschik:
fixed by PR #4569

PR #4569, opened 2019-07-03
fix(tekton): deriving build related meta information from pull refs

hferentschik:
- removing parameters which can be derived from step_create_meta_pipeline
- adding step to merge pull refs in meta pipeline
- making sure the right parameters are passed in step create task in meta pipeline

fixes issue #4551

### DIFF
--- a/pkg/cmd/step/create/step_create_meta_pipeline.go
+++ b/pkg/cmd/step/create/step_create_meta_pipeline.go
@@ -17,7 +17,6 @@ import (
 	"github.com/jenkins-x/jx/pkg/util"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-	"strings"
 	"time"
 
 	jxclient "github.com/jenkins-x/jx/pkg/client/clientset/versioned"
@@ -28,7 +27,6 @@ import (
 )
 
 const (
-	branchOptionName         = "branch"
 	contextOptionName        = "context"
 	defaultImageOptionName   = "default-image"
 	envOptionName            = "env"
@@ -36,9 +34,7 @@ const (
 	labelOptionName          = "label"
 	noApplyOptionName        = "no-apply"
 	outputOptionName         = "output"
-	pipelineKindOptionName   = "kind"
 	pullRefOptionName        = "pull-refs"
-	pullRequestOptionName    = "pr-number"
 	serviceAccountOptionName = "service-account"
 	sourceURLOptionName      = "source-url"
 
@@ -61,14 +57,10 @@ var (
 type StepCreatePipelineOptions struct {
 	*opts.CommonOptions
 
-	SourceURL    string
-	Branch       string
-	PullRefs     string
-	PipelineKind string
-	Job          string
-
-	PullRequestNumber string
-	Context           string
+	SourceURL string
+	PullRefs  string
+	Context   string
+	Job       string
 
 	CustomLabels []string
 	CustomEnvs   []string
@@ -100,8 +92,6 @@ func NewCmdCreateMetaPipeline(commonOpts *opts.CommonOptions) *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&options.SourceURL, sourceURLOptionName, "", "Specify the git URL for the source (required)")
-	cmd.Flags().StringVar(&options.Branch, branchOptionName, "", "The git branch to trigger the build in. Defaults to the current local branch name (required)")
-	cmd.Flags().StringVarP(&options.PipelineKind, pipelineKindOptionName, "k", "", "The kind of pipeline to create such as: "+strings.Join(jenkinsfile.PipelineKinds, ", "))
 	cmd.Flags().StringVar(&options.Job, jobOptionName, "", "The Prow job name in order to identify all pipelines belonging to a single trigger event (required)")
 
 	cmd.Flags().StringVar(&options.PullRefs, pullRefOptionName, "", "The Prow pull ref specifying the references to merge into the source")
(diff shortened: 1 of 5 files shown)


pkg/cmd/step/create/step_create_meta_pipeline.go
```diff
@@ -61,14 +57,10 @@ var (
 type StepCreatePipelineOptions struct {
 	*opts.CommonOptions
 
-	SourceURL    string
-	Branch       string
-	PullRefs     string
-	PipelineKind string
```
abayer: I'm not sure we can get rid of `PipelineKind`, `Context`, etc as options - but that may just be me thinking about things from the perspective of the `step_create_task` tests.

pkg/cmd/step/create/step_create_meta_pipeline.go
```diff
@@ -291,11 +267,46 @@ func (o *StepCreatePipelineOptions) getClientsAndNamespace() (tektonclient.Inter
...
-	var pr *prow.PullRefs
-	var err error
-	if pullRefs != "" {
-		pr, err = prow.ParsePullRefs(pullRefs)
+// determineBranchIdentifier finds a identifier for the branch which is build by the specified pull ref.
+// The name is either an actual git branch name (eg master) or a synthetic names  like PR-<number> for single pull requests
+// or 'batch' for Prow batch builds.
+// The method makes the decision purely based on the Prow pull ref. At this stage we don't have the full ProwJov spec
+// available.
+func (o *StepCreatePipelineOptions) determineBranchIdentifier(pullRef prow.PullRefs) string {
```
abayer: So all of this locks in to only kicking off builds via prow - that's not going to work. `jx start pipeline` is valid, e.g.

pkg/cmd/step/create/step_create_meta_pipeline.go
```diff
@@ -291,11 +267,46 @@ func (o *StepCreatePipelineOptions) getClientsAndNamespace() (tektonclient.Inter
...
+}
+
+func (o *StepCreatePipelineOptions) determinePipelineKind(pullRef prow.PullRefs) string {
+	var kind string
+
+	prCount := len(pullRef.ToMerge)
+	if prCount > 0 {
+		kind = jenkinsfile.PipelineKindPullRequest
+	} else {
+		kind = jenkinsfile.PipelineKindRelease
```
abayer: How about `feature` pipelines? Not used very often, but it is a supported option.

pkg/cmd/step/create/step_create_meta_pipeline.go
hferentschik: `Context` is still there, that can indeed not be derived. PipelineKind I derive from the pull ref atm, similar to what the pipeline runner does as well. If need be, this option could be brought back. It seems to me though that right now we only have release or pullrequest. Having two options you can also create inconsitent runs, by for example specifying a pull ref for a pull request, but use kind 'release'. That does not make much sense to me.

pkg/cmd/step/create/step_create_meta_pipeline.go
hferentschik: `jx start pipeline` does create a Prow job if prow is enabled and goes into a different branch if not - https://github.com/jenkins-x/jx/blob/65962ff5ef1a6d1c4776daee0163434c9c2cb566/pkg/cmd/start/start_pipeline.go#L156

I think we are fine there.

pkg/cmd/step/create/step_create_meta_pipeline.go
hferentschik: I meant to bring this up. I have it on my list of open questions. I am not quite sure what `feature` pipelines mean. That said, I don't see how our current code ever allows anything else than `release` or `pull request` atm - https://github.com/jenkins-x/jx/blob/fe66ed51204667e37ed067f65a7ebc67f7c75bbb/pkg/cmd/controller/controller_pipelinerunner.go#L205

Not saying this is correct, I also feel I would like to have at least one other option. I am not fully behind the idea of each merge to master creating a release. I'd like a "master" build occurring when the merge to master happens and then somehow trigger a release build based on some other condition. But right now, it seems we have these two options.

hferentschik: If we want we can bring the explicit pipeline kind option back and pass it through, but from how we currently integrate with Prow it does not really change anything.

pkg/cmd/step/create/step_create_meta_pipeline.go
hferentschik: BTW, this happens in similar form in pipeline runner - https://github.com/jenkins-x/jx/blob/fe66ed51204667e37ed067f65a7ebc67f7c75bbb/pkg/cmd/controller/controller_pipelinerunner.go#L322

hferentschik:
@abayer, I added some tests and removed the wip status. 

Now it is about addressing any outstanding issues/concerns. 

BTW, something like this should work for you as well to test this locally:
```
./build/jx --verbose step create pipeline --source-url https://github.com/hf-bee/test-go-project.git --job d7cf84b6-9cf8-11e9-8f83-46ab9c3d7824 --default-image hferentschik/jx:dc2b13425 --pull-refs master:79c7e4d,42:89866
```

pkg/cmd/step/create/step_create_meta_pipeline.go
abayer: Ah, I misread and just saw the line deleting `Context`, not the new line.

Maybe we should force `pullRequest` as the `Kind` if `PullRefs` is not empty, even if you specify a `Kind`? That would allow `feature` to still work. Or we can just ignore it for now and deal with it later if it comes up. =)

pkg/cmd/step/create/step_create_meta_pipeline.go
abayer: Ok, then we should be fine there.

pkg/cmd/step/create/step_create_meta_pipeline.go
abayer: Yeah, the more I think about it, the more I'm leaning towards ignoring `feature` for now. We can revisit it later if we need to.

abayer:
/lgtm

Let's do it!

hferentschik:
>Let's do it!

+1,  we can adjust later on. But this feels like a cleaner start.

PR #4578, opened 2019-07-04
fix: jx create quickstart "--org" and "--name" options doesn't work

### LINKED ISSUE #4549: jx create quickstart "--org" and "--name" options doesn't work
### Summary 
when run "jx create quickstart" with "--org" and "--name" options, still request to choose "? Which organisation do you want to use?" and request input "? Enter the new repository name"
run command line below:
 "jx create quickstart --git-username='sonyali03' --org='sonyali03' --name='demo070207'", 
### Expected behavior
don't need manually select organisation and input repository name because these already assigned in command line
### Actual behavior
the information below displayed:
"
Using Git provider GitHub at https://github.com
? Which organisation do you want to use? sonyali03
? Enter the new repository name: 
"
### Jx version

```
NAME               VERSION
jx                 2.0.372
jenkins x platform 2.0.330
Kubernetes cluster v1.13.2
kubectl            v1.14.0
helm client        Client: v2.11.0+g2e55dbe
git                git version 2.20.1 (Apple Git-117)
Operating System   Mac OS X 10.14.5 build 18F132
```

### Jenkins type

 Classic Jenkins

sonyafenge:
#### Submitter checklist

#### Description
Run command line below and found --org and --name parameters doesn't work.
"jx create quickstart --git-username='sonyafenge' --org='sonyafenge' --name='demo070207'"

Check the file "import.go" and found "--org" and "--name" stored in struct "ImportOptions", not in struct "GitGitRepositoryOptions", so config GitGitRepositoryOptions.Owner and GitGitRepositoryOptions.RepoName before func "GetGitRepositoryDetails" call func "PickNewOrExistingGitRepository".

Two rows code changed in function "GetGitRepositoryDetails":
	options.GitRepositoryOptions.Owner = options.Organisation
	options.GitRepositoryOptions.RepoName = options.Repository

This fix is to reduce the impact to minimum. another way is to fix "AddGitRepoOptionsArguments" in "git.go".

fixes #4549

### DIFF
--- a/pkg/cmd/importcmd/import.go
+++ b/pkg/cmd/importcmd/import.go
@@ -1446,6 +1446,9 @@ func (options *ImportOptions) GetGitRepositoryDetails() (*gits.CreateRepoData, e
 	if err != nil {
 		return nil, err
 	}
+	//config git repositoryoptions parameters: Owner and RepoName
+	options.GitRepositoryOptions.Owner = options.Organisation
+	options.GitRepositoryOptions.RepoName = options.Repository
 	details, err := gits.PickNewOrExistingGitRepository(options.BatchMode, authConfigSvc,
 		"", &options.GitRepositoryOptions, nil, nil, options.Git(), false, options.In, options.Out, options.Err)
 	if err != nil {


pkg/cmd/importcmd/import.go
```diff
@@ -1446,6 +1446,9 @@ func (options *ImportOptions) GetGitRepositoryDetails() (*gits.CreateRepoData, e
 	if err != nil {
 		return nil, err
 	}
+	//config git repositoryoptions parameters: Owner and RepoName
+	options.GitRepositoryOptions.Owner = options.Organisation
```
hferentschik: I am wondering whether this not has to happen much earlier? Basically in `Run`. If you look where and how `GetGitRepositoryDetails` is called, I wonder what happens when you specify `--batch` as well:

```go
	if !o.BatchMode {
		details, err = o.GetGitRepositoryDetails()
		if err != nil {
			return err
		}

		o.Filter.ProjectName = details.RepoName
	}
```

sonyafenge: thanks for your information, I doubled checked BatchMode and this change is not necessary for Batchmode because:
1. Batchmode use " --project-name" options ("o.Filter.ProjectName") as project/repo Name, only when batchmode equals false, the option "name" is used. you can also see this from: 

		o.Filter.ProjectName = details.RepoName
2. Batchmode also use "--org" options, but this already config at "ImportCreatedProject" in import.go, but organization popup happened before this config, so the issue only impact when batchmode equals false.
         Row #384 in Run() and Row #614 in func CreateNewRemoteRepository()

		o.Filter.ProjectName = details.RepoName
          
		options.GitRepositoryOptions.Owner = options.GetOrganisation()
3. Changing in function GetGitRepositoryDetails() also because there are total 3 files call this function, so this will fix same issue for create mlquickstart and create spring.

PR #4592, opened 2019-07-04
feat: Added an abstraction for logs and enabled long term storage in the CLI

### LINKED ISSUE #4413: Create an abstraction for a logs service so the UI and CLI can share it
### Summary 

There needs to be a way to make the UI / CLI share the code to fetch logs from different sources (pods / buckets).

This funcionality is currently working for the UI but not for `jx get build logs`.

The end goal is to have both the command the the UI show the same information when accessed regardless of build state.

The performance on `jx get build logs` should also be improved.

Lastly, the logs secrets masker should be applied to the logs on the CLI as well.

dgozalo:
Added an abstraction for logs and enabled long term storage in the CLI

feature: added changes to filtering

feature: fixed the lookup of pipeline activities

feature: renamed the logging file and fix after rebase

feature: finished refactor and added tests

feature: added tests

feature: added batch capabilities

chore: lint and fmt

#### Submitter checklist

#### Description
Added an abstraction layer on the logging so both the CLI and the UI can reuse the same logic.

This is done through the interface `LogWriter` which can be implemented by both to define how to 
deal with logs in different mediums.

The CLI's implementation of `LogWriter` is `CLILogWriter` and streams logs for running builds using `TailLogs`which calls Kubectl and single writes are done through the logger.

Also enabled the long term storage logs capabilities in the CLI. The logic has changed so it looks for `PipelineActivities` and their corresponding `PipelineRuns`. This was needed because we only used to let the user to see the logs of currently running builds instead of past builds which may have logs stored in buckets.

fixes #4413

### DIFF
--- a/pkg/builds/build_info.go
+++ b/pkg/builds/build_info.go
@@ -7,7 +7,7 @@ import (
 	"strings"
 	"time"
 
-	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	"github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/gits"
 	"github.com/jenkins-x/jx/pkg/kube"
 	"github.com/jenkins-x/jx/pkg/log"
--- a/pkg/cmd/get/get_build_logs.go
+++ b/pkg/cmd/get/get_build_logs.go
@@ -8,14 +8,15 @@ import (
 	"strings"
 	"time"
 
-	gojenkins "github.com/jenkins-x/golang-jenkins"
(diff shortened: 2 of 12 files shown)


hferentschik:
👍 for the intentions behind this.

abayer:
How does this handle multi-stage pipelines? That is, a single `PipelineRun` with multiple pods?

dgozalo:
> How does this handle multi-stage pipelines? That is, a single `PipelineRun` with multiple pods?

By iterating through the running building pods and validating that the labels match, you can show all the logs needed, ordered by creationTimestamp. this way you get the stages in order.

abayer:
@dgozalo - cool, just making sure - I assumed it was doing the right thing, but figured it was quicker to ask than to try to figure out the call paths exactly. =)

PR #4594, opened 2019-07-05
fix: git server url is hardcoded in pipelinerunner #4588

### LINKED ISSUE #4588: pipelinerunner doesnot support github domain other than "github.com"
### Summary 
My enterprise github is like "github.xxx.xxx.com".
Current pipelinerunner only supports "github.com" which is hard-coded.
[controller_pipelinerunner](https://github.com/jenkins-x/jx/blob/master/pkg/cmd/controller/pipelinerunner.go#L282)

### Expected behavior

By introducing new cmd option `--git-provider-url=xx.xxx.com`,  github domain can be set.
And the clone operation is successful.

### Actual behavior
When the app is created by `jx create  --git-provider-url=xx.xxx.com`, 
pipelinerrunner pod will exit and be restarted due to checkout error for more than 3 times.

### Jenkins type

- [+ ] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

rokii:
#### Submitter checklist

fixes #4588

### DIFF
--- a/pkg/cmd/controller/pipelinerunner.go
+++ b/pkg/cmd/controller/pipelinerunner.go
@@ -56,6 +56,7 @@ type PipelineRunnerOptions struct {
 	UseMetaPipeline      bool
 	MetaPipelineImage    string
 	SemanticRelease      bool
+	GitServerURL         string
 }
 
 // PipelineRunRequest the request to trigger a pipeline run
@@ -112,6 +113,7 @@ func NewCmdControllerPipelineRunner(commonOpts *opts.CommonOptions) *cobra.Comma
 	cmd.Flags().StringVarP(&options.Path, "path", "p", "/", "The path to listen on for requests to trigger a pipeline run.")
 	cmd.Flags().StringVarP(&options.ServiceAccount, "service-account", "", "tekton-bot", "The Kubernetes ServiceAccount to use to run the pipeline.")
 	cmd.Flags().BoolVarP(&options.NoGitCredentialsInit, "no-git-init", "", false, "Disables checking we have setup git credentials on startup.")
+	cmd.Flags().StringVarP(&options.GitServerURL, "git-provider-url", "", "https://github.com", "The default Git server URL. Currently only Github is supported.")
 	cmd.Flags().BoolVarP(&options.SemanticRelease, "semantic-release", "", false, "Enable semantic releases")
 
 	// TODO - temporary flags until meta pipeline is the default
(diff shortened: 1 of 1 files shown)


pkg/cmd/controller/pipelinerunner.go
```diff
@@ -112,6 +113,7 @@ func NewCmdControllerPipelineRunner(commonOpts *opts.CommonOptions) *cobra.Comma
 	cmd.Flags().StringVarP(&options.Path, "path", "p", "/", "The path to listen on for requests to trigger a pipeline run.")
 	cmd.Flags().StringVarP(&options.ServiceAccount, "service-account", "", "tekton-bot", "The Kubernetes ServiceAccount to use to run the pipeline.")
 	cmd.Flags().BoolVarP(&options.NoGitCredentialsInit, "no-git-init", "", false, "Disables checking we have setup git credentials on startup.")
+	cmd.Flags().StringVarP(&options.GitServerURL, "git-provider-url", "", "https://github.com", "The default Git server URL. Currently only Github is supported.")
```
hferentschik: I am a bit sceptical towards adding another flag? How do you use it then? Did you manually change the DeploymentConfig for the pipelinerunner? So you would need to do this on each install. If we go down this route, one probably would like to adjust the pipelinerunner Helm chart, passing in the git provider as specified during install. See https://github.com/jenkins-x-charts/prow/blob/master/prow/values.yaml#L47

That said, I think a better approach is to look up the provider URL. I think it is fair to assume that for each build request sent to the pipelinerunner, there should exist a SourceRepository CRD (`kubectl get kubectl get sourcerepositories`). If there is none, we can fail the request or default to 'github.com' for now. The [SourceRepository](https://github.com/jenkins-x/jx/blob/ee747c794323064ba0be029dcd6573a80cc7aef4/pkg/apis/jenkins.io/v1/types_extensions.go#L503) contains the provider information we are after.

I think the best would be to do a `sourceRepositoryInterface.List` passing a `ListOption` with a label selector for _owner_ and _repository_.

This is how a `SourceRepository` looks like:
```yaml
apiVersion: jenkins.io/v1
kind: SourceRepository
metadata:
  annotations:
    jenkins.io/last-build-number-for-pr-1: "1"
  creationTimestamp: "2019-06-27T16:24:59Z"
  generation: 1
  labels:
    owner: jenkins-x
    provider: git-github-com-jenkins-x-jx-git
    repository: jx
  name: jenkins-x-jx
  namespace: jx
  resourceVersion: "2702090"
  selfLink: /apis/jenkins.io/v1/namespaces/jx/sourcerepositories/jenkins-x-jx
  uid: 1b925fd1-98f8-11e9-a3f5-42010a840054
spec:
  description: Imported application for jenkins-x/jx
  org: jenkins-x
  provider: git@github.com:jenkins-x/jx.git
  providerName: github
  repo: jx
  scheduler:
    kind: ""
    name: ""
```

Does this make sense? Do you feel you want to continue working on this? That would be aweomse

hferentschik:
@abayer ^^ fyi, this one is interesting. Relates to our current efforts around meta pipeline as well

pkg/cmd/controller/pipelinerunner.go
rokii: Agree. Let me refine it.

hferentschik: Awesome, thanks. I close this PR in the meantime.

PR #4598, opened 2019-07-05
feat: use the docker image that we just built as step image in pipelines

fiunchinho:
#### Submitter checklist

#### Description
When running e2e tests, Instead of copying the jx binary with the PR changes into an old jx docker image, use the docker image build and pushed by the pipeline.

I've also replaced the references to Docker Hub registry with Google Cloud Registry.

### DIFF
--- a/jenkins-x-bdd.yml
+++ b/jenkins-x-bdd.yml
@@ -62,26 +62,26 @@ pipelineConfig:
 
               - name: build-and-push-image
                 command: /kaniko/executor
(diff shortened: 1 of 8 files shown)


abayer:
Error pushing:
```
error pushing image: failed to push to destination gcr.io/jenkinsxio/jx:covered-0.0.0-SNAPSHOT-PR-4598-4: no token in bearer response:
{"errors":[{"code":"DENIED","message":"Token exchange failed for project 'jenkinsxio'. Caller does not have permission 'storage.buckets.get'. To configure permissions, follow instructions at: https://cloud.google.com/container-registry/docs/access-control"}]}
```

fiunchinho:
Something like that @abayer ? I've splitted it into two stages.

abayer:
@fiunchinho Exactly! You can put the `environment` and `options.containerOptions` on the `pipeline` rather than on the individual stages, btw, to avoid duplication.

Anyway, now the builds are still failing, but they're not hanging, so that's a positive. =)

pmuir:
/lgtm

Very nice

Issue #4599, opened 2019-07-05
pipelinerunner needs to pass the correct prow job name to step create pipeline


hferentschik:
At the moment pipelinerunner uses the `Job` field of the `ProwJobSpec`, but that's not the unique Prow job id. 

```go
	createPipelineOption := &create.StepCreatePipelineOptions{}
	c := *o.CommonOptions
	createPipelineOption.CommonOptions = &c
	createPipelineOption.SourceURL = sourceURL
	createPipelineOption.PullRefs = pullRefs.String()
	createPipelineOption.Context = prowJobSpec.Context
	createPipelineOption.Job = prowJobSpec.Job

	createPipelineOption.ServiceAccount = o.ServiceAccount
	createPipelineOption.DefaultImage = o.MetaPipelineImage
```

In the pipelinerunner case the value to use is passed in `PipelineRunRequest.Labels` as `prowJobName`.

hferentschik:
fixed by pull request #4602

PR #4602, opened 2019-07-05
fix(tekton): making sure the right prow job id is passed to meta pipl…

hferentschik:
…ine creation

fixes issue #4599


abayer:
/test tekton

Think that was flakiness

Issue #4604, opened 2019-07-05
stepMergePullRefs in metapipeline.go needs to handle an empty base sha


hferentschik:
```go
func stepMergePullRefs(pullRefs prow.PullRefs) syntax.Step {
	args := []string{"--baseBranch", pullRefs.BaseBranch, "--baseSHA", pullRefs.BaseSha}
	for _, mergeSha := range pullRefs.ToMerge {
		args = append(args, "--sha", mergeSha)
	}

	step := syntax.Step{
		Name:      mergePullRefsStepName,
		Comment:   "Pipeline step merging pull refs",
		Command:   "jx step git merge",
		Arguments: args,
	}
	return step
}
```

The baseSha can be empty in which case `--baseSHA` should not be specifed.

hferentschik:
Currently when the `baseSha` is empty, the build-step-merge-pull-refs step will error with:

```
Error: flag needs an argument: --baseSHA
```

hferentschik:
fixed by pr #4608

PR #4608, opened 2019-07-05
fix(tekton): skiping git merge step if there are no pull refs to merge

hferentschik:
fixes issue #4604


hferentschik:
that was a quick merge

PR #4613, opened 2019-07-05
fix: Revert use of viper on step_create_task and step_create_meta_pipeline

abayer:
#### Submitter checklist

#### Description

For some reason I can't figure out, the viper addition here resulted in `--no-apply` and `--output=whatever` just not registering _at all_. So let's get rid of viper here until we can figure out what's happening.

/assign @hferentschik 

n/a

### DIFF
--- a/pkg/cmd/controller/pipelinerunner.go
+++ b/pkg/cmd/controller/pipelinerunner.go
@@ -59,6 +59,8 @@ type PipelineRunnerOptions struct {
 	UseMetaPipeline      bool
 	MetaPipelineImage    string
 	SemanticRelease      bool
+	noApply              bool
+	outDir               string
 }
 
 // PipelineRunRequest the request to trigger a pipeline run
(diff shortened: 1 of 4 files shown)


hferentschik:
> For some reason I can't figure out, the viper addition here resulted in --no-apply and --output=whatever just not registering at all. So let's get rid of viper here until we can figure out what's happening.

So you are saying that if you specify the options via the CLI they don't apply? I'll take a look. I'd rather fix the issue here. Being able to control this via env variables helps when integrating with this code.

abayer:
@hferentschik - exactly. It seems to be some interaction with viper usage elsewhere in jx, but I'm not sure where. I'm all for getting it working, but it may make sense to merge this for now.

pkg/cmd/controller/pipelinerunner.go
```diff
@@ -59,6 +59,8 @@ type PipelineRunnerOptions struct {
 	UseMetaPipeline      bool
 	MetaPipelineImage    string
 	SemanticRelease      bool
+	noApply              bool
```
hferentschik: For the pipelinerunner these options really don't make much sense. They are relevant for the transitive command which is called under the hood. Hence, the env variable approach is better.

hferentschik:
> exactly. It seems to be some interaction with viper usage elsewhere in jx, but I'm not sure where. I'm all for getting it working, but it may make sense to merge this for now.

I actually know where there rub is. Let me try something ...

hferentschik:
I suggest something like this - https://github.com/jenkins-x/jx/pull/4628. This way we are moving forward instead of backwards.

hferentschik:
Superseded by #4628.

PR #4616, opened 2019-07-06
feat: added pre-commit configuration and secrets basline

cagiti:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+repos:
+- repo: git@github.com:Yelp/detect-secrets
+  rev: v0.12.4
+  hooks:
+    - id: detect-secrets
+      args: ['--baseline', '.secrets.baseline']
+      exclude: .*/tests/.*
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,0 +1,803 @@
+{
+  "exclude": {
+    "files": "^pkg/jenkins/test_data/update_center.json.*$|^.secrets.baseline$|^.*test.*$",
+    "lines": null
+  },
+  "generated_at": "2019-07-06T09:33:07Z",
+  "plugins_used": [
+    {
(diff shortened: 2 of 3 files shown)


rawlingsj:
This is awesome @cagiti 

BTW what happens if someone doesn’t install the pre-commit hooks?  I'm just wondering if there's any checks we could but in place or do we expect if folks want the protection then they should install the pre-commit hooks?  I guess if this works nicely then we could roll out to all repos?  I'm happy to take that on, maybe a few of us can blitz it.

cagiti:
> This is awesome @cagiti
> 
> BTW what happens if someone doesn’t install the pre-commit hooks? I'm just wondering if there's any checks we could but in place or do we expect if folks want the protection then they should install the pre-commit hooks? I guess if this works nicely then we could roll out to all repos? I'm happy to take that on, maybe a few of us can blitz it.

Hey @rawlingsj, unfortunately the users would have to install the hook into their git repo using `pre-commit install` - they would also need pre-commit installed locally... however this is pre-installed into the [dev-env](https://github.com/jenkins-x/dev-env). The hook is installed into the repository local `.git/hooks/` directory

cagiti:
> > This is awesome @cagiti
> > BTW what happens if someone doesn’t install the pre-commit hooks? I'm just wondering if there's any checks we could but in place or do we expect if folks want the protection then they should install the pre-commit hooks? I guess if this works nicely then we could roll out to all repos? I'm happy to take that on, maybe a few of us can blitz it.
> 
> Hey @rawlingsj, unfortunately the users would have to install the hook into their git repo using `pre-commit install` - they would also need pre-commit installed locally... however this is pre-installed into the [dev-env](https://github.com/jenkins-x/dev-env). The hook is installed into the local `.git/hooks/` directory within their copy of the repository.

We could create/update a `jx step` command to clone and install the git hooks?

PR #4619, opened 2019-07-06
fix: Allow post preview jobs without ingress

### LINKED ISSUE #4591: jx post preview jobs presuppose that services are exposed externally
### Summary 
In preview.go, post preview jobs are set up so that, for example, a penetration testing framework like owasp-zap can be launched as a parallel container in the same environment and used to test the newly instantiated preview pod.

These jobs are passed a URL as an environment variable so that the job is aware of the published address of the service under test.

The URL that is being passed is that of the externally exposed Ingress and is the empty string if the service is not exposed.

I would assert that this presupposition is only true for the trivial case of default quickstarts and a minority of edge services. In production environments, it is to be expected that the majority of services should not be visible outside their own namespaces and thus will not have exposed external URLs.

The result of this assumption is that any service that has both a disabled ingress and a post preview job configured is likely to fail to build in preview since the post preview job will receive an empty string as a URL. This is certainly true in the case of owasp-zap jobs.

### Expected behavior
My expectation would be that preview services are tested locally to their namespace using their service name and not their external address. Having external addresses for preview environments feels like a security or intellectual property risk in general.

An argument could be made for passing an external URL if one exists or the local service URL if not. 

Generically, it would probably make most sense to pass three environment variables to post preview jobs:

1. The local service name
2. The external URL
3. A variable that equates to (2) if it exists, otherwise (1)

This would allow authors of post preview jobs maximal flexibility.

### Jx version

```
NAME               VERSION
jx                 2.0.372
Kubernetes cluster v1.12.8-gke.10
kubectl            v1.14.2
git                git version 2.21.0
Operating System   Mac OS X 10.13.6 build 17G6030
```

tdcox:

#### Submitter checklist

#### Description
This PR changes the behaviour of the setup for post preview jobs. The environment variable `JX_PREVIEW_URL` will now be set to the URL of the external ingress if one is configured, otherwise it will be set to the URL of the internal service local to the namespace.

Construction of the internal URL prefers https over http and common ports over named exotic ports.

Additionally, some new environment variables are created to permit more flexiblity for composing post preview jobs.

I am unable to test RunPostPreviewSteps in a build environment so I would be grateful if someone could smoke test this build or send me a copy of the executable so I can run manual tests?

This fix addresses problems with the owasp-zap addon. Consideration needs to be given to any other existing post preview jobs that I may not have visibility of.

fixes #4591

### DIFF
--- a/pkg/cmd/preview/preview.go
+++ b/pkg/cmd/preview/preview.go
@@ -25,7 +25,7 @@ import (
 
 	"github.com/jenkins-x/jx/pkg/kube/services"
 
-	"github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/cmd/opts"
 	"github.com/jenkins-x/jx/pkg/cmd/templates"
 	"github.com/jenkins-x/jx/pkg/config"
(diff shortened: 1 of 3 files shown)


tdcox:
@abayer Could you take a look at this one for me as well, please?

abayer:
Looks reasonable to me, but I'd appreciate someone who knows this area of the code better reviewing. =)

tdcox:
@rawlingsj Please could you also take a look when you have a moment?

Issue #4626, opened 2019-07-08
--no-apply and --out-dir don't work when commands are called programmatically 


hferentschik:
_step_create_task_ and _step_create_meta_pipeline_ are using now Viper to retrieve the _no-apply_ and _out-dir_ options. As a side effect, one cannot set these options programmatically on the underlying options structs.

hferentschik:
The problem is in this code:

```go
	if viper.GetBool(noApplyOptionName) {
		err := tektonCRDs.WriteToDisk(viper.GetString(outputOptionName), pipelineActivity)
		if err != nil {
			return errors.Wrapf(err, "failed to output Tekton CRDs")
		}
	} else {
		err := tekton.ApplyPipeline(jxClient, tektonClient, tektonCRDs, ns, gitInfo, branch, pipelineActivity)
		if err != nil {
			return errors.Wrapf(err, "failed to apply Tekton CRDs")
		}
		if o.Verbose {
			log.Logger().Infof("applied tekton CRDs for %s", tektonCRDs.PipelineRun().Name)
		}
	}
```

`viper.GetBool(noApplyOptionName)` requires that the value comes from the environment or that there are pflags set. However, if you programmatically populate the `StepCreatePipelineOptions`, this is not the case and you will always get `false`.

hferentschik:
merged via pull request #4628

PR #4628, opened 2019-07-08
fix(tekton): making sure no-apply and out-dir can also be set program…

hferentschik:
…matically

fixes issue #4626


hferentschik:
@abayer rebased, can you add another lgtm?

PR #4629, opened 2019-07-08
fix: git server url is hardcoded in pipelinerunner #4588

### LINKED ISSUE #4588: pipelinerunner doesnot support github domain other than "github.com"

rokii:
#### Submitter checklist

fixes #4588

### DIFF
--- a/pkg/cmd/controller/pipelinerunner.go
+++ b/pkg/cmd/controller/pipelinerunner.go
@@ -4,18 +4,20 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/jenkins-x/jx/pkg/prow"
-	"github.com/jenkins-x/jx/pkg/util"
-	"github.com/sirupsen/logrus"
 	"io/ioutil"
 	"net/http"
 	"net/http/httputil"
 	"os"
 	"os/signal"
 	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
+
+	"github.com/jenkins-x/jx/pkg/prow"
+	"github.com/jenkins-x/jx/pkg/util"
+	"github.com/sirupsen/logrus"
 
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
 	"github.com/jenkins-x/jx/pkg/cmd/step/create"
@@ -24,6 +26,7 @@ import (
 	"github.com/jenkins-x/jx/pkg/jenkinsfile"
 	"github.com/jenkins-x/jx/pkg/log"
 
+	jxclient "github.com/jenkins-x/jx/pkg/client/clientset/versioned"
 	"github.com/jenkins-x/jx/pkg/kube"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/controller/pipelinerunner.go
```diff
@@ -279,7 +283,10 @@ func (o *PipelineRunnerOptions) startPipeline(pipelineRun PipelineRunRequest) (P
 		return response, errors.Wrap(err, "failed to get env vars from prowjob")
 	}
 
-	sourceURL := fmt.Sprintf("https://github.com/%s/%s.git", prowJobSpec.Refs.Org, prowJobSpec.Refs.Repo)
+	sourceURL := o.getSourceURL(prowJobSpec.Refs.Org, prowJobSpec.Refs.Repo)
+	if sourceURL == "" {
```
hferentschik: +1 for falling back to github

pkg/cmd/controller/pipelinerunner.go
```diff
@@ -457,3 +464,42 @@ func (o *PipelineRunnerOptions) setupSignalChannel(cancel context.CancelFunc) {
 		cancel()
 	}()
 }
+
+func (o *PipelineRunnerOptions) getSourceURL(org, repo string) string {
+	jxClient, ns, err := o.getClientsAndNamespace()
```
hferentschik: You are missing an err check here. _jxClient_ might be nil here.

pkg/cmd/controller/pipelinerunner.go
```diff
@@ -457,3 +464,42 @@ func (o *PipelineRunnerOptions) setupSignalChannel(cancel context.CancelFunc) {
...
+
+	resourceInterface := jxClient.JenkinsV1().SourceRepositories(ns)
```
hferentschik: given that all you need _jxClient_ in this function is for getting the source repository interface, you might as well push that into `getClientsAndNamespace` and call it `getSourceRepositoryInterface`. Up to you really.

rokii:
I can not access the integration test dashboard. 
```
Failed to authenticate: github: user "rokii" not in required orgs or teams
```

pkg/cmd/controller/pipelinerunner.go
rokii: done

pkg/cmd/controller/pipelinerunner.go
rokii: `getClientsAndNamespace` is much general that may be used for some other case later.  So I wanna keep it.

hferentschik:
/lgtm

Thanks @rokii

PR #4633, opened 2019-07-08
fix(tekton): Handle multiple pipeline runs for a single "pipeline" in logs

abayer:
#### Submitter checklist

#### Description

We only get the list of pipelines to iterate over when we invoke `jx get build logs`. That doesn't work well in the metapipeline case - we want `jx get build logs` to keep going after the metapipeline finishes and pick up the logs for the generated pipeline as well. So when one pipeline finishes, let's refresh the pipeline listing and iterate.

/assign @dgozalo 
/assign @hferentschik 

n/a


abayer:
/hold

Waiting for #4592 so I stop stepping on @dgozalo =)

abayer:
No longer works after #4592 merged, working on a new version.

Issue #4615, opened 2019-07-05
"gcr.io/jenkinsxio/builder-go" as agent image always resolves to `latest` tag, but used to resolve otherwise


abayer:
until https://github.com/jenkins-x/jx/pull/4135, having `gcr.io/jenkinsxio/builder-go` as your `agent.container` in a build pack would (as of right now, early evening EDT on July 5) result in the image `gcr.io/jenkinsxio/builder-go:0.1.549` being used in steps. But since that change, `gcr.io/jenkinsxio/builder-go` (no tag) is used instead (meaning the `latest` tag is used in practice). Now, I'm not 100% sure we should "fix" this - if you want to not use a tag on a pod template image like this one, just use `go` as the image. But at least two build packs ([charts](https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes/blob/master/packs/charts/pipeline.yaml#L3) and [environment](https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes/blob/master/packs/environment/pipeline.yaml#L6)) use this approach, so it may be that others out there are as well.

I'm leaning towards just fixing the build packs and leaving this as is, but am open to persuasion.

abayer:
Yeah, this is an actual bug - it was removing https://github.com/jenkins-x/jx/pull/4135/files#diff-f19bbfe849f4cde14e8119e425507f62L844 that messed things up.

PR #4634, opened 2019-07-08
fix(tekton): Use pod template name for default image

abayer:
#### Submitter checklist

#### Description

For a while now, #4615 has resulted in `gcr.io/jenkinsxio/builder-maven` (and similarly image URLs without tags) always using `latest`, rather than the latest in the version stream. The desired result, however, is to use the pod template image, so let's just be cleaner and use the pod template name here. That'll always resolve correctly in practice.

/assign @hferentschik 

n/a

### DIFF
--- a/pkg/tekton/syntax/constants.go
+++ b/pkg/tekton/syntax/constants.go
@@ -14,5 +14,5 @@ const (
 	KanikoDockerImage = "gcr.io/kaniko-project/executor:9912ccbf8d22bbafbf971124600fbb0b13b9cbd6"
 
 	// DefaultContainerImage - the default image used for pipelines if none is specified.
-	DefaultContainerImage = "gcr.io/jenkinsxio/builder-maven"
+	DefaultContainerImage = "maven"
 )


pkg/tekton/syntax/constants.go
```diff
@@ -14,5 +14,5 @@ const (
 	KanikoDockerImage = "gcr.io/kaniko-project/executor:9912ccbf8d22bbafbf971124600fbb0b13b9cbd6"
 
 	// DefaultContainerImage - the default image used for pipelines if none is specified.
-	DefaultContainerImage = "gcr.io/jenkinsxio/builder-maven"
+	DefaultContainerImage = "maven"
```
hferentschik: I think in the next step we should switch from `maven` to `jx`. There is no conceivable reason why we would need a Java-based image.

jstrachan: the main reason historically is we use maven for updating the version numbers in maven project releases. For all other languages we can modify the code (e.g. package.json). I guess the meta pipeline can add the maven version in as a separate step?

PR #4610, opened 2019-07-05
WIP: Experimenting with prow changes for metapipeline. DO NOT MERGE.

abayer:
#### Submitter checklist

- [ ] Change is code complete and matches issue description.
- [ ] Change is covered by existing or new tests.

#### Description

Just don't merge this. =)

#### Which issue this PR fixes

fixes #

PR #4636, opened 2019-07-08
fix(tekton): Fixes for metapipeline run from prow

### LINKED ISSUE #4361: metapipeline: prow needs to treat both the CRD generator and actual execution PipelineRuns as one job
When the metapipeline work that @hferentschik is doing (see #4320) is done, we're going to want to change prow to wait for both the metapipeline `PipelineRun` (which will create the Tekton CRDs for the actual execution) and the actual execution `PipelineRun` to complete. I'm not sure how we'll do that. At first glance, I wonder if we should be propagating the `prowJobName` label to both `PipelineRun`s and then change https://github.com/jenkins-x/test-infra/blob/pipelinerunner/prow/cmd/pipeline/controller.go#L342-L363 to handle multiple `PipelineRun`s, marking the `ProwJob` as successful if all the `PipelineRun`s are complete and successful, and marking it as failed if any of the `PipelineRun`s are complete and a failure?

@ccojocar, @wbrefvem, @hferentschik - thoughts?

abayer:
#### Submitter checklist

#### Description

Fixes for a number of issues uncovered in the process of getting #4610 to pass.
* Env vars can contain commas, so pass as separate --env
* Pass build number to create task in metapipeline
* Handle logs for runs started after get build logs is called
* Labels should be specified individually as well
* Remove unused combineEnvVars function

/assign @hferentschik 
/assign @dgozalo 

fixes #4361

### DIFF
--- a/pkg/cmd/step/create/step_create_task.go
+++ b/pkg/cmd/step/create/step_create_task.go
@@ -720,27 +720,6 @@ func (o *StepCreateTaskOptions) combineLabels(labels map[string]string) error {
 	return nil
 }
 
-func (o *StepCreateTaskOptions) combineEnvVars(projectConfig *jenkinsfile.PipelineConfig) error {
-	// add any custom env vars
-	envMap := make(map[string]corev1.EnvVar)
-	for _, e := range projectConfig.Env {
-		envMap[e.Name] = e
-	}
-	for _, customEnvVar := range o.CustomEnvs {
-		parts := strings.Split(customEnvVar, "=")
-		if len(parts) != 2 {
-			return errors.Errorf("expected 2 parts to env var but got %v", len(parts))
-		}
-		e := corev1.EnvVar{
-			Name:  parts[0],
-			Value: parts[1],
-		}
-		envMap[e.Name] = e
-	}
-	projectConfig.Env = syntax.EnvMapToSlice(envMap)
-	return nil
-}
-
 func (o *StepCreateTaskOptions) getWorkspaceDir() string {
 	return filepath.Join("/workspace", o.SourceName)
 }
--- a/pkg/cmd/step/create/step_create_task.go
+++ b/pkg/cmd/step/create/step_create_task.go
@@ -269,6 +269,11 @@ func (o *StepCreateTaskOptions) Run() error {
 
 	pipelineName := tekton.PipelineResourceNameFromGitInfo(o.GitInfo, o.Branch, o.Context, tekton.BuildPipeline, tektonClient, ns)
 
+	err = o.setBuildValues()
+	if err != nil {
+		return err
+	}
+
 	exists, err = o.effectiveProjectConfigExists()
 	if err != nil {
 		return err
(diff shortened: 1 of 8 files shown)


abayer:
#4634 also needs to be merged for #4610 to pass without all my hacks. =)

pkg/cmd/step/create/step_create_task.go
```diff
@@ -390,15 +395,17 @@ func (o *StepCreateTaskOptions) createEffectiveProjectConfigFromOptions(tektonCl
...
 
-	if viper.GetBool(noApplyOptionName) || o.DryRun || o.InterpretMode {
-		o.BuildNumber = "1"
-	} else {
-		log.Logger().Debugf("generating build number...")
-		o.BuildNumber, err = tekton.GenerateNextBuildNumber(tektonClient, jxClient, ns, o.GitInfo, o.Branch, o.Duration, o.Context)
-		if err != nil {
-			return nil, err
+	if o.BuildNumber == "" {
+		if viper.GetBool(noApplyOptionName) || o.DryRun || o.InterpretMode {
```
hferentschik: Ahh, I see #4628 has not merged yet :-(

pkg/logs/tekton_logging.go
```diff
@@ -94,43 +94,79 @@ func findLegacyPipelineRunBuildNumber(pipelineRun *v1alpha12.PipelineRun) string
 	return buildNumber
 }
 
-// GetRunningBuildLogs obtains the logs of the provided PipelineActivity and streams the running build pods' logs using the provided LogWriter
-func GetRunningBuildLogs(pa *v1.PipelineActivity, buildName string, kubeClient kubernetes.Interface, writer LogWriter) error {
-	pods, err := builds.GetBuildPods(kubeClient, pa.Namespace)
+func getPipelineRunNamesForActivity(pa *v1.PipelineActivity, tektonClient tektonclient.Interface) ([]string, error) {
```
hferentschik: :+1

pkg/tekton/metapipeline/metapipeline.go
```diff
@@ -216,14 +215,15 @@ func stepCreateTektonCRDs(params CRDCreationParameters) syntax.Step {
 	args = append(args, "--service-account", params.ServiceAccount)
 	args = append(args, "--source", params.SourceDir)
 	args = append(args, "--branch", params.BranchIdentifier)
+	args = append(args, "--build-number", params.BuildNumber)
 	if params.Context != "" {
 		args = append(args, "--context", params.Context)
 	}
-	if len(params.Labels) > 0 {
-		args = append(args, "--label", strings.Join(params.Labels, ","))
+	for _, l := range params.Labels {
```
hferentschik: :+1, eventually I am wondering whether this should not be "passed" via the environment directly. Not sure what the best approach is. Passing command-line arguments as we do here makes things more explicit, on the other hand using env variables seems more idiomatic and probably more flexible. 

I think it is important though that essential/required values are passed the same way regardless of what we want to use.

pkg/tekton/metapipeline/metapipeline.go
```diff
@@ -216,14 +215,15 @@ func stepCreateTektonCRDs(params CRDCreationParameters) syntax.Step {
...
+		args = append(args, "--label", l)
 	}
-	if len(params.EnvVars) > 0 {
-		args = append(args, "--env", strings.Join(params.EnvVars, ","))
+	for _, e := range params.EnvVars {
```
hferentschik: This is ways better. Good catch, using the flags multiple times is the better approach.

pkg/tekton/metapipeline/metapipeline.go
```diff
@@ -216,14 +215,15 @@ func stepCreateTektonCRDs(params CRDCreationParameters) syntax.Step {
 	args = append(args, "--service-account", params.ServiceAccount)
 	args = append(args, "--source", params.SourceDir)
 	args = append(args, "--branch", params.BranchIdentifier)
+	args = append(args, "--build-number", params.BuildNumber)
```
hferentschik: 👍

PR #4646, opened 2019-07-09
fix: removing override of validate options

cagiti:


pkg/cmd/step/create/pr/step_create_pr_brew.go
```diff
@@ -75,7 +78,7 @@ func (o *StepCreatePullRequestBrewOptions) ValidateOptions() error {
 
 // Run implements this command
 func (o *StepCreatePullRequestBrewOptions) Run() error {
-	if err := o.ValidateOptions(); err != nil {
```
pmuir: you still want to call `o.ValidateOptions()` as that will cause the shared validator to run

cagiti: Now calling from validateBrewOptions

Issue #4642, opened 2019-07-09
Align label names in pipeline related resources


hferentschik:
All pipeline resources (pipelineactivity, pipeline, pipelinestructure, taskrun, pipelinerun) contain a repository label, however, the name is not aligned. PipelineActivity uses _repository_ whereas the others use _repo_. The label names should be aligned.

PR #4651, opened 2019-07-10
fix(test): allow test files to be updated by running the tests (golden files)

### LINKED ISSUE #4640: Use golden file testing pattern in step_create_task_test
I started to use the [golden file](https://medium.com/soon-london/testing-with-golden-files-in-go-7fccc71c43d3) testing patterns withing the JaCoCo App and I find it very useful. Given the amount of test pipeline configs the tests in _step_create_task_test_ are using, it would be nice to have an easy way to update them in case a change comes in (cc @abayer).

hferentschik:
fixes #4640

### DIFF
--- a/pkg/cmd/step/e2e/step_e2e_gc_test.go
+++ b/pkg/cmd/step/e2e/step_e2e_gc_test.go
@@ -46,10 +46,10 @@ func TestShouldDeleteOlderThanDuration(t *testing.T) {
 	o := e2e.StepE2EGCOptions{}
 	o.Duration = 2
 	cluster := getCluster("168", "jenkins-gkebdd", "159")
-	cluster.ResourceLabels["create-time"] = time.Now().Format("Mon-Jan-2-2006-15-04-05")
+	cluster.ResourceLabels["create-time"] = time.Now().UTC().Format("Mon-Jan-2-2006-15-04-05")
 	assert.Equal(t, false, o.ShouldDeleteOlderThanDuration(cluster))
 	cluster2 := getCluster("170", "jenkins-gkebdd", "159")
-	cluster2.ResourceLabels["create-time"] = time.Now().Add(-3 * time.Hour).Format("Mon-Jan-2-2006-15-04-05")
+	cluster2.ResourceLabels["create-time"] = time.Now().UTC().Add(-3 * time.Hour).Format("Mon-Jan-2-2006-15-04-05")
 	assert.Equal(t, true, o.ShouldDeleteOlderThanDuration(cluster2))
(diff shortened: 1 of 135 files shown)


pkg/cmd/step/e2e/step_e2e_gc_test.go
```diff
@@ -46,10 +46,10 @@ func TestShouldDeleteOlderThanDuration(t *testing.T) {
 	o := e2e.StepE2EGCOptions{}
 	o.Duration = 2
 	cluster := getCluster("168", "jenkins-gkebdd", "159")
-	cluster.ResourceLabels["create-time"] = time.Now().Format("Mon-Jan-2-2006-15-04-05")
+	cluster.ResourceLabels["create-time"] = time.Now().UTC().Format("Mon-Jan-2-2006-15-04-05")
```
hferentschik: @daveconde `ShouldDeleteOlderThanDuration` in _step_e2e_gc_ does take the time in UTC time (https://github.com/jenkins-x/jx/blob/c28614230d4e12925df0a7a0377c0ff479b6d4fb/pkg/cmd/step/e2e/step_e2e_gc.go#L144), so the tests need to do the same. For me this test was failing since I am two hours ahead of UTC.

pkg/cmd/controller/pipelinerunner_integration_test.go
```diff
@@ -102,7 +103,7 @@ var _ = Describe("Pipeline Runner Integration", func() {
 			pipelineRunner := PipelineRunnerOptions{
 				CommonOptions:        &commonOpts,
 				Path:                 "/",
-				BindAddress:          "0.0.0.0",
+				BindAddress:          "127.0.0.1",
```
hferentschik: @abayer - I am changing this to only bind to the loopback interface in the hope that this fixes the flakiness of the test. I tested this in a separate pull request and triggered the integration test multiple times and they passed.

If we continue seeing that this test is flaky we can take the next step and remove it.

cc @pmuir ^^

hferentschik:
@abayer once we get this in, I'll create the PR for #4642 on top of it

pkg/cmd/controller/pipelinerunner_integration_test.go
abayer: +1

hferentschik: Ok,  I removed it after all. The funny thing is that the unit test does the same test, but there it works. The difference is that the unit tests make no calls within the cluster.

hferentschik:
@abayer can you lgtm again. Needed to rebase

PR #4664, opened 2019-07-10
chore(tekton): Update to Tekton v0.5.1

abayer:
#### Submitter checklist

#### Description

Testing Tekton v0.5.0 (waiting for v0.5.1 to release shortly).

n/a

### DIFF
--- a/pkg/builds/build_info.go
+++ b/pkg/builds/build_info.go
@@ -74,7 +74,7 @@ func CreateBuildPodInfo(pod *corev1.Pod) *BuildPodInfo {
 	containers, _, isInit := kube.GetContainersWithStatusAndIsInit(pod)
 
 	for _, container := range containers {
-		if strings.HasPrefix(container.Name, "build-step-git-source") {
+		if strings.HasPrefix(container.Name, "build-step-git-source") || strings.HasPrefix(container.Name, "step-git-source") {
 			_, args := kube.GetCommandAndArgs(&container, isInit)
 			for i := 0; i <= len(args)-2; i += 2 {
 				key := args[i]
--- a/pkg/logs/tekton_logging.go
+++ b/pkg/logs/tekton_logging.go
@@ -46,13 +46,20 @@ func GetTektonPipelinesWithActivePipelineActivity(jxClient versioned.Interface, 
 	}
 
 	// This is a temporary solution until we add the "context" label to PipelineActivities
-	labelsFilter = strings.Replace(labelsFilter, "repository=", "repo=", 1)
 	if context != "" {
 		labelsFilter = modifyFilterForPipelineRun(labelsFilter, context)
 	}
 	tektonPRs, _ := tektonClient.TektonV1alpha1().PipelineRuns(ns).List(metav1.ListOptions{
 		LabelSelector: labelsFilter,
 	})
+
+	// Handle the old "repo" tag as well for legacy purposes.
+	if len(tektonPRs.Items) == 0 {
+		labelsFilter = strings.Replace(labelsFilter, "repository=", "repo=", 1)
+		tektonPRs, _ = tektonClient.TektonV1alpha1().PipelineRuns(ns).List(metav1.ListOptions{
+			LabelSelector: labelsFilter,
+		})
+	}
 
 	sort.Slice(tektonPRs.Items, func(i, j int) bool {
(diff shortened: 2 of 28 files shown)


abayer:
This is gonna fail - we need the `config-defaults` configmap in the tekton chart.

abayer:
(I manually hacked in the config map for right now)

abayer:
eh, tekton still failed due to `jx get build logs` timing out. Will dig later.

hferentschik:
how is this upgrade going?

abayer:
Poorly. =) I’ll circle back to it once the release process has restabilized.

abayer:
Assuming that the latest commit passes the `tekton` tests, this is good to go - I've verified it works against Tekton 0.5.1, now just to make sure it also works against Tekton 0.4.0 for compatibility reasons. =)

pkg/builds/build_info.go
```diff
@@ -74,7 +74,7 @@ func CreateBuildPodInfo(pod *corev1.Pod) *BuildPodInfo {
 	containers, _, isInit := kube.GetContainersWithStatusAndIsInit(pod)
 
 	for _, container := range containers {
-		if strings.HasPrefix(container.Name, "build-step-git-source") {
+		if strings.HasPrefix(container.Name, "build-step-git-source") || strings.HasPrefix(container.Name, "step-git-source") {
```
dgozalo: Wouldn't the second condition also be true for the first one? the first one contains "step-git-source".

pkg/logs/tekton_logging.go
```diff
@@ -46,14 +46,21 @@ func GetTektonPipelinesWithActivePipelineActivity(jxClient versioned.Interface,
...
 	if context != "" {
 		labelsFilter = modifyFilterForPipelineRun(labelsFilter, context)
 	}
 	tektonPRs, _ := tektonClient.TektonV1alpha1().PipelineRuns(ns).List(metav1.ListOptions{
 		LabelSelector: labelsFilter,
 	})
 
+	// Handle the old "repo" tag as well for legacy purposes.
+	if len(tektonPRs.Items) == 0 {
+		labelsFilter = strings.Replace(labelsFilter, "repository=", "repo=", 1)
```
dgozalo: It's going to have a "repository" tag instead of a "repo" one now? Nice.

pkg/tekton/pipeline_info.go
```diff
@@ -176,7 +176,7 @@ func CreatePipelineRunInfo(prName string, podList *corev1.PodList, ps *v1.Pipeli
 	}
 	containers, _, isInit := kube.GetContainersWithStatusAndIsInit(pod)
 	for _, container := range containers {
-		if strings.HasPrefix(container.Name, "build-step-git-source") {
+		if strings.HasPrefix(container.Name, "build-step-git-source") || strings.HasPrefix(container.Name, "step-git-source") {
```
dgozalo: Same as with the pod. The second condition is true for the first one.

pkg/builds/build_info.go
abayer: I thought `HasPrefix` looked for, in regex terms, `/^build-step-git-source.*/`?

pkg/logs/tekton_logging.go
abayer: @hferentschik did this last week. =)

pkg/builds/build_info.go
dgozalo: Nvm me! it's HasPrefix and not Contains :D :D :D

abayer: =) I actually considered going with `Contains` but decided to make sure we didn't get confused by someone actually naming a step something like `Step Git Source`, which would end up as a container named `step-step-git-source`, and wouldn't be the right container to look at, etc, etc...

pkg/logs/tekton_logging.go
hferentschik: Now one can find all resources related to a pipeline execution (including the pipeline activity) using the same label, something like:

```
for i in pipelineactivity pipeline pipelinestructure taskrun pipelinerun task pod; do 
   kubectl get "$i" --show-kind --no-headers --ignore-not-found -l repository=demo,build=<x>
  done
```

hferentschik:
@abayer looks good to me. Does merging this require syncing updates of the Tekton controller Helm charts as well?

Issue #4657, opened 2019-07-10
Create a Mocks interface for gcloud / gsutil components


dgozalo:
### Summary 
Currently, there is no Mocks interface for calls to the gcloud / gsutil CLIs, making it impossible to test any feature using them.

There needs to be an interface for the pkg/cloud/gke/gcloud.go component so we can generate mocks using Pegomock and efficiently test features using them.

garethjevans:
@dgozalo It's funny you mentioned this - https://github.com/jenkins-x/jx/pull/4666

garethjevans:
I just happened to be working on the same thing yesterday

PR #4666, opened 2019-07-10
fix: adding an interface around all gcloud commands

garethjevans:
this is the first step in allowing us to mock out gcloud in our unit tests.

more to follow


garethjevans:
Should fix #4657

pmuir:
@garethjevans this branch has conflicts...

PR #4667, opened 2019-07-10
fix: PipelineActivity doesn't have context label so skip in filter

abayer:
#### Submitter checklist

#### Description

`PipelineActivity` doesn't have the `context` label, so if you specify `--context=whatever` to `jx get build logs`, it never finds anything. Let's add a function to get the `PipelineActivity`-specific label selectors.

/assign @dgozalo 

n/a

### DIFF
--- a/pkg/builds/build_info.go
+++ b/pkg/builds/build_info.go
@@ -194,6 +194,21 @@ func CreateBuildPodInfo(pod *corev1.Pod) *BuildPodInfo {
 	return answer
 }
 
+// LabelSelectorsForActivity returns a slice of label selectors relevant to PipelineActivity corresponding to the filter
+func (o *BuildPodInfoFilter) LabelSelectorsForActivity() []string {
+	var labelSelectors []string
+	if o.Owner != "" {
+		labelSelectors = append(labelSelectors, "owner="+o.Owner)
+	}
+	if o.Repository != "" {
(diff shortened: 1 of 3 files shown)


pkg/builds/build_info.go
```diff
@@ -194,6 +194,21 @@ func CreateBuildPodInfo(pod *corev1.Pod) *BuildPodInfo {
...
 }
 
+// LabelSelectorsForActivity returns a slice of label selectors relevant to PipelineActivity corresponding to the filter
+func (o *BuildPodInfoFilter) LabelSelectorsForActivity() []string {
+	var labelSelectors []string
+	if o.Owner != "" {
+		labelSelectors = append(labelSelectors, "owner="+o.Owner)
+	}
+	if o.Repository != "" {
+		labelSelectors = append(labelSelectors, "repo="+o.Repository)
```
hferentschik: Is this really correct? I think the repository label is `repostitory` for PipelineActivity - https://github.com/jenkins-x/jx/blob/eea23e76f73a39fa75d12788e0df3d259d25980e/pkg/kube/activity.go#L445

That's what I try to align with https://github.com/hferentschik/jx/commit/bab9d3bb8ee7efb46ebeb381c123e99e88c4686e, once #4651 is merged. 

Unless I am missing something here!?

abayer: Good point - done.

Issue #4367, opened 2019-06-21
Perform full release of jx on master build of jenkins-x-versions


cagiti:
Once we have a pre-release of jx, we then need to trigger bdd tests to validate the release and publish it as a full release if the tests are successful.
Contributes towards #4120

rawlingsj:
I'm not sure we should run the BDD tests again on a prerelease because they must have passed for the initial PR to be merged.  I think if we're adding more BDD tests then we could add them as extra contexts and run on PRs pre merge to master?

Just my thoughts at least.

cagiti:
> I'm not sure we should run the BDD tests again on a prerelease because they must have passed for the initial PR to be merged. I think if we're adding more BDD tests then we could add them as extra contexts and run on PRs pre merge to master?
> 
> Just my thoughts at least.

On reflection I agree, I was following on from the way in which the release stage of jx executes the slow integration tests.

abayer:
@cagiti - this is done, right?

cagiti:
Completed

PR #4672, opened 2019-07-10
feat: add update release step

cagiti:
This is a new step designed to update the status (prerelease/release) using the release version, this ties into the version stream work #4120 where we only wish to release a test version of `jx` from the versions stream `jenkins-x-versions`. This step command will allow us to easily toggle the status of a version release.
Enables #4367


.gitignore
```diff
@@ -23,6 +23,7 @@ debug
 #Various temporary created while building/running jx.
 /build/
 release/
+!pkg/cmd/step/update/release/
```
pmuir: err, what?

cagiti: yeah, that was a quick fix... the step folder is 'release/' what was being ignored by git due to line above

cagiti: I'll look to update that to something more sensible

cagiti: I meant to go back and change it, but forgot... thanks @pmuir

pkg/cmd/step/update/release/step_update_release_github.go
```diff
@@ -0,0 +1,126 @@
...
+		Aliases: []string{"gh"},
+		Run: func(cmd *cobra.Command, args []string) {
+			options.Cmd = cmd
+			options.Args = args
+			err := options.Run()
+			helper.CheckErr(err)
+		},
+	}
+	AddStepUpdateReleaseFlags(cmd, &options.StepUpdateReleaseOptions)
+	cmd.Flags().BoolVarP(&options.PreRelease, "prerelease", "p", false, "The release state of that version release of the repository")
```
pmuir: I don't understand why this is specific to github...

pkg/cmd/step/update/release/step_update_release_github.go
```diff
@@ -0,0 +1,126 @@
...
+		return err
+	}
+	gitInfo, err := gits.ParseGitURL(gitURL)
+	if err != nil {
+		return err
+	}
+	o.State.GitInfo = gitInfo
+
+	gitKind, err := o.GitServerKind(gitInfo)
+	gitProvider, err := o.State.GitInfo.CreateProvider(o.InCluster(), authConfigSvc, gitKind, o.Git(), false, o.In, o.Out, o.Err)
```
pmuir: This looks wrong. Normally you would do `o.CreateGitProviderFromURLWithoutKind(url)` and that's enough

pmuir:
The use of the git providers is "unusual" and I think the step is not specific to github as it will work with gitea too as of now, but otherwise looks good to me.

pmuir:
The Semantic Pull Request is pending because you have a `wip` PR still in the list

PR #4696, opened 2019-07-12
feat: some improvements on jx get build logs filters

dgozalo:
#### Submitter checklist

#### Description

Improvements on the logs filters.

The context should be added to `PipelineActivity` in the future so we could avoid the first call on `PipelineRun` and the manipulation of filters.


abayer:
/lgtm

Nice!

PR #4710, opened 2019-07-12
fix: handle spaces properly

### LINKED ISSUE #4705: jenkins-x-builders updatebot-created PRs to jenkins-x-platform have malformed YAML
See https://github.com/jenkins-x/jenkins-x-platform/pull/5744/commits/9e4479cd9b66ec7fecab838e8f8351ddaa41069d - the images aren't indented.

pmuir:
This fixes #4705

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/cmd/step/create/pr/step_create_pr_chart.go
+++ b/pkg/cmd/step/create/pr/step_create_pr_chart.go
@@ -1,9 +1,11 @@
 package pr
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 
 	"github.com/jenkins-x/jx/pkg/gits"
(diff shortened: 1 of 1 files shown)


abayer:
/lgtm

...with the caveat that oh boy would I love to see a test. =) But that can wait.

dgozalo:
If we are dealing with yaml files there, wouldn't make sense to unmarshal the file and deal with objects? It would then form the resulting yaml with the correct formatting all the time.

pmuir:
@dgozalo yes, we could have unmarshalled to nested maps I think. TBH I originally copied the updatebot code (not very well apparently ;-) which did it with string replacements. Can you file an issue

@abayer yeah, should have done unit tests for all of these :-( I did them for quite a few of the commands. I'll try to add one in my next long break between builders builds ;-)

Issue #4349, opened 2019-06-20
Retrieve version of git merge image from version stream


hferentschik:
### Summary 

Following up on issue #4338 where the version of the image was updated. Instead of hard-coding the image version in a constant, the version should be retrieved from the version stream.

PR #4724, opened 2019-07-15
fix(tekton): making sure GitMergeImage gets resolved via version stream

hferentschik:
fix for issue #4349


pkg/version/version_data.go
```diff
@@ -255,7 +255,9 @@ func ForEachKindVersion(wrkDir string, kind VersionKind, callback Callback) erro
 	return nil
 }
 
-// ResolveDockerImage ensures the given docker image has a valid version if there is one in the version stream
+// ResolveDockerImage resolves the version of the specified image against the version stream defined in versionsDir.
+// If there is a version defined for the image in the version stream 'image:<version>' is returned, otherwise the
+// passed image name is returned as is.
```
hferentschik: TBH, I find the semantics behind this function a bit odd. I'd have a function which resolved and returns a version instead. Apart from the log messages, one cannot really determine easily whether there is a version for a given image.

Issue #4621, opened 2019-07-07
Improve logging of merge step in meta pipeline 


hferentschik:
When the meta pipeline is built, it adds a merge step to merge all pull refs. It would be nice if the output of the step contains enough information to verify what's HEAD of the current checkout and whether everything got merged.

PR #4744, opened 2019-07-16
 fix: adding info about merged commits in step git merge

hferentschik:
fix for issue #4621

### DIFF
--- a/pkg/cmd/step/create/step_create_task.go
+++ b/pkg/cmd/step/create/step_create_task.go
@@ -231,9 +231,16 @@ func (o *StepCreateTaskOptions) Run() error {
 		}
 	}
 
+	if o.VersionResolver == nil {
+		o.VersionResolver, err = o.CreateVersionResolver("", "")
+		if err != nil {
+			return errors.Wrap(err, "Unable to create version resolver")
+		}
+	}
+
 	pr, err := o.parsePullRefs()
 	if err != nil {
-		return errors.Wrapf(err, "Unable to find or parse PULL_REFS from custom environment")
+		return errors.Wrap(err, "Unable to find or parse PULL_REFS from custom environment")
 	}
 
 	exists, err := o.effectiveProjectConfigExists()
@@ -378,12 +385,6 @@ func (o *StepCreateTaskOptions) createEffectiveProjectConfigFromOptions(tektonCl
 		o.DefaultImage = syntax.DefaultContainerImage
 	}
 	log.Logger().Debugf("cloning git for %s", o.CloneGitURL)
-	if o.VersionResolver == nil {
-		o.VersionResolver, err = o.CreateVersionResolver("", "")
-		if err != nil {
-			return nil, err
-		}
-	}
 	if o.KanikoImage == "" {
 		o.KanikoImage = syntax.KanikoDockerImage
 	}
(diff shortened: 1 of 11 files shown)


pkg/cmd/step/create/step_create_task.go
```diff
@@ -231,9 +231,16 @@ func (o *StepCreateTaskOptions) Run() error {
 		}
 	}
 
+	if o.VersionResolver == nil {
```
hferentschik: @abayer we need to make sure the version resolver gets initialised also for the meta pipeline (when the effective pipeline config exists)

pkg/cmd/step/create/step_create_task_integration_test.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/jenkins-x/jx/pkg/util"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestStepCreateTask(t *testing.T) {
```
hferentschik: @abayer tried to add an integration test which runs _Run()_ in its simplest form in direct execution as well as in meta pipeline execution mode. Hopefully, this helps to catch errors earlier.

pkg/gits/git_cli.go
```diff
@@ -379,49 +379,37 @@ func (g *GitCLI) CommitIfChanges(dir string, message string) error {
...
 func (g *GitCLI) GetCommits(dir string, startSha string, endSha string) ([]GitCommit, error) {
-	args := []string{"log", "--format=raw", fmt.Sprintf("%s..%s", startSha, endSha)}
+	// use a custom format to get commits, using %x1e to separate commits and %x1f to separate fields
+	args := []string{"log", "--format=%H%x1f%an%x1f%ae%x1f%cn%x1f%ce%x1f%s%n%b%x1e", fmt.Sprintf("%s..%s", startSha, endSha)}
 	out, err := g.gitCmdWithOutput(dir, args...)
 	if err != nil {
 		return nil, errors.Wrapf(err, "running git %s", strings.Join(args, " "))
 	}
 	answer := make([]GitCommit, 0)
-	commits := strings.Split(out, "commit ")
```
hferentschik: @pmuir, I don't think splitting at 'commit' is going to work. It breaks if 'commit' appears in the message, eg merge commits.

pkg/gits/git_cli.go
```diff
@@ -379,49 +379,37 @@ func (g *GitCLI) CommitIfChanges(dir string, message string) error {
...
-				commit.SHA = strings.TrimPrefix(line, "commit ")
-			} else if strings.HasPrefix(line, "author ") {
-				name, email := parseAuthor(strings.TrimPrefix(line, "author "))
-				commit.Author = &GitUser{
-					Name:  name,
-					Email: email,
-				}
-			} else if strings.HasPrefix(line, "committer ") {
-				name, email := parseAuthor(strings.TrimPrefix(line, "committer "))
-				commit.Author = &GitUser{
```
hferentschik: @pmuir, I would think this should be commiter

hferentschik:
@abayer, needed to push an update. Forgot to include a test update.

hferentschik:
so why is this darn "Semantic Pull Request" check pending now!?

PR #4748, opened 2019-07-16
feat(tekton): Add support for specifying volumes to add to stage pods

### LINKED ISSUE #4665: We need to be able to declare kubernetes volumes in jenkins-x.yml / pipeline.yaml
### Summary 
Use-case: In order to inject training data into machine-learning builds, we need the ability to mount pre-existing kubernetes volumes into build containers against specified volumeMounts.

This is likely to need to be used as a user override to the pipeline from jenkins-x.yml so each training instance can be customised with a unique data set from storage.

The syntax for declaring the volumeMounts looks like it is in the containerOptions: but I can't seem to find where the matching volumes: declaration needs to go.

Initial requirement is to support gcePersistentDisk and PersistentVolumeClaim storage types with readOnly flags.

abayer:
#### Submitter checklist

#### Description

This will allow pipeline authors to add existing volumes to the pods created for their stages, which can then be mounted via `containerOptions`. Note that this will result in errors for volumes that don't exist, and could cause problems, depending on the volume's configuration, with parallel stages using the same volume.

Docs PR: https://github.com/jenkins-x/jx-docs/pull/1776

/assign @hferentschik 
/assign @dwnusbaum 
cc @tdcox 

fixes #4665

### DIFF
--- a/pkg/cmd/step/create/step_create_task_test.go
+++ b/pkg/cmd/step/create/step_create_task_test.go
@@ -315,6 +315,14 @@ func TestGenerateTektonCRDs(t *testing.T) {
 			kind:         "release",
 			useKaniko:    false,
 		},
+		{
+			name:         "volume-in-overrides",
+			language:     "maven",
+			repoName:     "jx-demo-qs",
+			organization: "abayer",
+			branch:       "master",
+			kind:         "release",
+		},
 	}
 
 	k8sObjects := []runtime.Object{
--- a/pkg/cmd/step/create/test_data/step_create_task/volume-in-overrides/jenkins-x.yml
+++ b/pkg/cmd/step/create/test_data/step_create_task/volume-in-overrides/jenkins-x.yml
@@ -1,0 +1,20 @@
+pipelineConfig:
+  env:
+    - name: FRUIT
+      value: BANANA
+    - name: GIT_AUTHOR_NAME
+      value: somebodyelse
+  pipelines:
+    overrides:
+      # Add a volume and mount it
+      - pipeline: release
+        stage: build
+        volumes:
+          - name: top-level-volume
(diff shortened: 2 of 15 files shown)


tdcox:
Thanks. Is this limited to PVCs at the moment?

abayer:
Yup, just PVCs. I'll try to get the tests/docs done tomorrow.

tdcox:
Ok. Will have to check if I can get what I need to do to work with PVCs. Simple volumes are easier for the use case.

abayer:
How would you be specifying a non-PVC in a pod in your use case, putting aside JX for the moment? I may be able to support that.

abayer:
ah, I gee, `GcePersistentDisk`. And you want to be able to mount them read-only. Lemme see what I can do.

abayer:
@tdcox - done!

abayer:
Ignore the `images` failure - that was just me verifying that gcr is hosed.

tdcox:
Good. Will that also work as an override to the builder pipeline in a jenkins-x.yml?

abayer:
Ah, no, it won't. Let me think about that.

tdcox:
Most likely use-case for this is that there is a standard pipeline in the buildpack that requires the user to override with a custom volume for their specific test data. The same probably applies to the associated volumeMounts, so might be an idea to check if these are supported in overrides too?

abayer:
Yeah - you can set `containerOptions` in `pipelineConfig.containerOptions` in the `jenkins-x.yml`, but that applies across all pipelines/stages/steps in the effective pipeline. We could add `volumes` there as well, but it does feel like some `overrides` syntax support for "define these `containerOptions` and/or `volumes` for this pipeline or stage" would be a good addition. Let me play around with that today.

Issue #4756, opened 2019-07-17
Failing test in helpers_test.go


hferentschik:
When executing the whole test suite `make test`, tests in `helpers_test.go` fail with errors of the form:

```
--- FAIL: TestFetchAndMergeOneSHA (0.31s)
    helpers_test.go:109:
        	Error Trace:	helpers_test.go:109
        	            				helpers_test.go:29
        	Error:      	Received unexpected error:
        	            	failed to run 'git commit -m Initial Commit' command in directory '/var/folders/25/wrys45bs1jscckzb3lj9rtb40000gn/T/remote898921712', output: 'fatal: empty ident name (for <>) not allowed'
```

This happens not when the test is run standalone, but only when the test suite is executed as a whole. There seems to be a dependency to `git_cli_test.go`.

PR #4757, opened 2019-07-17
fix(test): making sure git author and committer env variables won't b…

hferentschik:
…e set to empty string

fix for issue #4756


hferentschik:
I have seen test errors depending on the test execution order. This commit changes git_cli_test.go to be a bit more careful when re-setting the pre-test state.

abayer:
Holding so #4773 can get in.

PR #4778, opened 2019-07-18
fix: Update Garbage collector script for GKE

warrenbailey:
Determine if a service account is in use by using the prefix for the cluster name and not the full cluster name if the service account is related to a bdd test run.

This ensures that the vault token is not deleted whilst the cluster it is related to is still active.

#### Submitter checklist

fixes #


abayer:
Too many clusters running - I'm cleaning them up and hopefully this run will work.

PR #4819, opened 2019-07-22
fix(tekton): Make sure PipelineRuns exist and are running before logging.

abayer:
#### Submitter checklist

#### Description

This fixes issues around `jx get build log` thinking a `PipelineRun` is valid before it's actually started, tricking it into thinking that the `PipelineRun`'s pods have already been GCed.

/assign @dgozalo 

n/a

### DIFF
--- a/pkg/cmd/get/get_build_logs_test.go
+++ b/pkg/cmd/get/get_build_logs_test.go
@@ -63,6 +63,32 @@ func TestGetTektonLogsForRunningBuild(t *testing.T) {
 	}
 }
 
+func TestGetTektonLogsForRunningBuildWithPendingPod(t *testing.T) {
+	commonOpts := opts.NewCommonOptionsWithFactory(fake.NewFakeFactory())
+	commonOpts.BatchMode = true
+	testCaseDir := path.Join("test_data", "get_build_logs", "tekton_build_logs_pending")
+
+	activities := tekton_helpers_test.AssertLoadSinglePipelineActivity(t, testCaseDir)
+	jxClient := jxfake.NewSimpleClientset(activities)
+
+	tektonObjects := []runtime.Object{tekton_helpers_test.AssertLoadSinglePipelineRun(t, testCaseDir)}
+	tektonClient := tektonfake.NewSimpleClientset(tektonObjects...)
+
+	kubeClient := kubeMocks.NewSimpleClientset()
+
+	ns := "jx"
+
+	o := &GetBuildLogsOptions{
+		GetOptions: GetOptions{
+			CommonOptions: &commonOpts,
+		},
+	}
+
+	_, err := o.getTektonLogs(kubeClient, tektonClient, jxClient, ns)
+	assert.NotNil(t, err)
+	assert.Equal(t, "there are no build logs for the supplied filters", err.Error())
+}
+
 func TestGetTektonLogsForRunningBuildWithLegacyRepoLabel(t *testing.T) {
 	commonOpts := opts.NewCommonOptionsWithFactory(fake.NewFakeFactory())
 	commonOpts.BatchMode = true
--- a/pkg/logs/tekton_logging.go
+++ b/pkg/logs/tekton_logging.go
@@ -72,7 +72,7 @@ func GetTektonPipelinesWithActivePipelineActivity(jxClient versioned.Interface, 
 			prBuildNumber = findLegacyPipelineRunBuildNumber(&pr)
 		}
 		paName := createPipelineActivityName(pr.Labels, prBuildNumber)
-		if _, exists := paMap[paName]; exists {
+		if _, exists := paMap[paName]; exists && len(pr.Status.TaskRuns) > 0 {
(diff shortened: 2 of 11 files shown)


pkg/tekton/pipelines.go
```diff
@@ -261,6 +261,16 @@ func ApplyPipelineRun(tektonClient tektonclient.Interface, ns string, run *v1alp
 	resourceName := run.Name
 	resourceInterface := tektonClient.TektonV1alpha1().PipelineRuns(ns)
 
+	f := func() error {
+		_, err := tektonClient.TektonV1alpha1().Pipelines(ns).Get(run.Spec.PipelineRef.Name, metav1.GetOptions{})
+		return err
+	}
+	// Make sure the Pipeline has been created and registered before we try to create the PipelineRun.
+	err := util.RetryWithInitialDelay(2*time.Millisecond, 30*time.Second, f)
```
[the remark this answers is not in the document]
hferentschik: That said, I am a bit sceptical about the `CreateOrUpdatePipeline` implementation:

```golang
	answer, err := resourceInterface.Create(created)
	if err == nil {
		return answer, nil
	}
```

The assumption seems to be that if there is an error it must be that the pipeline existed. That's not the case. There could be many different errors. What's missing here is something like:

```
	if errors.IsAlreadyExists(err) {
		// let try an update
	}
```

pkg/logs/tekton_logging.go
```diff
@@ -72,7 +73,7 @@ func GetTektonPipelinesWithActivePipelineActivity(jxClient versioned.Interface,
 			prBuildNumber = findLegacyPipelineRunBuildNumber(&pr)
 		}
 		paName := createPipelineActivityName(pr.Labels, prBuildNumber)
```
hferentschik: TBH, I don't get what's happening in this method, but that's apart from the current changes

pkg/cmd/get/get_build_logs_test.go
```diff
@@ -63,6 +63,32 @@ func TestGetTektonLogsForRunningBuild(t *testing.T) {
 	}
 }
 
+func TestGetTektonLogsForRunningBuildWithPendingPod(t *testing.T) {
+	commonOpts := opts.NewCommonOptionsWithFactory(fake.NewFakeFactory())
+	commonOpts.BatchMode = true
+	testCaseDir := path.Join("test_data", "get_build_logs", "tekton_build_logs_pending")
+
+	activities := tekton_helpers_test.AssertLoadSinglePipelineActivity(t, testCaseDir)
```
hferentschik: IMO `AssertLoadSinglePipelineActivity` is a bad name here. Really this is about loading the pipeline activity `LoadSinglePipelineActivity`

pkg/cmd/get/get_build_logs_test.go
```diff
@@ -63,6 +63,32 @@ func TestGetTektonLogsForRunningBuild(t *testing.T) {
...
+	jxClient := jxfake.NewSimpleClientset(activities)
+
+	tektonObjects := []runtime.Object{tekton_helpers_test.AssertLoadSinglePipelineRun(t, testCaseDir)}
```
hferentschik: Same as above

hferentschik: OOI, would it be easier to write these types of tests as integration tests? I think there is nothing wrong with integration tests, they also run fast and you would have considerably less mocking going on. WDYT?

pkg/tekton/pipelines.go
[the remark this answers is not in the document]
abayer: So my theory, which is probably wrong, was that maybe the `Pipeline` isn’t getting returned by the API yet when the Tekton reconciler responds to the creation of the `PipelineRun`. I’m now pretty sure that’s not the case, and that the issue is a stale cache in the Tekton reconciler that needs to be addressed there. I’m going to remove this change.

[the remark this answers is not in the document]
hferentschik: >  I’m now pretty sure that’s not the case, and that the issue is a stale cache in the Tekton reconciler that needs to be addressed there. I’m going to remove this change.

that would make more sense. From the perspective of this code, the pipeline exists, otherwise you would not end up in  in this piece of code. That said, the error check we are doing on Create is wrong. If you are going to revert this change, could you please open a new issue to address this?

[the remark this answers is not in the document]
abayer: fwiw, we don't want to update if the `Pipeline` already exists - that means we've got a problem somewhere, like a reused build number, so erroring out is what we want.

pkg/tekton/pipelines_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	ns = "jx"
+)
+
+func TestPipelineRunIsNotPendingCompletedRun(t *testing.T) {
```
hferentschik: 👍

Issue #4789, opened 2019-07-19
Enable gitops for dev env for jx boot


pmuir:
### Jx version

The output of `jx version` is:

```
COPY OUTPUT HERE
```

### Jenkins type

- [ ] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

jstrachan:
@pmuir is this related to the gitops flag? https://github.com/jenkins-x/jx/blob/master/pkg/config/install_requirements.go#L155-L157

pmuir:
More generally that jx boot isn't well integrated with the dev env gitops workflows - e.g. it doesn't fork the upstream for you, it doesn't make PRs against it as it goes, it sets the wrong URL in the team settings, that flag is off by default. 

Right now the only way I could see to make it work was to manually fork the upstream repo, clone it locally, cd into the directory and then run `jx boot` there and then create PRs against that repo by hand.

pmuir:
I wasn't sure what your intent was here so we've worked around it using instructions for now.

pmuir:
It's the manual fork thing that's the critical bit I think, the PRs are less so.

PR #4823, opened 2019-07-22
fix: refactor ForkAndPullRepo

pmuir:
Previously this function branched pretty much at the
start which meant that it was very easy to end up
with fixes inconsistently applied. It is now a
single branch with the different paths done as
fork-joins which should be much easier to maintain.

I’ve also added unit tests for the common paths

Contributes to #4789


hferentschik:
@pmuir there seems to be a whole bunch of test failures. Is this WIP? It is worth reviewing in its current state?

pkg/gits/provider_fake.go
```diff
@@ -723,24 +768,65 @@ func (r *FakeRepository) Name() string {
 }
 
 // NewFakeRepository creates a new fake repository
-func NewFakeRepository(owner string, repoName string) *FakeRepository {
-	return &FakeRepository{
+func NewFakeRepository(owner string, repoName string, addFiles func(dir string) error, gitter Gitter) (*FakeRepository, error) {
```
hferentschik: Would it not be better to have a new function instead here? Something like `NewLocalRepository`? IMO this repo is not a fake per se anymore and by adding _addFiles_ and _gitter_ here you need to adjust a fair bit of existing tests passing effectively _nil_s.

pkg/cmd/controller/controller_workflow_integration_test.go
```diff
@@ -46,9 +47,9 @@ func TestSequentialWorkflow(t *testing.T) {
 	stagingRepoName := "environment-staging"
 	prodRepoName := "environment-production"
 
-	fakeRepo := gits.NewFakeRepository(testOrgName, testRepoName)
-	stagingRepo := gits.NewFakeRepository(testOrgName, stagingRepoName)
-	prodRepo := gits.NewFakeRepository(testOrgName, prodRepoName)
+	fakeRepo, _ := gits.NewFakeRepository(testOrgName, testRepoName, nil, nil)
```
hferentschik: Would it not be better to keep `NewFakeRepository` as is and introduce something like `NewLocal(Test)Repository` instead. Spares the retrofitting with `nil`s and is more accurate name wise.

pkg/gits/git_cli.go
```diff
@@ -220,7 +220,12 @@ func (g *GitCLI) RemoteUpdate(dir string) error {
 
 // Stash stashes the current changes from the given directory
 func (g *GitCLI) Stash(dir string) error {
-	return g.gitCmd(dir, "stash")
+	return g.gitCmd(dir, "stash", "push")
```
hferentschik: Rename to `StashPush` for symmetry and to make things more explicit when seeing this func called.

pkg/gits/git_cli.go
```diff
@@ -220,7 +220,12 @@ func (g *GitCLI) RemoteUpdate(dir string) error {
...
+}
+
+// StashPop applies the last stash
+func (g *GitCLI) StashPop(dir string) error {
```
hferentschik: What is the behaviour if there is no stash at the time `StashPop` is called?

I'd add some tests for `Stash` and `StashPop` to `TestGitCLI`

pkg/gits/git_cli.go
```diff
@@ -448,6 +453,10 @@ func (g *GitCLI) CreatePushURL(cloneURL string, userAuth *auth.UserAuth) (string
 		// already a git/ssh url?
 		return cloneURL, nil
 	}
+	// The file scheme doesn't support auth
```
hferentschik: 👍

pkg/gits/git_cli.go
```diff
@@ -949,3 +958,12 @@ func (g *GitCLI) nthTagSHA(dir string, n int) (string, string, error) {
 
 	return fields[0], fields[1], nil
 }
+
+// Remotes will list the names of the remotes
+func (g *GitCLI) Remotes(dir string) ([]string, error) {
```
hferentschik: I'd add some tests for this in `TestGitCLI`

pkg/gits/helpers.go
```diff
@@ -384,192 +379,185 @@ func PushRepoAndCreatePullRequest(dir string, gitInfo *GitRepository, base strin
 }
 
 // ForkAndPullPullRepo pulls the specified gitUrl into baseDir/org/repo using gitter, creating a remote fork if needed using the git provider
-func ForkAndPullPullRepo(gitURL string, baseDir string, baseRef string, branchName string, provider GitProvider, gitter Gitter, configGitFn ConfigureGitFn) (string, string, *GitRepository, error) {
+//
+// If there are existing files in dir (and dir is already a git clone), the existing files will pushed into the stash
+// and then popped at the end. If they cannot be popped then an error will be returned which can be checked for using
+// IsCouldNotPopTheStashError
+func ForkAndPullPullRepo(gitURL string, dir string, baseRef string, branchName string, provider GitProvider, gitter Gitter, configGitFn ConfigureGitFn) (string, string, *GitRepository, error) {
```
hferentschik: `ForkAndPullPullRepo` - two times `Pull`!?

pkg/gits/helpers.go
```diff
@@ -384,192 +379,185 @@ func PushRepoAndCreatePullRequest(dir string, gitInfo *GitRepository, base strin
 }
 
 // ForkAndPullPullRepo pulls the specified gitUrl into baseDir/org/repo using gitter, creating a remote fork if needed using the git provider
-func ForkAndPullPullRepo(gitURL string, baseDir string, baseRef string, branchName string, provider GitProvider, gitter Gitter, configGitFn ConfigureGitFn) (string, string, *GitRepository, error) {
+//
```
hferentschik: Are the docs correct here? I mean in particular "pulls the specified gitUrl into baseDir/org/repo". The checkout/clone dir us provided via `dir`, right?

pkg/gits/helpers.go
```diff
@@ -384,192 +379,185 @@ func PushRepoAndCreatePullRequest(dir string, gitInfo *GitRepository, base strin
...
+		// If the directory doesn't already exist, create it
+		err := os.MkdirAll(dir, util.DefaultWritePermissions)
 		if err != nil {
-			return "", "", nil, errors.WithStack(err)
+			return "", "", nil, fmt.Errorf("failed to create directory %s due to %s", dir, err)
 		}
-		err = gitter.Stash(dir)
+	}
+
+	stashed := false
```
hferentschik: So when are there really stashed changes in this checkout? I would not expect this really from a "fork and pull" function? Does this function not really try to do multiple things?

pmuir:
> It is worth reviewing in its current state?

Yes - test failures need resolving, but they aren't related to the redesign, just the knock on impact of the redesign (most of them are due to the test harness needing alterations to cope with this new structure as it excercises the code better).

pkg/gits/provider_fake.go
pmuir: That adjustment is now done anyway - that's what was needed to make the tests pass.

pkg/gits/helpers.go
pmuir: typo ;-)

pkg/gits/helpers.go
pmuir: We can rename the function I guess. What do you suggest?
 
The functionality is required to pull into an existing dir and not clobber existing files.

pmuir: This is covered in the comments.

rawlingsj:
If it helps I think this is the error in the tekton check logs:
```
(MISSING)STEP: parsing the output of jx get applications -e staging
STEP: validating that the application bdd-spring-1563919373 was returned by jx get applications -e staging
      failed to get application: No applications found. Output of jx get applications -e staging was No applications found in environments staging. Parsed applications map is map[]`

• Failure [951.272 seconds]
create spring

/tmp/jenkins-x/bdd-jx/test/spring/jx_create_spring.go:20
  Given valid parameters
  /tmp/jenkins-x/bdd-jx/test/spring/jx_create_spring.go:34
    when running jx create spring
    /tmp/jenkins-x/bdd-jx/test/spring/jx_create_spring.go:35
      creates a spring application and promotes it to staging
       [It]
      /tmp/jenkins-x/bdd-jx/test/spring/jx_create_spring.go:36

      get applications with a URL
      Unexpected error:
          <*errors.errorString | 0xc00009efd0>: {
              s: "No applications found",
          }
          No applications found
      occurred

      /tmp/jenkins-x/bdd-jx/test/helpers/test.go:147
```

So perhaps there's an issue with a the app deploying or starting?

pmuir:
The error is a bit higher up in the promotion.

pkg/gits/LICENSE
```diff
@@ -0,0 +1 @@
+TODO ;-)
```
hferentschik: Was this added by mistake? Why would we have a license file here?

pkg/gits/git_cli_test.go
```diff
@@ -394,3 +394,137 @@ func TestTags(t *testing.T) {
 	assert.Contains(t, tags, "v0.0.1")
 	assert.Contains(t, tags, "v0.0.2")
 }
+
+func TestGitCLI_Stash(t *testing.T) {
```
hferentschik: +1 nice with tests. Personally I was hoping to extend the Ginko test suite. IMO there would be several advantages:

* clearer test suite setup and teardown
* better error reports
* ability to focus on tests. 

However, I am just happy to have tests on this level.

pkg/gits/git_cli_test.go
```diff
@@ -394,3 +394,137 @@ func TestTags(t *testing.T) {
...
+				assert.True(t, os.IsNotExist(err))
+				return nil
+			},
+			wantPopErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := gits.NewGitCLI()
+			dir, err := ioutil.TempDir("", "")
```
hferentschik: your tmp dirs are not deleted

pkg/gits/git_cli_test.go
```diff
@@ -394,3 +394,137 @@ func TestTags(t *testing.T) {
...
+		initFn  func(dir string, gitter gits.Gitter) error
+		name    string
+		g       *gits.GitCLI
+		want    []string
+		wantErr bool
+	}{
+		{
+			name: "origin",
+			initFn: func(dir string, gitter gits.Gitter) error {
+				rDir, err := ioutil.TempDir("", "")
```
hferentschik: same as above

pkg/gits/helpers.go
hferentschik: > The functionality is required to pull into an existing dir and not clobber existing files.

Sure, I guess I would have done this differently. I'd check whether the dir does not exist or is not a git repo and in this case do a standard fork and clone and if it exist I'd add the remote and do the other bits required. 

Maybe a different name could be `ForkAndPullIntoLocalRepo`!?

pkg/gits/git_cli_test.go
pmuir: yeah, tbh I'm not sure why we added Ginko here. I would have preferred to stick with the standard idiom for testing within the project.

pkg/gits/LICENSE
pmuir: Yeah.

pkg/gits/git_cli_test.go
hferentschik: Fair enough. IMO Ginko has several advantages over the "standard idioms" and hence I prefer using Ginko, especially since it is already used at other places. 

Might be worth having an testing approach discussion across the team at some stage.

pkg/gits/git_cli_test.go
```diff
@@ -394,3 +394,146 @@ func TestTags(t *testing.T) {
...
+				return nil
+			},
+			wantPopErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := gits.NewGitCLI()
+			dir, err := ioutil.TempDir("", "")
+			defer func() {
```
hferentschik: 👍

PR #4835, opened 2019-07-23
fix(tekton): Look for more pods when running `jx get build logs`

### LINKED ISSUE #4548: jx get build logs doesn't pick up next stage if it hadn't started yet when command was run
`jx get build logs` will show you all stages' logs if the stages have all started, but if a stage starts after you've run `jx get build logs`, it'll just exit when the last stage that was running when the command started finishes. We should refresh the stages/pods whenever we complete the logs for a running stage.

abayer:
#### Submitter checklist

#### Description

We should be checking for logs for all stages in a pipeline, not just those which have launched at the time we start `jx get build logs`. So let's track which stages we've already seen in each `PipelineRun` and loop until we see them all.

/assign @hferentschik 
/assign @dgozalo 

fixes #4548


abayer:
Note - I'm not sure how to test the pods changing while `GetRunningBuildLogs` is running - any suggestions/pointers on that would be hugely appreciated. =)

abayer:
fwiw, I have tested this manually (on this PR's own `bdd` and `tekton` builds, since they're multi-stage!) and it works _perfectly_. =)

PR #4843, opened 2019-07-24
fix: Don't do release prepare for bdd and tekton contexts, copy pipeline

abayer:
#### Submitter checklist

#### Description

We're trying to run the `tekton` and `bdd` contexts on post-submit, but they've been failing forever due to there not being a `release` pipeline in either context. But if there _had_ been `release` pipelines, it would have tried to do new release versions for those contexts! So let's copy the `pullRequest` pipeline (until we're on a new enough version on prod that we can switch to `default` pipelines instead) and set `noReleasePrepare: true` to prevent the version number changing too.

/assign @pmuir 
/assign @warrenbailey 

n/a

### DIFF
--- a/jenkins-x-bdd.yml
+++ b/jenkins-x-bdd.yml
@@ -1,7 +1,133 @@
 buildPack: none
+noReleasePrepare: true
 pipelineConfig:
   pipelines:
+    # TODO: When the JX infra cluster is updated, switch to a default pipeline instead.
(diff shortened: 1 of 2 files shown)


abayer:
blech, there's an issue with `noReleasePrepare: true`...

abayer:
Yeah, this won't work. Ah well.

Issue #4851, opened 2019-07-25
jx root command contains flags which are specific to only a few commands


hferentschik:
Showing the options for the `jx` root command, shows that there are shared global flags which really should only be specific to a certain subset of commands:

```
jx options
The following options can be passed to any command:
  -b, --batch-mode=false: Runs in batch mode without prompting for user input
      --config-file='': Configuration file used for installation
      --install-dependencies=false: Enables automatic dependencies installation when required
      --no-brew=false: Disables brew package manager on MacOS when installing binary dependencies
      --skip-auth-secrets-merge=false: Skips merging the secrets from local files with the secrets from Kubernetes cluster
      --verbose=false: Enables verbose output
```

Onlt `batch-mode`, `verbose` and `version ` (not sure about the latter), should be available (inherited) to all sub commands. The remaining flags are specific to specific sub commands, in particular create and install.

hferentschik:
merged via PR #4852

PR #4852, opened 2019-07-25
fix: separating global flag options and install/create cluster flags

hferentschik:
fix for issue #4851

### DIFF
--- a/pkg/cmd/cmd.go
+++ b/pkg/cmd/cmd.go
@@ -71,7 +71,7 @@ import (
 // args used to determine binary plugin to run can be overridden (does not affect compiled in commands).
 func NewJXCommand(f clients.Factory, in terminal.FileReader, out terminal.FileWriter,
 	err io.Writer, args []string) *cobra.Command {
-	cmds := &cobra.Command{
+	rootCommand := &cobra.Command{
 		Use:              "jx",
 		Short:            "jx is a command line tool for working with Jenkins X",
 		PersistentPreRun: setLoggingLevel,
@@ -81,9 +81,7 @@ func NewJXCommand(f clients.Factory, in terminal.FileReader, out terminal.FileWr
 	features.Init()
 
 	commonOpts := opts.NewCommonOptionsWithTerm(f, in, out, err)
-
-	// commonOpts holds the global flags that will be shared/inherited by all sub-commands created bellow
-	commonOpts.AddCommonFlags(cmds)
+	commonOpts.AddBaseFlags(rootCommand)
 
 	addCommands := add.NewCmdAdd(commonOpts)
 	createCommands := create.NewCmdCreate(commonOpts)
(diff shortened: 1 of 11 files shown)


hferentschik:
I am wondering where these options really should be available. So far they were available via the root.

```golang
func (o *CommonOptions) AddSharedInstallFlags(cmd *cobra.Command) {
	cmd.PersistentFlags().StringVarP(&o.ConfigFile, "config-file", "", "", "Configuration file used for installation")
	cmd.PersistentFlags().BoolVarP(&o.NoBrew, OptionNoBrew, "", false, "Disables brew package manager on MacOS when installing binary dependencies")
	cmd.PersistentFlags().BoolVarP(&o.InstallDependencies, OptionInstallDeps, "", false, "Enables automatic dependencies installation when required")
	cmd.PersistentFlags().BoolVarP(&o.SkipAuthSecretsMerge, OptionSkipAuthSecMerge, "", false, "Skips merging the secrets from local files with the secrets from Kubernetes cluster")

	o.Cmd = cmd
}
```

I think they are only relevant for install and create cluster, right? If so, one might as well remove this function and add the flags to `InstallOptions#AddInstallFlags`. WDYT?

abayer:
Seems reasonable to me, but yeah, the potential ramifications make me wary of lgtm'ing it myself. =)

hferentschik:
> Seems reasonable to me, but yeah, the potential ramifications make me wary of lgtm'ing it myself. =)

Right. So I am sure about the changes to the flags in the root command. What I am not so sure about is where except create/install I would need to add _no-brew_, _config-file_, etc. I was hoping to get some feedback from the people working in this area. Was the config file changes not @cagiti's baby?

hferentschik:
or maybe it was @macox'es?

hferentschik:
@macox, @ccojocar - do you fancy to have a look at this?

macox:
@hferentschik I moved the config-file flag down in to common options on your request in this PR  https://github.com/jenkins-x/jx/pull/4308

hferentschik:
> I moved the config-file flag down in to common options on your request in this PR #4308

I see. Might have been a misunderstanding or I was just not thinking straight. As it stands now, the `--config-file` option is available for all commands, even though it only applies for create cluster and install Jenkins-X. 

I could envision that the config file also allows to persist flags one always want to set (eg always set the _verbose_ flag), but for now we don't do that and I don't think that something like that becomes a high priority in the near future. 

I also don't just move the `--config-file` option, but all of the options under `AddCommonFlags`, since they are no really common to all commands. Common are _verbose_ and _batch_ imo.

macox:
I'm happy with moving `--config-file` to install, some consensus on the others might be good though

hferentschik:
@jenkins-x/core ^^

hferentschik:
@garethjevans any take on this? ;-)

hferentschik:
@pmuir falls into our CLI discussion today. Current behaviour around global flags is just wrong. This is an attempt to rectify this.

PR #4857, opened 2019-07-25
fix(tekton): Properly handle non-step overrides for packs using jx syntax

abayer:
#### Submitter checklist

#### Description

Applying the overrides as we had been twice for build packs worked fine for `jenkins-x.yml`-defined pipelines or build packs that use the traditional build pack syntax. But for build packs using `jenkins-x.yml` syntax, it resulted in duplicate steps being added. So instead, let's split the step override from the non-step override.

/assign @hferentschik 

n/a

### DIFF
--- a/pkg/cmd/step/create/test_data/step_create_task/packs/jx-syntax-in-buildpack/pipeline.yaml
+++ b/pkg/cmd/step/create/test_data/step_create_task/packs/jx-syntax-in-buildpack/pipeline.yaml
@@ -1,0 +1,28 @@
+agent:
+  label: jenkins-maven
+  container: maven
+pipelines:
+  pullRequest:
+    pipeline:
+      agent:
+        image:
+          maven
+      stages:
+        - name: build
+          steps:
+            - command: source /root/.bashrc && flake8
+              name: flake8
+              image: maven
+              dir: /workspace/source
+  release:
+    pipeline:
+      agent:
+        image:
+          maven
+      stages:
+        - name: build
+          steps:
+            - command: source /root/.bashrc && flake8
+              name: flake8
+              image: maven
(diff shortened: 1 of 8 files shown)


pkg/cmd/step/syntax/step_syntax_effective_test.go
```diff
@@ -490,6 +490,53 @@ func TestCreateCanonicalPipeline(t *testing.T) {
 				},
 			},
 		},
+	}, {
+		name:         "overrides-with-buildpack-using-jenkins-x-syntax",
+		pack:         "jx-syntax-in-buildpack",
+		repoName:     "jx-demo-qs",
+		organization: "abayer",
+		branch:       "master",
+		expected: &config.ProjectConfig{
```
hferentschik: I have been wondering for quite some time whether for these tests it would not be better to also have the expected results as yaml directly and compare the expected yaml with the generated one. I find it hard to make sense of all these structs. It  would also move out a lot of code from this test file.  I have a feeling that it might be also easier to build the expected config in YAML directly. 

Not sure how much work this change would be. As said something I am wondering about.

pkg/cmd/step/create/test_data/step_create_task/packs/jx-syntax-in-buildpack/pipeline.yaml
```diff
@@ -0,0 +1,28 @@
+agent:
+  label: jenkins-maven
+  container: maven
```
hferentschik: I find it so confusing that we are dealing with 'agent', 'container', 'image'. I guess some things derive from trying to consolidate different models, but would it still not possible to have a single concept in the `jenkins-x.yml`? But maybe this ship has sailed?

PR #4872, opened 2019-07-26
fix: need to duplicate the repo first

pmuir:
#### Submitter checklist

#### Description
Rather than making a duplicate during the fork
we need to make a duplicate first and then use the
normal fork logic to handle the user/org scenario.

There are no tests for the duplicator right now, I'll do those next in a separate PR.

fixes #


pmuir:
/hold 

Not finished yet - just for sharing

pmuir:
This is missing tests right now - I'll get those in early next week.

pmuir:
/override tekton

As poor @daveconde is not allowed to

pmuir:
Note we checked the logs and this failure was transient, the previous run passed fine, and the interim change was just to a test file so unrelated.

PR #4924, opened 2019-07-31
fix: print the quickstarts that match

carlossg:
When more than one does in batch mode

#### Submitter checklist

#### Description

Since https://github.com/jenkins-x-quickstarts/golang-http-from-jenkins-x-yml was created now this command used in many examples fail

```
jx create quickstart \
  --language go \
  --project-name xxx \
  --batch-mode

error: More than one quickstart matches the current filter options. Try filtering based on other criteria (eg. Owner or Text)
```

With the patch it prints

```
error: More than one quickstart matches the current filter options. Try filtering based on other criteria (eg. Owner or Text): [golang-http golang-http-from-jenkins-x-yml]
```

/cc @vfarcic 

fixes #

### DIFF
--- a/pkg/quickstarts/model.go
+++ b/pkg/quickstarts/model.go
@@ -110,7 +110,7 @@ func (model *QuickstartModel) CreateSurvey(filter *QuickstartFilter, batchMode b
 		answer = names[0]
 	} else if batchMode {
 		// should not prompt for selection in batch mode so return an error
-		return nil, fmt.Errorf("More than one quickstart matches the current filter options. Try filtering based on other criteria (eg. Owner or Text)")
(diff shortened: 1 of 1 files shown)


carlossg:
Looks like the right way now to do the same example is 

```
jx create quickstart \
  --filter golang-http \
  --project-name xxx \
  --batch-mode
```

carlossg:
Hound is missing from the tests ?

hferentschik:
> Hound is missing from the tests ?

Go figure. I cannot override it either. Maybe it helps to close this PR and re-create it?

hferentschik:
Or one does it the old fashioned way ;-)

PR #4968, opened 2019-08-05
fix: added optional git reporting of pipelines for lighthouse

jstrachan:
longer term we're hoping to remove the build controller completely and replace with tekton `PullRequestResource` output reporting of git statuses. 

This PR adds a tactical workaround so we can get the version stream BDD tests going green with lighthouse - enabling a feature flag to report pipeline statuses to git in the `build controller`


abayer:
/assign @hferentschik 

I just put this on hold - I'd really like @hferentschik's thoughts.

jstrachan:
gonna split this into 2 PRs - I didn't mean to combine them

Issue #4971, opened 2019-08-05
Define interface for meta pipeline generation


hferentschik:
In order to avoid triggering the meta pipeline generation via `StepCreatePipelineOptions.Run` we want to extract an interface, eg `MetaPipelineGenerator` which defines all functions needed to create the meta pipeline. There already exist already `CreateMetaPipelineCRDs(params CRDCreationParameters)` in metapipeline.go, but there is still some pre-processing `step_create_meta_pipeline`.

The new interface should allow to create the meta pipeline by just passing the required parameters and it should be independent from from `jx step create pipeline` .  `jx step create pipeline` will itself call this new interface. Another potential client is Lighthouse - see https://github.com/jenkins-x/lighthouse/issues/61

This issue includes:

* Creation of interface
* Definition of required inputs 
* (optional) Extracting metapipline code/interface into a top level _metapipeline_ package in order to be able to easier consume meta pipeline as a dependency

deanesmith:
Approved but waiting for production cluster overhaul.

PR #4972, opened 2019-08-05
fix: add a metaclient

jstrachan:
* add a simple client which doesn't depend on the cobra / options /
    CommonOptions stuff

* helps lower the footprint of dependent code required for lighthouse

fixes https://github.com/jenkins-x/lighthouse/issues/61


pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+// The name is either an actual git branch name (eg master) or a synthetic names  like PR-<number> for single pull requests
+// or 'batch' for Prow batch builds.
+// The method makes the decision purely based on the Prow pull ref. At this stage we don't have the full ProwJov spec
+// available.
+func (o *MetaClient) determineBranchIdentifier(pullRef prow.PullRefs) string {
+	prCount := len(pullRef.ToMerge)
+	var branch string
+	switch prCount {
+	case 0:
+		{
```
ccojocar: I think the brackets can be removed around cases.

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+	if prCount > 0 {
+		kind = jenkinsfile.PipelineKindPullRequest
+	} else {
+		kind = jenkinsfile.PipelineKindRelease
+	}
+	log.Logger().Debugf("pipeline kind for pull ref '%s' : '%s'", pullRef.String(), kind)
+	return kind
+}
+
+func (o *MetaClient) gitCloneVersionStream(jxClient jxclient.Interface, ns string) (string, error) {
```
ccojocar: Do we really need the pipeline to be dependent to the version stream?

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+	if ref == "" {
+		ref = config.DefaultVersionsRef
+	}
+	log.Logger().WithField("url", url).WithField("ref", ref).Infof("shallow cloning version stream")
+	err = o.Git().ShallowClone(dir, url, ref, "")
+	return dir, err
+}
+
+// Git returns the git client
+func (o *MetaClient) Git() gits.Gitter {
```
ccojocar: Does this method need to be public?

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+	if o.git == nil {
+		o.git = gits.NewGitCLI()
+	}
+	return o.git
+}
+
+// SetGit sets the git client
+func (o *MetaClient) SetGit(git gits.Gitter) {
```
ccojocar: ditto

pkg/jxfactory/factory.go
```diff
@@ -0,0 +1,191 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+
+	// this is so that we load the auth plugins so we can connect to, say, GCP
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+)
+
+type factory struct {
+	Batch bool
```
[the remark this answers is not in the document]
jstrachan: deleted

pkg/jxfactory/factory.go
```diff
@@ -0,0 +1,191 @@
...
+	trace := os.Getenv("TRACE_KUBE_API")
+	if trace == "1" || trace == "on" {
+		config.WrapTransport = func(rt http.RoundTripper) http.RoundTripper {
+			return &Tracer{rt}
+		}
+	}
+	return config, nil
+}
+
+var kubeConfigCache *string
```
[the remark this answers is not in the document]
jstrachan: moved inside the factory

pkg/jxfactory/factory.go
```diff
@@ -0,0 +1,191 @@
...
+
+func createKubeConfig(offline bool) *string {
+	if offline {
```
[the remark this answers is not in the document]
jstrachan: lemme zap - looks like old code from testing the CLI factory

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+	"github.com/jenkins-x/jx/pkg/tekton/syntax"
+	"github.com/jenkins-x/jx/pkg/util"
+	"github.com/pkg/errors"
+	tektonclient "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
+	corev1 "k8s.io/api/core/v1"
+	kubeclient "k8s.io/client-go/kubernetes"
+)
+
+const (
+	jobOptionName       = "job"
```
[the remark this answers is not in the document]
jstrachan: wonder if we rename this `Pipeline`? Its usually `$owner/$repoName/$branch`

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+func (o *MetaClient) getPodTemplates(kubeClient kubeclient.Interface, ns string, containerName string) (map[string]*corev1.Pod, error) {
+	podTemplates, err := kube.LoadPodTemplates(kubeClient, ns)
+	if err != nil {
+		return nil, err
+	}
+	return podTemplates, nil
+}
+
+// GetFactory returns the factory
+func (o *MetaClient) GetFactory() jxfactory.Factory {
```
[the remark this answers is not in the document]
jstrachan: made private

pkg/tekton/metapipeline/metaclient/metaclient.go
jstrachan: no - made private

pkg/tekton/metapipeline/metaclient/metaclient.go
jstrachan: I left this one public so we can create test cases with fake gitters in other packages

jstrachan:
@ccojocar think I've got all those issues fixed now

pkg/tekton/metapipeline/metaclient/metaclient.go
jstrachan: yes - we need this for replacing docker image versions inside the meta pipeline

pkg/tekton/metapipeline/metaclient/metaclient.go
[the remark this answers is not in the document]
jstrachan: ah - looks like its not even used ;) have removed it

pkg/tekton/metapipeline/metaclient/metaclient.go
jstrachan: lemme make that one private until we need it

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+	OutDir      string
+	Verbose     bool
+	NoApply     *bool
+	versionsDir string
+	git         gits.Gitter
+	factory     jxfactory.Factory
+}
+
+// Run creates the meta pipeline
+func (o *MetaClient) Run() error {
```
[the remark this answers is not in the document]
jstrachan: done

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+
+	// resourceName is shared across all builds of a branch, while the pipelineName is unique for each build.
+	resourceName := tekton.PipelineResourceNameFromGitInfo(gitInfo, branchIdentifier, o.Context, tekton.MetaPipeline, nil, "")
+	pipelineName := tekton.PipelineResourceNameFromGitInfo(gitInfo, branchIdentifier, o.Context, tekton.MetaPipeline, tektonClient, ns)
+	buildNumber, err := tekton.GenerateNextBuildNumber(tektonClient, jxClient, ns, gitInfo, branchIdentifier, retryDuration, o.Context)
+	if err != nil {
+		return errors.Wrap(err, "unable to determine next build number")
+	}
+
+	if o.Verbose {
```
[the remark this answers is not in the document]
jstrachan: its what was there before - but I've removed `Verbose` and switched to `Debugf`

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+	versionsDir string
+	git         gits.Gitter
+	factory     jxfactory.Factory
+}
+
+// Run creates the meta pipeline
+func (o *MetaClient) Run() error {
+	err := o.validateArguments()
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
jstrachan: the error already has reasonable error messages around what arguments are missing etc - but wrapped as well for good measure

pkg/tekton/metapipeline/metaclient/metaclient.go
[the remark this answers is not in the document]
jstrachan: done

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+		if err != nil {
+			return errors.Wrapf(err, "failed to output Tekton CRDs")
+		}
+	} else {
+		err := tekton.ApplyPipeline(jxClient, tektonClient, tektonCRDs, ns, gitInfo, branch, pipelineActivity)
+		if err != nil {
+			return errors.Wrapf(err, "failed to apply Tekton CRDs")
+		}
+		if o.Verbose {
+			log.Logger().Infof("applied tekton CRDs for %s", tektonCRDs.PipelineRun().Name)
```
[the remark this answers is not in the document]
jstrachan: done

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+		if err != nil {
+			return err
+		}
+
+		defer os.RemoveAll(o.versionsDir)
+	}
+
+	crdCreationParams := metapipeline.CRDCreationParameters{
+		Namespace:        ns,
+		Context:          o.Context,
```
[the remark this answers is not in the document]
jstrachan: I don't follow - they need to be public to pass them in from outside this package?

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+	retryDuration      = time.Second * 30
+	defaultCheckoutDir = "source"
+)
+
+// MetaClient contains the arguments to create the meta pipeline
+type MetaClient struct {
+	SourceURL      string
+	PullRefs       string
+	Context        string
+	Job            string
```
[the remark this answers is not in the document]
jstrachan: Job has been removed

hferentschik:
I see where you are coming with this now and as mentioned yesterday it relates to issue #4971. 

I am concerned that this is basically just a cut&paste of various bits and pieces of existing code w/o updating/adjusting the whole codebase. I also think the extracted client falls off short of what I was hoping from the meta pipeline client. 

I'll add a few comments to the pull request itself.

pkg/cmd/namespace/namespace.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
 	"github.com/jenkins-x/jx/pkg/cmd/opts"
 	"github.com/jenkins-x/jx/pkg/cmd/templates"
-	"github.com/prometheus/common/log"
+	"github.com/jenkins-x/jx/pkg/log"
```
hferentschik: good catch

pkg/jxfactory/factory.go
```diff
@@ -0,0 +1,183 @@
+package jxfactory
```
hferentschik: I like the idea of splitting out the jxfactory. This will make it also easier for Apps which need a client to modify Jenkins-X CRDs. That said, I think we should adjust the whole codebase in this case opposed to duplicate code.

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
+package metaclient
```
hferentschik: I don't think this additional pkg is needed. I'd just add _client.go_ to the parent package. It is also a bit too much _meta_ in this path - metapipeline/metaclient/metaclient.go

pkg/tekton/metapipeline/metaclient/metaclient.go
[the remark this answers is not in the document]
ccojocar: yeah, that's right!

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+const (
+	pullRefOptionName   = "pull-refs"
+	sourceURLOptionName = "source-url"
+
+	retryDuration      = time.Second * 30
+	defaultCheckoutDir = "source"
+)
+
+// MetaClient contains the arguments to create the meta pipeline
+type MetaClient struct {
```
hferentschik: I would think _Client_ would be enough. No need to repeat _meta_. Also, I was hoping to lock the interface down more. In its current form, this is just what we have in `step create pipeline` without Cobra. I was thinking of creating an actual interface, adding a factory method and a default implementation. 

Instead of having this exported properties on the struct, I'd like to define a couple of functions like `createMetaPipelineCRDs` and `applyMetaPipleCRDs` which take as parameters the required input. Certain options are specific to the client instance, whereas others are the actual parameters needed for the CRD generation. I'd like to separate this and also discuss with @ccojocar what type of parameters we want to pass between Lighthouse and meta pipeline client. For example, do we really want to keep Prows PullRef syntax?

pkg/tekton/metapipeline/metaclient/metaclient.go
```diff
@@ -0,0 +1,324 @@
...
+	Results     tekton.CRDWrapper
+	OutDir      string
+	NoApply     bool
+	versionsDir string
+	gitter      gits.Gitter
+	factory     jxfactory.Factory
+}
+
+// Create creates the meta pipeline
+func (c *MetaClient) Create() error {
```
hferentschik: From the Create method, I would expect that it actually returns the created resources instead of applying them directly. This way a client can decide what to do with the resources. For example, apply them to the cluster or write them to disk or generate some shell script for the local execution of the commands in the pipeline (aka interpret mode).

pkg/tekton/metapipeline/metaclient/metaclient.go
hferentschik: As with jxfactory it worries me that this is duplicating the code.

pkg/jxfactory/interface.go
```diff
@@ -0,0 +1,38 @@
...
+	"k8s.io/client-go/rest"
+
+	tektonclient "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
+
+	// this is so that we load the auth plugins so we can connect to, say, GCP
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+)
+
+// Factory is the interface defined for Kubernetes, Jenkins X, and Tekton REST APIs
+//go:generate pegomock generate github.com/jenkins-x/jx/pkg/jxfactory Factory -o mocks/factory.go
```
hferentschik: Interesting that the lint task passed, since the pegomocks are not part of the pull request. I think the problem is in the Makefile and _verify-generation-complete_ rule:

```Makefile
	$(eval CHANGED = $(shell git ls-files --modified -- . ':(exclude)pkg/client/openapi/k8s_io_apimachinery_intstr_unversioned/openapi_generated.go'))
```

It only bails on modified files. One probably needs to add `--others` as well to handle non tracked file:

```Makefile
	$(eval CHANGED = $(shell git ls-files --modified --others -- . ':(exclude)pkg/client/openapi/k8s_io_apimachinery_intstr_unversioned/openapi_generated.go'))
```

jstrachan:
@hferentschik this was a first iteration to remove the options/cobra dependency. We can keep improving + refactoring

pkg/tekton/metapipeline/metaclient/metaclient.go
jstrachan: I added jxfactory as we have other uses for it outside of metaclient (e.g. for JXUI) so its a handy way to have factory methods to create kube / jx / tekton clients.

We can easily refactor the clients package to reuse jxfactory

pkg/tekton/metapipeline/metaclient/metaclient.go
jstrachan: @hferentschik feel free to keep refactoring - this was literally just the first step along the road to decouple lighthouse from the options/cobra stuff

pkg/tekton/metapipeline/metaclient/metaclient.go
jstrachan: BTW actually there are a few changes with jxfactory to remove the CLI handling code loading kubernetes configurations; so its probably best if we have a minor bit of duplication for now for loading the kube configs between the CLI + the jxfactory. The factory code is fairly simple though; sometimes a small bit of duplication is OK

pkg/tekton/metapipeline/metaclient/metaclient.go
jstrachan: I've moved the MetaClient into metapipeline as `Client`

pkg/jxfactory/factory.go
jstrachan: as commented elsewhere in the review; the code in clients is dependent on CLI flags + a static variable + works a little different; so its probably better for now to keep both factory implementations and over time try remove the old factory with the new factory where we can

pkg/jxfactory/interface.go
jstrachan: I've checked in the generated mocks now

jstrachan:
@hferentschik I think I've addressed all of your concerns now in the latest push

pkg/tekton/metapipeline/client.go
```diff
@@ -0,0 +1,323 @@
...
+const (
+	pullRefOptionName   = "pull-refs"
+	sourceURLOptionName = "source-url"
+
+	retryDuration      = time.Second * 30
+	defaultCheckoutDir = "source"
+)
+
+// Client contains the arguments to create the meta pipeline
+type Client struct {
```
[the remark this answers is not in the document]
hferentschik: Would it not be better to also define a proper interface and then switch to that interface, opposed to switching to this "client" and then once more once we refactor the code?

My understanding is that the current Lighthouse code using `StepCreatePipelineOptions` works and we are just trying to make the import easier and reduce the final binary size. Is there another reason to move so fast on this?

Also, have we really landed on a solution regarding https://github.com/jenkins-x/lighthouse/issues/61 yet?

pkg/tekton/metapipeline/client.go
```diff
@@ -0,0 +1,323 @@
...
+func (c *Client) handleResult(tektonClient tektonclient.Interface,
+	jxClient jxclient.Interface,
+	tektonCRDs *tekton.CRDWrapper,
+	buildNumber string,
+	branch string,
+	pullRefs prow.PullRefs,
+	ns string,
+	gitInfo *gits.GitRepository) error {
+
+	pipelineActivity := tekton.GeneratePipelineActivity(buildNumber, branch, gitInfo, &pullRefs, tekton.MetaPipeline)
```
hferentschik: This is the part we can improve by having a more fine-grained interface  opposed to a single _Create_.

pkg/tekton/metapipeline/client.go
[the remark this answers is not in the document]
jstrachan: I'm not sure we need an interface given there's only 1 implementation any time soon; we need the struct anyway to pass the arguments so not really sure it buys us much. There's already an interface for all this in lighthouse anyway.

We'd like to minimise the `jx` dependency on both lighthouse but other downstream repositories too ASAP; hence the incremental refactor. 

Note we can keep refactoring code for the rest of time so why the need to slow down and do all possible refactors you can think of on 1 PR?

pkg/tekton/metapipeline/client.go
jstrachan: sounds fine - though the immediate need is a single `Create`. We can add more fine grained methods later right?

hferentschik:
@jstrachan closing this PR for now. I'll take it from there and try to develop this a bit further. I'll open a new pull request asap.

PR #4976, opened 2019-08-05
fix: switch jx step create pr versions to new logic

pmuir:
#### Submitter checklist

#### Description
This commit changes jx step create pr to use the
new logic built for creating PRs

It also moves the version stream stuff to it’s own package

fixes #


pmuir:
@cagiti @dgozalo not sure I got all the tests to pass yet, but this is now conceptually correct so well worth a look I think.

Issue #4970, opened 2019-08-05
GC Previews not working in vault cluster


warrenbailey:
### Summary 

The GC previews job fails in a vault enabled cluster

### Steps to reproduce the behavior

1. Create a new K8 cluster 
1. Install JX via boot with **vault** enabled - ensure the GC service account fix is apply to that cluster (https://github.com/jenkins-x/jenkins-x-platform/pull/5619)

1. Create a quick start project (e.g. jx create spring)

1. Create a PR for that project

1. Once preview environment is available

1. Run the GC preview job

### Expected behavior

The GC job should remove the PR environment

### Actual behavior

The GC job errors with the following output

error: no "jx-vault-wbailey-boot1" vault found in namespace "jx"

### Jx version

The output of `jx version` is:

NAME               VERSION
jx                 2.0.551
Kubernetes cluster v1.12.8-gke.10
kubectl            v1.15.0
helm client        Client: v2.14.2+ga8b13cc
git                git version 2.22.0
Operating System   Mac OS X 10.14.5 build 18F203

### Jenkins type

- [x] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

### Kubernetes cluster

GKE

### Operating system / Environment

Mac O/S

warrenbailey:
Unfortunately the vault code hides the error. I've added some additional logging and it looks like it's missing another service account permission.

{"level":"error","msg":"Vault jx-vault-wbailey-boot1 not found in namespace %!n(string=jx)s, err is vaults.vault.banzaicloud.com \"jx-vault-wbailey-boot1\" is forbidden: User \"system:serviceaccount:jx:jenkins-x-gcpreviews\" cannot get resource \"vaults\" in API group \"vault.banzaicloud.com\" in the namespace \"jx\"","time":"2019-08-05T13:58:59Z"}

warrenbailey:
I've raised a PR to add the necessary vault permissions here: https://github.com/jenkins-x/jenkins-x-platform/pull/5822

Unfortunately that doesn't fully fix the issue, as then you can see the gcpreview job can't actually unseal the vault

error: wait for vault to be initialized and unsealed: reading vault health: Get /v1/sys/health?drsecondarycode=299&performancestandbycode=299&sealedcode=299&standbycode=299&uninitcode=299: unsupported protocol scheme ""

warrenbailey:
[...]

PR #4986, opened 2019-08-06
chore: add debug logging for vault

warrenbailey:
Whilst fixing a gcpreview issue #4970 I added some useful logging for debugging vault issues

Signed-off-by: warrenbailey <warren@warrenbailey.net>

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/cmd/clients/factory.go
+++ b/pkg/cmd/clients/factory.go
@@ -416,6 +416,7 @@ func (f *factory) ResetSecretsLocation() {
 
 // CreateSystemVaultClient gets the system vault client for managing the secrets
 func (f *factory) CreateSystemVaultClient(namespace string) (vault.Client, error) {
+	log.Logger().Debugf("creating system vault for namespace %s", namespace)
 	name, err := f.getVaultName(namespace)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


warrenbailey:
this outputs a lot of errors in `jx boot`

hferentschik:
@warrenbailey your first commit is not in conventional commit message format. Maybe you could update the message or imo even better squash the two commits.

pkg/cmd/clients/factory.go
```diff
@@ -416,6 +416,7 @@ func (f *factory) ResetSecretsLocation() {
 
 // CreateSystemVaultClient gets the system vault client for managing the secrets
 func (f *factory) CreateSystemVaultClient(namespace string) (vault.Client, error) {
+	log.Logger().Debugf("Creating system vault for namespace %s", namespace)
```
hferentschik: one thing I keep wondering about is capitalization and punctuation of our logging messages. We seem to be all over the place atm. IMO it would be nice to establish some guidelines. 

(personally, I favour not capitalization for logging and errors)

warrenbailey: I agree, some logging standards would be a good idea. I'm happy to remove the capitalisation if that's what we're going with.

warrenbailey:
@hferentschik I usually squash my commits but forgot, I'll do it now

pkg/cmd/clients/factory.go
hferentschik: > I agree, some logging standards would be a good idea. I'm happy to remove the capitalisation if that's what we're going with.

I guess for now it does not matter. Hopefully, we have a plan forward after the CLI overhaul discussion.

PR #4989, opened 2019-08-06
fix: use the correct identifier to refer to the PipelineResource

jstrachan:
when creating a second PR the `pipelineIdentifier != resourceIdentifier` which then breaks the creation of the pipeline for 2nd PR on the same repo.

e.g. without this fix the 2nd PullRequest fails with message

```
PipelineRun jx/meta-dev1-environment-pr-373-20-9jcwz-1 doesn''t bind
      Pipeline jx/meta-dev1-environment-pr-373-20-9jcwz-1''s PipelineResources correctly:
      PipelineRun bound resources didn''t match Pipeline: Didn''t provide required
      values: [meta-dev1-environment-pr-373-20]
```

### DIFF
--- a/pkg/tekton/syntax/pipeline.go
+++ b/pkg/tekton/syntax/pipeline.go
@@ -1519,7 +1519,7 @@ func (j *ParsedPipeline) GenerateCRDs(pipelineIdentifier string, buildIdentifier
 		Spec: tektonv1alpha1.PipelineSpec{
 			Resources: []tektonv1alpha1.PipelineDeclaredResource{
 				{
-					Name: resourceIdentifier,
+					Name: pipelineIdentifier,
 					Type: tektonv1alpha1.PipelineResourceTypeGit,
 				},
 			},


abayer:
We should be using the same resource identifier for all builds of a branch, but a distinct resource per branch. I don’t think this change actually will fix that properly - where are you calling GenerateCRDs from?

jstrachan:
I tested creating a 2nd PR without this fix - the above error was generated. I added the fix and can generate many PRs on the same repo. 

The issue is related to inconsistency between the name of the git resource on the `Pipeline` versus the name used to create the git resource

wbrefvem:
Ran into this trying to run BDD tests against lighthouse.

hferentschik:
> I tested creating a 2nd PR without this fix - the above error was generated. I added the fix and can generate many PRs on the same repo.

Any chance we can make this testable?

abayer:
#4991 fixes this more appropriately - the issue was that I missed changing what's passed to generate the resource in one place. Whoops.

PR #4991, opened 2019-08-06
fix(tekton): Metapipeline resource name should be consistent.

abayer:
#### Submitter checklist

#### Description

It should be the same in the `PipelineRun` as in the `Pipeline`, generated in the same way as in the generated pipeline. This means we should be passing the resource name to `GenerateSourceRepoResource` here as we do in `generateTektonCRDs`.

/assign @wbrefvem 
/assign @jstrachan 
/assign @hferentschik 

Replaces #4989 

n/a

### DIFF
--- a/pkg/tekton/metapipeline/metapipeline.go
+++ b/pkg/tekton/metapipeline/metapipeline.go
@@ -80,7 +80,7 @@ func CreateMetaPipelineCRDs(params CRDCreationParameters) (*tekton.CRDWrapper, e
 	if revision == "" {
 		revision = params.PullRef.BaseBranch
 	}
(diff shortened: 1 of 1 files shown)


abayer:
And yes, @hferentschik, we should really have a BDD test with multiple PRs.

abayer:
(if we don't already)

abayer:
ooooooh, that's it - this would only show up when using the metapipeline, which is why it didn't appear until now.

jstrachan:
the version stream BDD tests should fail with this with lighthouse to; as there are multiple pipeline runs on the staging repo

PR #4999, opened 2019-08-06
fix: allow the parameters.schema.json to be templated for 'jx boot'

### LINKED ISSUE #4995: allow the `parameters.schema.json` to be templated for different git providers
different git providers have different validation rules for tokens; plus we want to generate different descriptions of how to create a token for different providers. 

So lets allow the boot `parameters.schema.json` to be generated via a go template so we can use a different schema for different kinds of git or kubernetes provider

jstrachan:
so that we can parameterise some schema values based on different requirements like git server/kinds or kubernetes providers

fixes #4995


pkg/cmd/step/create/step_create_values.go
```diff
@@ -194,3 +203,66 @@ func (o *StepCreateValuesOptions) CreateValuesFile() error {
 	}
 	return nil
 }
+
+func (o *StepCreateValuesOptions) templateSchemaFile(schemaFileName string, requirements *config.RequirementsConfig) error {
```
pmuir: can you move this to the `surveyutils` package...

pkg/cmd/step/create/step_create_values.go
```diff
@@ -194,3 +203,66 @@ func (o *StepCreateValuesOptions) CreateValuesFile() error {
...
+	}
+	if requirements.Cluster.GitServer == "" {
+		switch requirements.Cluster.GitKind {
+		case "bitbucketcloud":
+			requirements.Cluster.GitServer = "https://bitbucket.org"
+		case "github":
+			requirements.Cluster.GitServer = "https://github.com"
+		case "gitlab":
+			requirements.Cluster.GitServer = "https://gitlab.com"
+		}
```
pmuir: Is there a way to not do this in code so that we can keep the step generic and not tied to the schema?

jstrachan: its more a bit of validation of the requirements object to ensure we populate git server URLs for the github/gitlab host names if none exists in the `jx-requirements.yml` - not sure how else we could do this really

pmuir: Not sure there is a good way around it tbh.

PR #5001, opened 2019-08-06
fix: making meta pipeline the default in pipelinerunner

### LINKED ISSUE #4992: Make meta pipeline the default
After the Prow changes required for meta pipeline were merged we can now enable meta pipeline per default.

hferentschik:
fixes #4992

### DIFF
--- a/pkg/logs/tekton_logging.go
+++ b/pkg/logs/tekton_logging.go
@@ -218,6 +218,10 @@ func GetRunningBuildLogs(pa *v1.PipelineActivity, buildName string, kubeClient k
 		if !foundLogs {
 			break
 		}
+		pipelineRunNames, err = getPipelineRunNamesForActivity(pa, tektonClient)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get PipelineRun names for activity %s in namespace %s", pa.Name, pa.Namespace)
+		}
 	}
 	if !foundLogs {
 		return errors.New("the build pods for this build have been garbage collected and the log was not found in the long term storage bucket")
--- a/pkg/cmd/controller/pipelinerunner.go
+++ b/pkg/cmd/controller/pipelinerunner.go
@@ -121,7 +121,7 @@ func NewCmdControllerPipelineRunner(commonOpts *opts.CommonOptions) *cobra.Comma
 	cmd.Flags().BoolVarP(&options.SemanticRelease, "semantic-release", "", false, "Enable semantic releases")
 
 	// TODO - temporary flags until meta pipeline is the default
-	cmd.Flags().BoolVarP(&options.UseMetaPipeline, "use-meta-pipeline", "", false, "Uses the meta pipeline to create the pipeline.")
+	cmd.Flags().BoolVarP(&options.UseMetaPipeline, "use-meta-pipeline", "", true, "Uses the meta pipeline to create the pipeline.")
 	cmd.Flags().StringVar(&options.MetaPipelineImage, "meta-pipeline-image", "", "Specify the docker image to use if there is no image specified for a step.")
 	return cmd
 }


hferentschik:
@abayer I guess once this is merged and a new builder image is available we need to create a pull request against the chart repo as well -  https://github.com/jenkins-x-charts/prow/blob/master/prow/values.yaml#L51, right?

abayer:
The chart will be updated automatically.

We need to keep it on the maven builder to deal with https://github.com/jenkins-x-buildpacks/jenkins-x-classic/blob/master/packs/maven/pipeline.yaml#L20

abayer:
Looks like there's an issue with the `jx get build logs` call terminating when the metapipeline finishes, rather than keeping going - I think it's a bug in `GetRunningBuildLogs`...

abayer:
@hferentschik I think maybe I fixed it with https://github.com/jenkins-x/jx/pull/5001/commits/c8e71e3aabae3f0913dc1b786422c936187e990a

abayer:
/lgtm 

That seems to have done the trick, so I'm lgtm'ing and putting it on hold for @hferentschik to review my change. =)

abayer:
/assign @dgozalo 

...assigning Daniel as well since he's familiar with that mess of code.

abayer:
Blech! Failure on the PR build for the `golang-http` quickstart - `failed to run 'jx -test.coverprofile=build/reports/jx_step_git_merge.5440c917-5442-4b11-9d69-cd7e1b1acc07.out step git merge --verbose --baseBranch master --baseSHA 212ca55d2bc5b08296f14c0ec27f1001ded80e42 --sha e1d7493b758bb3d6d25c7fdf9d4fae4e97604041' command in directory '', output: ''`

Could just be transient noise, so I'm rerunning.

abayer:
Oh, and https://github.com/jenkins-x/jx/pull/5004 will let you see the logs for completed builds that still have their pods around. =)

abayer:
Error could maybe have been due to coverage whackiness - `testing: open build/reports/jx_step_git_merge.5440c917-5442-4b11-9d69-cd7e1b1acc07.out: no such file or directory` shows up in the same step as the failure. So once #4997 is merged, this might not be an issue anyway.

abayer:
Oh, and https://github.com/jenkins-x/bdd-jx/pull/56 will make issues like the log not continuing or a problem in the pipeline or metapipeline execution a little more clear.

abayer:
Well, same thing happened again, so it might be a legitimate problem. Rerunning again now that #4997 has merged just to see if that makes a difference.

abayer:
Well, huh - it could still fail at some point later, but without coverage enabled, it got past the point it had been barfing on. I'll keep watching for a while longer before clocking out.

abayer:
yaaaaaaaay passed

hferentschik:
> The chart will be updated automatically.

Well, I thought so as well, but looking at the repo I am not so sure it does. The last update is 7 days old. Unless I am missing something. Let's wait and see.

> We need to keep it on the maven builder to deal with https://github.com/jenkins-x-buildpacks/jenkins-x-classic/blob/master/packs/maven/pipeline.yaml#L20

For the non-meta pipeline yes, but with meta pipeline, the pipelinerunner does not set the version anymore. There is happens in the meta pipeline. All that's needed for pipelinerunner is `jx`. That said, it might make sense to keep the maven builder image so that one can for now still easily switch back to non meta pipeline.
Eventually, it would be nice so to just use the jx builder image and make it minimal as well, aka have some small (eg alpine) base image.

pkg/logs/tekton_logging.go
```diff
@@ -218,6 +218,10 @@ func GetRunningBuildLogs(pa *v1.PipelineActivity, buildName string, kubeClient k
 		if !foundLogs {
 			break
 		}
+		pipelineRunNames, err = getPipelineRunNamesForActivity(pa, tektonClient)
```
hferentschik: So when the meta pipeline starts we get the first set of pipeline runs which are getting streamed and once they are all streamed before we exit the loop we get the names again which will capture the pipeline runs for execution pipeline. This works since the last step of the meta pipeline is the creation of the execution pipeline CRDs. Right?

It is still all very implicit not based on non explicit assumptions. I also wonder whether there is not a potential race condition. I wonder whether we should bake the notion of a meta pipeline better into the code. One thing we could for example to is add a label which differentiates between pipeline run for meta vs execution pipeline. This way, given a pipeline activity one can explicity ask first for the meta pipeline pipeline runs and then for the execution pipeline runs. WDYT? Worth opening an issue? @dgozalo?

hferentschik: See #5007.

PR #5003, opened 2019-08-06
feat(tekton): Add flag to distribute parallel stages across nodes

### LINKED ISSUE #5002: Add flag to automatically attempt to distribute parallel stages across nodes
A problem we hit in Jenkins X's own production builds for the `jenkins-x-builders` repo is that Kaniko, particularly with very large images and layers, can cause problems with the Kubernetes node's container GC, particularly if there are multiple Kaniko pods running on the node at the same time. This is annoying. =)

Something that might help us with this - and that will hopefully be useful in other contexts - would be to put an affinity policy on the `PipelineRun` saying "if possible, put the pods for stages in this run on different nodes from each other". It wouldn't be perfect, but it would definitely help.

Now, this is not saying I'm going to add the ability to configure full affinity policies explicitly on the pipeline - first, the only way I can make this work requires using the `PipelineRun` name (to use in an anti-affinity policy that says "avoid nodes with a pod already on them with the label `tekton.dev/pipelineRun=THIS_PIPELINERUN_NAME`, and since the name is unique per run, you can't exactly hardcode the name into the `jenkins-x.yml`. Second, I'm not sure there's enough value in allowing full affinity policy configuration to be worth the added complexity. So I'm just going to add an option (say, `distributeParallelAcrossNodes: true`) that can be set on a `pipeline`'s `options` (but not a stage, since we only have control over the affinity policy on the `PipelineRun`).

abayer:
#### Submitter checklist

#### Description

See #5002 for more detail.

/assign @hferentschik 

fixes #5002


pkg/tekton/metapipeline/metapipeline.go
```diff
@@ -81,7 +81,7 @@ func CreateMetaPipelineCRDs(params CRDCreationParameters) (*tekton.CRDWrapper, e
 		revision = params.PullRef.BaseBranch
 	}
 	resources := []*pipelineapi.PipelineResource{tekton.GenerateSourceRepoResource(params.ResourceName, &params.GitInfo, revision)}
-	run := tekton.CreatePipelineRun(resources, pipeline.Name, pipeline.APIVersion, labels, params.ServiceAccount, nil, nil)
+	run := tekton.CreatePipelineRun(resources, pipeline.Name, pipeline.APIVersion, labels, params.ServiceAccount, nil, nil, nil)
```
hferentschik: So here we don't need the affinity? Would it not make sense here as well?

pkg/tekton/pipelines.go
```diff
@@ -254,7 +256,8 @@ func CreatePipelineRun(resources []*pipelineapi.PipelineResource,
 			Resources: resourceBindings,
 			Params:    pipelineParams,
 			// TODO: We shouldn't have to set a default timeout in the first place. See https://github.com/tektoncd/pipeline/issues/978
-			Timeout: timeout,
+			Timeout:  timeout,
+			Affinity: affinity,
```
hferentschik: I guess that's safe even if affinity is nil, right?

pkg/tekton/metapipeline/metapipeline.go
hferentschik: Would the aim not also be to separate between say build nodes and app nodes?

hferentschik: I guess issue #5002 provides the answer. This is just about distributing parallel stages. For meta pipeline generation that makes no sense I guess.

hferentschik:
@abayer I guess this needs to be reflected in the docs as well.

PR #5004, opened 2019-08-06
fix: Get logs from pod(s) if possible if BuildLogsURL starts with http

abayer:
#### Submitter checklist

#### Description

Right now, on prod, if you try to run `jx get build logs` on a completed run, you're out of luck: it sees `BuildLogsURL` set, but it's set to an `https` URL (or maybe an `http` URL, I didn't actually check) and so just fails out. If we're not allowing streaming of logs from non-bucket sources in the first place, let's at least just skip `http*` cases and go back to the pods.

/assign @dgozalo 

n/a

### DIFF
--- a/pkg/cmd/get/get_build_logs.go
+++ b/pkg/cmd/get/get_build_logs.go
@@ -430,7 +430,7 @@ func (o *GetBuildLogsOptions) getTektonLogs(kubeClient kubernetes.Interface, tek
 		return true, errors.New("there are no build logs for the supplied filters")
 	}
 
-	if pa.Spec.BuildLogsURL != "" {
+	if pa.Spec.BuildLogsURL != "" && !strings.HasPrefix(pa.Spec.BuildLogsURL, "http") {
 		return false, logs.StreamPipelinePersistentLogs(logWriter, pa.Spec.BuildLogsURL)
 	}
 


pkg/cmd/get/get_build_logs.go
```diff
@@ -430,7 +430,7 @@ func (o *GetBuildLogsOptions) getTektonLogs(kubeClient kubernetes.Interface, tek
 		return true, errors.New("there are no build logs for the supplied filters")
 	}
 
-	if pa.Spec.BuildLogsURL != "" {
+	if pa.Spec.BuildLogsURL != "" && !strings.HasPrefix(pa.Spec.BuildLogsURL, "http") {
```
hferentschik: Looking at `logs.StreamPipelinePersistentLogs` I am wondering why we cannot just get the content of the URL in the case of http/https? Are these URLs not pointing to actual logs? My understanding was that if no bucket is provided, logs can be stashed into a GitHub repo. At least for some time I seem to recall looking at logs in a GitHub repo. Assuming this are the kind of HTTP URLs we are talking about here, why don't we just read their content?

hferentschik: Here is what I mean https://github.com/jenkins-x/environment-tekton-mole-dev/tree/gh-pages/jenkins-x/logs.

abayer: As I mentioned offline, https://github.com/jenkins-x/jx/blame/0548010f5597c7e92bbdb6196454878d894cb38f/pkg/logs/tekton_logging.go#L278-L281 is relevant here - @dgozalo can speak more to this.

dgozalo: My understanding is that we were moving the storage of persistent logs to buckets at all times, meaning we wouldn't be supporting the git storage anymore. The case in production right now is that the storage configuration gets wiped and it goes back to git storage, but this shouldn't really happen.

The thing is, if buckets storage is not enabled, there shouldn't be any logs sent to GitHub.

Maybe we need to enable getting the logs from GH Pages as a fallback?

abayer:
Think this gets obsoleted by #5021 and changes @dgozalo is gonna make.

pkg/cmd/get/get_build_logs.go
hferentschik: > Maybe we need to enable getting the logs from GH Pages as a fallback?

+1

dgozalo:
I'm pretty sure we can discard this one.

Fyi: https://github.com/jenkins-x/jx/pull/5030

abayer:
Yup!

PR #5006, opened 2019-08-07
fix: jx delete preview returns error

macox:

#### Submitter checklist

The `jx delete preview` cmd is returning the error `error: no environment with name 'x' found` since this PR https://github.com/jenkins-x/jx/pull/4382.

### DIFF
--- a/pkg/cmd/preview/preview.go
+++ b/pkg/cmd/preview/preview.go
@@ -865,10 +865,11 @@ func (o *PreviewOptions) DefaultValues(ns string, warnMissingName bool) error {
 			}
 		}
 	}
-	o.Name = naming.ToValidName(o.Name)
 	if o.Name == "" {
 		return fmt.Errorf("No name could be defaulted for the Preview Environment. Please supply one!")
 	}
+	o.Name = naming.ToValidName(o.Name)
+
 	if o.Namespace == "" {
 		prefix := ns + "-"
 		if len(prefix) > 63 {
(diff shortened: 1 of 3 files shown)


pkg/cmd/preview/preview.go
```diff
@@ -865,10 +865,11 @@ func (o *PreviewOptions) DefaultValues(ns string, warnMissingName bool) error {
 			}
 		}
 	}
-	o.Name = naming.ToValidName(o.Name)
 	if o.Name == "" {
 		return fmt.Errorf("No name could be defaulted for the Preview Environment. Please supply one!")
 	}
+	o.Name = naming.ToValidName(o.Name)
```
hferentschik: Is moving this code not just patching up an underlying problem in `naming.ToValidName`? The semantics used to be that `ToValidName` of the empty string is the empty string, but now it is `x`. Whether the original semantics is correct might be arguable, but returning 'x' seems even worse. 

So imo to fix the issue we should change `naming.ToValidName` and handle empty string explicitly. While on it we could add a test for this case and update the docs.

pkg/kube/naming/names.go
```diff
@@ -35,6 +35,9 @@ func ToValidNameTruncated(name string, maxLength int) string {
 }
 
 func toValidName(name string, allowDots bool, maxLength int) string {
+	if name == "" {
```
hferentschik: +1

Issue #4943, opened 2019-08-01
The step Validate Git fails on the master build of a fresh jx boot dev environment


dgozalo:
### Summary 

After successfully creating a fresh jx boot cluster in GitOps mode, the dev environment will execute its master branch build.
This build fails on the step `Validate Git` due to: 

`error: No Git user.name is defined. Please run the command: git config --global --add user.name "MyName"` 

### Steps to reproduce the behavior
Create a new cluster, add an app or quickstart, see the master build of the dev environment.

### Expected behavior
The step shouldn't fail. 

### Jenkins type

- [X] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

cagiti:
@dgozalo is this only an issue for clusters created with `gitops: true`?

dgozalo:
@cagiti yes, only in GitOps mode we execute the dev environment pipelines. I'll update the initial description.

macox:
Now that the possibility of failure has been removed it would be good to configure the correct git user (bot user) when the pipeline is run in cluster, and not default to jenkins-x-bot

garethjevans:
Is this still the case @macox? am hoping that once the fix has propagated through the builders and we get jenkins-x-versions releasing properly that this will go away

PR #5014, opened 2019-08-07
feat: allow gitconfig to be configured from the environment

garethjevans:
Uses the env variables GIT_COMMITER_NAME / GIT_COMMITER_EMAIL


cagiti:
contributes towards #4943

PR #5015, opened 2019-08-07
[wip] chore: changing GINKGO_ARGS from -v to --progress

hferentschik:
playing with the GINKGO output options

### DIFF
--- a/jx/scripts/ci.sh
+++ b/jx/scripts/ci.sh
@@ -15,7 +15,7 @@ export REPORTS_DIR="${BASE_WORKSPACE}/build/reports"
 # for BDD tests
 export GIT_PROVIDER_URL="https://github.beescloud.com"
 export GHE_TOKEN="$GHE_CREDS_PSW"
-export GINKGO_ARGS="-v"
+export GINKGO_ARGS="--progress --slowSpecThreshold=900"
 
 export JX_DISABLE_DELETE_APP="true"
 export JX_DISABLE_DELETE_REPO="true"


jx/scripts/ci.sh
```diff
@@ -15,7 +15,7 @@ export REPORTS_DIR="${BASE_WORKSPACE}/build/reports"
 # for BDD tests
 export GIT_PROVIDER_URL="https://github.beescloud.com"
 export GHE_TOKEN="$GHE_CREDS_PSW"
-export GINKGO_ARGS="-v"
+export GINKGO_ARGS="--progress"
```
abayer: You'll want to do this in `jx/bdd/tekton/ci.sh` as well.

abayer:
So in the bdd run, where the flag is changed, I'm now just seeing 
```
GO111MODULE=on ginkgo --progress -r test/spring --slowSpecThreshold=50000
Running Suite: Jenkins X E2E tests: create_spring_application
=============================================================
Random Seed: 1565189889
Will run 1 of 1 specs
```
for a while now with no further output. I wonder if there's some way we can get just the `By(fmt.Sprintf("finding the activity for %s in %v", activityKey, activities), ...) {` message to show up without also getting the `jx get build log` output too?

hferentschik:
> for a while now with no further output. I wonder if there's some way we can get just the By(fmt.Sprintf("finding the activity for %s in %v", activityKey, activities), ...) { message to show up without also getting the jx get build log output too?

Right, that's what I would be after as well. It seems --progress also writes to GinkgoWriter and there you either have the default to be silent and only dump on failure or to log everything. Hmmm

abayer:
whoops, wrong pr.

PR #5021, opened 2019-08-07
fix: Fix gh-pages for log storage in a couple ways

abayer:
#### Submitter checklist

#### Description

A few things were just completely borked in using gh-pages for log storage for a while now. If the repo didn't already have a gh-pages branch, we tried to reclone without deleting the existing attempted clone, which was bad, and then when there was a gh-pages, we tried to push `HEAD` without specifying that we actually want to push it to `gh-pages`. This fixes both of those things.

/assign @dgozalo 
/assign @pmuir 

n/a

### DIFF
--- a/pkg/collector/git_collector.go
+++ b/pkg/collector/git_collector.go
@@ -154,7 +154,7 @@ func (c *GitCollector) CollectData(data []byte, outputPath string) (string, erro
 	if err != nil {
 		return u, err
 	}
-	err = gitClient.Push(ghPagesDir, "origin", false, false, "HEAD")
+	err = gitClient.Push(ghPagesDir, "origin", false, false, "HEAD:"+c.gitBranch)
 	return u, err
 }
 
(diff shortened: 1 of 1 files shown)


abayer:
fyi, I changed `jenkins-x-controllerbuild` on the prod cluster to use the image from this PR and it's working very, very smoothly. Once this is merged/released, I'll update `environment-tekton-mole-dev`.

pkg/collector/git_collector.go
```diff
@@ -154,7 +154,7 @@ func (c *GitCollector) CollectData(data []byte, outputPath string) (string, erro
 	if err != nil {
 		return u, err
 	}
-	err = gitClient.Push(ghPagesDir, "origin", false, false, "HEAD")
+	err = gitClient.Push(ghPagesDir, "origin", false, false, "HEAD:"+c.gitBranch)
```
hferentschik: Provided you are on the gh-pages branch `git push HEAD` is just a shortcut for `git push gh-pages:gh-pages`. Not sure whether this makes a difference here unless there is more to it than meets the eye.

PR #5022, opened 2019-08-07
fix: Push an updated PR before updating its title and body

### LINKED ISSUE #4785: master sometimes builds twice on merge
This isn’t consistent, but a nontrivial portion of the time, I’ll see `jenkins-x-builders` kick off two builds after a PR is merged. I think I’ve seen this for `jx` as well, but don’t quote me on that. This creates issues not just in terms of resources used, but if the two builds both succeed, they overwrite each other in the downstream PRs, or even worse, as in the case of https://github.com/jenkins-x/jenkins-x-platform/pull/5776 until I fixed it by hand, you can even end up with a mismatch of versions for different builders. This is not a good thing.

abayer:
#### Submitter checklist

#### Description

~~For whatever reason, GitHub seems to get goofy when you take an existing PR, update its title and body, and _then_ immediately push a new commit to the PR. That ordering results in GitHub sending two `synchronize` events for the exact same commit, triggering two builds in Prow, etc. Which isn't ideal.~~

~~But if we flip the ordering to push first and _then_ update the title and body, only one `synchronize` is sent. So let's do that thing.~~

UPDATE: @pmuir pointed out that we actually had two pushes, which is probably the root cause here. So let's get rid of the redundant extra one instead of just moving things around.

I've tested this manually by messing around with https://github.com/jenkins-x/jenkins-x-platform/pull/5835, calling the exact same `jx step create pr chart ...` that the `jx` release does, but with bogus version numbers, and it _works_. =)

/assign @pmuir 

fixes #4785

### DIFF
--- a/pkg/gits/helpers.go
+++ b/pkg/gits/helpers.go
@@ -345,10 +345,6 @@ func PushRepoAndCreatePullRequest(dir string, upstreamRepo *GitRepository, forkR
 		if err != nil {
 			return nil, errors.Wrapf(err, "updating pull request %s", existingPr.URL)
 		}
-		err = gitter.Push(dir, forkPushURL, true, false, fmt.Sprintf("%s:%s", "HEAD", prDetails.BranchName))
-		if err != nil {
-			return nil, errors.WithStack(err)
-		}
 		log.Logger().Infof("Updated Pull Request: %s", util.ColorInfo(pr.URL))
 	} else {
 		gha.Head = headPrefix + prDetails.BranchName


pkg/gits/helpers.go
```diff
@@ -341,14 +341,14 @@ func PushRepoAndCreatePullRequest(dir string, upstreamRepo *GitRepository, forkR
...
 		var err error
-		pr, err = provider.UpdatePullRequest(gha, *existingPr.Number)
-		if err != nil {
-			return nil, errors.Wrapf(err, "updating pull request %s", existingPr.URL)
-		}
 		err = gitter.Push(dir, forkPushURL, true, false, fmt.Sprintf("%s:%s", "HEAD", prDetails.BranchName))
 		if err != nil {
 			return nil, errors.WithStack(err)
 		}
+		pr, err = provider.UpdatePullRequest(gha, *existingPr.Number)
```
hferentschik: might be worth to add a comment here that order matters

pmuir: I've actually just spotted that we push on 311-315 and that this second push shouldn't even be there. Do you want to try just deleting it @abayer

abayer: Well, yeesh. That’s hilarious. Updating.

abayer: Done.

PR #5045, opened 2019-08-12
fix: switch jx step create pr versions to new logic

pmuir:
This commit changes jx step create pr to use the
new logic built for creating PRs

It also moves the version stream stuff to it’s own package

#### Submitter checklist

fixes #


pkg/versionstream/resolver.go
```diff
@@ -0,0 +1,74 @@
+package versionstream
```
dgozalo: Love this feature btw. It makes using the versions stream much cleaner.

Issue #4993, opened 2019-08-06
Remove legacy code from 'step create task'


hferentschik:
Once meta pipeline is enabled and tested (#4992), we can review the existing code and clean it up by removing the pre meta pipeline code which we kept to be able to have both approaches during development.

In particular this is code in `step_create_task.go` as well as the pipelinerunner.

hferentschik:
In fact I think we should delete _step_create_task_ and just provide a client/API to execute this functionality similar to #4971.

There is really no good reason to have a command for this.

Issue #4581, opened 2019-07-04
Separate logging from required CLI output to the user


hferentschik:
In some commands, we use info level logging for expected/required command output. We configure our logger in case of a TTY to not add any extra logging details (eg the log level, etc), to make it more look like default user output.

Logging and UI in-/output are two different concerns though and we should separate them. Where user output it expected it should be written to stdout, preferably from within `Run`. In other cases it might make sense to pass a handle to some sub-system which can be used to write user output.

hferentschik:
This issue is tightly related to #4581. IMO `--verbose` is not the same as changing the log level. It increases the default output to the user, making it more verbose. This is not the same as setting a log level. Logging and in- and output of a command are separate things which we merged together, but treating the info logger as default writer for user stdout.

PR #5051, opened 2019-08-12
feat: introducing a meta pipeline client

### LINKED ISSUE #4971: Define interface for meta pipeline generation

hferentschik:
fixes #4971, https://github.com/jenkins-x/lighthouse/issues/61

- adding metadata.Client and metadata.ClientFactory for controlling meta pipeline creation
- removing step_create_meta_pipeline
- adjusting code to use new meta pipeline client instead of 'step create pipeline'
- creating jxfactory package to reduce import size when using kube, jx and tekton clients


pkg/tekton/metapipeline/client.go
```diff
@@ -0,0 +1,24 @@
+package metapipeline
+
+import (
+	"github.com/jenkins-x/jx/pkg/kube"
+	"github.com/jenkins-x/jx/pkg/tekton"
+)
+
+// Client defines the interface for meta pipeline creation and application.
+type Client interface {
```
hferentschik: @ccojocar, this would be my take on a meta pipeline client. You see it in action [here](https://github.com/jenkins-x/jx/pull/5051/files#diff-cfadf679d93b353852ef9097113f3e53R404)

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -0,0 +1,295 @@
...
+	defaultImage     string
+	serviceAccount   string
+	versionStreamURL string
+	versionStreamRef string
+}
+
+// NewMetaPipelineClient creates a new client for the creation and application of meta pipelines.
+// The responsibility of the meta pipeline is to prepare the execution pipeline and to allow Apps to contribute
+// the this execution pipeline.
+func NewMetaPipelineClient() (*ClientFactory, error) {
```
hferentschik: @ccojocar default impl of the interface. 

The `jx step create pipeline` is removed. So no command for this anymore.

pkg/cmd/controller/pipelinerunner.go
```diff
@@ -144,6 +145,51 @@ func (o *PipelineRunnerOptions) Run() error {
...
+	} else {
+		if !o.NoGitCredentialsInit {
+			err := o.InitGitConfigAndUser()
+			if err != nil {
+				return err
+			}
+
+			err = o.stepGitCredentials()
+			if err != nil {
+				logger.Warn(err.Error())
```
ccojocar: Shouldn't this fail here instead of just logging the error and continue?

pkg/cmd/controller/pipelinerunner.go
```diff
@@ -144,6 +145,51 @@ func (o *PipelineRunnerOptions) Run() error {
...
+			}
+		}
+	}
+	return nil
+}
+
+func (o *PipelineRunnerOptions) setLogLevel() error {
+	level := os.Getenv("JX_LOG_LEVEL")
```
ccojocar: Shouldn't this take into account also some CLI log level options? At least we should document this env variable.

pkg/cmd/controller/pipelinerunner.go
```diff
@@ -168,9 +214,12 @@ func (o *PipelineRunnerOptions) startWorkers(ctx context.Context, wg *sync.WaitG
 		for {
 			select {
 			case <-ctx.Done():
-				logger.Info("Shutting down HTTP server")
+				logger.Info("shutting down HTTP server")
 				ctx, cancel := context.WithTimeout(ctx, shutdownTimeout*time.Second)
 				_ = srv.Shutdown(ctx)
+				if o.metaPipelineClient != nil {
+					_ = o.metaPipelineClient.Close()
```
ccojocar: Maybe we should log this error in debug.

pkg/cmd/start/start_pipeline.go
```diff
@@ -401,3 +422,30 @@ func (o *StartPipelineOptions) startJenkinsJob(name string) error {
 		time.Sleep(time.Second)
 	}
 }
+
+func (o *StartPipelineOptions) prowToMetaPipelinePullRef(sourceURL string, prowPullRef *prow.PullRefs) metapipeline.PullRef {
```
ccojocar: These methods seems to be duplicated in pipelinerunner. Shouldn't we move them in `prow` package since they seems related to it?

pkg/jxfactory/factory.go
```diff
@@ -0,0 +1,189 @@
...
+	cfg, err := f.CreateKubeConfig()
+	if err != nil {
+		return nil, "", err
+	}
+	client, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, "", err
+	}
+	if client == nil {
+		return nil, "", fmt.Errorf("Failed to create Kubernetes Client")
```
ccojocar: error are not capitalised

pkg/tekton/crds.go
```diff
@@ -124,6 +125,30 @@ func (crds *CRDWrapper) ObjectReferences() []kube.ObjectReference {
 	return resources
 }
 
+func (crds *CRDWrapper) String() string {
+	var allResources []interface{}
+	allResources = append(allResources, crds.pipeline)
+	for _, task := range crds.tasks {
+		allResources = append(allResources, task)
```
ccojocar: You can replace the entire `for` with `append(allResources, task...).

pkg/tekton/crds.go
```diff
@@ -124,6 +125,30 @@ func (crds *CRDWrapper) ObjectReferences() []kube.ObjectReference {
...
+	}
+	for _, resource := range crds.resources {
+		allResources = append(allResources, resource)
```
ccojocar: ditto

pkg/tekton/metapipeline/client.go
```diff
@@ -0,0 +1,24 @@
...
+	Create(pullRef PullRef, pipelineType PipelineKind, context string, envs map[string]string, labels map[string]string) (kube.PromoteStepActivityKey, tekton.CRDWrapper, error)
+
+	// Apply takes the given CRDs for processing, usually applying them to the cluster.
+	Apply(pipelineActivity kube.PromoteStepActivityKey, crds tekton.CRDWrapper) error
+
+	// Close cleans up the resources use by this client.
+	Close() error
+
+	//DefaultImage returns the default image used for pipeline tasks if no other image is specified.
+	DefaultImage() string
```
ccojocar: Is it expected that these are hard-coded? I don't see how users of this client can configure them over the interface?

pkg/tekton/metapipeline/client.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"github.com/jenkins-x/jx/pkg/kube"
+	"github.com/jenkins-x/jx/pkg/tekton"
+)
+
+// Client defines the interface for meta pipeline creation and application.
+type Client interface {
+	// Create creates the Tekton CRDs needed for executing the pipeline as defined by the input parameters
+	Create(pullRef PullRef, pipelineType PipelineKind, context string, envs map[string]string, labels map[string]string) (kube.PromoteStepActivityKey, tekton.CRDWrapper, error)
```
ccojocar: I would extract the list of arguments into an `options` struct. This will allow us to extend the arguments easier in future.

pkg/tekton/metapipeline/client.go
```diff
@@ -0,0 +1,24 @@
...
+	Apply(pipelineActivity kube.PromoteStepActivityKey, crds tekton.CRDWrapper) error
+
+	// Close cleans up the resources use by this client.
+	Close() error
+
+	//DefaultImage returns the default image used for pipeline tasks if no other image is specified.
+	DefaultImage() string
+
+	// ServiceAccount returns the service account under which to execute the pipeline.
+	ServiceAccount() string
```
ccojocar: Consumer of this pipeline might want to control which `service account` and `namespace` is used. I would make this available in the create `options`.

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -0,0 +1,295 @@
...
+	retryDuration         = time.Second * 30
+	defaultCheckoutDir    = "source"
+)
+
+var (
+	logger = log.Logger().WithFields(logrus.Fields{"component": "meta-pipeline-client"})
+)
+
+// ClientFactory implements the interface methods to create and apply the meta pipeline.
+type ClientFactory struct {
```
ccojocar: This name is wired to me. `ClientFactory` is an implementation of `Client` interface.  Does it really need to be public, and exposed form the package? We can just return the `Client` interface from the constructor and you can name this just `client`.

pkg/tekton/metapipeline/clientfactory.go
ccojocar: I would return the `Client` instead to encapsulate the effective implementation.

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -0,0 +1,295 @@
...
+	tektonClient, jxClient, kubeClient, ns, err := getClientsAndNamespace()
+	if err != nil {
+		return nil, err
+	}
+
+	return NewMetaPipelineClientWithClientsAndNamespace(jxClient, tektonClient, kubeClient, ns)
+}
+
+// NewMetaPipelineClientWithClientsAndNamespace creates a new client for the creation and application of meta pipelines using the specified parameters.
+func NewMetaPipelineClientWithClientsAndNamespace(jxClient versioned.Interface, tektonClient tektonclient.Interface, kubeClient kubernetes.Interface, ns string) (*ClientFactory, error) {
```
ccojocar: ditto

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -0,0 +1,295 @@
...
+
+	// resourceName is shared across all builds of a branch, while the pipelineName is unique for each build.
+	resourceName := tekton.PipelineResourceNameFromGitInfo(gitInfo, branchIdentifier, context, tekton.MetaPipeline, nil, "")
+	pipelineName := tekton.PipelineResourceNameFromGitInfo(gitInfo, branchIdentifier, context, tekton.MetaPipeline, c.tektonClient, c.ns)
+	buildNumber, err := tekton.GenerateNextBuildNumber(c.tektonClient, c.jxClient, c.ns, gitInfo, branchIdentifier, retryDuration, context)
+	if err != nil {
+		return kube.PromoteStepActivityKey{}, tekton.CRDWrapper{}, errors.Wrap(err, "unable to determine next build number")
+	}
+
+	logger.WithField("repo", gitInfo.URL).WithField("buildNumber", buildNumber).Info("creating meta pipeline CRDs ")
```
ccojocar: shouldn't be this `debug`?

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -0,0 +1,295 @@
...
+	return *pipelineActivity, *tektonCRDs, nil
+}
+
+// Apply takes the given CRDs to process them, usually applying them to the cluster.
+func (c *ClientFactory) Apply(pipelineActivity kube.PromoteStepActivityKey, crds tekton.CRDWrapper) error {
+	err := tekton.ApplyPipeline(c.jxClient, c.tektonClient, &crds, c.ns, &pipelineActivity)
+	if err != nil {
+		return errors.Wrapf(err, "failed to apply Tekton CRDs")
+	}
+	logger.WithField("pipeline", crds.PipelineRun().Name).Info("applied tekton CRDs")
```
ccojocar: `debug`

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -0,0 +1,295 @@
...
+		return nil, err
+	}
+
+	return podTemplates, nil
+}
+
+func (c *ClientFactory) determineBranchIdentifier(pipelineType PipelineKind, pullRef PullRef) (string, error) {
+	var branch string
+	switch pipelineType {
+	case ReleasePipeline:
```
ccojocar: You can remove the outer brackets in every case statement.

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -0,0 +1,295 @@
...
+func (c *ClientFactory) cloneVersionStream(url string, ref string) (string, error) {
+	dir, err := ioutil.TempDir("", "jx-version-repo-")
+	if err != nil {
+		return "", errors.Wrap(err, "unable to create temp dir for version stream")
+	}
+
+	logger.Infof("cloning version stream url: %s ref: %s into %s", url, ref, dir)
+
+	// Not using GitCLi Clone/ShallowClone atm, since it does not work with tags (HF)
+	args := []string{"clone", "--depth", "1", "--branch", ref, url, "."}
```
ccojocar: I think this is introducing an indirect dependency to `git` CLI into the consumers of this client. :( This is not nice if I want to have a pure go process.

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -0,0 +1,295 @@
...
+	resourceName := tekton.PipelineResourceNameFromGitInfo(gitInfo, branchIdentifier, context, tekton.MetaPipeline, nil, "")
+	pipelineName := tekton.PipelineResourceNameFromGitInfo(gitInfo, branchIdentifier, context, tekton.MetaPipeline, c.tektonClient, c.ns)
+	buildNumber, err := tekton.GenerateNextBuildNumber(c.tektonClient, c.jxClient, c.ns, gitInfo, branchIdentifier, retryDuration, context)
+	if err != nil {
+		return kube.PromoteStepActivityKey{}, tekton.CRDWrapper{}, errors.Wrap(err, "unable to determine next build number")
+	}
+
+	logger.WithField("repo", gitInfo.URL).WithField("buildNumber", buildNumber).Info("creating meta pipeline CRDs ")
+
+	extendingApps, err := GetExtendingApps(c.jxClient, c.ns)
```
ccojocar: I suppose the `GetExtendingApps` method is defined in the `metapipeline` package. Does it need to be publicly exposed from the package?

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -0,0 +1,295 @@
...
+		PodTemplates:     podTemplates,
+		ServiceAccount:   c.serviceAccount,
+		Labels:           labels,
+		EnvVars:          envs,
+		DefaultImage:     c.defaultImage,
+		Apps:             extendingApps,
+		VersionsDir:      c.versionDir,
+		GitInfo:          *gitInfo,
+	}
+	tektonCRDs, err := CreateMetaPipelineCRDs(crdCreationParams)
```
ccojocar: ditto

pkg/cmd/clients/factory.go
```diff
@@ -36,31 +32,29 @@ import (
...
 	Batch bool
 
-	kubeConfig      kube.Kuber
-	impersonateUser string
-	bearerToken     string
-	secretLocation  secrets.SecretLocation
-	offline         bool
+	secretLocation secrets.SecretLocation
+	offline        bool
+	jxFactory      jxfactory.Factory
```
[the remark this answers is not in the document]
hferentschik: ahh, good point. Let me try this.

pkg/cmd/controller/pipelinerunner.go
```diff
@@ -144,6 +145,51 @@ func (o *PipelineRunnerOptions) Run() error {
 	return nil
 }
 
+func (o *PipelineRunnerOptions) configure() error {
+	err := o.setLogLevel()
+	if err != nil {
+		return errors.Wrap(err, "unable to set log level")
+	}
+
+	logger.Infof("meta pipeline enabled: %t", o.UseMetaPipeline)
```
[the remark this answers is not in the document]
hferentschik: Why? This comes out once at startup of the controller. I had several times the case where I was not sure in which mode the pipelinerunner is running and I had to refer back to the deployment config. That's actually the reason I added this log statement. 

That said, once we do #4993, the pipelinerunner would always run in meta mode in which case this lines would not be required. For now though, I'd prefer info.

pkg/cmd/controller/pipelinerunner.go
```diff
@@ -144,6 +145,51 @@ func (o *PipelineRunnerOptions) Run() error {
...
+	if o.UseMetaPipeline {
+		logger.Info("creating meta pipeline client")
```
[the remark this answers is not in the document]
hferentschik: same as above. I also find when it comes to logging, more usually does not hurt. Definitely helps when troubleshooting issue based on logs. IMO the only place to really watch out for is logging in loops.

pkg/cmd/controller/pipelinerunner.go
hferentschik: +1

pkg/cmd/controller/pipelinerunner.go
hferentschik: Thing is with the CLI options I cannot fully control the log level anymore. This is captured in issue #5080 and #4581. 

So yes, I kind of behind the back introduce `JX_LOG_LEVEL`. Preferably this should be done as part of #5080 and be handled in the log package. 

If we agree that it makes sense to have this env variable, I can remove this from this PR and create a new for for #5080. WDYT?

pkg/cmd/controller/pipelinerunner.go
hferentschik: makes sense

pkg/cmd/start/start_pipeline.go
hferentschik: yes, I was thinking about this. I went for duplication for now, for two reasons. I did not want to depend on another package (even though the prow package is probably safe) and secondly because eventually we will move away from Prow. Once we fully embrace lighthouse it does not make sense anymore to use a Prow pull ref. 

I can take another look.

pkg/tekton/metapipeline/client.go
hferentschik: My thinking was that even though we need the service account and default image, they are more client specific options, hence I did not want to add it to single invocation of _Create_. If we introduce an options struct for _Create_ this might be different. I'd probably add them there.

pkg/tekton/metapipeline/clientfactory.go
hferentschik: It was related to the default image and service account. And yes, we should just return Client. Let me sort this out.

pkg/tekton/metapipeline/clientfactory.go
hferentschik: I'd go for info. What are your concerns. Logs getting too big? So you would suggest that all the controller logs is that it started and then nothing more, except there is an error or warning? Just trying to understand where you are coming from?

pkg/tekton/metapipeline/clientfactory.go
hferentschik: probably not, let me check

pkg/tekton/metapipeline/clientfactory.go
hferentschik: true, but that's also the case if I use gits.ShallowClone. It also shells out a git command execution (or in fact multiple). 

What's the alternative? We don't have a Go impl of git, right? I agree in a perfect world we would just have a single go process.

hferentschik:
@ccojocar, thanks for the great review. I'll push an update shortly. T

pkg/cmd/controller/pipelinerunner.go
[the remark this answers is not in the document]
ccojocar: Agree

pkg/cmd/controller/pipelinerunner.go
ccojocar: sgm!

pkg/cmd/start/start_pipeline.go
ccojocar: fair enough for now. Thanks!

pkg/tekton/metapipeline/clientfactory.go
ccojocar: This is more my contemplation because I was thinking that we should get rid of this dependency in `Lighthouse` to reduce the docker image size.

There is `go-git` but yeah we shouldn't start to introduce a new dependency just for this.

pkg/tekton/metapipeline/clientfactory.go
ccojocar: It would be great if we do not export the struct from the package by just using `client` without capital letter. 

I believe this is easy achievable if we return the interface in the constructor as suggested bellow. Also it is a static type check which ensure at compile time that the `client` type really implements the `Client` interface.

pkg/cmd/clients/factory.go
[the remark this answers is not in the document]
hferentschik: Hmm, on second thought, does embedding really help. If I do this:

```go
type factory struct {
	jxfactory.Factory
	Batch bool
	secretLocation secrets.SecretLocation
	offline        bool
}
```

I am just saying that the `factory` defines the methods of `jxfactory.Factory`, I still need to define implementations. So I could do:

```
type factory struct {
	jxfactory.Factory

	Batch bool

	secretLocation secrets.SecretLocation
	offline        bool
	jxFactory      jxfactory.Factory
}
```

This way it is guaranteed that I am not miss to any of the  `jxfactory.Factory` functions in `factory`, but I still use `jxFactory` to delegate these functions.

Or do you have something else in mind?

pkg/cmd/start/start_pipeline.go
```diff
@@ -202,37 +201,59 @@ func (o *StartPipelineOptions) createMetaPipeline(jobname string) error {
...
 	}
 
-	po := create.StepCreatePipelineOptions{
-		SourceURL:    sourceURL,
-		Job:          jobname,
-		PullRefs:     pullRefs,
-		Context:      o.Context,
-		CustomLabels: o.CustomLabels,
-		CustomEnvs:   o.CustomEnvs,
+	log.Logger().Info("creating meta pipeline client")
```
hferentschik: actually, here is should be Debug, since we are in a user facing command and info logging is equal to user feedback

pkg/cmd/start/start_pipeline.go
hferentschik: Actually I could remove the duplication. I don't really need to go via a Prow pull ref, I can create the meta pipeline one directly.

pkg/tekton/crds.go
hferentschik: Now I remember, I tried this and it did not work. I was surprised as well. It gives you a compile error:
```
pkg/tekton/crds.go:131:23: cannot use crds.tasks (type []*"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1".Task) as type []interface {} in append
```

pkg/tekton/metapipeline/clientfactory.go
hferentschik: done

pkg/tekton/metapipeline/clientfactory.go
hferentschik: ahh, I guess now I understand why you want to use Debug logging. This is also called from `jx start pipeline` and from the CLI  I don't really want this to be visible per default.  

I only had the pipeline runner in mind as client when I wrote this where I would expect these type of things being logged per default. 

I'll change all logging in this file to Debug and then we just need to make sure we start the pipelinerunner per default in verbose mode.

pkg/tekton/metapipeline/clientfactory.go
hferentschik: made private

pkg/tekton/metapipeline/clientfactory.go
hferentschik: I like the idea to do rely less on external tools in general. However, trying to get away from calling git is definitely more than we can chew off here. 

That said, this is just another point pro `pipelinerunner` imo. If we were to keep `pipelinerunner` which would clone the version stream and I find it more reasonable that our direct controller images have git backed in. If lighthouse where to make an HTTP request to the `pipelinerunner` you would not have to care about this.

hferentschik:
@ccojocar I pushed an update to the PR

pkg/tekton/metapipeline/clientfactory.go
hferentschik: Makes me also wonder a bit, why version stream is something which we clone a repo for. I would the version stream more envision as something living in the cluster and which I can query for versions of things. Whether this in cluster thing itself gets its state/updates from a repo is a different question, but I find the needs to clone odd.

hferentschik:
@ccojocar anything else you want me to look at?

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -0,0 +1,280 @@
...
+	defaultCheckoutDir = "source"
+)
+
+var (
+	logger = log.Logger().WithFields(logrus.Fields{"component": "meta-pipeline-client"})
+)
+
+// clientFactory implements the interface methods to create and apply the meta pipeline.
+type clientFactory struct {
+	Client
```
ccojocar: I don't get why the `Client` interface needs to be embedded into `clientFactory` type. The type is implementing already this interface. 

This pattern is a form of composition in go (https://rakyll.org/typesystem/). Normally, it is used to embed methods of other types into a type. 

In this case you are save to remove it.

hferentschik: I removed it

hferentschik: Prior to when the constructor was returning a concrete implementation instead of the interface adding `Client` would ensure that you implement all methods. But this is moot now anyway since the constructor returns the interface.

hferentschik:
@ccojocar fyi, I rebased this PR. Needs a new lgtm.

abayer:
The integration test failure was a flaky test, which will be fixed by #5177, and I've just pushed https://github.com/jenkins-x/jx/pull/5051/commits/629c880cce2a0116602e36698e169306b8718ec8 to fix the Tekton test failure, which was completely legit. =)

hferentschik:
I guess I might need to rebase this on #5172 to get a clean run eventually.

abayer:
It's in the merge pool, so we'll see...

abayer:
dangit, currently running one failed with a scheduler test burp. Rerunning integration.

abayer:
...and barring a surprise, #5172 is going to get merged before this gets done anyway.

hferentschik:
here we go, another rebase.

abayer:
argh, it was failing because it was missing my fix to actually specify the metapipeline client. Cherrypicked it over and pushed.

abayer:
yaaaaaaay merged! cc @wbrefvem as an fyi.

hferentschik:
@abayer, thanks for merging and fixing my merge faux pas in pipelinerunner.

PR #5056, opened 2019-08-12
feat: refactor of LogWriters and skip logging steps after failure

### LINKED ISSUE #4936: Build logs are currently confusing
### Summary 
The new log format is confusing and very difficult for developers to interpret accurately. There appear to be a number of issues. For example, in the following log we can see several problems:
https://gist.github.com/terrycoda/0c5652fa05a9e3069e377b410e3b5b76

Firstly, the ultimate problem that terminates the build is here: https://gist.github.com/terrycoda/0c5652fa05a9e3069e377b410e3b5b76#file-gistfile1-txt-L53 however the log contains spurious output that implies that 8 subsequent steps executed but produced no log output. Logging should have terminated at line 54.

The message `WARNING: pod project-coda-coda-han-0004-trai-bj6bj-14-build-2kh4n-pod-131710 has failed` is not a warning, it's an ERROR as execution is terminated.

That, combined with the issue that log monitoring with `jx get build logs` does not terminate, implies that subsequent pipeline stages are still in progress, but are silent.

Even more confusingly, these warnings start to appear at line 33: https://gist.github.com/terrycoda/0c5652fa05a9e3069e377b410e3b5b76#file-gistfile1-txt-L33
This is prior to the output from several earlier steps in the pipeline being displayed, making the actual point of failure indeterminate.

Worse, these warnings are colorized and the actual errors are not.

As a result, it is almost impossible for anyone who does not understand how the pipelines are implemented under the hood to work out where in the pipeline their specific error occurred. I would suggest that this should not be required knowledge for the average user of the platform.

We should probably remove all but one ERROR showing the failure of the build and this should occur after the actual point of failure. All subsequent log output should be suppressed and monitoring should terminate.

Furthermore, this build has been executed 14 times and the only log output recorded in the LTS bucket is from the final container of the last stage of the first execution of the build.

dgozalo:
#### Submitter checklist

#### Description
Various fixes to the formatting of logs, it also now controls the flow of the streaming, avoiding to show empty steps after a previous one has failed.

Made a big refactor around the writers, before this, each implementation was responsible to get the logs from the pods / containers, which led to discrepancies. This was wrong as the implementation should only focus on showing the logs the way they need and not on how to obtain them.

Now we make use of channels to send / receive instances of LogLine, which also contains some metadata (in this case, whether the line should be masked or not) this also leaves the masking ( and how to do it ) to the implementation's choice.

Making use of LogWriters, I modified the implementation of `controllerbuild` to obtain the logs in order to store them the same way as we do for running builds, this way we keep the same formatting for both methods (this should also be true for the UI).

Last but not least, the performance improved a lot by using channels.

Should fix #4936, but let's confirm with @tdcox first.


dgozalo:
/cc @pmuir 
/cc @abayer 

Might need to override Semantic Pull Request btw, something weird happened when rebasing.

pmuir:
for the semantic pr, fixup the merge commit and it will be sorted - you've just written real changes into the merge when resolving.

PR #5061, opened 2019-08-12
fix: warn the user if TLS is not enabled

### LINKED ISSUE #5044: boot: register by default the webhooks over HTTP instead of HTTPS
### Summary 

`jx boot` command creates by default the webhook endpoints as HTTP endpoints and register them with the git provider including setting the shared secret.

This is not a secure setup and should be enabled by default since the webhook is configured and exchanged in plain text. 

Install jx using `jx boot` following the instructions. 

```
kubectl get ing

 k get ing
NAME                   HOSTS                                 ADDRESS         PORTS     AGE
chartmuseum            chartmuseum-jx.34.76.209.178.nip.io   34.76.209.178   80        6d22h
hook                   hook-jx.34.76.209.178.nip.io          34.76.209.178   80        6d22h
```

### Expected behavior

Enable TLS for webhooks by default when using `jx boot` to ensure that the shared secret is not exchanged in clear text.

### Jx version

```
COPY OUTPUT HERE
```

### Jenkins type

- [ ] Classic Jenkins

### LINKED ISSUE #5048: Ensure users are aware that Vault does not have a secure (https) connection
Vault is surfaced via ingress despite not yet being secure (https). Minimally users need to be aware of this as a priority.

There are other options such as the one described in jenkins-x/jx#4925.

Show warning to users that Vault does not have a secure (https) connection and allowing the ability to opt out of Vault if needed.

We can validate in the pre-install verification step if the vault is configured and the tls flag is disabled.

All this information is available in the jx-requirements.yaml when secretStorage: vault (https://github.com/jenkins-x/jenkins-x-boot-config/blob/a5bef54de7357a99eb66b491fd69cd96eca851d2/jx-requirements.yml#L20) and ingress.tls.enabled: false and ingress.tls.production: false (https://github.com/jenkins-x/jenkins-x-boot-config/blob/a5bef54de7357a99eb66b491fd69cd96eca851d2/jx-requirements.yml#L17).

prompt for user consent in that case at least user is aware of the insecure combination.

pmuir:
#### Submitter checklist

#### Description
Warn the user that continuing without TLS is not
recommended if they are using vault or web hooks.

Note that right now you can’t not use web hooks.

We present a confirmation dialog to continue, with
`N` being the default after logging warnings.

fixes #5044 fixes #5048

### DIFF
--- a/pkg/cmd/step/verify/step_verify_preinstall.go
+++ b/pkg/cmd/step/verify/step_verify_preinstall.go
@@ -99,6 +99,11 @@ func (o *StepVerifyPreInstallOptions) Run() error {
 		return err
 	}
 
+	err = o.verifyTLS(requirements)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
 	o.SetDevNamespace(ns)
 
 	log.Logger().Infof("verifying the kubernetes cluster before we try to boot Jenkins X in namespace: %s", info(ns))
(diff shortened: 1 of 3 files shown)


pkg/cmd/step/verify/step_verify_preinstall.go
```diff
@@ -488,6 +493,33 @@ func (o *StepVerifyPreInstallOptions) verifyStorage(requirements *config.Require
...
+		if requirements.Webhook != config.WebhookTypeNone {
+			log.Logger().Warnf("TLS is not enabled so your webhooks will be called using HTTP. This means your webhook secret will be sent to your cluster in the clear. We do not recommend you do this.")
+			confirm = true
+		}
+		if confirm {
+			if o.BatchMode {
+				return errors.Errorf("cannot continue because TLS is not enabled.")
+			}
+			message := fmt.Sprintf("Do you wish to continue?")
+			help := fmt.Sprintf("Jenkins X needs TLS enabled to send secrets securely. We strongly recommend enabling TLS.")
```
ccojocar: I think we need some docs to indicate how to enable TLS in the requirements to avoid having users fiddling around with this config. This is just a follow up on this.

pkg/cmd/step/verify/test_data/preinstall/no_kaniko_or_terraform/jx-requirements.yml
```diff
@@ -14,7 +14,7 @@ ingress:
   namespaceSubDomain: -jx.
   tls:
     email: ""
-    enabled: false
+    enabled: true
```
ccojocar: I suppose that there is already a test to cover the case when TLS is enabled.

pmuir: I'm not sure I follow - this test is for TLS enabled and I added one for TLS disabled.

pkg/cmd/step/verify/step_verify_preinstall.go
pmuir: We start to cover it at https://jenkins-x.io/getting-started/boot/#ingress - what do you think is missing?

PR #5065, opened 2019-08-12
fix: don't force --global check for git name & email

garethjevans:


garethjevans:
the gitter check was using the --global flag when checking for the configured name & email.  This will fail on repos within a tekton pipeline that only use a local configuration.

PR #5091, opened 2019-08-14
feat: added a flag to skip the jx install for terraform clusters

garethjevans:
--skip-jx-install can be used to stop the jx install into the dev cluster.  The generated cluster can then be used with jx boot

### DIFF
--- a/pkg/cmd/create/create_terraform.go
+++ b/pkg/cmd/create/create_terraform.go
@@ -41,7 +41,7 @@ type Flags struct {
 	SkipLogin                   bool
 	ForkOrganisationGitRepo     string
 	SkipTerraformApply          bool
-	SkipJxInstall               bool
+	SkipInstallation            bool
 	NoActiveCluster             bool
 	IgnoreTerraformWarnings     bool
 	JxEnvironment               string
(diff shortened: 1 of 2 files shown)


pkg/cmd/create/create_terraform.go
```diff
@@ -127,6 +128,7 @@ func (options *CreateTerraformOptions) addFlags(cmd *cobra.Command, addSharedFla
 	cmd.Flags().StringVarP(&options.Flags.ClusterName, optionClusterName, "", "", "The name of a single cluster to create - cannot be used in conjunction with --"+optionCluster)
 	cmd.Flags().StringVarP(&options.Flags.CloudProvider, optionCloudProvider, "", "", "The cloud provider (currently gke only) - cannot be used in conjunction with --"+optionCluster)
 	cmd.Flags().BoolVarP(&options.Flags.SkipTerraformApply, "skip-terraform-apply", "", false, "Skip applying the generated Terraform plans")
+	cmd.Flags().BoolVarP(&options.Flags.SkipJxInstall, "skip-jx-install", "", false, "Skip installing Jenkins X into the generated cluster")
```
pmuir: we should make this consistent (name, help text) with the gke command https://jenkins-x.io/commands/jx_create_cluster_gke/

pkg/cmd/create/create_terraform.go
```diff
@@ -444,9 +446,13 @@ func (options *CreateTerraformOptions) createOrganisationGitRepo() error {
 		if err != nil {
 			log.Logger().Infof("Skipping jx install")
 		} else {
-			err = options.installJx(devCluster, clusterDefinitions)
-			if err != nil {
-				return err
+			if !options.Flags.SkipJxInstall {
```
hferentschik: I am confused, why is this flag needed? The code seems to imply that `SkipTerraformApply` is already just about that, skipping the jx install. What's the semantic difference between `skip-terraform-apply` and  `skip-jx-install`? 

What's about the use of both flags in combination?

Would something like the following not be better:

```go
	if options.Flags.SkipTerraformApply || options.Flags. SkipJxInstall {
            log.Logger().Infof("Skipping jx install")
            return nil
	} 

        devCluster, err := options.findDevCluster(clusterDefinitions)
	if err != nil {
		return err
	} 
	
         err = options.installJx(devCluster, clusterDefinitions)
	if err != nil {
		return err
	}

       return nil
```

garethjevans: SkipTerraformApply allows you to generate terraform but not apply it
SkipJxInstall (SkipInstallation) allows you to generate terraform, apply it, but then use something like `jx boot` to install the cluster

cagiti: The `skip-installation` flag for terraform is the same as the `skip-installation` when creating a cluster without terraform. If the description of the `skip-terraform-apply` options isn't clear then we could update that in a separate PR?

hferentschik: Sure, I guess my point here is also about the imo quite twisted logic and nested conditionals. What I suggest is afaict semantically equivalent and easier to take in.

PR #5099, opened 2019-08-16
feat(cmd): bind use meta pipeline and custom image option to viper

### LINKED ISSUE #5085: Make it possible to disable metapipelines with an environment variable
### Summary 
It should be possible to disable metapipelines wihtout having to modify the Prow chart.

A good way would be to disable it with an environment variable that could be added when building the values of the chart.

This way we could disable for the next CJXD release.

hferentschik:
- split pipelinerunner into command and controller
- updated tests and made it a non integration test

fixes #5085


hferentschik:
I would have loved to get PR #5051 in first and rebase on top of that, but I guess the other way around is ok as well.

abayer:
sorry, @hferentschik! I was prioritizing `area/fox` PRs.

PR #5160, opened 2019-08-21
fix(test): use log.CaptureOutput instead os.Pipe() to capture log output

### LINKED ISSUE #5158: TestGetRunningBuildLogsWithMultipleStage fails in integration context
```
=== Failed
=== FAIL: pkg/logs TestGetRunningBuildLogsWithMultipleStages (0.03s)
Failed to write to log, write |1: file already closed
14
    tekton_logging_test.go:593: 
            Error Trace:    tekton_logging_test.go:593
            Error:          Not equal: 
                            expected: 15
                            actual  : 14
            Test:           TestGetRunningBuildLogsWithMultipleStages


DONE 1112 tests, 4 skipped, 1 failure in 1728.279s
Step: step-stash-test-results:
```

hferentschik:
fixes #5158

### DIFF
--- a/pkg/logs/tekton_logging_test.go
+++ b/pkg/logs/tekton_logging_test.go
@@ -8,7 +8,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
-	"os"
 	"path"
 	"regexp"
 	"testing"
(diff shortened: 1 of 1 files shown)


hferentschik:
Let's see whether this makes a difference. IMO it is nicer to read regardless, but I think it might actually also prevent the error. Maybe there are flushing issues when using the pipe?

PR #5121, opened 2019-08-19
fix: exit with a diff in pipeline

pmuir:
#### Submitter checklist

- [ ] Change is code complete and matches issue description.
- [ ] Change is covered by existing or new tests.

#### Description
This fixes jx step verify environments to not create a PR in a pipeline but instead run git diff and error if there is a diff

#### Which issue this PR fixes

fixes #5120

PR #5162, opened 2019-08-21
fix: added test around failing multiple context scheduler

garethjevans:
expecting this to fail, using the PR as a question on how to configure multiple
contexts within a scheduler


pmuir:
Gareth's test fails, but so foes the pipelinerunner_controller, which we are looking at in #5121

pmuir:
Now included in #5166

PR #5168, opened 2019-08-22
chore: wrap flaky test in retry

pmuir:

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/cmd/controller/pipeline/pipelinerunner_controller.go
+++ b/pkg/cmd/controller/pipeline/pipelinerunner_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+
 	"github.com/jenkins-x/jx/pkg/cmd/clients"
 	"github.com/jenkins-x/jx/pkg/cmd/opts"
 
@@ -115,6 +116,11 @@ func (c *controller) startWorkers(ctx context.Context, wg *sync.WaitGroup, cance
 				logger.Infof("using custom pipeline image: %s", c.metaPipelineImage)
 			}
 			if err := srv.ListenAndServe(); err != nil {
+				if err == http.ErrServerClosed {
+					logger.Debugf("server closed")
+				} else {
+					logger.Errorf(errors.Wrapf(err, "starting http server on %s port %d", c.bindAddress, c.port).Error())
+				}
 				cancel()
 				return
 			}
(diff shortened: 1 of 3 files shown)


pkg/cmd/controller/pipeline/pipelinerunner_controller.go
```diff
@@ -97,6 +98,7 @@ func (c *controller) Start() {
 
 func (c *controller) startWorkers(ctx context.Context, wg *sync.WaitGroup, cancel context.CancelFunc) {
 	wg.Add(1)
+	logger.Debugf("about to spawn creating HTTP server mux on %s port %d", c.bindAddress, c.port)
```
hferentschik: that seems not needed

pkg/cmd/controller/pipeline/pipelinerunner_controller.go
```diff
@@ -108,13 +110,19 @@ func (c *controller) startWorkers(ctx context.Context, wg *sync.WaitGroup, cance
 			Handler: mux,
 		}
 
+		logger.Debugf("about to spawn starting HTTP server on %s port %d", c.bindAddress, c.port)
```
hferentschik: see above

pmuir: We *definitely* need more logging here, it's impossible to see what is going on otherwise. In particularly silently dropping errors is not good.

pkg/cmd/controller/pipeline/pipelinerunner_controller.go
pmuir: Agreed. removed it.

pmuir:
> I don't think we need the additional logging in pipelinerunner_controller.go. I'd leave this file unchanged and just skip the tests.

I don't agree. Silently dropping errors is a really bad idea IMO. Makes it very hard to understand what is going on.

pkg/cmd/controller/pipeline/pipelinerunner_controller.go
pmuir: I think here due to the nested go routines that some extra logging is very helpful to aid debugging.

pmuir: Some comments would also be a good idea ;-)

pmuir: Ok, removed that debugging, it's the silent dropping of errors that bother me

hferentschik:
> I don't agree. Silently dropping errors is a really bad idea IMO. Makes it very hard to understand what is going on.

Right, logging the error for `ListenAndServe` makes sense. I was talking about the multiple logging of the host and port. 

That said, I created https://github.com/jenkins-x/jx/pull/5172 in an attempt to "fix" the test. I added the additional error handling there as well.
I think the issue is that the request is made to early and the port is not open yet. I added an explicit wait for the port to be open.

PR #5170, opened 2019-08-22
fix: add test for multiple additional queries in scheduler

pmuir:
#### Submitter checklist

fixes #


abayer:
Should pass now - it was the pesky logging flakiness, which is now worked around.

PR #5172, opened 2019-08-22
fix(test): waiting for HTTP to be open prior to making requests

### LINKED ISSUE #5167: pipelinecontroller tests are flaky
### Summary 

They intermittently fail connecting the http server. This happens more when under load. I have reproduced it in a devpod.

https://github.com/jenkins-x/jx/blob/master/pkg/cmd/controller/pipeline/pipelinerunner_controller_test.go#L73

hferentschik:
fixes #5167

### DIFF
--- a/pkg/cmd/controller/pipeline/pipelinerunner_controller_test.go
+++ b/pkg/cmd/controller/pipeline/pipelinerunner_controller_test.go
@@ -3,22 +3,22 @@ package pipeline
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
+	"github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	"github.com/jenkins-x/jx/pkg/client/clientset/versioned/fake"
+	"github.com/jenkins-x/jx/pkg/log"
 	"io/ioutil"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"net"
 	"net/http"
 	"sync"
 	"testing"
+	"time"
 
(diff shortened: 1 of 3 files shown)


hferentschik:
/hold 

let's see whether tests are passing and trigger them a few times

pkg/cmd/controller/pipeline/pipelinerunner_controller_test.go
```diff
@@ -240,3 +253,29 @@ func getFreePort() (int, error) {
 	}()
 	return l.Addr().(*net.TCPAddr).Port, nil
 }
+
+func waitForOpenPort(host string, port int, timeOut time.Duration) error {
```
ccojocar: Is there any reason to not use `httptest` instead of opening directly the ports?

https://golang.org/pkg/net/http/httptest/

abayer:
Passed, rekicking.

pkg/cmd/controller/pipeline/pipelinerunner_controller_test.go
hferentschik: hmm, maybe just ignorance. How would that look like in this case? 

What's happening is that we select a "random" free port and pass it to the controller. The controller then starts its HTTP server waiting for requests. We then make a few HTTP requests to test various things.

The issue I am trying to fix is that the way the test was written, the test request would in some cases be made w/o the HTTP server being ready (at least that's what I believe atm). SO for this reason, I added this _waitForOpenPort_ method which will wait until the HTTP server is listening or the timeout occurs.

abayer:
Test failure wasn't real - a node crashed. Rerunning.

hferentschik:
> Test failure wasn't real - a node crashed. Rerunning.

lol

PR #5173, opened 2019-08-22
test: add an option in the makefile to disable the test caching only when is set

ccojocar:
The test caching seems to be disable by default in all test without any option to
turn it on when test are executed locally. This option will allow to keep the
test caching on and disable it explicitly only when required.

This will speed up the test execution at least when run locally.

#### Submitter checklist

fixes #

### DIFF
--- a/jenkins-x-integration.yml
+++ b/jenkins-x-integration.yml
@@ -25,7 +25,7 @@ pipelineConfig:
               - name: PARALLEL_BUILDS
                 value: "2"
               - name: DISABLE_TEST_CACHING
-                value: "ture"
+                value: "true"
             options:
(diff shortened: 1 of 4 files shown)


jenkins-x-integration.yml
```diff
@@ -24,6 +24,8 @@ pipelineConfig:
                 value: "true"
               - name: PARALLEL_BUILDS
                 value: "2"
+              - name: DISABLE_TEST_CACHING
+                value: "ture"
```
pmuir: That's not how you spell `true` ;-)

ccojocar: Fixed. Thanks!

PR #5194, opened 2019-08-23
fix(tekton): removing LogWriter.StreamLog since it causes synchronisa…

### LINKED ISSUE #5171: Fix flaky tekton logging tests
`TestGetRunningBuildLogsWithMatchingBuildPods` and `TestGetRunningBuildLogsWithMultipleStages` failing  with e.g.

```Pipeline failed on stage 'from-fakebranch' : container 'build-step-promote-**-promote'. The execution of the pipeline has stopped.
    tekton_logging_test.go:425: 
        	Error Trace:	tekton_logging_test.go:425
        	Error:      	Not equal: 
        	            	expected: 21
        	            	actual  : 20
        	Test:       	TestGetRunningBuildLogsWithMatchingBuildPods


DONE 1112 tests, 5 skipped, 1 failure in 1486.637s
Step: step-stash-test-results:
```

hferentschik:
…tion issues

fixes #5171

### DIFF
--- a/pkg/cmd/controller/controller_build.go
+++ b/pkg/cmd/controller/controller_build.go
@@ -70,27 +70,6 @@ func (w *LongTermStorageLogWriter) WriteLog(logLine logs.LogLine) error {
 	return nil
 }
 
-// StreamLog will receive a logs channel and an errors channel which the logs producer will send
-// it will mask the lines marked as ShouldMask then it will append the line's bytes to the already stored ones
-func (w *LongTermStorageLogWriter) StreamLog(lch <-chan logs.LogLine, ech <-chan error) error {
-	for {
-		select {
-		case l, ok := <-lch:
-			if !ok {
-				return nil
-			}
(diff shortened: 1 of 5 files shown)


hferentschik:
This is a "simple" fix of the issues outlined in #5171. It does not use any additional goroutines and hence avoids the need for additional synchronisation on the tekton logger side as well as in the implementation of the `LogWriter`s. For the use cases within the `jx` codebase, this is imo sufficient and simpler. 

I am not sure what can be gained by having `LogWriter.StreamLog` be executed in a goroutine. If one wants to buffer parts of the logs in the LogWriter, that's imo an implementation detail.

That said, using `WaitGroup`s to make sure that the main goroutine cannot exit until the streaming thread has completed would be a possible solution if one wants to keep the current approach. This would need to be paired, however, with a solution on the LogWriter side to avoid out of order writes caused by `WriteLog` and `StreamLog`.

Issue #5169, opened 2019-08-22
Improve output of `jx get build logs`


hferentschik:
`jx get build logs` displays a header row for each container of the pipeline, followed by the plain logs. This makes it sometimes hard to know the context of the a given line:

![Screenshot 2019-08-22 at 11 11 07](https://user-images.githubusercontent.com/296354/63507685-94cdb500-c4d8-11e9-84b4-9cf7d469a03a.png)

I suggest we do something similar to the `tkn` log view, where each line gets prefixed with the contextual information:

![Screenshot 2019-08-22 at 11 18 11](https://user-images.githubusercontent.com/296354/63507730-b169ed00-c4d8-11e9-957c-50c01176b587.png)

We can make this also partly configurable via an option.

tdcox:
That makes it very difficult to read the logs because most of the useful data is shifted off the screen to the right or wrapped. Also, the start of the important info will be staggered due to the variable length of the header, making it even worse to read. 

Colour should be reserved for extremely important items only.

hferentschik:
This is a matter of taste I think. For just tailing the log in order to see the progress I much prefer the suggested new format. In other cases it the current format might be better. I think it is a matte of taste as well as use case. 

It does not have to be either or. We can add an option for it.

PR #5201, opened 2019-08-23
feat: improve synchronization on logging so we don't lose or mix trac…

dgozalo:
…es (like headers)

#### Submitter checklist

#### Description
There was a problem with logging where we could lose the last few traces when logging through the CLI.

This was due to miss synchronization between channels.
We could also mix headers and traces in certain situations.

I've added synchronization via WaitGroup so we always wait for the printing goroutine to finish before leaving the logging functions.

Also, the func `LogWriter.WriteLog` should now put the line in the given channel instead of writing it itself, this means it will be in order when printing the log.

### DIFF
--- a/pkg/logs/tekton_logging.go
+++ b/pkg/logs/tekton_logging.go
@@ -12,6 +12,7 @@ import (
 	"net/url"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/jenkins-x/jx/pkg/cloud/gke"
@@ -39,13 +40,16 @@ type TektonLogger struct {
 	KubeClient        kubernetes.Interface
 	LogWriter         LogWriter
 	Namespace         string
-	LogsRetrieverFunc retrieverFunc
+	logsRetrieverFunc retrieverFunc
+	logsChannel       chan LogLine
+	errorsChannel     chan error
+	wg                *sync.WaitGroup
 }
 
 // LogWriter is an interface that can be implemented to define different ways to stream / write logs
 // it's the implementer's responsibility to route those logs through the corresponding medium
 type LogWriter interface {
-	WriteLog(line LogLine) error
+	WriteLog(line LogLine, lch chan<- LogLine) error
 	StreamLog(lch <-chan LogLine, ech <-chan error) error
 	BytesLimit() int
 }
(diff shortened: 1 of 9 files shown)


pkg/logs/tekton_logging.go
```diff
@@ -247,52 +251,64 @@ func (t TektonLogger) GetRunningBuildLogs(pa *v1.PipelineActivity, buildName str
...
-			})
+			}, t.logsChannel)
+			if err != nil {
+				return err
+			}
+			break
 		}
 	}
+	// We are done using the logs and errors channels, any message in the channels should still be read before calling wg.Done()
+	t.closeLoggingChannels()
```
hferentschik: so what happens if the code above returns an error? Then we neither sens the closing signal to the channel nor do we wait for the completion of the goroutine. 

This might in most cases not matter, but I think synchronisation is still not correct.

I was also hoping to get the test cases updated to what I had in PR #5194 which made more concise assertions

pkg/logs/tekton_logging_test.go
```diff
@@ -421,16 +424,80 @@ func TestGetRunningBuildLogsWithMatchingBuildPods(t *testing.T) {
...
+		},
+		Spec: v1.PipelineActivitySpec{
+			Build:         "1",
+			GitBranch:     "fakebranch",
+			GitRepository: "fakerepo",
+			GitOwner:      "fakeowner",
+		},
+	}
+
+	bytesF, err := ioutil.ReadFile("/Users/daniel-gozalo/go/src/github.com/jenkins-x/jx/pkg/logs/test_data/multiple_stages/pipelinerun.yml")
```
hferentschik: I had removed this from PR #5194. I assume that is some local debug/test code which is not meant to be committed.

pkg/logs/tekton_logging_test.go
```diff
@@ -570,19 +636,14 @@ func TestGetRunningBuildLogsWithMultipleStages(t *testing.T) {
...
 	}
 
-	logOutput := log.CaptureOutput(func() {
-		err := tl.GetRunningBuildLogs(pa, "abayer/js-test-repo/master/1")
-		assert.NoError(t, err)
-	})
-
-	aORb := regexp.MustCompile("Pod logs...")
-	n := aORb.FindAllStringIndex(logOutput, -1)
+	err := tl.GetRunningBuildLogs(pa, "abayer/js-test-repo/master/1")
```
hferentschik: There is a good reason to capture the whole log output, the test output get silent and does not pollute the logs.

pkg/logs/tekton_logging.go
```diff
@@ -40,12 +41,15 @@ type TektonLogger struct {
...
+	logsChannel       chan LogLine
+	errorsChannel     chan error
+	wg                *sync.WaitGroup
 }
 
 // LogWriter is an interface that can be implemented to define different ways to stream / write logs
 // it's the implementer's responsibility to route those logs through the corresponding medium
 type LogWriter interface {
-	WriteLog(line LogLine) error
+	WriteLog(line LogLine, lch chan<- LogLine) error
```
hferentschik: I find this interface very odd. What is this function really saying? Hey LogWriter, I want you do log a line and btw, here is the channel to write it to? Would you in this case not just send the log message to the channel yourself and be done with it? Looking at the implementations of this interface this is pretty much what is happening now.

I still argue if one really wants to decouple the actual log writing from what else is happening in tekton logger, it should be the responsibility of the LogWriter to provide a channel and tekton_logging just sends everything down that sink. This way you don't have to deal with all the goroutines and synchronisation in tekton_logigng.

Or maybe alternatively you have a single Stream function which takes a channel and reads from it until it closes. In this case, tekton_logging would create the channel and call Stream, but then just this channel, no differentiation between WriteLog and StreamLog. 

Either way, if you want to use channel there should only be a single function in the interface. One needs to decide who would be responsible for the channel creation LogWriter or TektonLogger. Either way, all logging would need to go through the channel.

hferentschik: Talking about removing `WriteLog`. I think it would make sense to not print the header line for the container at all:

```
err = t.LogWriter.WriteLog(LogLine{
			Line: fmt.Sprintf("\nShowing logs for build %v stage %s and container %s\n",				Line: fmt.Sprintf("\nShowing logs for build %v stage %s and container %s",
				infoColor.Sprintf(buildName), infoColor.Sprintf(stageName), infoColor.Sprintf(ic.Name)),					infoColor.Sprintf(buildName), infoColor.Sprintf(stageName), infoColor.Sprintf(ic.Name)),
```

I am thinking about https://github.com/jenkins-x/jx/issues/5169 in particular as well and I think it makes sense for the UI as well. 

Instead, we could enhance `LogLine` with contextual information about the current pod, container, stage, etc. Then it becomes a concern of the `LogWriter` to decide how it wants to render the logs from the various containers and how it wants to separate the logs. The `LogWriter` can decide to write a header logline as we do now, but it could also prepend contextual information to each log line it prints. One would get a solution for #5169 pretty much for free by just plugging in a different implementation of `LogWriter`. 

We can add to the contextual information whether a given logline is from a failed step/stage. So again it becomes the `LogWriter` who is responsible for adding colour escape sequences. When refactoring the tests and matching expected log lines, I, for example, had to account escape sequences which was odd. 

IMO we would end up with a more flexible solution, a `LogWriter` which actually does have responsibilities as well as better separation of concerns. @dgozalo, @abayer WDYT?

abayer: Hmm, that seems reasonable to me. One thing I know I'd like eventually is timestamping on the log lines, so contextual info may be the only way to get there.

dgozalo: Yeah, I eventually wanted to surfase the tekton logging format and see if it would fit us better.

The idea of sending a `LogLine` struct as an event was exactly that, we'd probably want to add some metadata to the traces at some point, and that was just the way to set it up.

I think we should try to get the opinion of the community on the formatting, though.

dgozalo: Also, I kind of agree on the interface being a bit weird right now, and it all comes down to the fixes that were needed due to synchronization.

I think it would make sense to just keep the StreamLog function. You can get every single line sent to the channel there too, in case you want to add some extra metadata.

I think the creation of the channels and synchronization should be in tekton_logging, that way you make the Writers a lot of simple to plug and use.

pkg/logs/tekton_logging_test.go
dgozalo: oops, didn't notice that.

pkg/logs/tekton_logging.go
hferentschik: > Also, I kind of agree on the interface being a bit weird right now, and it all comes down to the fixes that were needed due to synchronization.

Cool

> I think it would make sense to just keep the StreamLog function. You can get every single line sent to the channel there too, in case you want to add some extra metadata.

+1

> I think the creation of the channels and synchronization should be in tekton_logging, that way you make the Writers a lot simpler to plug and use.

I guess that might be a matter of taste. The points above are more important to agree on.

PR #5204, opened 2019-08-24
chore: refactor stuff out of CommonOptions

pmuir:
#### Submitter checklist

#### Description
An attempt to refactor some of the really obvious candidates in `opts`.

I tried to move anything that didn't have a meaningful dependency on `CommonOptions` - so I tried to get anything which had no depdencies, and any of the methods calling it that had no other depdencies.

The goal here is on first pass to unwind `CommonOptions` a bit, and get stuff into better locations and then at a later date clean up the code properly.

IOW don't boil the ocean :-)

fixes #


pmuir:
Suggest we ignore the Hound on this, as I don't want to let the scope creep to actually fixing things - let's use subsequent PRs for that!

PR #5223, opened 2019-08-27
fix: make helm templates skip applying an empty chart

ccojocar:
#### Submitter checklist

fixes #


pmuir:
Yep, that broke something!

ccojocar:
I am looking into them to check if they can be fixed.

ccojocar:
@pmuir @jstrachan I think this solution to skip the empty charts installation does not quite work to solve the issue in boot. There are some empty charts such as https://github.com/jenkins-x/jenkins-x-boot-config/tree/master/systems/jxing which needs to be installed. They just reference other charts.

PR #5232, opened 2019-08-28
chore: remove misleading warning message

ccojocar:
This warning message is misleading when vault is not configured as a secret
engine. It shouldn't be a warning since there is falling back on local secrets in that case.

#### Submitter checklist

fixes #


ccojocar:
@pmuir I changed the warning message into a debug. Please could you have a look? Thanks!

PR #5245, opened 2019-08-28
fix(logs): Set step start time based on previous step's finish time

### LINKED ISSUE #5243: Step duration in `PipelineActivity` is wrong
Since Tekton uses non-init containers for its steps now, the start time for all steps in a stage is the same - the actual command gets executed once the previous step finishes, but the containers are all started at once using an entrypoint wrapper to determine when to run the command. This messes up duration etc badly.

abayer:
#### Submitter checklist

#### Description

Also realized that we weren't actually testing things right, after the move to `cmd/controller`, and that any changes we made to steps in `UpdateForStage` were getting discarded anyway!

/assign @dgozalo 

fixes #5243


pkg/cmd/controller/controller_build.go
```diff
@@ -789,30 +789,38 @@ func updateForStage(si *tekton.StageInfo, a *v1.PipelineActivity) {
...
 	if si.Pod != nil {
+		var stageSteps []v1.CoreActivityStep
 		pod := si.Pod
 		_, containerStatuses, _ := kube.GetContainersWithStatusAndIsInit(pod)
 		containersTerminated = len(containerStatuses) > 0
-		for _, c := range containerStatuses {
-			name := strings.Replace(strings.TrimPrefix(c.Name, "build-step-"), "-", " ", -1)
-			name = strings.Replace(strings.TrimPrefix(c.Name, "step-"), "-", " ", -1)
-			title := strings.Title(name)
+		for i, c := range containerStatuses {
```
hferentschik: Personally I am not a big fan of single character variables except for indeces. _container.Name_ is imo so much easier to make sense of than c.Name.
Just saying.

pkg/cmd/controller/controller_build.go
```diff
@@ -789,30 +789,38 @@ func updateForStage(si *tekton.StageInfo, a *v1.PipelineActivity) {
...
+			title := getStepTitle(c.Name)
 			step, _ := kube.GetOrCreateStepInStage(stage, title)
 			running := c.State.Running
 			terminated := c.State.Terminated
 
 			var startedAt metav1.Time
 			var finishedAt metav1.Time
+			if startedAt.IsZero() {
```
hferentschik: Could this not be extracted in a function which describes what's happening here? Not only do we end up with less nested code, but hopefully one also understands what's happening here.

pkg/cmd/controller/controller_build.go
```diff
@@ -967,6 +977,13 @@ func updateForStage(si *tekton.StageInfo, a *v1.PipelineActivity) {
 	}
 }
 
+// getStepTitle translates the step container's name into the title for the step used in PipelineActivity
+func getStepTitle(containerName string) string {
+	name := strings.Replace(strings.TrimPrefix(containerName, "build-step-"), "-", " ", -1)
```
hferentschik: hmm, so there is no other way to get the actual step name? That seems to be quire fragile code. Maybe we step name should be label on the container?

pkg/kube/activity.go
```diff
@@ -571,18 +571,18 @@ func GetOrCreateStage(a *v1.PipelineActivity, stageName string) (*v1.PipelineAct
 }
 
 // GetOrCreateStepInStage gets or creates the step for the given name in the given stage
-func GetOrCreateStepInStage(stage *v1.StageActivityStep, stepName string) (*v1.CoreActivityStep, bool) {
+func GetOrCreateStepInStage(stage *v1.StageActivityStep, stepName string) (v1.CoreActivityStep, bool) {
```
hferentschik: What do you try to achieve here? I usually also think returning by value is better, but in this particular case I am a bit worried. Given the function name as well as the other GetOrCreate functions in _activity.go_, I'd expect a pointer. 
Now I get a copy of the step and need to somehow remember to update it in the _StageActivityStep_. In your usecase it works afaict, because you rebuild all steps in a loop.

hferentschik:
Maybe move the build controller into the build package since we touch the file anyways?

pmuir:
Just trying to Cosmin's fix in, as it sorts out a flaky test

PR #5265, opened 2019-08-29
fix: register jenkins jobs as part of boot

jstrachan:
this lets us use `jx boot` with Jenkins servers


jstrachan:
note that we also need this PR merging on the boot config: https://github.com/jenkins-x/jenkins-x-boot-config/pull/49

PR #5274, opened 2019-08-30
feat: add a step command to verify if a URL returns an expected HTTP status code

ccojocar:
This command waits with exponential backoff for a URL to return an expected status code. It will
fail when the timeout is exceeded.

#### Submitter checklist

#### Description

Add a new command `step verify url` which waits with exponential backoff for a URL to return an expected HTTP status code. 

This is particularly useful to check when a URL configured by the external DNS is ready during the boot sequence.

fixes #

### DIFF
--- a/pkg/cmd/step/verify/step_verify_url.go
+++ b/pkg/cmd/step/verify/step_verify_url.go
@@ -1,0 +1,106 @@
+package verify
+
+import (
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/jenkins-x/jx/pkg/cmd/helper"
+	"github.com/jenkins-x/jx/pkg/cmd/opts"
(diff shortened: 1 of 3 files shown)


pkg/cmd/step/verify/step_verify_url.go
```diff
@@ -0,0 +1,107 @@
...
+	return nil
+}
+
+func (o *StepVerifyURLOptions) logError(err error) error {
+	log.Logger().Infof("Retrying due to: %s", err)
+	return err
+}
+
+// Run waits with exponential backoff for an endpoint to return an expected HTTP status code
+func (o *StepVerifyURLOptions) Run() error {
```
cagiti: Should we a utility function here or re-use an existing backoff strategy?

ccojocar: It's using a utility function. See L91.

PR #5277, opened 2019-08-30
fix: use the version stream to find the latest jx version

### LINKED ISSUE #5276: jx version should use the recommended jx version in the version stream
when verifying the users version of jx we should check with the dependency matrix in the version stream - rather than the latest binary on github

jstrachan:
rather than using whatever is on github, lets use the version in the version stream dependency matrix.

This means that `jx version` and `jx upgrade cli` will now use the version stream instead

also added `o.GetVersionResolver()` as a helper method to lazily create and cache the creation of the version resolver to avoid multiple git clones + filling the log with duplicate clone text

fixes #5276

### DIFF
--- a/pkg/cmd/opts/version.go
+++ b/pkg/cmd/opts/version.go
@@ -22,6 +22,20 @@ func (o *CommonOptions) CreateVersionResolver(repo string, gitRef string) (*vers
 	}, nil
 }
 
+// GetVersionResolver gets a VersionResolver, lazy creating one if required so we can reuse it later
+func (o *CommonOptions) GetVersionResolver() (*versionstream.VersionResolver, error) {
+	var err error
+	if o.versionResolver == nil {
+		o.versionResolver, err = o.CreateVersionResolver("", "")
+	}
+	return o.versionResolver, err
+}
+
+// SetVersionResolver gets a VersionResolver, lazy creating one if required
+func (o *CommonOptions) SetVersionResolver(resolver *versionstream.VersionResolver) {
+	o.versionResolver = resolver
+}
+
 // GetPackageVersions returns the package versions and a table if they need to be rendered
 func (o *CommonOptions) GetPackageVersions(ns string, helmTLS bool) (map[string]string, table.Table) {
 	info := util.ColorInfo
--- a/pkg/cmd/step/verify/step_verify_packages.go
+++ b/pkg/cmd/step/verify/step_verify_packages.go
@@ -1,12 +1,19 @@
 package verify
 
 import (
+	"fmt"
+
+	"github.com/jenkins-x/jx/pkg/cmd/create"
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
 	"github.com/jenkins-x/jx/pkg/cmd/opts"
 	"github.com/jenkins-x/jx/pkg/cmd/opts/step"
 	"github.com/jenkins-x/jx/pkg/cmd/templates"
+	"github.com/jenkins-x/jx/pkg/cmd/upgrade"
 	"github.com/jenkins-x/jx/pkg/config"
 	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/util"
+	"github.com/jenkins-x/jx/pkg/version"
+	"github.com/jenkins-x/jx/pkg/versionstream"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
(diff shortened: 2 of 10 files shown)


pkg/cmd/upgrade/upgrade_cli.go
```diff
@@ -19,7 +16,9 @@ import (
 
 var (
 	upgradeCLILong = templates.LongDesc(`
-		Upgrades the Jenkins X command line tools if there is a newer release
+		Upgrades the Jenkins X command line tools if there is a new version available in the version stream
+
+		For more information on Version Streams see: [https://jenkins-x.io/architecture/version-stream/](https://jenkins-x.io/architecture/version-stream/)
```
pmuir: Probably want to say that the reference to the version stream is stored in team settings somewhere?

pkg/cmd/upgrade/upgrade_cli.go
```diff
@@ -63,9 +62,8 @@ func NewCmdUpgradeCLI(commonOpts *opts.CommonOptions) *cobra.Command {
 
 // Run implements the command
 func (o *UpgradeCLIOptions) Run() error {
-	if config.LoadActiveInstallProfile() == config.CloudBeesProfile {
-		o.NoBrew = true
-	}
+	// upgrading to a specific version is not yet supported in brew so lets disable it for upgrades
+	o.NoBrew = true
```
pmuir: This I think is a problem as you can end up with a brew and non-brew install, which has the potential to screw up your path. Looks like brew does support versions https://stackoverflow.com/questions/3987683/homebrew-install-specific-version-of-formula/9832084#9832084 so we perhaps want to use that. Or disable brew install entirely.

pkg/cmd/version.go
```diff
@@ -73,46 +75,58 @@ func (o *VersionOptions) Run() error {
...
-	newVersion, err := o.GetLatestJXVersion()
+	newVersion, err := o.GetLatestJXVersion(resolver)
 	if err != nil {
 		return errors.Wrap(err, "getting latest jx version")
 	}
 	if currentVersion.LT(newVersion) {
 		app := util.ColorInfo("jx")
-		log.Logger().Warnf("\nA new %s version is available: %s", app, util.ColorInfo(newVersion.String()))
+		log.Logger().Info("\n")
+		log.Logger().Warnf("A different %s version %s is available in the version stream. We highly recommend you upgrade to it.", app, util.ColorInfo(newVersion.String()))
```
pmuir: "different" seems wrong here. Perhaps just `%s version %s is...`

pkg/cmd/opts/version.go
```diff
@@ -22,6 +22,20 @@ func (o *CommonOptions) CreateVersionResolver(repo string, gitRef string) (*vers
 	}, nil
 }
 
+// GetVersionResolver gets a VersionResolver, lazy creating one if required so we can reuse it later
+func (o *CommonOptions) GetVersionResolver() (*versionstream.VersionResolver, error) {
```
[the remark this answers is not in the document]
jstrachan: I added the cache as it results in about 10 lines of text output as it re-clones each time

pkg/cmd/step/verify/step_verify_packages.go
```diff
@@ -93,13 +100,67 @@ func (o *StepVerifyPackagesOptions) Run() error {
...
+				},
+			}
+			options.Version = versionText
+			options.NoBrew = true
+			err = options.Run()
+			if err != nil {
+				return err
+			}
+			log.Logger().Info("\n")
+			log.Logger().Warnf("the version of %s has been updated to %s. Please re-run: %s", info("jx"), info(versionText), info("jx boot"))
```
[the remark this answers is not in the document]
jstrachan: its a little more complex than that unfortunately; `jx step verify preinstall` is a child proces in an interpreted pipeline triggered by a parent process (running `jx boot` which we don't know the exact CLI parameters to easily - so whatever happens we need to terminate this process with an error; to terminate the parent boot process.

one day we could return a magic error code and have the interpreted pipeline detect it, return the error code to the `jx boot` process and get `jx boot` to automatically restart itself. 

Thats a fair bit of work and pretty hard to test; so figured a simpler first option would be to tell the user to just re-run `jx boot`?

pkg/cmd/upgrade/upgrade_cli.go
jstrachan: its already an issue for CJXD too btw which took the same approach so I followed suit. 

how about we raise a separate issue of how to try get `jx upgrade cli` to work with specific versions and brew? its probably a thorny issue to get right while we figure out how to get the brew formula upgraded to a given jx version after we merge a version stream (with a timing issue if a user uses a new version stream before we update the brew formula). I guess we can just fail if the brew formula has not been updated yet?

pmuir: CJXD doesn't use brew at all (the instructions are to download from the website) so this is not an issue for CJXD.

jstrachan: I've raised this issue to track upgrading `jx` to a version via brew https://github.com/jenkins-x/jx/issues/5290

jstrachan: wonder if we should just ditch brew ;)

PR #5282, opened 2019-08-30
fix: better defaulting of the git kind + webhook in boot

jstrachan:
based on folks using standard public git servers

### DIFF
--- a/pkg/config/install_requirements.go
+++ b/pkg/config/install_requirements.go
@@ -444,7 +444,8 @@ func (c *RequirementsConfig) addDefaults() {
 		if c.Cluster.GitServer == "https://github.com" || c.Cluster.GitServer == "https://github.com/" {
 			c.Webhook = WebhookTypeProw
 		} else {
-			c.Webhook = WebhookTypeLighthouse
+			// TODO when lighthouse is GA lets default to it
+			// c.Webhook = WebhookTypeLighthouse
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/config/install_requirements.go
```diff
@@ -425,24 +425,28 @@ func (c *RequirementsConfig) IsLazyCreateSecrets(flag string) (bool, error) {
...
 		c.Cluster.GitName = c.Cluster.GitKind
 	}
 	if c.Ingress.NamespaceSubDomain == "" {
 		c.Ingress.NamespaceSubDomain = "-" + c.Cluster.Namespace + "."
 	}
+	if c.Webhook == WebhookTypeNone {
+		if c.Cluster.GitServer == "https://github.com" || c.Cluster.GitServer == "https://github.com/" {
+			c.Webhook = WebhookTypeProw
+		} else {
+			c.Webhook = WebhookTypeLighthouse
```
pmuir: I'm not sure this is a good idea without a strong warning to users that lighthouse is experimental. Can we add a confirmation dialog (override in batch mode) that says that we don't support their server type but that we do have an experimental feature called lighthouse that they could try out and make it really clear it's experimental!. Probably should be an error message to start with for the wrong type of git.

jstrachan: this function has no way to know if its got batch mode enabled so gonna just omit defaulting for lighthouse for now until its GA

PR #5284, opened 2019-08-30
[wip] fix: making all repository generation private

### LINKED ISSUE #5186: jx import creates git repos as public, they should be private by default instead
### Summary
We should assume repos should be private by default else code and secrets could be accidentally leaked.
### Expected behaviour
When importing a project, create Git repo as PRIVATE
When creating a new cluster environment repos are PRIVATE
### Actual behaviour
When importing a project, Git repo is PUBLIC and not unsecured.

hferentschik:
fixes #5186


docs/apidocs/openapi-spec/openapiv2.yaml
```diff
@@ -2420,7 +2420,7 @@ definitions:
         type: string
       forkBranchPatterns:
         type: string
-      gitPrivate:
+      gitPublic:
```
ccojocar: Is this change backward compatible?

hferentschik: On the CRD level yes, thanks to a custom UnmarshallJSON - https://github.com/jenkins-x/jx/pull/5284/commits/298486187de88b7b614856838c7710f0708d5a88#diff-2f0b17514b6caac80feb96e8fa27a5a5R428

hferentschik:
@pmuir, afaict the changes regarding making repos private per default look good. The reason I put this on hold is the issue I am having with the dev gitops repo which does not get created. It might be an unrelated bug, but I'd like to find out what's going on first. 

It seems to happen when one selects a user (opposed to org) as the owner of the dev repo. With the switch to private repos per default, more people will will try that as well, since organizations on the free plan won't allow creating private repos.

hferentschik:
Blocked on #5341

hferentschik:
@ccojocar @pmuir - I pushed another commit with addresses the unauthenticated fetching from the private repo (#5341). I did for now not implement the use of the git credential helper as outlined in the issue. We can address this in another issue. Instead, I did the same as for pushing namely making sure the fetch URL is fully specified and contains the credentials. 

I tested this locally and all repositories created by `jx [import|create|boot]` are now private. 

One thing I still would like to do is to upgrade the `jenkins-x-test` GitHub organization. If we get the basic paid subscription we can create private repositories and run the BDD test in default mode. Right now I am explicitly making them public in `ci.sh` via `--git-public=true`.

Not sure whether we should go ahead and merge or whether we should wait for this.

abayer:
fyi, failures seem to be legit - for example, you need to run codegen. =)

hferentschik:
> fyi, failures seem to be legit - for example, you need to run codegen

codegen yes, but the other test failures are test related. Not quite sure what to do with this.

pkg/gits/helpers.go
```diff
@@ -499,14 +499,24 @@ func ForkAndPullRepo(gitURL string, dir string, baseRef string, branchName strin
 		}
 	}
 
-	err = gitter.FetchBranch(dir, originRemote, branchName)
+	userDetails := provider.UserAuth()
+	originFetchURL, err := gitter.CreateAuthenticatedURL(originURL, &userDetails)
+	if err != nil {
+		return "", "", nil, nil, errors.Wrapf(err, "failed to create authenticated fetch URL for %s", originURL)
+	}
+	err = gitter.FetchBranch(dir, originFetchURL, branchName)
```
hferentschik: @pmuir so this code does something similar as the push code we have been talking about. It effectively uses the fully qualified remote URL + credentials for the fetch. This works fine over HTTP, but fails for file:// URLs. Hence all the test failures. Not quite sure how to move forward with this. 

Any suggestions? I cannot figure out how to adjust the tests, adding another if statement to detect whether we deal with http vs file seems fugly and when I tried did not resolve all test failures. 

After all the credential helper might be after all the way forward. WDYT?

hferentschik: I think I found a way, let's see whether tests are passing now

pkg/config/install_requirements.go
```diff
@@ -206,6 +207,45 @@ type ClusterConfig struct {
 	KanikoSAName string `json:"kanikoSAName,omitempty"`
 }
 
+// UnmarshalJSON method handles the rename of EnvironmentGitPrivate to EnvironmentGitPublic.
+func (t *ClusterConfig) UnmarshalJSON(data []byte) error {
```
ccojocar: i am wondering if this method can be moved into a helper and provide the keys as parameters. It seems to be duplicated above.

pkg/gits/helpers.go
```diff
@@ -590,12 +600,16 @@ func ForkAndPullRepo(gitURL string, dir string, baseRef string, branchName strin
 		}
 	}
 
+	log.Logger().Info("foo")
```
ccojocar: Probably this logger should be removed.

pkg/gits/helpers.go
```diff
@@ -590,12 +600,16 @@ func ForkAndPullRepo(gitURL string, dir string, baseRef string, branchName strin
...
+
 	if stashed {
 		err = gitter.StashPop(dir)
 		if err != nil && !IsNoStashEntriesError(err) { // Ignore no stashes as that's just because there was nothing to stash
 			return "", "", nil, nil, errors.Wrapf(err, "unable to pop the stash")
 		}
 	}
+
+	log.Logger().Info("bar")
```
ccojocar: ditto

pkg/gits/helpers.go
hferentschik: ups, yes.

pkg/gits/helpers.go
hferentschik: removed. Had it there to for debugging purposes

pkg/config/install_requirements.go
hferentschik: Potentially, but where would you place this extracted code. It would not be a generic helper. Even though the technique is identical, the structs and the field names are different. 

For now, I think I prefer duplication.

hferentschik:
@pmuir, @ccojocar - I pushed another update, removing the debug logging I accidentally added. 

It still seems to be a bit of a discussion if this is the direction we want to take. AFAIU there are concerns that evaluation of Jenkins-X (especially on the OSS side) gets harder. Using GitHub, a user cannot use the free GitHub organization longer to set up the recommended Jenkins X config. He either needs to explicitly set the "git public" option or use a paid subscription or use a GitHub user as the owner of the repositories. 

I can see that more users fall into the trap that the owner of the repository cannot be the Prow bot user, but imo this is a shortcoming we should fix in Prow. Overall, I think it is better to lean more towards the safe side, then the opposite.

hferentschik:
@ccojocar needs a new approval. FAICT it seems to work now

PR #5285, opened 2019-09-01
fix: Fixed boolean logic around creating dev environment repo

### LINKED ISSUE #5336: Dev repo does not get created for boot + gitops
When installing Jenkins-X with gitops enabled using boot, the dev repo does not get created.

```
$ jx boot
...
STEP: verify-env command: /bin/sh -c jx step verify env in dir: .

validating git repository for dev at URL https://github.com/hf-bee-bot/environment-hardy-jx-private-env-dev.git
creating environment dev git repository for URL: https://github.com/hf-bee-bot/environment-hardy-jx-private-env-dev.git to namespace jx
validating git repository for production at URL https://github.com/hf-bee-bot/environment-hardy-jx-private-env-production.git
creating environment production git repository for URL: https://github.com/hf-bee-bot/environment-hardy-jx-private-env-production.git to namespace jx-production
Using Git provider github at https://github.com
? Using Git user name: hf-bee-bot
? Using organisation: hf-bee-bot
Creating repository hf-bee-bot/environment-hardy-jx-private-env-production
Creating Git repository hf-bee-bot/environment-hardy-jx-private-env-production
Pushed Git repository to https://github.com/hf-bee-bot/environment-hardy-jx-private-env-production

validating git repository for staging at URL https://github.com/hf-bee-bot/environment-hardy-jx-private-env-staging.git
creating environment staging git repository for URL: https://github.com/hf-bee-bot/environment-hardy-jx-private-env-staging.git to namespace jx-staging
Using Git provider github at https://github.com
? Using Git user name: hf-bee-bot
? Using organisation: hf-bee-bot
Creating repository hf-bee-bot/environment-hardy-jx-private-env-staging
Creating Git repository hf-bee-bot/environment-hardy-jx-private-env-staging
Pushed Git repository to https://github.com/hf-bee-bot/environment-hardy-jx-private-env-staging
```

The log says "creating environment dev git repository", but then the repo is not actually created.

chrismellard:
This was failing during jx boot cli.

#### Submitter checklist

#### Description

Unless I'm missing a change in design looks like some boolean logic was incorrectly refactored a couple days ago. `jx boot` now no longer creates a dev repo because of it.

fixes #5336

### DIFF
--- a/pkg/cmd/step/verify/step_verify_environments.go
+++ b/pkg/cmd/step/verify/step_verify_environments.go
@@ -278,7 +278,7 @@ func (o *StepVerifyEnvironmentsOptions) createEnvGitRepository(name string, requ
 	}
 
 	if name == kube.LabelValueDevEnvironment || environment.Spec.Kind == v1.EnvironmentKindTypeDevelopment {
-		if !o.InCDPipeline() && !o.InCluster() {
+		if !o.InCluster() {
 			err := o.prDevEnvironment(gitInfo.Name, gitInfo.Organisation, privateRepo, userAuth, requirements, server)
 			if err != nil {
 				return errors.Wrapf(err, "creating dev environment for %s", gitInfo.Name)


chrismellard:
@dgozalo Think you're probably best placed to comment on this PR

dgozalo:
This is a tricky one. The logic expecting both to be false to go ahead and create the dev environment is correct - we only want to create the environment when we are executing boot in a local computer instead of a pipeline to avoid creating it in the master pipeline.

The problem here is that one of the functions is not working as expected which leads to it not creating the dev environment.

I'll investigate more on this and get back to you soon.

chrismellard:
@dgozalo In that case its the BUILDNUMBER being set that incorrectly identifies this as being part of a CICD pipeline

dgozalo:
We should get rid of the call to `o.InCDPipeline` because as it stands right now, it will always return true even outside an actual pipeline when executing boot so it doesn't really add much.

hferentschik:
> We should get rid of the call to `o.InCDPipeline` because as it stands right now, it will always return true even outside an actual pipeline when executing boot so it doesn't really add much.

Exactly, this is my conclusion as well. Unless there is something else than "are we running local or in cluster" , there is no point in `InCDPipeline`. The impl itself seems questionable as well.

I'll be trying to remove `o.InCDPipeline` as part of PR  #5284, since I hit the problem there as well.

pmuir:
I just fixed the commit message to meet standards

hferentschik:
I created issue #5336. Might be worth referencing this in the commit message as well.

PR #5292, opened 2019-09-02
fix: increase the vault health retry timeout to 10 minutes and add some more logging

ccojocar:
#### Submitter checklist

fixes #

### DIFF
--- a/pkg/kube/vault/vault_factory.go
+++ b/pkg/kube/vault/vault_factory.go
@@ -24,7 +24,7 @@ const (
 	maxRetries = 2
 
 	// healthReadyTimeout define the maximum duration to wait for vault to become initialized and unsealed
-	healthhRetyTimeout = 5 * time.Minute
+	healthhRetyTimeout = 10 * time.Minute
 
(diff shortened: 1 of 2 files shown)


pkg/kube/vault/vault_factory.go
```diff
@@ -167,6 +169,7 @@ func waitForVault(vaultClient *api.Client, initialDelay, timeout time.Duration)
 func waitForKVEngine(vaultClient *api.Client, initialDelay, timeout time.Duration) error {
 	return util.RetryWithInitialDelaySlower(initialDelay, timeout, func() error {
 		if _, err := vaultClient.Logical().Read(kvEngineConfigPath); err != nil {
+			log.Logger().Info("Waiting for KV engine to be configured...")
```
pmuir: This doesn't seem right - it's just before returning an error

ccojocar: The retry takes place on error. This log is intended to show some action when a new retry is being performed.

pmuir: oic

PR #5303, opened 2019-09-02
feat: Add context to PipelineActivity, filter gc by context

### LINKED ISSUE #5302: PipelineActivity should include pipeline's context
I was playing with `jx gc activities` to try to get it to treat multiple PR context builds better - right now, it sees a PR that's been triggered once and has five contexts as having five builds, so that the PR history limit default of 2 kicks in and three of the contexts' builds will get deleted. I'm thinking of changing it to track PR builds not just by owner/repo/branch but by context well, so that, in the default setup, two builds of each context for each PR will be preserved (until hitting the max age, of course).

But turns out I couldn't do that quite yet - we don't track the context on `PipelineActivity`! Certainly seems like we should. =) That'll allow for better GC as I described above, and more options for filtering, display, etc going forward.

### LINKED ISSUE #5334: jx get build logs shouldn't require PipelineRun exists for completed PipelineActivity
As of now, we generate the list of pipeline names for `jx get build logs` via listing `PipelineRun`s. This means we need to keep `PipelineRun`s around as long as we keep `PipelineActivity`s around. I don't think we need to do this - well, right now we do because there isn't a context on `PipelineActivity`, but https://github.com/jenkins-x/jx/pull/5303 will fix that. Once that's in, I think we can change the logic to do something like this:

* Include every `PipelineActivity` with an associated `PipelineRun` that isn't `Pending` in `names`.
* Include every `PipelineActivity` without an associated `PipelineRun` in `names` if the `PipelineActivity` has completed.

This'll involve rewriting some logic in `tekton_logging.go`'s `GetTektonPipelinesWithActivePipelineActivity` function, probably to create a map of `PipelineRun`s and then iterate over the list of `PipelineActivity`s to check in that map first, and then check if the `PipelineActivity` is completed, and if either are valid, add to `names`.

abayer:
#### Submitter checklist

#### Description

Adds `Context` to `PipelineActivitySpec`, and includes the context in `jx gc activities` filtering for build count limit.
Stop using `PipelineActivity` for `PipelineRun` owner reference, and GC `PipelineRun`s that completed at least two hours ago in `gc_activities`.

fixes #5302
fixes #5334


abayer:
Looks like the BDD spring PR build doesn't get picked up by `jx get build logs` properly - 
```
STEP: checking that there is a job built successfully by calling jx get build logs --wait cb-kubecd/bdd-spring-1567685295/PR-1
The selected pipeline didn't start, let's wait a bit
```
The build does get kicked off, I ran the same command in the cluster and it worked, so it seems something isn't refreshing correctly when the pipeline hasn't started at initial call.

abayer:
Ah, no, it picked it up eventually.

PR #5304, opened 2019-09-02
fix: lets load quickstarts from the version stream

### LINKED ISSUE #5283: support loading the quickstarts from the version stream
a common cause of confusion for folks using bitbucket / gitlab is that we currently use the github client to query the available quickstarts which means we need a github.com client + token.

It would simplify things greatly if we just loaded them from the version stream instead. That would also mean it would be easier to keep the quickstarts + their versions in sync in the version stream to improve quality

jstrachan:
fixes #5283

### DIFF
--- a/pkg/cmd/step/create/pr/step_create_pr_quickstarts.go
+++ b/pkg/cmd/step/create/pr/step_create_pr_quickstarts.go
@@ -7,7 +7,10 @@ import (
 	"github.com/jenkins-x/jx/pkg/auth"
 	"github.com/jenkins-x/jx/pkg/cmd/opts/step"
 	"github.com/jenkins-x/jx/pkg/config"
+	"github.com/jenkins-x/jx/pkg/gits/operations"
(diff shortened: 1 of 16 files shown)


pkg/cmd/step/create/pr/step_create_pr_quickstarts.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/jenkins-x/jx/pkg/cmd/helper"
+
+	"github.com/jenkins-x/jx/pkg/cmd/opts"
+	"github.com/jenkins-x/jx/pkg/cmd/templates"
+	"github.com/spf13/cobra"
+)
+
+var (
+	createPullRequestQuickStartsLong = templates.LongDesc(`
+		Creates a Pull Request on a 'jx boot' git repository to mirror all the SourceRepository CRDs into the quickStarts Chart
```
pmuir: copy and paste ;-)

jstrachan:
@pmuir think I've fixed those issues now thanks

PR #5305, opened 2019-09-02
chore: using the 'latest' tagged images in bdd tests

garethjevans:
#### Submitter checklist

fixes #

### DIFF
--- a/Dockerfile.builder-base
+++ b/Dockerfile.builder-base
@@ -1,3 +1,3 @@
-FROM jenkinsxio/builder-base:latest
+FROM gcr.io/jenkinsxio/builder-jx:latest
 
 COPY build/linux/jx /usr/bin/jx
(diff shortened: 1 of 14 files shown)


rawlingsj:
BTW I think we've seen issues running images with the `:latest` tag before.  If a node or laptop already has an image of tag latest then that is used rather than a pull being performed, this can cause issues if a new or updated layer has been pushed to the latest tag.

I'm not sure if that will affect this usecase with the BDD test though?  Figured I'd drop the thought here anyway just in case.

garethjevans:
Ta @rawlingsj, I think for this usecase we should be ok, it should only be a problem if we start building these images locally

rawlingsj:
> Ta @rawlingsj, I think for this usecase we should be ok, it should only be a problem if we start building these images locally

yeah I guess folks just need to remember to do a `gcr.io/jenkinsxio/builder-go:latest` for example when building locally, I tend to a bit but that's cool.  Definitely agree it's better than having random versions! :)

/lgtm

PR #5347, opened 2019-09-05
fix: population of organisation + project on gitlab

### LINKED ISSUE #5345: `jx promote` fails on gitlab
seems when we query the Repository we don't populate the organisation so the create of the promote PR fails

jstrachan:
fixes #5345


abayer:
Something's busted in waiting for `jx get build logs` on pipelines with a context, presumably thanks to my change that went in earlier today. I'm digging into it.

PR #5407, opened 2019-09-10
fix: build controller should restart if it fails to list/watch pods

### LINKED ISSUE #5395: controllerbuild sporadically loses watch on pods with metapipeline enabled
Haven't nailed this down exactly, but `controllerbuild` seems to go silent sporadically with metapipeline enabled. We've seen this in the Lighthouse BDD tests and multiple times in the most recent attempt to turn metapipeline on for prod, so I don't think it's a coincidence.

jstrachan:
fix up a simpler and nicer retry logic for watching resources like pods so
we can handle losing a watch; we try to rerun things if a watch fails
otherwise we restart the pod

fixes #5395

### DIFF
--- a/pkg/cmd/controller/controller_build.go
+++ b/pkg/cmd/controller/controller_build.go
@@ -11,7 +11,9 @@ import (
 	"time"
 
 	"github.com/jenkins-x/jx/pkg/cmd/step/git"
+	"github.com/jenkins-x/jx/pkg/kube/watcher"
 	"github.com/jenkins-x/jx/pkg/logs"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/ghodss/yaml"
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
@@ -29,17 +31,14 @@ import (
 	"github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/builds"
 	"github.com/jenkins-x/jx/pkg/client/clientset/versioned"
+	"github.com/jenkins-x/jx/pkg/kube"
 	"github.com/jenkins-x/jx/pkg/log"
 	"github.com/jenkins-x/jx/pkg/util"
 	"github.com/spf13/cobra"
 	tektonclient "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
-	"k8s.io/client-go/tools/cache"
-
-	"github.com/jenkins-x/jx/pkg/kube"
 )
 
 // ControllerBuildOptions are the flags for the commands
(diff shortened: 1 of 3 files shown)


pkg/kube/watcher/watcher.go
```diff
@@ -0,0 +1,94 @@
...
+
+	watcher, err := w.ListWatch.Watch(w.Namespace, w.ListOptions)
+	if err != nil {
+		log.Logger().Fatalf("failed to watch resources in namespace %s due to %s", w.Namespace, err.Error())
+	}
+
+	resources, err := w.ListWatch.List(w.Namespace, w.ListOptions)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			log.Logger().Fatalf("failed to list resources in namespace %s due to %s", w.Namespace, err.Error())
```
hferentschik: Is this mitigating an actual problem and how does a restart of the controller help?

pkg/cmd/controller/controller_build.go
```diff
@@ -195,56 +194,40 @@ func (o *ControllerBuildOptions) Run() error {
...
+				o.onPod(obj, kubeClient, jxClient, ns)
+			},
+			DeleteFunc: func(obj runtime.Object) {
+			},
+		}
 	}
 
-	// Wait forever
-	select {}
+	watcher := watcher.NewPodWatcher(kubeClient, ns, metav1.ListOptions{}, handlers)
```
hferentschik: OOI, apart of abstracting some of the watching logic into `PodWatcher`, how does this all fix #5395? What was the underlying issue?

jstrachan: the old code wrapped all of the handling watch errors, so if a watcher failed the pod just sat there forever, never attempting a reconnect.

now we break the run loop if a watch fails: https://github.com/jenkins-x/jx/blob/d7f746d25566ac9d0e7a469083702fd883aeae51/pkg/kube/watcher/watcher.go#L65-L68

which we then recreate everything again - or if the list/watch fail we terminate the pod (e.g. if the API server is down due to an upgrade of k8s versions)

pkg/kube/watcher/watcher.go
jstrachan: if the API server is down, we terminate the pod - so that it will keep retrying until a list/watch works

ccojocar: I think this will crash the build controller POD completely making Kuberntes scheduler to restart the Pod and eventually fail based on the liveness probe setting when something is wrong in the cluster.

In the same time, there is an infinite for loop a the end of the build controller which keeps executing the Run if an error occurs passed this point. This looks like a second retrying strategy in case the code execution passes this point. 

I think some times, the result channel can be just closed without returning an error. From docs:
```
	// Returns a chan which will receive all the events. If an error occurs
	// or Stop() is called, this channel will be closed, in which case the
	// watch should be completely cleaned up.
```

I might be useful to have only one retrying mechanism in all use cases. I would rather propagate the error out of Run and let the outer for to decide based on error type how to handle the situation instead of using a `log.Fatalf`.

pkg/kube/watcher/watcher.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+	}
+
+	for _, resource := range resources {
+		w.onAdd(resource)
+	}
+
+	// TODO we could ensure that watch events are always newer than the first list just in case?
+	ch := watcher.ResultChan()
+	for event := range ch {
```
ccojocar: The channel might be closed directly when the connection is lost without propagating an error. In that case the for loop terminates and we return a `nil` error to the caller which likely is fine in the current implementation because Run is wrapped in an infinite loop without error check.

abayer:
I’m unsure if this is a good idea - we’ve only seen the controller hang sporadically with metapipeline enabled, and we don’t know if it’s losing the watch or if something else is wrong. Given that we’re currently using standard informer logic, and it’s only having issues with metapipeline enabled, I feel like it’s more likely to be an issue specific to how we’re handling metapipeline in the controller than that we’re losing the watch.

pkg/kube/watcher/watcher.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// ListWatch an interface for something that can list and watch resources
+type ListWatch interface {
+	List(ns string, listOptions metav1.ListOptions) ([]runtime.Object, error)
+	Watch(ns string, listOptions metav1.ListOptions) (watch.Interface, error)
+}
+
+// Run watches the resources until the connection is lost
+func (w *Watcher) Run() error {
```
ccojocar: It might be helpful to have a graceful termination for the watcher. An `ctx context.Context` can be defined as an argument of `Run` and then in the select the termination can be handled:
```Go
select {
    case <-ctx.Done():
	watcher.Stop()
	return
}
```

pkg/cmd/controller/controller_build.go
```diff
@@ -195,56 +194,40 @@ func (o *ControllerBuildOptions) Run() error {
...
+	for {
+		err = watcher.Run()
+		if err != nil {
```
ccojocar: I think this loop retries endlessly. I might be useful to handle the case when an error is fatal and crash the pod to let the kubernetes scheduler to restart it.

jstrachan: the tricky bit is knowing which errors are fatal and which are not - thats why I went with `log.Fatal` - to make explicit fatal errors

pkg/kube/watcher/watcher.go
jstrachan: we could switch the `log.Fatal` if we cannot List/Watch and return an error. The tricky decision is knowing when to terminate the container or not from outside

jstrachan:
@abayer I've seen the build controller lose watches silently and stop processing all pods without being restarted. There's a definite issue with losing watches permanently

pkg/cmd/controller/controller_build.go
jstrachan: the loop is designed loop endlessly - to recreate lists/watches and process the watches until watch errors - and terminate with `log.Fatal` if the watcher cannot be created

abayer:
@jstrachan I guess my question is how we know it's specifically that we lost the watch, and not that something hung in the actual logic of the controller. Maybe I'm missing something, but I don't see how enabling metapipeline would cause the watch to stop working sporadically.

pkg/cmd/controller/controller_build.go
ccojocar: I think you can define some custom error types in watcher package such as `FatalError` and `RetryableError` and wrap all error accordantly. On the caller side you can use a type switch to handle each case:
```
 switch e := err.(type) {
    case *RetryableError:
        // retry
    case *FatalError:
        // crash the pod
    default:
    }
```

ccojocar: This will give the caller the chance to handle these cases in its own way. Maybe we want to reuse this code in other areas which want to handle this cases differently.

abayer:
I'm digging into the logs from the hangs I saw yesterday morning and am seeing a pattern of all three cases having their last log line seem to be https://github.com/jenkins-x/jx/blob/e820d616812cc33dd41859c8fd7e4489f809f017/pkg/cmd/controller/controller_build.go#L759 for a metapipeline. Trying to verify now.

abayer:
Ah, no, it's something within `reportStatus`...

abayer:
Ok, it's not `reportStatus`, it's `logJobCompletedState`- https://github.com/jenkins-x/jx/blob/e820d616812cc33dd41859c8fd7e4489f809f017/pkg/cmd/controller/controller_build.go#L1389 is the last line logged. We shouldn't even be getting there, since this is firing for a metapipeline...

abayer:
Either we're getting to https://github.com/jenkins-x/jx/blob/e820d616812cc33dd41859c8fd7e4489f809f017/pkg/cmd/controller/controller_build.go#L729 despite being on a metapipeline or we're somehow ending up at https://github.com/jenkins-x/jx/blob/e820d616812cc33dd41859c8fd7e4489f809f017/pkg/cmd/controller/controller_build.go#L613 despite being a `PipelineRun`, not a Knative Build. I'm guessing it's the former.

abayer:
So https://github.com/jenkins-x/jx/pull/5411 fixes why we were getting to that point for a metapipeline in the first place. I'm not yet sure that's why this is happening, but I strongly suspect it is.

abayer:
See #5411 for more discussion, but I'm fairly certain that the bug fixed there is exposing an issue with `tekton.GetRunningBuildLogs` that only shows up at some point between completion of the metapipeline and starting of the first pod for the generated pipeline.

jstrachan:
@ccojocar I've refactored the code to separate creating the channel (the list/watch) from processing the channel so we can avoid the `log.Logger().Fatal` from the watcher library

ccojocar:
@jstrachan Thanks for fixing the review comments! It looks good!

PR #5409, opened 2019-09-10
feat: avoid passing arguments down the chain with create cluster

### LINKED ISSUE #5000: jx CLI failed to generate Jenkins API token over REST API due to: logging in: cookies did not work; bad login or not using legacy security realm: 403 Forbidden
### Summary 

The jx CLI fails to generate the Jenkins API token while creating an EKS cluster on AWS.

```
mkdir -p ~/.jx/bin

curl -L https://github.com/jenkins-x/jx/releases/download/v2.0.558/jx-linux-amd64.tar.gz | tar xzv

mv jx ~/.jx/bin

jx --version

export AWS_DEFAULT_REGION=us-west-2

echo "nexus:
  enabled: false
" | tee myvalues.yaml

jx create cluster eks \
    --cluster-name jx-rocks \
    --region $AWS_DEFAULT_REGION \
    --node-type t2.large \
    --nodes 3 \
    --nodes-min 3 \
    --nodes-max 6 \
    --default-admin-password=admin \
    --default-environment-prefix jx-rocks \
    --git-provider-kind github \
    --static-jenkins true

```

### Expected behavior

The appropriate Jenkins resources/jobs should have been created.

### Actual behavior

The warning message "**retrying after error:Failed to create the tmp-cd folder in Jenkins: 500 Server Error**" is printed several times, but the output includes the message "**Jenkins X installation completed successfully**" even though the Jenkins resources/jobs were not created.

### Jx version

```
NAME               VERSION
jx                 2.0.558
jenkins x platform 2.0.911
Kubernetes cluster v1.13.8-eks-a977ba
kubectl            v1.13.2
git                git version 2.17.1
Operating System   Ubuntu 18.04.2 LTS
```

### Jenkins type

- [X] Classic Jenkins

### Kubernetes cluster

EKS cluster created with the command listed above.  See attached log file (hostname, username, e-mail address, etc were replaced by dummy values).
 
[jx_create_cluster_eks_log.txt](https://github.com/jenkins-x/jx/files/3473673/jx_create_cluster_eks_log.txt)

Ubuntu 18.04 LTS

dgozalo:
#### Submitter checklist

- [] Change is covered by existing or new tests.

#### Description
Neither `jx create cluster eks` nor `jx install` receive any arguments as part of its execution and every data needed is passed by flags.

Introducing a flag the wrong way: 
`--static-jenkins true` was causing the value to be passed as an argument instead of being the part of the flag like `--static-jenkins=true`.

This was being passed down the commands calls as an argument and ended up calling `jx create jenkins token "true"` which was causing errors when loging into the Jenkins server.

fixes #5000

### DIFF
--- a/pkg/cmd/create/create_cluster_eks.go
+++ b/pkg/cmd/create/create_cluster_eks.go
@@ -124,6 +124,11 @@ func (o *CreateClusterEKSOptions) Run() error {
 		os.Exit(-1)
 	}
 
+	// We don't want to pass any args that were introduced by mistake because they are passed to all other commands down the chain
+	if len(o.Args) > 0 {
+		log.Logger().Warn("We detected arguments being passed to the command, these arguments will be ignored and only flags will be considered")
+		o.Args = []string{}
+	}
 	flags := &o.Flags
 
 	zones := flags.Zones


pkg/cmd/create/create_cluster_eks.go
```diff
@@ -124,6 +124,8 @@ func (o *CreateClusterEKSOptions) Run() error {
 		os.Exit(-1)
 	}
 
+	// We don't want to pass any args that were introduced by mistake because they are passed to all other commands down the chain
+	o.Args = []string{}
```
ccojocar: I would also log a warning message before reseting the Args just to make the user aware of this case.

dgozalo: Check it again, I added a warning log message.

abayer:
Batching is broken - opening a issue for that. Will reopen this once it's no longer batchable.

PR #5411, opened 2019-09-10
fix: Handling of completion of metapipeline in build controller was flawed

### LINKED ISSUE #5395: controllerbuild sporadically loses watch on pods with metapipeline enabled

abayer:
#### Submitter checklist

#### Description

This fixes a number of inconsistencies/shoddy logic/odd behavior around how metapipelines are handled in `controller_build`. First, we weren't properly detecting the `PipelineRun` type due to using the wrong prefix, which was dumb on my part. =) That on its own led to us assuming the `PipelineActivity` was complete as soon as the metapipeline finished, which led us down code paths that result in #5395. Additionally, we weren't consistent with our `PipelineRunInfo` names, which may not have been a functional bug, but was definitely a logging bug that made it hard for me to figure out what was going on.

Along the way, I realized that even with those things dealt with, we still ended up processing the metapipeline's pod, albeit after the completion of the generated pipeline, so it didn't actually do anything, but that's still annoying, so I've fixed the logic for deciding whether to call `logJobCompleteState` and try to populate the long-term storage logs URL, etc to be smarter about how to handle metapipeline.

Oh, and we weren't actually using the image we just built in the `jenkins-x-controllerbuild` pod in the Tekton BDD tests, which took me a couple runs to realize, so let's do that.

fixes #5395

### DIFF
--- a/pkg/tekton/pipeline_info.go
+++ b/pkg/tekton/pipeline_info.go
@@ -153,7 +153,7 @@ func CreatePipelineRunInfo(prName string, podList *corev1.PodList, ps *v1.Pipeli
 		Name:        PipelineResourceName(pr.Labels[LabelOwner], pr.Labels[LabelRepo], pr.Labels[LabelBranch], pr.Labels[LabelContext], pipelineType, nil, "") + "-" + pr.Labels[LabelBuild],
 		PipelineRun: pr.Name,
 		Pipeline:    pr.Spec.PipelineRef.Name,
-		Type:        BuildPipeline,
+		Type:        pipelineType,
 		CreatedTime: pr.CreationTimestamp.Time,
 	}
 
--- a/pkg/tekton/pipeline_info.go
+++ b/pkg/tekton/pipeline_info.go
@@ -151,7 +151,7 @@ func CreatePipelineRunInfo(prName string, podList *corev1.PodList, ps *v1.Pipeli
 		CreatedTime: pr.CreationTimestamp.Time,
 	}
 
-	if strings.HasPrefix(pr.Name, "metapipeline-") {
+	if strings.HasPrefix(pr.Name, MetaPipeline.String() + "-") {
 		pri.Type = MetaPipeline
 	}
 
--- a/pkg/tekton/pipeline_info.go
+++ b/pkg/tekton/pipeline_info.go
@@ -143,16 +143,18 @@ func CreatePipelineRunInfo(prName string, podList *corev1.PodList, ps *v1.Pipeli
 		return nil, errors.New(fmt.Sprintf("PipelineRun %s cannot be found", prName))
 	}
 
+	pipelineType := BuildPipeline
+
+	if strings.HasPrefix(pr.Name, MetaPipeline.String()+"-") {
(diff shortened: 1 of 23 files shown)


pkg/tekton/pipeline_info.go
```diff
@@ -151,7 +151,7 @@ func CreatePipelineRunInfo(prName string, podList *corev1.PodList, ps *v1.Pipeli
 		CreatedTime: pr.CreationTimestamp.Time,
 	}
 
-	if strings.HasPrefix(pr.Name, "metapipeline-") {
+	if strings.HasPrefix(pr.Name, MetaPipeline.String() + "-") {
```
hferentschik: +1, definitely a bug. Any idea how this would manifest itself in the controller at least seemingly not making any progress?

abayer: Trying to figure that out now. =) https://github.com/jenkins-x/jx/blob/e820d616812cc33dd41859c8fd7e4489f809f017/pkg/cmd/controller/controller_build.go#L1066 is the next line I'm sure should be logged after https://github.com/jenkins-x/jx/blob/e820d616812cc33dd41859c8fd7e4489f809f017/pkg/cmd/controller/controller_build.go#L729, the last line we do see, so something must be going wrong between those.

abayer:
Current thinking is that `tekton.GetRunningBuildLogs` is hanging in some race condition around completion of the metapipeline and starting of the generated pipeline.

pkg/tekton/pipeline_info.go
ccojocar: Shouldn't you add the pipeline type also as a label into the PipelienRun? Encoding it into the name seems a bit fragile to me.

hferentschik:
> Current thinking is that tekton.GetRunningBuildLogs is hanging in some race condition around completion of the metapipeline and starting of the generated pipeline.

My prime candidate as well, since this is the only place where we make use of go routines and where I could see the observed behaviour happening if the log streaming channel does not get closed properly.

abayer:
Note that this is still messy - depending on how my afternoon goes, I may rewrite more of the logic in `updatePipelineActivityForRun` here. Otherwise, I'll open an issue to refactor it down the road.

abayer:
Also, fwiw, I think there's a good chance that the proximate cause of #5395 is some combination of trying to collect build logs for long term storage after the metapipeline finishes but before the generated pipeline has actually spun up its pod completely, specifically when using buckets for the storage, since it has never showed up in tests, but BDD tests are using GitHub pages. Regardless, it's a path we won't cross again now.

abayer:
Putting back on hold until I write unit tests for `ControllerBuildOptions.onPipelinePod`. Which I will do my best to get done today.

abayer:
Ok, some `onPipelinePod` unit tests added.

PR #5429, opened 2019-09-11
wip

### LINKED ISSUE #5117: Make system vault name and its associated cloud resources configurable in jx boot
### Summary 

The vault name and the names of its associated resources are currently derived from the cluster name. This make recovering a vault for a cluster which was deleted quite difficult. It requires to create a vault with the exact same name, which is not easily possible to wire up to a new cluster which has a different name.

### Expected behavior

It should be possible to install jenkins-x using an existing vault (previously created). 

jx boot command configuration should be extended to allow users to define:
- vault name 
- vault resources name (bucket name, kms key ring name)
- other configurations parameters such as disable ingress, erase vault content etc.

### Actual behavior

It is not possible to reuse an older vault this make recovering secrets quite difficult. 

### Jx version

```
COPY OUTPUT HERE
```

### Jenkins type

- [ ] Classic Jenkins

pmuir:

#### Submitter checklist

I can't find any tests to add to for this. @ccojocar if I missed some, please let me know :-)

Just finishing some manual testing now.

fixes #5117


pmuir:
Also requires https://github.com/jenkins-x/jenkins-x-boot-config/pull/65

.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 repos:
-- repo: git@github.com:Yelp/detect-secrets
+- repo: https://github.com/Yelp/detect-secrets.git
```
pmuir: @cagiti is this valid? I don't use ssh for git at all so this seems more universal, as everyone has http these days.

PR #5437, opened 2019-09-11
fix: add 'jx step cluster lock'

jstrachan:
to lock a test cluster and connect to it to run commands such as BDD tests on one of a pool of test clusters


abayer:
/hold

Holding to avoid batches. Will unhold when the coast is clear.

PR #5440, opened 2019-09-11
fix: check when the vault kv secrets engine is ready for writing

ccojocar:
It seems that only checking if the configuration is accessible is not enough to ensure
that vault is ready to be used for writing secrets during the boot process.

There seem that error `no handler for route`.
See https://www.vaultproject.io/intro/getting-started/secrets-engines for more details.

#### Submitter checklist

fixes #


abayer:
I'll rekick bdd and tekton once https://github.com/jenkins-x/jx/pull/5428 is in

ccojocar:
Excellent! Thanks!

Issue #5310, opened 2019-09-03
TLS seems to be broken on applications when using boot


ccojocar:
### Summary 

- Enable TLS when installing with boot via the cert-manager and external DNS
- Create an application with either `create spring` or `create quickstart` and deploy it with Jenkins X
- The certificate for the application public endpoint form staging environment seems to be invalid 

The application ingress resource seems to still have the expose controller annotations:
```
apiVersion: v1
items:
- apiVersion: extensions/v1beta1
  kind: Ingress
  metadata:
    annotations:
      fabric8.io/generated-by: exposecontroller
      kubernetes.io/ingress.class: nginx
      kubernetes.io/tls-acme: "true"
    creationTimestamp: 2019-09-03T08:13:40Z
    generation: 1
    labels:
      provider: fabric8
    name: bdd-spring-1567497978
    namespace: jx-staging
    ownerReferences:
    - apiVersion: v1
      kind: Service
      name: bdd-spring-1567497978
      uid: ba295232-ce22-11e9-bb9b-42010a84003c
    resourceVersion: "8133"
    selfLink: /apis/extensions/v1beta1/namespaces/jx-staging/ingresses/bdd-spring-1567497978
    uid: bcc72597-ce22-11e9-bb9b-42010a84003c
  spec:
    rules:
    - host: bdd-spring-1567497978.jx-staging.boot.bdd.jenkins-x.rocks
      http:
        paths:
        - backend:
            serviceName: bdd-spring-1567497978
            servicePort: 80
    tls:
    - hosts:
      - bdd-spring-1567497978.jx-staging.boot.bdd.jenkins-x.rocks
      secretName: tls-bdd-spring-1567497978
  status:
    loadBalancer:
      ingress:
      - ip: 
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

No cert-manger issuer seems to be installed in the stating namespace. The cert-manger fails with the following error when trying to acquire the certificate for newly deployed application:

```
I0903 08:13:40.282576       1 base_controller.go:193] cert-manager/controller/ingress-shim "level"=0 "msg"="finished processing work item" "key"="jx-staging/bdd-spring-1567497978"
I0903 08:14:04.218156       1 base_controller.go:187] cert-manager/controller/ingress-shim "level"=0 "msg"="syncing item" "key"="jx-staging/bdd-spring-1567497978"
[...]

PR #5442, opened 2019-09-11
fix: make the cloud DNS secret name configurable in the jx requirements

ccojocar:
Use the default name when is not configured and update the value back into the requirements file.

#### Submitter checklist

part of #5310
fixes #


abayer:
The task not found error will stop showing up once the latest versions PR merges and I update weasel. =)

PR #5452, opened 2019-09-12
fix: avoid failing boot if folks want TLS and are not using GKE

### LINKED ISSUE #5451: jx boot is unnecessarily restrictive for GKE right now
`jx boot` failing OOTB if folks are not using GKE doesn't sound terribly useful. Instead lets just give a warning that we've not yet verified it works on that platform.

jstrachan:
we log a warning that the combination has not been tested though

fixes #5451


jstrachan:
is this a duplicate of https://github.com/jenkins-x/jx/pull/5452

PR #5461, opened 2019-09-12
chore: Delete metapipeline PipelineRuns more aggressively

abayer:
#### Submitter checklist

#### Description

With metapipeline on, you end up with a lot more pods and `PipelineRun`s - literally double, in fact! There's not a lot of point in keeping them around, so let's GC all metapipeline `PipelineRun`s 10 minutes after they've completed, rather than two hours like with generated/standalone `PipelineRun`s.

/assign @hferentschik 

n/a


hferentschik:
> I'm a little nervous about this the more I think about it. Will get back to it later.

Why?

Issue #5454, opened 2019-09-12
add a jx step to make it easy to import/trigger/run the bdd tests on a cluster


jstrachan:
lets make it super easy for folks to run the BDD tests on any Jenkins X cluster

e.g. so users can:

* create a k8s cluster
* run `jx boot`
* test it out via `jx bdd` - which then runs the BDD tests on the cluster to verify the cluster is good

So rather than lots of icky shell scripts that are hard to maintain, we can easily run/trigger a bdd pipeline in any cluster. 

It should help users/partners who - say - setup Jenkins X + a kubernetes cluster in a certain way - to easily run the BDD tests themselves

PR #5466, opened 2019-09-12
fix: a `jx step verify behavior` command to trigger the bdd tests

jstrachan:
so folks can easily run the BDD tests inside any Jenkins X installation

fixes https://github.com/jenkins-x/jx/issues/5454


pmuir:
/hold 

I think we need to think through the name of this and its relationship to the `jx step bdd`

jstrachan:
@pmuir I'm open to ideas on a better name.

BTW am hoping we can move away from `jx step bdd` really and just use `jx boot` / jx create cluster / jx install` to setup clusters when required - then run `jx bdd`

pmuir:
Put some ideas in the slack thread

pmuir:
Suggestion is to use `jx step verify behavior`

jstrachan:
@pmuir renamed to `jx step verify behavior`

Issue #5472, opened 2019-09-13
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.


alexvolsh:
### Summary 
Warning message appeared after Jenkins X CLI upgrade.

### Steps to reproduce the behavior
```
brew update
brew upgrade
jx version
```

### Expected behavior
```
NAME               VERSION
jx                 2.0.731
jenkins x platform 2.0.1094
Kubernetes cluster v1.15.3
kubectl            v1.15.3
helm client        Client: v2.14.3+g0e7f3b6
git                2.23.0
Operating System   Mac OS X 10.15 build 19A558d
```

### Actual behavior
```
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.
NAME               VERSION
jx                 2.0.731
jenkins x platform 2.0.1094
Kubernetes cluster v1.15.3
kubectl            v1.15.3
helm client        Client: v2.14.3+g0e7f3b6
git                2.23.0
Operating System   Mac OS X 10.15 build 19A558d


verifying packages
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.
WARNING: GitPrivate specified in TeamSettings. GitPrivate is deprecated use GitPublic instead.
```

### Jx version

The output of `jx version` is:

```
NAME               VERSION
jx                 2.0.731
jenkins x platform 2.0.1094
Kubernetes cluster v1.15.3
kubectl            v1.15.3
[...]

PR #5482, opened 2019-09-13
fix(boot): adding additional user verification when creating private repos

### LINKED ISSUE #5476: Improve user experience around creating private repositories
With issue #5186, repositories get per default created in _private_ mode. This can lead to some late failures during Jenkins-X installation, were early in the process we ask for the owner of the various environment repositories, but we don't create the repositories until much later. This can lead to errors when the user is using, for example, a free GitHub organisation as the owner of these repositories, since one cannot create private repositories in these.

To improve the user experience (in interactive mode) we require the user to explicitly confirm the choice if private repositories with the info that private repos can only be used for organisations with a paid service.

hferentschik:
fixes #5476 and #5472

### DIFF
--- a/pkg/cmd/step/verify/step_verify_preinstall.go
+++ b/pkg/cmd/step/verify/step_verify_preinstall.go
@@ -531,24 +531,25 @@ func (o *StepVerifyPreInstallOptions) gatherRequirements(requirements *config.Re
 		}
 	}
 
-	if requirements.Cluster.EnvironmentGitOwner == "" {
-		requirements.Cluster.EnvironmentGitOwner, err = util.PickValue(
-			"Git Owner name for environment repositories",
-			"",
-			true,
-			"Jenkins X leverages GitOps to track and control what gets deployed into environments.  This "+
-				"requires a Git repository per environment.  This question is asking for the Git Owner where these"+
-				"repositories will live",
-			o.In, o.Out, o.Err)
-		if err != nil {
-			return nil, errors.Wrap(err, "getting GKE Zone")
-		}
-	}
-
 	requirements.Cluster.Provider = strings.TrimSpace(strings.ToLower(requirements.Cluster.Provider))
 	requirements.Cluster.ProjectID = strings.TrimSpace(requirements.Cluster.ProjectID)
 	requirements.Cluster.Zone = strings.TrimSpace(strings.ToLower(requirements.Cluster.Zone))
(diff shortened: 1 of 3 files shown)


pkg/cmd/step/verify/step_verify_preinstall.go
```diff
@@ -558,18 +559,44 @@ func (o *StepVerifyPreInstallOptions) gatherRequirements(requirements *config.Re
...
+}
 
-	return requirements, nil
+func (o *StepVerifyPreInstallOptions) verifyPrivateRepos(requirements *config.RequirementsConfig) error {
+	log.Logger().Infof("Environment repos will be %s, if you want to create %s environment repos, please set %s to %s in jx-requirements.yaml", util.ColorInfo("private"), util.ColorInfo("public"), util.ColorInfo("environmentGitPublic"), util.ColorInfo("true"))
+
+	if requirements.Cluster.GitKind == "github" {
+		message := fmt.Sprintf("If '%s' is an GitHub organisation it needs to have a paid subscription to create private repos. Do you wish to continue?", requirements.Cluster.EnvironmentGitOwner)
+		help := fmt.Sprint("GitHub organisation on a free plan cannot create private repositories. You either need to upgrade, use a GitHub user instead or use public repositories.")
+		confirmed := util.Confirm(message, false, help, o.In, o.Out, o.Err)
```
hferentschik: So here we have the additional confirmation we talked about for boot. Do we need something for `jx install` as well?

@jstrachan, not sure how I would do something special for batch mode here? Something needs to be set in the requirements file if in batch mode.

hferentschik:
This is not going to make a difference to the failing BDD tests in the version stream repo. For that we need to decide what we want to use:

* Use a different org
* Let a GitHub user own the repos
* Use public repos

pkg/cmd/step/verify/step_verify_preinstall.go
ccojocar: I think this will prompt now always when `EnvironmentGitPublic` is set `false` which is a problem when boot is executed in the pipeline. You can use the `InCluster()` to detect when this is running in a pipeline and skip the `confirm`.

hferentschik: I think if o.BatchMode might be even a better condition to check. If batch mode is enabled, don't confirm, wdyt?

hferentschik:
@ccojocar I pushed an update

hferentschik:
Closing this PR. Changes are part of #5523

PR #5511, opened 2019-09-17
fix: Revert making all created repos private by default

abayer:
#### Submitter checklist

#### Description

Reverts #5284. As seen in #5507, we weren't ever actually testing this correctly (due to `--git-public false` actually resulting in `--git-public` being true and `false` being treated as additional arguments to `jx create cluster`). Fixing that showed that promote-to-production fails when using the GitHub user as the env repo owner as well, which is problematic in and of itself, but more immediately, we can't get tests to pass without creating a second user to own the private env repos or an org that allows private repos. This is blocking anything from getting merged in `jenkins-x-versions`, which means more and more things can be slipping into the periodic PRs there that could be introducing more failure cases we can't detect, etc.

So...we should revert and revisit.

/assign @daveconde 
/assign @pmuir 
/assign @hferentschik 

n/a


abayer:
This will require followup changes in `jenkins-x-boot-config` and `jenkins-x-versions` to go back to using `cb-kubecd` as the repo owner.

hferentschik:
it hurts, but here it goes:

/lgtm

Issue #5341, opened 2019-09-05
ForkAndPullRepo does not work with private repositories


hferentschik:
`ForkAndPullRepo` in `helpers.go` does not work with private repositories, since it does not pass credentials when accessing the repository.

```
$ jx boot
...
validating git repository for dev at URL https://github.com/hf-bee-bot/environment-hardy-jx-private-env-dev.git
creating environment dev git repository for URL: https://github.com/hf-bee-bot/environment-hardy-jx-private-env-dev.git to namespace jx
Duplicated Git repository https://github.com/jenkins-x/jenkins-x-boot-config to https://github.com/hf-bee-bot/environment-hardy-jx-private-env-dev
Setting upstream to https://github.com/hf-bee-bot/environment-hardy-jx-private-env-dev
error: creating dev environment for environment-hardy-jx-private-env-dev: forking and pulling https://github.com/hf-bee-bot/environment-hardy-jx-private-env-dev.git: fetching origin master: git output: remote: Repository not found.
fatal: repository 'https://github.com/hf-bee-bot/environment-hardy-jx-private-env-dev.git/' not found: failed to run 'git fetch origin master' command in directory '/Users/hardy/tmp/issue-5186/boot2/boot', output: 'remote: Repository not found.
fatal: repository 'https://github.com/hf-bee-bot/environment-hardy-jx-private-env-dev.git/' not found'
error: failed to interpret pipeline file jenkins-x.yml: failed to run '/bin/sh -c jx step verify env' command in directory '.', output: ''
```

hferentschik:
Relates to #5186

hferentschik:
So issue is that the method in the Gitter interface do not deal with authentication. In some cases this can be worked around, for example as done with the `Push` command in `DuplicateGitRepoFromCommitish` of `helpers.go`:

```go
	userDetails := provider.UserAuth()
		duplicatePushURL, err := gitter.CreatePushURL(duplicateInfo.CloneURL, &userDetails)
		if err != nil {
			return nil, errors.Wrapf(err, "failed to create push URL for %s", duplicateInfo.CloneURL)
		}
		err = gitter.SetRemoteURL(dir, "origin", duplicateInfo.CloneURL)
		if err != nil {
			return nil, errors.Wrapf(err, "failed to set remote url to %s", duplicateInfo.CloneURL)
		}
		err = gitter.Push(dir, duplicatePushURL, true, false, fmt.Sprintf("%s:%s", "HEAD", toBranch))
[...]

Issue #5515, opened 2019-09-17
sourcerepository label in PipelineActivity can exceed 63 char limit


hferentschik:
### Summary 

_PipelineActivity_ instances contain a _sourcerepository_ label which is a concatenation of repository owner and name. Given that there is no similar length restriction on the actual repository, one can get into a situation where _PipelineActivity_ instances cannot be submitted due to an exceeding label length.

### Steps to reproduce the behavior

The problem occurs for example when running BDD tests. When using _jenkins-x-bot-test_ opposed to _cb-kubecd_, one gets this error:

```bash
{"branch":"PR-1","component":"pipelinerunner","context":"promotion-build","level":"info","meta":true,"msg":"triggering pipeline","revision":"2a770c6685ece4a2f0aea95226269aa60c2d4e85","sourceURL":"https://github.com/jenkins-x-bot-test/environment-pr-5507-12-tekton-gkebdd-production.git","time":"2019-09-17T11:21:10Z"}
{"level":"info","msg":"PipelineActivity for jenkins-x-bot-test-environment-pr-5507-12-tekton-gkebdd-production-PR-1-1","time":"2019-09-17T11:21:10Z"}
{"component":"pipelinerunner","level":"info","msg":"unable to apply Tekton CRDs: failed to apply Tekton CRDs: PipelineActivity.jenkins.io \"jenkins-x-bot-test-environment-pr-5507-12-tekton-gkebdd-production-pr-1-1\" is invalid: metadata.labels: Invalid value: \"jenkins-x-bot-test-environment-pr-5507-12-tekton-gkebdd-production\": must be no more than 63 characters could not start pipeline: unable to apply Tekton CRDs: failed to apply Tekton CRDs: PipelineActivity.jenkins.io \"jenkins-x-bot-test-environment-pr-5507-12-tekton-gkebdd-production-pr-1-1\" is invalid: metadata.labels: Invalid value: \"jenkins-x-bot-test-environment-pr-5507-12-tekton-gkebdd-production\": must be no more than 63 characters","time":"2019-09-17T11:21:10Z"}
```

We need to revisit the use of _sourcerepository_ as a label.

hferentschik:
The problem is in activity.go

```go
func updateActivity(k *PipelineActivityKey, activity *v1.PipelineActivity) {
	if activity.Labels == nil {
		activity.Labels = make(map[string]string, 4)
	}

	updateActivitySpec(k, &activity.Spec)

[...]

PR #5523, opened 2019-09-17
fix: making all repository generation private per default

### LINKED ISSUE #5186: jx import creates git repos as public, they should be private by default instead

hferentschik:
Fixes #5186, #5341, #5476, #5472

Following up on #5511 which reverted the defaulting to private repos


hferentschik:
will need a fix for #5515 first.

pkg/cmd/deletecmd/delete_application.go
```diff
@@ -481,16 +482,18 @@ func (o *DeleteApplicationOptions) init() error {
 	return nil
 }
 
-func (o *DeleteApplicationOptions) deletePipelineActivitiesForSourceRepository(jxClient versioned.Interface, ns string, srName string) error {
+func (o *DeleteApplicationOptions) deletePipelineActivitiesForSourceRepository(jxClient versioned.Interface, ns string, gitOwner string, gitRepo string) error {
+	gitOwnerSelector := fields.OneTermEqualSelector("metadata.spec.gitOwner", gitOwner)
```
hferentschik: @abayer see whether this is going to cut it. If this gets us past the error with labels in the BDD tests, I can create a dedicated PR for #5515

hferentschik:
@abayer, so your fix for the BDD command line as well as the changes I added for removing the sourcerepository label seem to work. I verified that the created test repos were private:

```
jenkins-x-bot-test/environment-pr-5523-20-tekton-gkebdd-production created 2019-09-18 07:24:50 UTC private true
jenkins-x-bot-test/environment-pr-5523-20-tekton-gkebdd-staging created 2019-09-18 07:24:42 UTC private true
```

hferentschik:
@abayer now rebased on #5526. Is there anything else you think we should address before we can get this back in?

abayer:
Looks reasonable to me - the tests are using private repos, so if the tekton tests pass, we're good.

PR #5530, opened 2019-09-18
feat: add an --end-step flag into `jx boot` & `jx step create task`

garethjevans:


hferentschik:
What's the driving motivation behind that? Is there an issue giving a bit more context?

jstrachan:
@hferentschik  so we can support partial installs (e.g. install vault + ingress + cert manager + external-dns - but not tekton/prow) - so that we can spin up a cluster, then populate vault via some other tool (UI / CLI) - then run a regular full boot afterwards

PR #5557, opened 2019-09-20
fix: move activity_integration_test.go into dedicated package

### LINKED ISSUE #5556: Ginko tests in kube package run integration and non integration tests at the same time
### Summary 

Ginkgo has a single test suite per package. Even if there are multiple test files with Gingko tests, they are getting combined into a single test suite. 

This is per se not an issue, but one needs to remember that these tests share their _BeforeSuite_ and _AfterSuite_. 

Also if one is a unit test files and one an integration test file, they are getting combined. See also https://github.com/onsi/ginkgo/issues/130

In the _kube_ package, we have two test files, one a unit test file and one an integration test file. To ensure they run separately, the easiest is to put them into separate packages.

hferentschik:
fixes #5556

### DIFF
--- a/pkg/kube/test_integration/activity_integration_test.go
+++ b/pkg/kube/test_integration/activity_integration_test.go
@@ -1,6 +1,6 @@
 // +build integration
 
-package kube_test
+package test_integration_test
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


hferentschik:
@abayer I think splitting into multiple packages is the right approach here. See also discussion in the referenced issue in #5556

PR #5563, opened 2019-09-23
fix: default vault client to using internal kubernetes service dns na…

rawlingsj:
…me when running incluster

### DIFF
--- a/pkg/kube/vault/vault.go
+++ b/pkg/kube/vault/vault.go
@@ -2,6 +2,8 @@ package vault
 
 import (
 	"fmt"
+	"os"
+
 	"github.com/jenkins-x/jx/pkg/log"
 
 	"github.com/banzaicloud/bank-vaults/operator/pkg/apis/vault/v1alpha1"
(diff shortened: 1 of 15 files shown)


pkg/kube/vault/vault.go
```diff
@@ -430,17 +432,22 @@ func GetVaults(client kubernetes.Interface, vaultOperatorClient versioned.Interf
...
-		if err != nil {
-			log.Logger().Warnf("error finding vault service url setting to empty string, err: %s", err)
-			vaultURL = ""
-		}
-		if vaultURL == "" {
-			vaultURL, err = services.FindIngressURL(client, ns, vaultName)
+		// default to using internal kubernetes service dns name
+		vaultURL := fmt.Sprintf(defaultInternalVaultURL, vaultName)
+		// is there a better way to do this?
+		if os.Getenv("KUBERNETES_SERVICE_HOST") == "" {
```
rawlingsj: wonder if this is ewnough to check that we're running incluster seeing that we dont have the common options factory in these functions?

pmuir: It should be easy to refactor the `IsInCluster()` method off the factory and into a util method as it depends on precisely nothing

rawlingsj:
gonna add an interface and mock around `cluster.IsInCluster()` so it's easier to test as we will get different behaviour when running tests in or outside of a pipeline

/hold

Issue #5297, opened 2019-09-02
Log handling and historical data needs to be redesigned


abayer:
JX's log handling is problematic - there are issues with how we display logs, lack of metadata around the log lines (timestamps, for example), performance issues due to requiring a `PipelineActivity` in the cluster for each build that can have its logs viewed, related historical data issues since logs and activity isn't available to users for any builds that have had their `PipelineActivity` deleted (which we have to do regularly to keep `jx get build logs` performance from becoming unworkable), long term storage logs are for the whole pipeline without anything for individual steps, etc. 

In general, we need to revisit how logs and build history work in Jenkins X. The first step is defining what we want the behavior/available data/etc to be for users, and then we need to determine the optimal way to implement that. This is not something we can expect to "fix" in the short term with tweaks around the edges - the entire system for logging and build history needs to be reevaulated and redesigned.

abayer:
cc @hferentschik @dgozalo

codebryo:
Just had a convo with @daveconde about this. I actually made a design that I mostly intended for the UI team of things and wanted to verify if we are on the same track - https://whimsical.com/QdCozbq3C4UszvgycM2DR
On that node, CodeShip has a ton of Go Code around logs. We bascially spent a year in building a logging infrastructure and streaming servicem so I would expect there is stuff to salvage.

msavy:
@abayer Not sure if this is a function of the issues you're describing, but with the latest jx tekton gitops, there seems to be no way via UI such as Prow (or even the PR bot) to view the jx build logs in the browser in the same way that you would via Jenkins classic build system web interface.  

Prow just has very basic status info - no logs, AFAICT.

Happy to write up a feature request ticket for this if it is not obviously already encapsulated by this. Do let me know.

PR #5566, opened 2019-09-23
fix: Don't include successful metapipeline in logs

### LINKED ISSUE #5455: Metapipeline should be removed from PipelineActivity and logs once the generated pipeline starts
We should remove the metapipeline from the `PipelineActivity` (and no longer include it in `jx get build logs` or in the long term storage logs) once it's passed and the generated pipeline has started. There's no point in continuing to show the metapipeline info if it passed - it just clutters logs/the UI/etc.

abayer:
#### Submitter checklist

#### Description

If you start running `jx get build logs` before the metapipeline succeeds, it should still include the metapipeline steps in the logs, but if you start afterwards, the metapipeline steps shouldn't be in the logs, and they shouldn't be included in the long-term storage logs for the build.

Changes to `PipelineActivity` are still to come.
/assign @hferentschik 
/assign @dgozalo 

fixes #5455

### DIFF
--- a/pkg/cmd/controller/test_data/controller_build/completed_generated/cluster_state/pipelineruns.yml
+++ b/pkg/cmd/controller/test_data/controller_build/completed_generated/cluster_state/pipelineruns.yml
@@ -10,6 +10,7 @@ items:
       build: "1"
       created-by-prow: "true"
       owner: cb-kubecd
+      pipelineType: meta
       prowJobName: 6bae4c2b-d3ed-11e9-b0c6-72256f93bb5c
       repository: bdd-spring-1568135191
       tekton.dev/pipeline: meta-cb-kubecd-bdd-spring-15681-1
@@ -97,6 +98,7 @@ items:
       build: "1"
       created-by-prow: "true"
       owner: cb-kubecd
+      pipelineType: build
       prowJobName: 6bae4c2b-d3ed-11e9-b0c6-72256f93bb5c
       repository: bdd-spring-1568135191
       tekton.dev/pipeline: cb-kubecd-bdd-spring-1568135191-1
--- a/pkg/cmd/controller/test_data/controller_build/completed_generated/cluster_state/pipelineruns.yml
+++ b/pkg/cmd/controller/test_data/controller_build/completed_generated/cluster_state/pipelineruns.yml
@@ -10,7 +10,7 @@ items:
       build: "1"
       created-by-prow: "true"
       owner: cb-kubecd
-      pipelineType: meta
+      jenkins.io/pipelineType: meta
       prowJobName: 6bae4c2b-d3ed-11e9-b0c6-72256f93bb5c
       repository: bdd-spring-1568135191
       tekton.dev/pipeline: meta-cb-kubecd-bdd-spring-15681-1
@@ -98,7 +98,7 @@ items:
       build: "1"
       created-by-prow: "true"
       owner: cb-kubecd
-      pipelineType: build
+      jenkins.io/pipelineType: build
       prowJobName: 6bae4c2b-d3ed-11e9-b0c6-72256f93bb5c
       repository: bdd-spring-1568135191
       tekton.dev/pipeline: cb-kubecd-bdd-spring-1568135191-1
--- a/pkg/cmd/get/get_build_logs.go
+++ b/pkg/cmd/get/get_build_logs.go
@@ -18,7 +18,6 @@ import (
 	"github.com/jenkins-x/jx/pkg/kube"
 	"github.com/jenkins-x/jx/pkg/logs"
 	"github.com/jenkins-x/jx/pkg/tekton"
-	"github.com/knative/pkg/apis"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
@@ -28,7 +27,6 @@ import (
 	"github.com/jenkins-x/jx/pkg/cmd/templates"
 	"github.com/jenkins-x/jx/pkg/log"
 	"github.com/jenkins-x/jx/pkg/util"
-	tektonv1alpha1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	tektonclient "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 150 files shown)


abayer:
I'm debating whether to go with removing the metapipeline steps from `PipelineActivity` after all - the logic needed to support that would be rather messy, and it'd be a purely cosmetic change, unlike the log change, which does have functional implications (i.e., it stops us from trying to archive metapipeline logs if the metapipeline didn't fail, which should sneak around some edge case issues with the metapipeline pod getting deleted early). Since this is just a temporary fix until we rework logging completely (see https://github.com/jenkins-x/jx/issues/5297), I'm not sure it's worth messing around with `PipelineActivity` logic in `controller_build` at this point. @hferentschik, any thoughts?

abayer:
Yeah, decided to not mess with `PipelineActivity` here.  This is ready to go.

pkg/cmd/get/get_build_logs.go
```diff
@@ -490,36 +488,6 @@ func (o *CLILogWriter) BytesLimit() int {
 	return 0
 }
 
-func checkForStagePod(kubeClient kubernetes.Interface, ns string, pr *tektonv1alpha1.PipelineRun, pri *tekton.PipelineRunInfo, stage *tekton.StageInfo) error {
```
hferentschik: why can this be removed?

pkg/cmd/step/create/test_data/step_create_task/add-env-to-default-in-buildpack/tasks.yml
```diff
@@ -20,6 +21,7 @@ items:
         name: version
       resources:
       - name: workspace
+        outputImageDir: ""
```
hferentschik: where is that coming from? That's not related to the new label for meta and build pipeline, right?

hferentschik: OOI, did you use the golden file mechanism to update the test files?

pkg/cmd/step/create/test_data/step_create_task/js_build_pack/pipelinerun.yml
```diff
@@ -22,6 +23,4 @@ spec:
       name: abayer-js-test-repo-build-pack
   serviceAccount: tekton-bot
   timeout: 240h0m0s
-  trigger:
```
hferentschik: that's gone?

pkg/cmd/step/create/test_data/step_create_task/loop-in-buildpack-syntax/pipeline.yml
```diff
@@ -27,7 +28,6 @@ spec:
       inputs:
       - name: workspace
         resource: abayer-jx-demo-qs-master
-    retries: 0
```
hferentschik: ?

pkg/logs/tekton_logging.go
```diff
@@ -194,79 +195,98 @@ func getPipelineRunNamesForActivity(pa *v1.PipelineActivity, tektonClient tekton
 		return tektonPRs.Items[i].CreationTimestamp.Before(&tektonPRs.Items[j].CreationTimestamp)
 	})
 
-	var names []string
+	runs := make(map[string]*v1alpha12.PipelineRun)
```
hferentschik: why is the tekton pipeline import 'v1alpha12'. At best I would expect 'v1alpha1'. That said, I think it might be better to import it as something like 'pipeline' or 'tekton'. That helps me understand the code and is does not change with version upgrades. 'v1alpha12' is in this context just a "useless" version string which does not really help me making sense of the code.

pkg/logs/tekton_logging.go
```diff
@@ -194,79 +195,98 @@ func getPipelineRunNamesForActivity(pa *v1.PipelineActivity, tektonClient tekton
...
-				if err != nil {
-					return errors.Wrapf(err, "failed to get pipeline structure for %s in namespace %s", prName, pa.Namespace)
+	for !loggedAllRunsForActivity {
+		runsByType, err := getPipelineRunsForActivity(pa, t.TektonClient)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get PipelineRun names for activity %s in namespace %s", pa.Name, pa.Namespace)
+		}
+
+		var runToLog *v1alpha12.PipelineRun
+		metaPr := runsByType[tekton.MetaPipeline.String()]
```
hferentschik: 👍

pkg/logs/tekton_logging.go
```diff
@@ -194,79 +195,98 @@ func getPipelineRunNamesForActivity(pa *v1.PipelineActivity, tektonClient tekton
...
+			stagesSeen := make(map[string]bool)
 
-				// Repeat until we've seen pods for all stages
-				for len(allStages) > len(stagesSeen) {
-					pods, err := builds.GetPipelineRunPods(t.KubeClient, pa.Namespace, prName)
-					if err != nil {
-						return errors.Wrapf(err, "failed to get pods for pipeline run %s in namespace %s", prName, pa.Namespace)
-					}
+			// Repeat until we've seen pods for all stages
+			for len(allStages) > len(stagesSeen) {
```
hferentschik: hmm, that's a fair amount of nesting here. By just looking at the code I have a hard time understanding what's happening here? Any chance this would be split up a bit more?

pkg/cmd/controller/test_data/controller_build/completed_generated/cluster_state/pipelineruns.yml
```diff
@@ -10,6 +10,7 @@ items:
       build: "1"
       created-by-prow: "true"
       owner: cb-kubecd
+      pipelineType: meta
```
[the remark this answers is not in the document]
abayer: Oh hey, you're right!

...actually, this should probably be prefixed by `jenkins.io/`. Lemme change that now so that we don't have to deal with it in the future.

pkg/cmd/get/get_build_logs.go
abayer: It was unused, so while I was here, I cleaned things up.

pkg/cmd/step/create/test_data/step_create_task/loop-in-buildpack-syntax/pipeline.yml
abayer: No longer needs to be specified in `Pipeline`s. =)

hferentschik:
> So if you run jx get build logs while the metapipeline is running, it'll give you the metapipeline logs and then continue onto the build/execution logs as well. If you run jx get build logs after the metapipeline has finished successfully, you'll just get the build/execution logs. If the metapipeline failed, you'll still get the metapipeline logs.

> And the long-term storage logs will just have the build/execution logs, assuming the metapipeline passes - again, if the metapipeline fails, the long-term storage will have the metapipeline logs.

Seems a bit odd/random, but if it helps us, for now, that's fine with me.

pkg/cmd/step/create/test_data/step_create_task/loop-in-buildpack-syntax/pipeline.yml
hferentschik: ok

abayer:
The main practical reason is the long term storage - if the metapipeline pod was gone by the time the executed pipeline finished, things go very wrong the way things are set up currently. I also don't think there's a reason to clutter the logs with the successful metapipeline - eventually, I'd like to trim out more of the "generic" steps (like the init containers, the steps inserted by Tekton to copy the workspace from one stage to the next, etc) when they're successful too, and, as I said, it'd be nice to keep the successful metapipeline out of `PipelineActivity` so that it doesn't add ignorable noise to the UI etc. But that's for later. =)

hferentschik:
> The main practical reason is the long term storage - if the metapipeline pod was gone by the time the executed pipeline finished, things go very wrong the way things are set up currently. I also don't think there's a reason to clutter the logs with the successful metapipeline 

It will be, since eventually it might show important app extension logs

> eventually, I'd like to trim out more of the "generic" steps (like the init containers, the steps inserted by Tekton to copy the workspace from one stage to the next, etc) when they're successful too

I am a bit doubtful about this, at least if it is done under the hood. If certain parts of the logs magically disappear (even if they were successful) it could people onto the wrong trail. If I can see certain parts of the logs if a build fails, but not if it succeeds, my alarm bells would already go off. 

That said, having an explicit filter option like "don't show Tekton generic steps" or "only show build logs" makes sense. But imo it needs to be an active choice and if I wanted to there should always be a way to see the full logs, meta pipeline, generic tasks et all.

> and, as I said, it'd be nice to keep the successful metapipeline out of PipelineActivity so that it doesn't add ignorable noise to the UI etc. But that's for later. =)

IMO this is really an UI issue. The backend should have everything available at all time, and the UI decides how and what to display.

pkg/logs/tekton_logging.go
```diff
@@ -26,8 +26,9 @@ import (
 	"github.com/jenkins-x/jx/pkg/client/clientset/versioned"
 	"github.com/jenkins-x/jx/pkg/kube"
 	"github.com/jenkins-x/jx/pkg/log"
+	knativeapis "github.com/knative/pkg/apis"
 	"github.com/pkg/errors"
-	v1alpha12 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
+	tektonapis "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
```
hferentschik: :-)

hferentschik:
What we also should do is to look over the generated names/labels. For example "fromBuildPack" is a bit confusing in the UI.

Issue #5595, opened 2019-09-25
Use the jx version in the version of the jenkins-x-builders


cagiti:
### Summary 
It's not easy to detect what version of jx is within a builders image by looking at the version of the builder. Would be excellent to version the builders based off the jx version, something like:
jx version: `1.2.3`
builders version: `1.2.3-<builders_build_number>`

abayer:
Need to figure out how we can override the next version logic in `pipelinerunner`/the metapipeline to let us update the whole version/tag based on the `jx` version and an incrementing build number.

abayer:
Looks like if we set `pipelineConfig.pipelines.release.setVersion`, we can change the `jx step next-version` call to use `--version (something)`, where `(something)` would need to include the `jx` version...

abayer:
@cagiti Do you know if we have some convenient way of extracting a version for something from `dependency-matrix/matrix.yaml`? The "new" `jx` version is in there, so that seems like the best place to pull it from.

cagiti:
> @cagiti Do you know if we have some convenient way of extracting a version for something from `dependency-matrix/matrix.yaml`? The "new" `jx` version is in there, so that seems like the best place to pull it from.

Not as far as I know, I've used some crude `sed` command for the interim, feels like we should write a step for that to read the yaml version.

abayer:
Exactly what I just decided. =)

axsaucedo:
@cagiti @abayer is it possible to stop the tag being set alltogether? My objective is to stop the tags from being pushed to the github repo with:
```
      setVersion:
        steps:
        - name: next-version
          sh: echo "ignoring..."
```
However when running  a pipeline with this, it would still error out with the following:
```
error: failed to set the version on release pipelines: failed to read file /workspace/source/VERSION
```
I have tried just creating the file to see if that would be enough but it still errors out if the file is empty, and if it has any value it will push a release with that tag. It would be very useful for us to be able to supress the tag being pushed without the error.

PR #5598, opened 2019-09-25
feat: Add a step to get a dependency version from the dependency matrix

abayer:
#### Submitter checklist

#### Description

Can be used either with the version stream or a local directory.

First step of #5595

/assign @cagiti 
/assign @pmuir 

n/a


abayer:
So the idea here is that we'll have an override of `pipelineConfig.pipelines.release.setVersion` in `jenkins-x-builders` that will have something like `jx step next-version --version $(jx step get dependency-version --host=github.com --owner=jenkins-x --repo=jx --dir . | sed 's/.*: \(.*\)/\1/') --use-git-tag-only --tag`.

The output of `jx step get dependency-version --host=github.com --owner=jenkins-x --repo=jx --dir .` will be something like `Version for host github.com, owner jenkins-x, repo jx in matrix at . is: 2.0.784`

cagiti:
👍

abayer:
Moved things around a little and added a test - should be good to go.

PR #5601, opened 2019-09-25
chore: add extra build pipeline env vars that lowercase branch and pr…

rawlingsj:
…eview so we dont need to lowecase them in bash when in a pipeline


abayer:
/lgtm

But I expect `step_create_task_test` to fail due to the new env var - you can automatically update the test files by running `make test1-pkg PKG=./pkg/cmd/step/create TEST=TestGenerateTektonCRDs UPDATE_GOLDEN=1`

rawlingsj:
> But I expect step_create_task_test to fail due to the new env var - you can automatically update the test files by running make test1-pkg PKG=./pkg/cmd/step/create TEST=TestGenerateTektonCRDs UPDATE_GOLDEN=1

awesome - thanks @abayer, I had no idea of that one :)

abayer:
/lgtm

=)

PR #5608, opened 2019-09-26
feat: update the compliance to use sonobuoy v0.16.0

### LINKED ISSUE #5603: K8s 1.16  jx compliance fails
FAILED [sig-cli] Kubectl client [k8s.io] Kubectl run deployment should create a deployment from an image  [Conformance]

ccojocar:
Also remove the dependency to our fork of sonobuoy.

#### Submitter checklist

fixes #5603

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,8 @@ require (
 	github.com/banzaicloud/bank-vaults v0.0.0-20190508130850-5673d28c46bd
 	github.com/beevik/etree v1.0.1
 	github.com/blang/semver v3.5.1+incompatible
+	github.com/briandowns/spinner v1.7.0 // indirect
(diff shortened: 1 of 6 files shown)


ccojocar:
@hferentschik @abayer As a side note. I really like how `sonobuoy` runs the compliance tests. Basically the compliance tests are a large number of ginkgo tests packed in a worker container image, which can be started in parallel. There are different flavours of this image depends which kind the tests you are running. It is possible to get the status, live log stream and the JUnit report when he tests are completed. 

The workflow is something like:
 `jx compliance run` - start the compliance tests in a job in cluster
` jx compliance status|logs` - can be called at any time to check the logs and status 
` jx compliance results` - will collect the rests results as a JUnit report when tests are completed

It would be interesting to see if could benefit form some of these concepts in our BDD tests.

PR #5612, opened 2019-09-26
chore: remove the --file flag from 'jx step credential'

garethjevans:
this command allows a pipeline to download a secret and store it in
a file on the file system.  This is not a recommended approach, secrets 
should be mounted outside of the workDir

### DIFF
--- a/pkg/cmd/step/step_credential.go
+++ b/pkg/cmd/step/step_credential.go
@@ -2,7 +2,6 @@ package step
 
 import (
 	"fmt"
-	"io/ioutil"
 	"sort"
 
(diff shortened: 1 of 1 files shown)


abayer:
/hold

@garethjevans - putting this on hold briefly so that we can get #5613 in ASAP. I'll unhold it once that merges.

garethjevans:
no worries @abayer

abayer:
(this got unheld right after I added `boot-vault` as required, so it went and kicked that off, but it also needs to run all the other contexts since there'd been a merge to master, so I just launched them manually)

abayer:
...and then `boot-vault` failed but that was due to a burp fetching go modules.

PR #5613, opened 2019-09-26
fix: allow vault URL lookup to be avoided and always use Ingress so we can run in CI and immitate what a user does with boot

rawlingsj:


pkg/cmd/clients/factory.go
```diff
@@ -491,7 +493,18 @@ func (f *factory) CreateVaultClient(name string, namespace string) (vault.Client
 	if err != nil {
 		return nil, errors.Wrap(err, "creating vault client")
 	}
-	vaultClient, err := clientFactory.NewVaultClient(name, namespace)
+	// if there's an issue loading a requirements yaml lets just default to automatic
+	useIngressURL := false
+	requirements, _, _ := config.LoadRequirementsConfig("")
+
+	// allows us to override using the default lookup URL for vault and ensure we always use the ingress. Used in CI.
+	if requirements.Vault.DisableURLDiscovery {
```
abayer: Just checking 'cos I'm not clear enough on how all this works - will this mean that `jx install --ng` (or `jx create cluster ... --ng` or `--vault --gitops` etc) gets `useIngressURL` as `true`?

rawlingsj: Just had a chat with @abayer on slack, we're thinking we should remove the `--ng` and `--gitops` flags + disable tests, trying to support those flags is a pita as they never quite worked anyways, `jx boot` should be the way folks install with vault etc

Andrew is going to start raising PRs to avoid the version stream tests for these contexts and we can do more PRs tomorrow to deprecate the flags.

abayer:
/override tekton

One of the nodes on the tekton context's test cluster got itself into a busted state, hanging one of the pipelines. It got very very far along before that happened, and we neeeeeeed this to merge, so I'm overriding.

PR #5616, opened 2019-09-27
chore: add extra build pipeline env vars that lowercase branch and preview so we dont need to lowecase them in bash when in a pipeline

rawlingsj:


pkg/cmd/step/create/step_create_task.go
```diff
@@ -974,6 +978,10 @@ func (o *StepCreateTaskOptions) modifyEnvVars(container *corev1.Container, globa
 			Name:  "PREVIEW_VERSION",
 			Value: "${inputs.params.version}",
 		})
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  "PREVIEW_VERSION_LOWER",
+			Value: strings.ToLower("${inputs.params.version}"),
```
abayer: This isn’t going to do what you want - it’s just lowercasing the already lowercase `${inputs.params.version}`. I think you want to add an addition parameter at https://github.com/jenkins-x/jx/blob/master/pkg/cmd/step/create/step_create_task.go#L1298 and then have this env var reference that parameter.

rawlingsj: ok cool - thanks!  will try that

rawlingsj:
closing until I get time to update it

PR #5628, opened 2019-09-27
feat: add kuberhealthy check option to jx diagnose

macox:

#### Description

Adds a option to `jx diagnose` that checks the health status of a cluster by what is returned from an installed https://github.com/Comcast/kuberhealthy service


abayer:
fyi, the lint failure is legit - code gen needs to be run.

hferentschik:
Do we have an issue for this with some more context?
I am in particular wondering how this command relates to existing commands like `jx status` or `jx diagnose`. How does `jx health` fit in there and when should I use it as opposed to any of the others? Could we not roll this type say into `diagnose`. 

Personally I am concerned about this command proliferation.

garethjevans:
@hferentschik This is going to be part of the Arcalos flow when performing upgrades, we want to be able to assert that "things" in the cluster are healthy before and after an upgrade has run. `jx diagnose` doesn't really do that, although it may make sense to add the output of this command into `jx diagnose`.

garethjevans:
It also relies on having kuberhealthy installed which may not be running in all clusters

macox:
I'll change the cmd to be `jx kuberhealthy` to make it clearer that is uses that tool

hferentschik:
> This is going to be part of the Arcalos flow when performing upgrades, we want to be able to assert that "things" in the cluster are healthy before and after an upgrade has run.

Sounds reasonable. I am not arguing against the functionality that's added. I am concerned about the CLI as a whole. If I were are user and I had commands like `status`, `diagnose` and `health` I would scratch my head wondering what the differences are. I'd just find it confusing.  
 
> jx diagnose doesn't really do that, 

Sure, my question was more whether it could be made part of `diagnose`. Either per default or via a switch? 

Also, would it make sense to use hidden commands? If it is purely Arcalos specific, maybe it should for now be a hidden command.  Cobra allows making commands hidden so that one can call them, but they don't appear in the help system.

pmuir:
Yeah, I think generally having one command here makes, not three. I would suggest we enhance `jx diagnose` to cover everything that `jx status` and this new `jx health` do. `jx diagnose` already calls `jx status`. We can add an options to `jx diagnose` such as `--show` and have it default to `["version", "status", "pods", "ingresses", "secrets", "health"]` so that you can control the output. Obviously if there is no `kuberhealthy` app installed it shouldn't error but perhaps just print an info message telling people how to add it.

hferentschik:
> Yeah, I think generally having one command here makes, not three. I would suggest we enhance jx diagnose to cover everything that jx status and this new jx health do.

I think that's the best way forward

Issue #5432, opened 2019-09-11
Enable Long Term Storage for S3 - boot


dgozalo:
### Summary 
We need to enable the creation and use of `s3` buckets for `LongTermStorage` when booting in `EKS`.

It should generate new bucket names if none were provided or use the provided ones, then assign them to logs / reports / repository storage.

dgozalo:
Blocked while I work on https://github.com/jenkins-x/jx/issues/5645. This will enable fine grained permissions for every AWS service.

PR #5629, opened 2019-09-27
feat: add the requirements file to a ConfigMap on every pipeline run so it can be obtained in other contexts

dgozalo:
#### Submitter checklist

#### Description
When using AWS, you need to provide the AWS_REGION variable in multiple parts of the code (among others).

I needed a way to reliably get this from the cluster configuration so we may as well save the current state of the requirements yaml in a config map so we can access it from everywhere in the code.

This allows for some utility code that can be used to return an implementation based on cloud provider. 

part of #5432


abayer:
I think this may prove handy as a better solution for what #5630 is trying to do...

PR #5630, opened 2019-09-27
fix: looking up vault URL when incluster

rawlingsj:


rawlingsj:
@abayer 
/hold

abayer:
/hold cancel
/lgtm

I’d still like a better solution for the long term, but for the immediate, I can’t come up with something I’m more comfortable with than this, so hey. =)

rawlingsj:
> I’d still like a better solution for the long term, but for the immediate, I can’t come up with something I’m more comfortable with than this, so hey. =)

@abayer thanks for following up with this while I was out, I apreciate it.  I agree with the above BTW.

PR #5640, opened 2019-09-30
chore(test): making boot-vault context gitops enabled

### LINKED ISSUE #5638: Enable gitops managed dev environment for boot-vault context
Enabling gitops allows to write test which are closer to the recommended Jenkins X setup.

hferentschik:
fixes #5638

### DIFF
--- a/jenkins-x-bdd.yml
+++ b/jenkins-x-bdd.yml
@@ -122,7 +122,7 @@ pipelineConfig:
                 image: gcr.io/jenkinsxio/builder-go:${inputs.params.version}
                 command: ./jx/scripts/ci.sh
 
-              - name: stash_xml_report
+              - name: stash-xml-report
                 image: gcr.io/jenkinsxio/builder-go:${inputs.params.version}
                 command: jx
                 args:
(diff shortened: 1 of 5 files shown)


hferentschik:
@abayer this should be good to merge, wdyt? 
the next step then is to fix #5672

jx/bdd/boot-vault/jx-requirements.yml
```diff
@@ -1,6 +1,7 @@
+gitops: true
```
hferentschik: That's the key here. Enable git managed env repos.

jx/bdd/tekton/ci.sh
```diff
@@ -44,7 +44,6 @@ jx step bdd --versions-repo https://github.com/jenkins-x/jenkins-x-versions.git
   --no-delete-app \
   --no-delete-repo \
   --tekton \
-  --tests install \
```
hferentschik: the install task is deprecated. No use to have it anymore

jenkins-x-bdd.yml
```diff
@@ -122,7 +122,7 @@ pipelineConfig:
                 image: gcr.io/jenkinsxio/builder-go:${inputs.params.version}
                 command: ./jx/scripts/ci.sh
 
-              - name: stash_xml_report
+              - name: stash-xml-report
```
hferentschik: using dashes to be consistent with the other names and also because of the UI splitting on '-' (which seems odd)

PR #5651, opened 2019-09-30
fix: Don't use pointers for getPipelineRunsForActivity return

### LINKED ISSUE #5652: Long-term storage of logs fails with multiple existing PipelineRuns of the same owner/repo/branch
This is due to me being dumb with go pointers.

abayer:
#### Submitter checklist

#### Description

I'm not good with pointers, basically. If there are multiple runs for an owner/repo/branch, without this, the pointers returned by `getPipelineRunsForActivity` end up pointing to the last run we looked at, not the one that actually matched. Sigh. I modified an existing test to fail without this fix.

/assign @hferentschik 

fixes #5652

### DIFF
--- a/pkg/logs/tekton_logging.go
+++ b/pkg/logs/tekton_logging.go
@@ -168,7 +168,7 @@ func findLegacyPipelineRunBuildNumber(pipelineRun *tektonapis.PipelineRun) strin
 	return buildNumber
 }
 
-func getPipelineRunsForActivity(pa *v1.PipelineActivity, tektonClient tektonclient.Interface) (map[string]*tektonapis.PipelineRun, error) {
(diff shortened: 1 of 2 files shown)


pkg/logs/tekton_logging.go
```diff
@@ -168,7 +168,7 @@ func findLegacyPipelineRunBuildNumber(pipelineRun *tektonapis.PipelineRun) strin
 	return buildNumber
 }
 
-func getPipelineRunsForActivity(pa *v1.PipelineActivity, tektonClient tektonclient.Interface) (map[string]*tektonapis.PipelineRun, error) {
+func getPipelineRunsForActivity(pa *v1.PipelineActivity, tektonClient tektonclient.Interface) (map[string]tektonapis.PipelineRun, error) {
```
hferentschik: by value is better anyways ;-)

PR #5673, opened 2019-10-01
fix: upgrade pipeline requires updated builder image

### LINKED ISSUE #5658: jx boot upgrade: PR pipeline needs to use upgraded builder
## Summary

During the `jx upgrade boot` process a PR is created containing all the upgrade commits. The PR pipeline can fail as the builder it's using is the existing version, but it actually needs to use the version that we are upgrading to,  ensuring compatibility with changes in the PR.

## Acceptance Criteria
- During the run of the upgrade pipeline the builder image that is in the version stream that is specified in the jx-requirements.yaml should be used

macox:

fixes #5658

### DIFF
--- a/pkg/cmd/upgrade/upgrade_boot.go
+++ b/pkg/cmd/upgrade/upgrade_boot.go
@@ -17,6 +17,7 @@ import (
 	"github.com/spf13/cobra"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 )
 
@@ -36,6 +37,10 @@ var (
 		# create pr for upgrading a jx boot gitOps cluster
 		jx upgrade boot
 `)
+)
+
+const (
+	upgradeVersionStreamRef = "master"
 )
 
 // NewCmdUpgradeBoot creates the command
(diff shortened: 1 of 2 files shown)


pkg/cmd/upgrade/upgrade_boot.go
```diff
@@ -462,3 +472,36 @@ func (o *UpgradeBootOptions) gitProvider(gitInfo *gits.GitRepository) (gits.GitP
...
+		return errors.Wrapf(err, "failed to create version resolver for %s", builder)
+	}
+	imageVersion, err := resolver.ResolveDockerImage(builder)
+	if err != nil {
+		return errors.Wrapf(err, "failed to resolve image %s", builder)
+	}
+
+	log.Logger().Infof("Updating pipeline agent images to %s", util.ColorInfo(imageVersion))
+	projectConf.PipelineConfig.Pipelines.PullRequest.Pipeline.Agent.Image = imageVersion
+	projectConf.PipelineConfig.Pipelines.Release.Pipeline.Agent.Image = imageVersion
```
rawlingsj: wonder if we need to check the individual steps and see if they specify an image and update those too?

rawlingsj: also I think we can have a pipeline that only specifies a `PullRequest` or a `Release` so I think we could hit a `nil` pointer above?  Not 100% sure but a quick test with a couple of sample `jenkins-x.yml` might be worth it to check?

daveconde:
I'll pick this one up

PR #5676, opened 2019-10-02
fix: increase linter deadline to 15 minutes

ccojocar:
#### Submitter checklist

fixes #

### DIFF
--- a/hack/linter.sh
+++ b/hack/linter.sh
@@ -24,7 +24,7 @@ golangci-lint run \
     -E gosec \
     -E gofmt \
     --skip-dirs vendor \
-    --deadline 5m0s \
+    --deadline 15m0s \
     --verbose 
 
 # -E errcheck \


hack/linter.sh
```diff
@@ -24,7 +24,7 @@ golangci-lint run \
     -E gosec \
     -E gofmt \
     --skip-dirs vendor \
-    --deadline 5m0s \
+    --deadline 15m0s \
```
hferentschik: 15min !? Does linting take so long?

ccojocar: It's about 5 min right now. And it seems to timeout occasionally resulting in a build failure.

abayer:
/lgtm

Seems reasonable. We’ll want to watch to see if timeouts still happen with the increased duration, of course - that would be a sign of a legit hanging problem that increasing the timeout further would be unlikely to fix.

Issue #1207, opened 2018-07-04
VERSION is still at 1.0.1


JacobAMason:
Building jx from the latest source always yields a binary which believes it is version 1.0.1 because the VERSION file is behind. Is there a good reason for this?

i0n:
Hello!

I believe this is because the VERSION currently gets injected into the binary as part of the build process, so any self compiled binary will always use 1.0.1.

Maybe there's a better way of doing this, any ideas?

JacobAMason:
That (version string getting injected) was my assumption, seeing the Jenkins bot issuing releases. I hope the release process is powered by Jenkins-x, because that'd be cool.

I'm assuming there is a one-liner that the bot is executing to build this that includes the newest version. My recommendation is to disclose a similar `make` command in the build-from-source instructions that injects a version number (`export VERSION=` or whatnot), and then remove the VERSION file--I'm guessing it's completely ignored by CD, is never incremented by anyone making PRs, and is not going to be noticed by most people building from source until they run `jx version` and get an auto update warning.

As an aside, auto update is neat, but might make this hard to be used with a package manager. The reason I stumbled onto this was because I wasted to install jx through the aur and noticed it didn't exist so I've been trying to set up the package at https://aur.archlinux.org/packages/jx/ I'm not entirely sure how to resolve the two different update strategies and am open to suggestions and assistance on that.

i0n:
This is fixed by https://github.com/jenkins-x/jx/pull/1455 :-)

PR #5720, opened 2019-10-06
fix: Adding clarity to git access error message

### LINKED ISSUE #5718: Wrong GitAccess better message
### Summary 

When we have this kind of error 
```
WARNING: Querying webhooks on enterprise/app-server: GET https://api.github.com/repos/enterprise/app-server/hooks: 404 Not Found []
Creating GitHub webhook for enterprise/app-server for url https://hook-for enterprise.domain.io/hook
error: POST https://api.github.com/repos/enterprise/app-server/hooks: 404 Not Found [] 
```

### Expected behavior

Display a better message to user to say that the account don't have the correct access.

### Kubernetes cluster

EKS 

```
NAME               VERSION
jx                 2.0.827
Kubernetes cluster v1.13.10-eks-5ac0f1
kubectl            v1.16.1
git                2.21.0 (Apple Git-122)
```

nehagup:
fixes #5718

### DIFF
--- a/pkg/gits/github.go
+++ b/pkg/gits/github.go
@@ -445,7 +445,12 @@ func (p *GitHubProvider) CreateWebHook(data *GitWebHookArguments) error {
 	}
 
 	log.Logger().Infof("Creating GitHub webhook for %s/%s for url %s", util.ColorInfo(owner), util.ColorInfo(repo), util.ColorInfo(webhookUrl))
-	_, _, err = p.Client.Repositories.CreateHook(p.Context, owner, repo, hook)
+	_, resp, err := p.Client.Repositories.CreateHook(p.Context, owner, repo, hook)
+	if err != nil {
+		if resp.StatusCode == 404 || resp.StatusCode == 403 {
+			return errors.Wrapf(err, "unable to create webhook for %s/%s - permission denied", owner, repo)
+		}
+	}
 	return err
 }
 


pkg/gits/github.go
```diff
@@ -445,7 +445,13 @@ func (p *GitHubProvider) CreateWebHook(data *GitWebHookArguments) error {
 	}
 
 	log.Logger().Infof("Creating GitHub webhook for %s/%s for url %s", util.ColorInfo(owner), util.ColorInfo(repo), util.ColorInfo(webhookUrl))
-	_, _, err = p.Client.Repositories.CreateHook(p.Context, owner, repo, hook)
+	_, resp, err := p.Client.Repositories.CreateHook(p.Context, owner, repo, hook)
+	if err != nil {
+		if resp.StatusCode == 404 || resp.StatusCode == 403 {
+			log.Logger().Error("Permission denied")
```
ccojocar: Why don't you wrap the error with a message in this case? 

```Go
errors.Wrap(err, "Permission denied")
```

an the let the caller the log the error

hferentschik: That's more consistent with the rest of the code base.

hferentschik: Maybe we can find a way to add a test as well? ;-)

nehagup: Yes, seems better.

nehagup: I'm keeping the statusCode just to make sure it appears only in case of 404 and 403.

pkg/gits/github.go
```diff
@@ -445,7 +445,13 @@ func (p *GitHubProvider) CreateWebHook(data *GitWebHookArguments) error {
 	}
 
 	log.Logger().Infof("Creating GitHub webhook for %s/%s for url %s", util.ColorInfo(owner), util.ColorInfo(repo), util.ColorInfo(webhookUrl))
-	_, _, err = p.Client.Repositories.CreateHook(p.Context, owner, repo, hook)
+	_, resp, err := p.Client.Repositories.CreateHook(p.Context, owner, repo, hook)
+	if err != nil {
+		if resp.StatusCode == 404 || resp.StatusCode == 403 {
+			return errors.Wrap(err, "Permission denied")
```
hferentschik: +1, maybe we can provide a bit more context `errors.Wrapf(err, "unable to create webhook for %s/%s - permission denied", owner, repo)`.
This way when you get the error you have a bit more context and you don't need to figure out what caused a 404.

Do you mind adding some tests as well?

nehagup: Hey, @hferentschik how can I mock a Github provider for the test case?

hferentschik: On the second thought, a `MockGitProvider` won't cut it here. You would need to create the `GitHubProvider` test instance and set its `Client` instance to something you can control, aka some test server from which you can return what you need/want (opposed to talking to GitHub directly).

The way to do this is shown here - https://github.com/google/go-github/blob/master/github/github_test.go#L34
You need this _setup_ function as a helper for your tests. It allows you to spin off an HTTP server for which you can control what requests to handle and how to respond. 

You can see how this is used here - https://github.com/google/go-github/blob/master/github/repos_hooks_test.go

I think it would make for a nice test and open the doors for better testing of `GitHubProvider`. Once the helper method is there we can write more tests using it.

WDYT, do you want to give it a go? Happy to help in case you need some help.

nehagup: Sure, Can I please have another PR for that? It seems it may take some time to learn it.

nehagup:
I'm not sure why this is failing, the build says : 

`k8s.io/client-go@v2.0.0-alpha.0.0.20190115164855-701b91367003+incompatible: invalid pseudo-version: preceding tag (v2.0.0-alpha.0) not found`

seems to be related to #1207

hferentschik:
> Sure, Can I please have another PR for that? It seems it may take some time to learn it.

Fair enough, we can merge this PR and you can have a go at learning the test framework and adding a unit test for it. This will be a good exercise as well.

To merge it though, you first have to fix the Go formatting. We use and enforce the default Go formatting style and it seems that there is an issue with the change you made. You can run `make  fmt` locally and it should fix all formatting violations for you. IMO it is then the best to amend the existing commit and do a forced push to this pull request. Something along these lines:

```bash
$ make fmt
$ git add .
$ git commit --amend --no-edit 
$ git push -f <remote> HEAD
```

Once the linting issue is fixed, I'll merge the PR.

hferentschik:
> I'm not sure why this is failing, the build says :

> k8s.io/client-go@v2.0.0-alpha.0.0.20190115164855-701b91367003+incompatible: invalid 
> pseudo-version: preceding tag (v2.0.0-alpha.0) not found

Where and how do you see this problem?

nehagup:
> > I'm not sure why this is failing, the build says :
> 
> > k8s.io/client-go@v2.0.0-alpha.0.0.20190115164855-701b91367003+incompatible: invalid
> > pseudo-version: preceding tag (v2.0.0-alpha.0) not found
> 
> Where and how do you see this problem?

I get this on `make build`
Even on `make fmt` I get `go: k8s.io/client-go@v2.0.0-alpha.0.0.20190115164855-701b91367003+incompatible: invalid pseudo-version: preceding tag (v2.0.0-alpha.0) not found`

hferentschik:
> I get this on make build
Even on make fmt I get go: k8s.io/client-go@v2.0.0-alpha.0.0.20190115164855-701b91367003+incompatible: invalid pseudo-version: preceding tag (v2.0.0-alpha.0) not found

Which version of Go are you using?

nehagup:
I was using 1.13.x . 
Thanks!

PR #5722, opened 2019-10-06
fix: Removing duplicate jx chart

### LINKED ISSUE #5721: Duplicate presence of Charts.yaml
### Summary 
charts.yaml is present in jx/charts/jx and root.

nehagup:
Signed-off-by: neha gupta <gneha21@yahoo.in>

fixes #5721


hferentschik:
@nehagup, thanks for the pull request. I agree this having to _Chart.yaml_. If possible I would also just like to have one.

That said, I somehow seem to recall that both files are actually used. I am not sure whether we can just delete it. It might not fail directly within `jx`, but there are several downstream repositories which need to get build as well.

@abayer, @pmuir - do you know where this _Chart.yaml_ is potentially used.

hferentschik:
@ccojocar do you know whether _Chart.yaml_ is needed?

ccojocar:
I am not sure where it is used, but if you check the `templates` folder there is only a Release CRD and a service. The release CRD contains a really old version of jx. 

I suppose that also the `templates` folder needs to be remove since it belongs to the Chart file defined in the root directory.

hferentschik:
@nehagup, could you update the PR and also delete the template directory?

nehagup:
@hferentschik  can we also remove values.yaml since we're removing charts.yaml from the root?

hferentschik:
@nehagup, how it is going? Just wondering whether you need a hand getting this ready to be merged?

nehagup:
hey @hferentschik apologies I couldn't catch up(was travelling). I'm working on this actively now.

MarckK:
This looks great, @nehagup!  🎉
@hferentschik is off on PTO; let's ask @pmuir, @daveconde, @abayer for a review.

nehagup:
Thank you all for the support. :)

PR #5730, opened 2019-10-07
test: enable more lint checkers and fixed their warnings

ccojocar:
#### Submitter checklist

fixes #


Makefile.codegen
```diff
@@ -22,6 +22,7 @@ test-codegen: ## Test the code geneator
 # Anything generated by this target should be checked in
 generate: build-codegen install-generate-deps generate-mocks generate-openapi generate-client ## Generate the Go code (crds, mocks, openapi, client)
 	@$(MAKE) fmt
+	@$(MAKE) importsfmt
```
hferentschik: +1

Makefile
```diff
@@ -256,6 +257,12 @@ fmt: ## Format the code
       		echo "Fixed formatting for: $(FORMATTED)"; \
       	fi
 
+.PHONY: importfmt
+importsfmt:
+	# $(GO_NOMOD) get golang.org/x/tools/cmd/goimports
+	@echo "Formatting the imports..."
+	goimports -w $(GO_DEPENDENCIES)
```
hferentschik: where is goimports coming from? Don't we need a target to install it?

hack/generate.sh
```diff
@@ -6,6 +6,11 @@ RED='\033[0;31m'
 GREEN='\033[0;32m'
 RESET='\033[0m'
 
+if ! [ -x "$(command -v goimports)" ]; then
+	echo "Installing goimports"
+    go get golang.org/x/tools/cmd/goimports
```
hferentschik: why not install this as part of the Makefile? Now you are forced to run this script  before you could run code generation targets.

ccojocar: I can add it also there.

ccojocar: We don't seems to install any tools with `go` in the makefile.

Makefile
ccojocar: It is installed in `generate.sh` for pipeline.

Makefile
```diff
@@ -247,6 +247,7 @@ release-distro:
 clean: ## Clean the generated artifacts
 	rm -rf build release dist
 
+.PHONY: clean
```
hferentschik: +1

ccojocar: fixed

hack/generate.sh
hferentschik: We used to and imo should also. The alternative is to let the Makefile rules somehow call the scripts which is not so nice imo.

The problem is that the Makefile is not self-contained anymore. You cannot call all the target without error. For example, you would need to call the _generate.sh_ script to then call `make generate`. If you don't know that it is very hard to figure out. Most likely you will assume the build is broken.

I think our mixture and separation of Makefile and shell scripts gets more and more confusing.

ccojocar: I can create an issue to make the `make geneate` self-contained.

ccojocar: I get this stuff locally:
```
CGO_ENABLED=0 GO111MODULE=on go build  -o build/codegen cmd/codegen/codegen.go
GO111MODULE=on go get github.com/petergtz/pegomock/...@v2.2.0
go: finding github.com/petergtz/pegomock v2.2.0
Generating Mocks using pegomock
GO111MODULE=on go generate ./...
All Go files properly formatted
Generating OpenAPI structs for Kubernetes Clients
./build/codegen --generator-version 36ebc4887cdc openapi --output-package=pkg/client --input-package=github.com/jenkins-x/jx/pkg/apis --group-with-version=jenkins.io:v1
INFO[0000] installing k8s.io/kube-openapi/cmd/openapi-gen with version 36ebc4887cdc via 'go get' to /Users/cosmin/.jx/codegen/go  app=codegen
error installing kubernetes openapi tools: error running go get -u -d k8s.io/kube-openapi/cmd/openapi-gen/..., output # cd /Users/cosmin/.jx/codegen/go/src/golang.org/x/tools; git pull --ff-only
fatal: Not possible to fast-forward, aborting.
package golang.org/x/tools/imports: exit status 128: failed to run 'go get -u -d k8s.io/kube-openapi/cmd/openapi-gen/...' command in directory '', output: '# cd /Users/cosmin/.jx/codegen/go/src/golang.org/x/tools; git pull --ff-only
fatal: Not possible to fast-forward, aborting.
package golang.org/x/tools/imports: exit status 128'
make: *** [codegen-openapi] Error 1
```

And is due to fact that code generator doesn't like a newer version of the `tools`. It happens manly with a Go version newer than the one from the pipeline.

PR #5731, opened 2019-10-07
fix: Switch bdd context to create a cluster

abayer:
#### Submitter checklist

#### Description

The `bdd` context has been flaky lately in the shared cluster, so let's move it to new clusters for each run.

n/a


abayer:
/hold

Holding until I've verified that the contents of `jenkins-x-static.yml` actually work, at which point I'll revert the overwrite of `jenkins-x-bdd.yml`.

abayer:
/hold cancel

Unholding - not worth the hassle of context renaming, really.

ccojocar:
@abayer I guess you need to add `static-e2e-tests` to weasel cluster dev env. It doesn't seem to trigger.

PR #5740, opened 2019-10-08
feat: make step boot vault use the version stream defined in jx-requirements.yaml

### LINKED ISSUE #5660: jx boot upgrade: Vault operator uses existing version stream from team settings
## Summary

During the build pipeline for the PR created by the upgrade process the vault operator fails as version stream ref in the team settings is updated after the vault operator is updated.

## Acceptance Criteria
- When upgrading in a boot pipeline the version stream ref should be updated in the team settings before the vault operator is update.

dgozalo:

#### Submitter checklist

#### Description
Right now we are using the versions repo and ref defined in the team settings in the `dev` environment CRD.

We need to use the VersionsResolver created from the current `jx-requirements.yaml` instead so we can boot using the new version when upgrading the cluster. 

fixes #5660

### DIFF
--- a/pkg/cmd/step/verify/step_verify_preinstall.go
+++ b/pkg/cmd/step/verify/step_verify_preinstall.go
@@ -258,7 +258,13 @@ func (o *StepVerifyPreInstallOptions) Run() error {
 		} else {
 			log.Logger().Info("Running in cluster, not recreating permissions")
 		}
+	}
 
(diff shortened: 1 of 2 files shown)


abayer:
Looks very similar to #5741? cc @macox and @ccojocar as well

ccojocar:
I am not sure if there is something more in the PR. I approved the  #5741.

dgozalo:
Modified the code to include Mark's change, added the update to the versions ref in `TeamSettings`.

PR #5741, opened 2019-10-08
fix: specify version stream for vault operator

macox:

#### Submitter checklist

#### Description

The vault operator is currently always using the default version stream to lookup its image tag.  This can lead to the incorrect one being used (or a failure occurring as the ref of the incorrect version stream being doesn't exist)


abayer:
/assign @ccojocar 

Assigning Cosmin since I know he touched this recently.

PR #5756, opened 2019-10-09
[wip] fix: push modifications to boot repo directly to master

### LINKED ISSUE #5672: Initial local 'jx boot' execution should push config directly to master
Currently, when executing the initial `jx boot` locally, a boot config (based on the input of the user) is applied and a PR created against the dev repository. This PR does not fulfil any purpose since all the changes are already applied. It created two unnecessary builds (one for the PR and one master build when the PR merges). The outcome of this is, however, a noop in terms of the cluster configuration.

Instead of creating the PR, the changes should directly be merged into master. 

A side effect of this is, that the _boot-vault_ context in jx, does create a git managed cluster, but it does not merge the initial PR. If one then writes a BDD test which tries to verify some gitops operations involving the dev repo (eg installing an app), you get an error of the form:

```
Showing logs for build ******************/environment-pr-5643-26-boot-vault-dev/master #1 promotion stage release and container step-verify-preinstall
error: please specify 'project' in **-requirements when running  in  batch mode
Pipeline failed on stage 'release' : container 'step-verify-preinstall'. The execution of the pipeline has stopped.
     RUNNER:Execution completed with exit code 0
```

hferentschik:
fixes #5672


pkg/cmd/step/verify/step_verify_environments.go
```diff
@@ -312,20 +252,119 @@ func (o *StepVerifyEnvironmentsOptions) createEnvGitRepository(name string, requ
...
+	if err != nil {
+		return err
+	}
+
+	dir, err := filepath.Abs(o.Dir)
+	if err != nil {
+		return errors.Wrapf(err, "resolving %s to absolute path", o.Dir)
+	}
+
+	environmentRepo, err := o.createDevEnvironmentRemote(envGitInfo, dir, fromGitURL, fromBaseRef, !public, requirements, provider, gitter)
```
hferentschik: @pmuir one thing I don't understand is why we create the remote and push the original repo data to it, but ony in the `jx boot` case add a commit with the modifications? 

I also find it odd that we push something to the remote, then fork and pull, add a commit and then push again. Could we not do everything locally on then do a single push? 

Apart from this, afaict the refactored code does not what the original code did, except that the changes to config are pushed to master instead of creating a pull request. 

Still cannot shake off the feeling that we only should push once.

PR #5787, opened 2019-10-11
feat: enable S3 as a LongTermStorage target and refactor code to work with two providers

### LINKED ISSUE #5432: Enable Long Term Storage for S3 - boot

dgozalo:
feat: rebased to get requirements from TeamSettings

feat: rebased and modified code

feat: enable S3 as a LongTermStorage target and refactor code to work with two providers

feat: rebased to get requirements from TeamSettings

feat: finalize work on legacy bucket provider

feat: added tests and interfaces

feat: added some tests around buckets

feat: added buckets mocks

feat: added the last tests

chore: remove unsused code

chore: linting and formatting

#### Submitter checklist

#### Description
- Added a way to create S3 buckets when booting EKS.
- Added two more functions to `buckets.Provider` so we can use it as the base for all others.
- Fixed S3 Bucket Provider and implemented the two new functions.
- Implemented the two new functions in the GKE bucket provider.
- Created `LegacyBucketProvider` for `gocloud` library code that we used to use. This will be used by non boot clusters and/or by providers that we haven't created an implementation of - `buckets.Provider` for. 
- Added a way to obtain the bucket provider based on the `jx-requirements.yaml` file stored as a blob in the `dev` environment's `TeamSettings` or return the default if we can't find it (non boot cluster).
- Added several tests for providers and collectors.

fixes #5432


dgozalo:
/hold

Holding it until we cut a release candidate for CJXDv4

dgozalo:
/hold cancel

We can now go ahead and merge this.

dgozalo:
/override boot-vault

For some reason it's never reporting back as OK, but the BDD tests pass and the cluster is already destroyed.

dgozalo:
/hold 
Need to figure out the problem with boot-vault

abayer:
We now believe that whatever's causing `boot-vault` to hang on a `jx step stash` is a legit issue - we're already messing up our stashes for `bdd`, `boot-vault`, and `tekton` clusters due to them stashing while in the cluster context of the test cluster, resulting in stashing to things like `gs://jx-prod-logs/jenkins-x/e2e-tests/jenkins-x/jx/gh-pages/31/workspace/source/build/reports/junit.html`. #5915 nukes the `~/.kube/config` before stashing so that it's actually looking at the prod cluster's storage config, rather than the test cluster's, but the suspicion is that changes in this PR are resulting in `jx step stash` hanging when using boot and a cluster with git repo storage set up. So...yeah. More tests needed. =)

hferentschik:
@dgozalo @abayer - so what the state on this PR?

dgozalo:
/hold cancel

Tested it works with all contexts

pkg/cloud/amazon/storage/bucket_provider.go
```diff
@@ -22,12 +26,11 @@ import (
...
 	api          s3iface.S3API
+	uploader     s3manageriface.UploaderAPI
+	downloader   s3manageriface.DownloaderAPI
 }
 
-func (b *AmazonBucketProvider) s3() (s3iface.S3API, error) {
-	if b.api != nil {
-		return b.api, nil
-	}
+func (b AmazonBucketProvider) createAWSSession() (*session.Session, error) {
```
ccojocar: nit: missing `*` from `func (b *AmazonBucketProvider)`

pkg/cloud/buckets/interface.go
```diff
@@ -1,8 +1,13 @@
...
 
+import "bufio"
+
 // Provider represents a bucket provider
+//go:generate pegomock generate github.com/jenkins-x/jx/pkg/cloud/buckets Provider -o mocks/buckets_interface.go
 type Provider interface {
 	// CreateNewBucketForCluster creates a new dynamically named bucket
 	CreateNewBucketForCluster(clusterName string, bucketKind string) (string, error)
 	EnsureBucketIsCreated(bucketURL string) error
+	UploadFileToBucket(bytes []byte, outputName string, bucketURL string) (string, error)
```
ccojocar: I think using a `Reader` interface might be more scalable than `[]bytes`.

pkg/cloud/amazon/storage/bucket_provider.go
dgozalo: That's actually fine, don't need to modify anything from the struct in that function.

pkg/cloud/buckets/interface.go
dgozalo: Good idea, just changed it.

pkg/cloud/amazon/storage/bucket_provider.go
ccojocar: Yeah, it's fine because the compiler will take care of and covert it into a pointer when it is required, but usually linters do complain about this inconsistency.

PR #5795, opened 2019-10-12
fix: lookup boot config ref unless it's specified

macox:
#### Description

Currently if you specify a git-url for boot config but no git-ref it will use the master boot config. This change ensures that it will lookup the boot config version that is specified in the version stream.


abayer:
Looks right to me, but would like @ccojocar and/or @dgozalo to take a look since they're more familiar with this particular bit than me. =)

PR #5823, opened 2019-10-15
fix: only create/update dev environment repo when within `jx boot`

### LINKED ISSUE #5819: step_verify_environments fails when run in pipeline
When running the pipeline builds for the boot config repository, the following error occurs:

```
Validating git repository for dev environment at URL https://github.com/ccojocar-org/environment-central-dev.git
error: [the environment variable BASE_CONFIG_REF must be specified]

Pipeline failed on stage 'release' : container 'step-verify-jenkins-x-environment'. The execution of the pipeline has stopped.
```

The problem is that _BASE_CONFIG_REF_ is only set in the case of executing `jx boot`. The variable is not set when `verify env` is called as part of a pipeline build. 

Currently, the code looks like this:

```golang
func (o *StepVerifyEnvironmentsOptions) createDevEnvironmentRepository(envGitInfo *gits.GitRepository, public bool, provider gits.GitProvider, gitter gits.Gitter, requirements *config.RequirementsConfig) error {
	fromGitURL, fromBaseRef, err := o.readEnvironment()
	if err != nil {
		return err
	}

	dir, err := filepath.Abs(o.Dir)
	if err != nil {
		return errors.Wrapf(err, "resolving %s to absolute path", o.Dir)
	}

	environmentRepo, err := o.createDevEnvironmentRemote(envGitInfo, dir, fromGitURL, fromBaseRef, !public, requirements, provider, gitter)
	if err != nil {
		return errors.Wrapf(err, "creating remote for dev environment %s", envGitInfo.Name)
	}

	if o.isJXBoot() {
		err = o.pushDevEnvironmentUpdates(environmentRepo, dir, provider, gitter)
		if err != nil {
			return errors.Wrapf(err, "error updating dev environment for %s", envGitInfo.Name)
		}

		// Add a remote for the user that references the boot config that they originally used
		err = gitter.SetRemoteURL(dir, "jenkins-x", fromGitURL)
		if err != nil {
			return errors.Wrapf(err, "setting jenkins-x remote to boot config %s", fromGitURL)
		}
	}
	return nil
}
```

The error occurs in 

```golang
fromGitURL, fromBaseRef, err := o.readEnvironment()
```

If _BASE_CONFIG_REF_ is not set `step verify` will fail. 

There are several options to potentially resolve this:

* Make sure that _BASE_CONFIG_REF_ is also set when the pipeline runs. Questions is what it should be set to and where this value is coming from.
* Move the check wether we are running `jx boot` if o.isJXBoot() {` higher up and only execute the whole function if we are in `jx boot`. This implies that when `step verify env` is called in the pipeline we are not recreating and re-pushing to the dev repo.

hferentschik:
fixes #5819

* only call `handleDevEnvironmentRepository` is in `jx boot`. This effectively pulls up ` o.isJXBoot()` one level 
* Let helpers#DuplicateGitRepoFromCommitish return an error if the target repo exists which is more in line with one would expect from the function name

### DIFF
--- a/pkg/cmd/step/verify/step_verify_environments.go
+++ b/pkg/cmd/step/verify/step_verify_environments.go
@@ -259,12 +259,12 @@ func (o *StepVerifyEnvironmentsOptions) createEnvironmentRepository(name string,
 	gitter := o.Git()
 
 	if name == kube.LabelValueDevEnvironment || environment.Spec.Kind == v1.EnvironmentKindTypeDevelopment {
-		if requirements.GitOps {
+		if o.isJXBoot() && requirements.GitOps {
 			provider, err := envGitInfo.CreateProviderForUser(server, userAuth, gitKind, gitter)
 			if err != nil {
 				return errors.Wrap(err, "unable to create git provider")
 			}
-			err = o.createDevEnvironmentRepository(envGitInfo, public, provider, gitter, requirements)
+			err = o.handleDevEnvironmentRepository(envGitInfo, public, provider, gitter, requirements)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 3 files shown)


hferentschik:
Note, this is semantically equivalent to the original.

Issue #3713, opened 2019-04-23
jx get preview --current --url-only


vfarcic:
### Summary 

`jx get preview` cannot retrieve only the address, instead of a log entry.

### Steps to reproduce the behavior

```bash
jx get preview --current -o jsonpath="{.msg}"
```

### Expected behavior

Only the message (it contains only the URL).

### Actual behavior

Log output. `-o` is ignored (not implemented).

deanesmith:
Hi @vfarcic, is this still an issue?

vfarcic:
Now it's worse and I cannot even execute `jx get preview --current`. Now I'm getting `error: No $JOB_NAME defined for the current pipeline job to use`.

hferentschik:
@vfarcic, where and how do you want to run `jx get preview --current`? Have you seen the help for this command:

```
# View the current preview environment URL
# inside a CI pipeline
jx get preview --current
```

The intent of the command is to run in the pipeline where it has the available context information, aka the "current" build. 

What is the semantics you would expect from `jx get preview --current`? `jx get preview` can be run from anywhere listing all previews, what do you expect from `--current`.

My guess is that you expect this to be run from a given checkout of a project and you want to get the latest (in terms of time) deployed preview for this project.

hferentschik:
@vfarcic could you also explain:

> jx get preview cannot retrieve only the address, instead of a log entry.

With "address", you mean the preview URL right? And with log entry a row in the default output of `jx get preview`?

> jx get preview --current -o jsonpath="{.msg}"

I am not following on the -o option? I guess you envision `jx get preview --current ` to display something in the same form as the default output, but filtered by 'current' (see discussion above) and then you'd like an additional -o option to select a specific value/column of this output, right?

hferentschik:
Ahh, and in the title you mention `--url-only`. Is this an alternative for `-o`? Would you prefer a specific `--url-only` or a more generic `-o`?

vfarcic:
We can close this issue. It's been over 6 months since I created it and I do not even remember anymore what's it about.

PR #5824, opened 2019-10-15
test: adds get preview test

mauromorales:
I'm trying to follow up on issue #3713. With this initial PR I'd like to add test cases for the current state of the implementation. Currently I managed to create two cases for the failed scenarios, however I wasn't sure how to create an environment and a preview to test the positive case. Guidance on how to do this would be appreciated and since this is my first PR also any comments and suggestions on the added tests.

Once tests are in place my plan would be to fix the issue reported by the user.

### DIFF
--- a/pkg/cmd/get/get_preview_test.go
+++ b/pkg/cmd/get/get_preview_test.go
@@ -1,0 +1,149 @@
+package get_test
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/jenkins-x/jx/pkg/kube"
+
+	"github.com/jenkins-x/jx/pkg/cmd/get"
+	"github.com/jenkins-x/jx/pkg/cmd/opts"
+	"github.com/jenkins-x/jx/pkg/cmd/testhelpers"
+	"github.com/jenkins-x/jx/pkg/gits"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	jenkinsv1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	clientmocks "github.com/jenkins-x/jx/pkg/cmd/clients/mocks"
+	helm_test "github.com/jenkins-x/jx/pkg/helm/mocks"
+	resources_test "github.com/jenkins-x/jx/pkg/kube/resources/mocks"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestGetPreview(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Get Preview Suite")
+}
+
+var _ = Describe("get preview", func() {
+	Describe("CurrentPreviewUrl()", func() {
+		var (
+			originalRepoOwner  string
+			originalRepoName   string
+			originalJobName    string
+			originalBranchName string
+
+			devEnv *jenkinsv1.Environment
+
+			err    error
+			stdout []byte
+		)
+
+		BeforeSuite(func() {
+			originalRepoOwner = os.Getenv("REPO_OWNER")
+			originalRepoName = os.Getenv("REPO_NAME")
+			originalJobName = os.Getenv("JOB_NAME")
+			originalBranchName = os.Getenv("BRANCH_NAME")
+		})
+
+		AfterSuite(func() {
+			os.Setenv("REPO_OWNER", originalRepoOwner)
+			os.Setenv("REPO_NAME", originalRepoName)
+			os.Setenv("JOB_NAME", originalJobName)
+			os.Setenv("BRANCH_NAME", originalBranchName)
+		})
+
+		JustBeforeEach(func() {
+			commonOpts := &opts.CommonOptions{}
+			commonOpts.SetDevNamespace("jx")
+
+			factory := clientmocks.NewMockFactory()
+			user := jenkinsv1.User{
+				ObjectMeta: v1.ObjectMeta{
+					Namespace: "jx",
+					Name:      "test-user",
+				},
+				Spec: jenkinsv1.UserDetails{
+					Name:     "Test",
+					Email:    "test@test.com",
+					Accounts: make([]jenkinsv1.AccountReference, 0),
+				},
+			}
+
+			testhelpers.ConfigureTestOptionsWithResources(commonOpts,
+				[]runtime.Object{},
+				[]runtime.Object{&user, devEnv},
+				&gits.GitFake{CurrentBranch: "job"},
+				&gits.FakeProvider{},
+				helm_test.NewMockHelmer(),
+				resources_test.NewMockInstaller(),
+			)
+
+			commonOpts.SetFactory(factory)
+
+			options := &get.GetPreviewOptions{
+				GetEnvOptions: get.GetEnvOptions{
+					GetOptions: get.GetOptions{
+						CommonOptions: commonOpts,
+					},
+				},
+			}
+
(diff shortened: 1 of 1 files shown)


abayer:
Yay! Tests! =) I'll give this a more detailed look in a bit.

abayer:
Don't worry about that first batch of failures - that's just 'cos your new commit was pushed before the builds triggered by the previous commit had fully started.

pkg/cmd/get/get_preview_test.go
```diff
@@ -0,0 +1,160 @@
...
+			Name:      "test-user",
+		},
+		Spec: jenkinsv1.UserDetails{
+			Name:     "Test",
+			Email:    "test@test.com",
+			Accounts: make([]jenkinsv1.AccountReference, 0),
+		},
+	}
+
+	testhelpers.ConfigureTestOptionsWithResources(commonOpts,
```
abayer: I think you need something like this in `TestCurrentPreviewUrlWithoutEnvVars` too.

pkg/cmd/get/get_preview_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	options := &get.GetPreviewOptions{
+		GetEnvOptions: get.GetEnvOptions{
+			GetOptions: get.GetOptions{
+				CommonOptions: commonOpts,
+			},
+		},
+	}
+
+	want := "No Preview for name: jx-testing-jx-testing-job"
```
abayer: This is failing because it's still using `$JOB_NAME` behind the scenes - I think you probably want to set that to `job` for this test, reverting it back afterwards.

pkg/cmd/get/get_preview_test.go
```diff
@@ -0,0 +1,160 @@
...
+			GetOptions: get.GetOptions{
+				CommonOptions: commonOpts,
+			},
+		},
+	}
+
+	want := "http://example.com"
+
+	if got, err := options.CurrentPreviewUrl(); err != nil {
+		t.Errorf("CurrentPreviewUrl() didn't expect error but errored with: `%s'", err)
```
abayer: If I'm understanding the test failure correctly, I think this also needs `JOB_NAME` to be overridden like in the above test.

pkg/cmd/get/get_preview_test.go
abayer: Actually, it may be that you need to unset `JOB_NAME` or set it to `""`...

dgozalo: In fact, $JOB_NAME is set in the container running the tests I guess?

abayer: Yup, hence the test failing with `jx-testing-jx-testing-pr-5824` as the preview name it's looking for.

mauromorales: @abayer @dgozalo is $JOB_NAME set in the container only? I think this is part of the problem the user is reporting. How is this value supposed to be set when calling it from the command line from a users' machine which is interacting with a cluster?

abayer: In that context, `$JOB_NAME` would be unset...

mauromorales: @abayer is there a plan to let the user set it via a flag in the cli or some other way? I'm wandering what is the right flow here. E.g. user sees the error:  `error: No $JOB_NAME defined for the current pipeline job to use`. is he now supposed to call `JOB_NAME=x jx get preview --current`? and will the user know what job name stands for?

abayer: @mauromorales I honestly have no idea. =)

abayer:
I'll try to play with the tests that are failing myself later today.

mauromorales:
Thanks @abayer I think I found the missing piece, I needed to also set the BRANCH_NAME env var. I'll try now to refactor it and remove the WIP tag

pkg/cmd/get/get_preview_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+	jenkinsv1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	clientmocks "github.com/jenkins-x/jx/pkg/cmd/clients/mocks"
+	helm_test "github.com/jenkins-x/jx/pkg/helm/mocks"
+	resources_test "github.com/jenkins-x/jx/pkg/kube/resources/mocks"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestCurrentPreviewUrlWithoutEnvVars(t *testing.T) {
+	ro := os.Getenv("REPO_OWNER")
```
hferentschik: There seems to be a lot of duplidated setup code in each test. I'd extract a setup and teardown function which move this out of the test function. teardown can be called in the defer clause. 

This helps draw attention to the actual test.

hferentschik: Personally, I'd write Ginkgo test, since imo it more expressive, but that might be a matter of case.

pkg/cmd/get/get_preview_test.go
hferentschik: I think the ''--current" flag is not supposed to be used from the local machine. This is also implied by the help for this command:

> \# View the current preview environment URL
> \# inside a CI pipeline
> jx get preview --current

Locally we don't have the information we need. It's not just the _JOB_NAME_ which is a problem, but also the repo owner and repo name. We cannot extract them from the environment. You would either have to add explicit options for this or assume that this command gets run from the repository for which you want to retrieve the preview information in which case you could get the repo information from the git info. 

Also, what does 'current' even mean in this context?  The issue is imo not clearly defined and we need to discuss first on the issue what the path forward is. 

The '--current' flag is unfortunate imo, since it does not fit the semantics of the command. `jx get previews` is intended to be independent of the directory you run it from. It accesses the env settings in the cluster. A '--current' flag needs more context and would need to be run from an actual project directory.

I'd remove the whole flag. This might be a bit dangerous in case this flag is used in some existing build packs of ours. If we need to keep it, we could check whether we are in Cluster and if not return a warning to the user that this option is not available locally.

hferentschik: I added several comments on the issue https://github.com/jenkins-x/jx/issues/3713 and removed the "good first issue" label. This is not trivial at all.

mauromorales:
Thanks for the comments @hferentschik. I've dried the tests using Ginkgo as you suggested. Please have a look.

pkg/cmd/get/get_preview_test.go
mauromorales: Makes sense, let's have the conversation on the issue then and I'll focus on having the tests just reflect the current implementation

hferentschik:
> Thanks for the comments @hferentschik. I've dried the tests using Ginkgo as you suggested. Please have a look.

Will do.

Could you in the meantime maybe squash the commits and use a commit message using the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.3/) format? We are enforcing this format since we use the commit message meta data to create changelogs. 

See also https://jenkins-x.io/docs/contributing/code/#the-commit-message

pkg/cmd/get/get_preview_test.go
```diff
@@ -0,0 +1,149 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestGetPreview(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Get Preview Suite")
+}
+
+var _ = Describe("get preview", func() {
```
hferentschik: 👍 Looks nice. IMO much more concise than the earlier tests.

abayer:
/meow space

Hooray!

PR #5831, opened 2019-10-15
fix(logs): For completed builds, only check for pods that ran

### LINKED ISSUE #5830: controllerbuild hangs after a few minutes
I'm seeing this in prod - it seems to be somehow due to https://github.com/jenkins-x/jx/commit/694a6e5d456b8782f654f74dcdf191e6f08642f7 but I'm not sure how exactly yet. Working on it.

abayer:
#### Submitter checklist

#### Description

Without this, we loop forever trying to find the pod(s) for stages that were never actually run.

fixes #5830


abayer:
dockerhub is 503ing, so that's fun. =)

PR #5838, opened 2019-10-16
fix: better API generation

jstrachan:
which also supports generation of reference docs for `jx-requirements.yml` and `jenkins-x.yml`


abayer:
yay!

abayer:
/hold

Holding briefly so that #5833 gets through.

abayer:
/hold cancel

And #5833 is in, so unholding. Sorry for the delay. =)

PR #5845, opened 2019-10-16
fix: Add support for eks arns as context names

### LINKED ISSUE #5841: `jx install` fails to configure docker registry on EKS
### Summary

I'm trying to install a new instance of Jenkins X on an EKS cluster. I'm using the same process I've been using for months that's been working fine, but now I'm encountering a cryptic error.

1. `jx install --provider=eks --no-default-environments --git-provider-url=*** --static-jenkins`
1. 'No' to wildcard DNS
1. 'Yes' to wait for IP resolution
1. 'Yes' to magic DNS name
1. Configure git options

### Expected behavior

Install completes successfully

### Actual behavior

```
WARNING: unable to calculate docker registry values - parsing the current Kubeernetes context kubernetes: unable to parse kubernetes as <cluster_name>.<region>.*
```

happens immediately after starting `jx install`, and

```
error: configuring the docker registry: configure cloud provider docker registry: parsing the current Kubeernetes context kubernetes: unable to parse kubernetes as <cluster_name>.<region>.*
```

happens right after finishing git server configuration.

### Jx version

```
NAME               VERSION
jx                 2.0.876
Kubernetes cluster v1.13.11-eks-5876d6
kubectl            v1.16.1
git                2.23.0
Operating System   Ubuntu 16.04.6 LTS
```

also tested on

```
NAME               VERSION
jx                 2.0.876
Kubernetes cluster v1.13.11-eks-5876d6
kubectl            v1.15.1
helm client        Client: v2.12.3+geecf22f
git                2.21.0 (Apple Git-122)
Operating System   Mac OS X 10.14.6 build 18G95
```

### Jenkins type

- [x] Classic Jenkins

### Kubernetes cluster

EKS via custom creation scripts

Ubuntu 16.04 and macOS 10.14.6

### LINKED ISSUE #5844: Assumes wrong format for EKS cluster context
So I'm trying to run `jx boot` on an EKS cluster we created via Terraform. We also set up our context file by running `aws eks update-kubeconfig`. 
This tool sets up the context to the cluster where the context name is the ARN of the eks cluster.
`jx boot` seems to expect the clusters context name to be in the format of `<cluster_name>.<region>.*` so i keep running into the following error:
```
error: there was a problem obtaining the current cluster name and region: parsing the current Kubeernetes context arn:aws:eks:us-east-1:account-id:cluster/andrew-test-dev: unable to parse arn:aws:eks:us-east-1:account-id:cluster/andrew-test-dev as <cluster_name>.<region>.*
```
I think the function in question is https://github.com/jenkins-x/jx/blob/874ec4a52a22e6147ef04365d6001d0ea7ed7790/pkg/cloud/amazon/common.go#L68

My jx-requirements.yaml has the following block in it:
```
cluster:
  azure: {}
  clusterName: andrew-test-dev
  gitKind: github
  gitName: github
  gitServer: https://github.com
  namespace: jx
  provider: eks
```

I'm running:
- jx 2.0.871
- kubectl client 1.16.1
- kubectl server 1.14.7

Note: Apologies for not following the standard Issues template, but Github never gave it to me when i used their "Create issue from code" feature.

andrewhertog:
#### Submitter checklist

#### Description
The regex to determine the EKS cluster name and region does not support the EKS ARN which is the format of the context name generated by `aws eks update-kubeconfig` 
This issue appears when running `jx boot` and `jx install`

Added functionality where an attempt is made to retrieve the eks cluster name from AWS via the AWS eks sdk looping through all the eks clusters until one is found that has an endpoint matching the one in the current kubeconfig context.

Also includes a small typo fix

I was unable to run `make build` as i received the following errors:
```
make build
CGO_ENABLED=0 GO111MODULE=on go build -ldflags " -X github.com/jenkins-x/jx/pkg/version.Version=2.0.882-dev+66cb8bdc8 -X github.com/jenkins-x/jx/pkg/version.Revision='66cb8bdc8' -X github.com/jenkins-x/jx/pkg/version.Branch='andrewhertog/eks-arn-support' -X github.com/jenkins-x/jx/pkg/version.BuildDate='20191016-14:52:57' -X github.com/jenkins-x/jx/pkg/version.GoVersion='1.13.1'" -o build/jx cmd/jx/jx.go
go: k8s.io/client-go@v2.0.0-alpha.0.0.20190115164855-701b91367003+incompatible: invalid pseudo-version: preceding tag (v2.0.0-alpha.0) not found
make: *** [build] Error 1
```
This is my first time ever working on a go project so any feedback is welcome

fixes #5844, fixes #5841

Update: Fixed this by changing from Go1.13 to Go1.11
Update: Added AWS sdk lookup for cluster name

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -171,5 +171,3 @@ replace github.com/Azure/azure-sdk-for-go => github.com/Azure/azure-sdk-for-go v
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v10.15.5+incompatible
 
 replace github.com/banzaicloud/bank-vaults => github.com/banzaicloud/bank-vaults v0.0.0-20190508130850-5673d28c46bd
-
-go 1.13
--- a/go.mod
+++ b/go.mod
@@ -171,3 +171,5 @@ replace github.com/Azure/azure-sdk-for-go => github.com/Azure/azure-sdk-for-go v
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v10.15.5+incompatible
 
 replace github.com/banzaicloud/bank-vaults => github.com/banzaicloud/bank-vaults v0.0.0-20190508130850-5673d28c46bd
+
+go 1.13
--- a/pkg/cloud/amazon/common.go
+++ b/pkg/cloud/amazon/common.go
@@ -9,6 +9,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/eks"
 	"github.com/jenkins-x/jx/pkg/kube"
 	"github.com/pkg/errors"
 )
@@ -65,21 +66,57 @@ func ResolveRegionWithoutOptions() (string, error) {
 	return ResolveRegion("", "")
 }
 
+// GetClusterNameFromAWS uses the AWS SDK to parse through each EKS cluster until it finds one that matches the endpoint in
+// the kubeconfig. From there it will retrieve the cluster name
+func GetClusterNameFromAWS(profileOption string, regionOption string, kubeEndpoint string) (string, error) {
+	sess, err := NewAwsSession(profileOption, regionOption)
+	if err != nil {
+		return "", errors.Wrapf(err, "Error creating AWS Session")
+	}
+	svc := eks.New(sess)
+
+	input := &eks.ListClustersInput{}
+	result, err := svc.ListClusters(input)
+	if err != nil {
+		return "", errors.Wrapf(err, "Error calling Eks List Clusters")
+	}
+
+	for _, cluster := range result.Clusters {
+		input := &eks.DescribeClusterInput{
+			Name: aws.String(*cluster),
+		}
+		result, err := svc.DescribeCluster(input)
+		if err != nil {
+			return "", errors.Wrapf(err, "Error calling Describe Cluster on "+*cluster)
+		}
+
+		if *result.Cluster.Endpoint == kubeEndpoint {
+			return *result.Cluster.Name, nil
+		}
+	}
+
+	return "", errors.Errorf("Unable to get cluster name from AWS")
+}
+
 // ParseContext parses the EKS cluster context to extract the cluster name and the region
 func ParseContext(context string) (string, string, error) {
+	// First check if context name matches <cluster-name>.<region>.*
 	reg := regexp.MustCompile(`([a-zA-Z][-a-zA-Z0-9]*)\.((us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d)\.*`)
 	result := reg.FindStringSubmatch(context)
-	// Also check if the context name matchesAWS ARN format as defined:
-	// https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazonelasticcontainerserviceforkubernetes.html#amazonelasticcontainerserviceforkubernetes-resources-for-iam-policies
-	arnReg := regexp.MustCompile(`arn:aws:eks:((?:us(?:-gov)?|ap|ca|cn|eu|sa)-(?:central|(?:north|south)?(?:east|west)?)-\d):[0-9]*:cluster\/([a-zA-Z][-a-zA-Z0-9]*)`)
-	arnResult := arnReg.FindStringSubmatch(context)
 	if len(result) >= 3 {
 		return result[1], result[2], nil
-	} else if len(arnResult) >= 3 {
-		return arnResult[2], arnResult[1], nil
-	} else {
-		return "", "", errors.Errorf("unable to parse %s as <cluster_name>.<region>.* or arn:aws:<region>:account-id:cluster/<cluster_name>", context)
 	}
+
+	// or else if the context name matchesAWS ARN format as defined:
+	// https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html
+	reg = regexp.MustCompile(`arn:aws:eks:((?:us(?:-gov)?|ap|ca|cn|eu|sa)-(?:central|(?:north|south)?(?:east|west)?)-\d):[0-9]*:cluster\/([a-zA-Z][-a-zA-Z0-9]*)`)
+	result = reg.FindStringSubmatch(context)
+	if len(result) >= 3 {
+		return result[2], result[1], nil
+	}
+
+	return "", "", errors.Errorf("unable to parse %s as <cluster_name>.<region>.* or arn:aws:<region>:account-id:cluster/<cluster_name>", context)
+
 }
 
 // GetCurrentlyConnectedRegionAndClusterName gets the current context for the connected cluster and parses it
@@ -89,10 +126,22 @@ func GetCurrentlyConnectedRegionAndClusterName() (string, string, error) {
 	if err != nil {
 		return "", "", errors.Wrapf(err, "loading kubeconfig")
 	}
+
 	context := kube.Cluster(kubeConfig)
-	currentClusterName, currentRegion, err := ParseContext(context)
+	server := kube.CurrentServer(kubeConfig)
+	currentClusterName, err := GetClusterNameFromAWS("", "", server)
 	if err != nil {
-		return "", "", errors.Wrapf(err, "parsing the current Kubernetes context %s", context)
+		currentClusterName, currentRegion, err := ParseContext(context)
+		if err != nil {
+			return "", "", errors.Wrapf(err, "parsing the current Kubernetes context %s", context)
+		}
+
+		return currentClusterName, currentRegion, nil
+	}
+
+	currentRegion, err := ResolveRegionWithoutOptions()
+	if err != nil {
+		return "", "", errors.Wrapf(err, "Error resolving Region")
 	}
 	return currentClusterName, currentRegion, nil
 }
--- a/pkg/cloud/amazon/common.go
+++ b/pkg/cloud/amazon/common.go
@@ -66,19 +66,19 @@ func ResolveRegionWithoutOptions() (string, error) {
 	return ResolveRegion("", "")
 }
 
-// GetClusterNameFromAWS uses the AWS SDK to parse through each EKS cluster until it finds one that matches the endpoint in
+// GetClusterNameAndRegionFromAWS uses the AWS SDK to parse through each EKS cluster until it finds one that matches the endpoint in
 // the kubeconfig. From there it will retrieve the cluster name
-func GetClusterNameFromAWS(profileOption string, regionOption string, kubeEndpoint string) (string, error) {
-	sess, err := NewAwsSession(profileOption, regionOption)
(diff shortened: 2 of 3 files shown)


hferentschik:
@andrewhertog, fyi:

```
# github.com/jenkins-x/**/pkg/cloud/amazon
pkg/cloud/amazon/common.go:72:2: arnReg declared and not used
note: module requires Go 1.13
make: *** [build] Error 2
```

pkg/cloud/amazon/common.go
```diff
@@ -69,10 +69,15 @@ func ResolveRegionWithoutOptions() (string, error) {
 func ParseContext(context string) (string, string, error) {
```
hferentschik: TBH,  I don't think this is a valid approach regardless. There is no guarantee on how the context name looks like. A user might change this at will and there is no guarantee that the provider changes this at will.

That said, there is  an `aws ecs describe-clusters ` which gives you cluster information. It contains the cluster arn which I think is more reliable to parse. Maybe there is even another method one can call to get the dedicated information one needs. 

So arn might be the way to go, but imo we also need to get away from relying on the context name altogether.

@dgozalo WDYT?

go.mod
```diff
@@ -171,3 +171,5 @@ replace github.com/Azure/azure-sdk-for-go => github.com/Azure/azure-sdk-for-go v
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v10.15.5+incompatible
 
 replace github.com/banzaicloud/bank-vaults => github.com/banzaicloud/bank-vaults v0.0.0-20190508130850-5673d28c46bd
+
+go 1.13
```
hferentschik: I don't think we should change this. This would force everyone to upgrade to Go 1.13. This is really another topic altogether.

pkg/cloud/amazon/common.go
```diff
@@ -69,10 +69,15 @@ func ResolveRegionWithoutOptions() (string, error) {
...
 	reg := regexp.MustCompile(`([a-zA-Z][-a-zA-Z0-9]*)\.((us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d)\.*`)
 	result := reg.FindStringSubmatch(context)
-	if len(result) < 3 {
-		return "", "", errors.Errorf("unable to parse %s as <cluster_name>.<region>.*", context)
+	arnReg := regexp.MustCompile(`arn:aws:eks:((us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d):[0-9]*:cluster/([a-zA-Z][-a-zA-Z0-9]*)`)
```
hferentschik: Do you have a reference to the definition of the cluster arn. A link as a comment would be great and seeing it's definition would help to judge this regexp

pkg/cloud/amazon/common.go
```diff
@@ -85,7 +90,7 @@ func GetCurrentlyConnectedRegionAndClusterName() (string, string, error) {
 	context := kube.Cluster(kubeConfig)
 	currentClusterName, currentRegion, err := ParseContext(context)
 	if err != nil {
-		return "", "", errors.Wrapf(err, "parsing the current Kubeernetes context %s", context)
+		return "", "", errors.Wrapf(err, "parsing the current Kubernetes context %s", context)
```
hferentschik: good catch

pkg/cloud/amazon/common.go
dgozalo: The problem with `aws ecs/eks describe-clusters` is that you may get every single cluster created in a given region unless you provide a cluster name, which is exactly what we are trying to find here.

I do agree that parsing the ARN should be the way to go as it's not subject to change at all, the problem is getting the ARN of the currently connected cluster.

I think we can keep this solution until we find a more reliable way to obtain this information. We are also parsing the context in the GKE clusters and I can't find of any other solution right now.

hferentschik: > The problem with aws ecs/eks describe-clusters is that you may get every single cluster created in a given region unless you provide a cluster name, which is exactly what we are trying to find here.

Maybe you could call `list-clusters` to get the list of the clusters and then iterate over them using `describe cluster` and matching on the endpoint? This way you would get to the ARN.

How else do you want to progress on this issue?

go.mod
dgozalo: Yeah, we would need some more changes to make this work reliably.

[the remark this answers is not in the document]
andrewhertog: This has been removed. Sorry, when i read the contributing docs, i just read the latest Go version, not the specific version of Go 1.11. I've updated my local dev and updated this branch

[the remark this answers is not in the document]
andrewhertog: For some reason this line keeps getting added back to my project... I've removed it from the PR though

pkg/cloud/amazon/common.go
andrewhertog: I've added a link in a comment in the code, not sure if thats where you want it.

- https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazonelasticcontainerserviceforkubernetes.html#amazonelasticcontainerserviceforkubernetes-resources-for-iam-policies
- https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html

go.mod
andrewhertog: Removed. This was a result of the wrong version of Go installed on my dev machine

pkg/cloud/amazon/common.go
andrewhertog: I've finished the implementation of ARN regex testing, but I'm going to work on the list/describe clusters option to see if i can get that to work. My knowledge of Go is somewhat limited, but i'm going to give it a try anyway.

hferentschik: > but I'm going to work on the list/describe clusters option to see if i can get that to work.

Nice. If we could get it to work that would be awesome.

pkg/cloud/amazon/common.go
```diff
@@ -69,10 +69,17 @@ func ResolveRegionWithoutOptions() (string, error) {
 func ParseContext(context string) (string, string, error) {
 	reg := regexp.MustCompile(`([a-zA-Z][-a-zA-Z0-9]*)\.((us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d)\.*`)
 	result := reg.FindStringSubmatch(context)
-	if len(result) < 3 {
-		return "", "", errors.Errorf("unable to parse %s as <cluster_name>.<region>.*", context)
+	// Also check if the context name matchesAWS ARN format as defined:
+	// https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazonelasticcontainerserviceforkubernetes.html#amazonelasticcontainerserviceforkubernetes-resources-for-iam-policies
+	arnReg := regexp.MustCompile(`arn:aws:eks:((?:us(?:-gov)?|ap|ca|cn|eu|sa)-(?:central|(?:north|south)?(?:east|west)?)-\d):[0-9]*:cluster\/([a-zA-Z][-a-zA-Z0-9]*)`)
+	arnResult := arnReg.FindStringSubmatch(context)
+	if len(result) >= 3 {
```
hferentschik: I find the logic a bit odd here. We are creating and matching both regexp every time regardless. I would more expect that we test one regex (`reg` since it is the default for when we create the context) and if it matches we return directly. If there is no match, we try the next.

One could just move this if statement up above the creation and match of `arnReg`.

That's for the case where we stick to this parsing of the context name. Would much prefer a more reliable approach.

pkg/cloud/amazon/common.go
andrewhertog: I've merged in changes to support this workflow, and reorganized the regex test.

pkg/cloud/amazon/common.go
```diff
@@ -82,10 +126,22 @@ func GetCurrentlyConnectedRegionAndClusterName() (string, string, error) {
 	if err != nil {
 		return "", "", errors.Wrapf(err, "loading kubeconfig")
 	}
+
 	context := kube.Cluster(kubeConfig)
-	currentClusterName, currentRegion, err := ParseContext(context)
+	server := kube.CurrentServer(kubeConfig)
+	currentClusterName, err := GetClusterNameFromAWS("", "", server)
+	if err != nil {
+		currentClusterName, currentRegion, err := ParseContext(context)
```
hferentschik: So the fallback is still ParseContext? Fair enough I guess.

pkg/cloud/amazon/common.go
```diff
@@ -65,14 +66,57 @@ func ResolveRegionWithoutOptions() (string, error) {
 	return ResolveRegion("", "")
 }
 
+// GetClusterNameFromAWS uses the AWS SDK to parse through each EKS cluster until it finds one that matches the endpoint in
+// the kubeconfig. From there it will retrieve the cluster name
+func GetClusterNameFromAWS(profileOption string, regionOption string, kubeEndpoint string) (string, error) {
```
hferentschik: 👍 That's pretty much what I had in mind :-)

@dgozalo, WDYT? To me, that seems more reliable.

pkg/cloud/amazon/common.go
andrewhertog: Yeah, i guess it might be a little redundant now though.

pkg/cloud/amazon/common.go
```diff
@@ -82,10 +126,22 @@ func GetCurrentlyConnectedRegionAndClusterName() (string, string, error) {
...
+	if err != nil {
+		currentClusterName, currentRegion, err := ParseContext(context)
+		if err != nil {
+			return "", "", errors.Wrapf(err, "parsing the current Kubernetes context %s", context)
+		}
+
+		return currentClusterName, currentRegion, nil
+	}
+
+	currentRegion, err := ResolveRegionWithoutOptions()
```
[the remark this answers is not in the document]
andrewhertog: The big problem with trying to obtain the region/clustername from the context name is the fact that the context name could possibly be anything. This is why this issue was opened in the first place. I don't think its possible in every case to assume that we can use the context to obtain this information.
I could reorder the logic to attempt to use the context name first rather than the aws-sdk, if you prefer, but i believe that the sdk logic should remain.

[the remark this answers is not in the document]
andrewhertog: The thing is that the region value in the ARN will always match the region value used with the aws sdk's config. I think it adds unneeded code complexity by trying to parse the arn just for the sake of the region.

Maybe what i could do is use the active aws sdk session used for `DescribeCluster` and get the region from that sessions config, rather than calling `ResolveRegionWithoutOptions` and having that instantiate a new sdk session, just for the sake of pulling the region from the session config.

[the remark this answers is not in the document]
dgozalo: Keep in mind that the person who created the cluster in a certain region with a certain configuration might not be the same person that is going to run install / boot later.

We need to be idempotent here and always return the region where the cluster was created and not the region that the user calling the command has configured.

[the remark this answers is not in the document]
andrewhertog: I understand, But if you are running `DescribeClusters` within aws region us-east-1 for an example, you will not be able to see any clusters that are in different aws regions. 
Therefore with that logic, the region the user has configured, must be the region that the cluster exists in. 
I think to be idempotent as you are suggesting, we need to have a command line flag or configuration value to allow reading the region from there that can override the users configuration value.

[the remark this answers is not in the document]
dgozalo: You are right, that should be fine. I overlooked that part.

PR #5847, opened 2019-10-16
Fix tests execution with 1.13

ilpianista:
#### Submitter checklist

#### Description
`go test` fails with Go 1.13 due to golang/go#32246.

```
$ cd pkg/cmd
$ go test
# golang.org/x/xerrors
/home/andrea/go/pkg/mod/golang.org/x/xerrors@v0.0.0-20190513163551-3ee3066db522/adaptor_go1_13.go:16:14: undefined: errors.Frame
/home/andrea/go/pkg/mod/golang.org/x/xerrors@v0.0.0-20190513163551-3ee3066db522/format_go1_13.go:12:18: undefined: errors.Formatter
FAIL	github.com/jenkins-x/jx/pkg/cmd [build failed]
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -130,7 +130,7 @@ require (
 	golang.org/x/oauth2 v0.0.0-20190604053449-0f29369cfe45
 	golang.org/x/sync v0.0.0-20190423024810-112230192c58
 	golang.org/x/sys v0.0.0-20190616124812-15dcb6c0061f
-	golang.org/x/xerrors v0.0.0-20190513163551-3ee3066db522 // indirect
+	golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7 // indirect
 	google.golang.org/api v0.1.0 // indirect
 	google.golang.org/genproto v0.0.0-20190219182410-082222b4a5c5 // indirect
 	gopkg.in/AlecAivazis/survey.v1 v1.8.3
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -171,3 +171,7 @@ replace github.com/Azure/azure-sdk-for-go => github.com/Azure/azure-sdk-for-go v
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v10.15.5+incompatible
 
 replace github.com/banzaicloud/bank-vaults => github.com/banzaicloud/bank-vaults v0.0.0-20190508130850-5673d28c46bd
+
+replace github.com/spf13/cobra v0.0.0-00000000000000-c46add8a6528 => github.com/spf13/cobra v0.0.0-20170711120833-c46add8a6528
```
hferentschik: So why the Cobra replace now? If anything we should try to reduce the amount of replace statements needed instead of increasing them.  How is this related to the xerror problem?

go.mod
```diff
@@ -171,3 +171,7 @@ replace github.com/Azure/azure-sdk-for-go => github.com/Azure/azure-sdk-for-go v
...
+
+go 1.13
```
hferentschik: Not sure whether we can enforce 1.13 at this point. Effectively we are forcing everyone to go to 1.13 now, right?

@rawlingsj, I seem to recall you having a discussion around upgrading to Golang 1.13 and basically holding it back for now. Am I just imagining this?

hferentschik:
@ilpianista, thanks for the PR. Did you have a concrete problem you wanted to solve with an upgrade to 1.13?

ilpianista:
> @ilpianista, thanks for the PR. Did you have a concrete problem you wanted to solve with an upgrade to 1.13?

Nothing serious :-) Since I'm on 1.13, I was unable to run the tests until I updated `go.mod` and thus I thought it would make sense to share it.

go.mod
ilpianista: I had to because Go 1.13 is more strict about [version validations](https://golang.org/doc/go1.13#version-validation).

rawlingsj: Yeah I had a little look a couple of weeks back and ran into a number of issues with dependencies and mod `replace ->`.  We might need to hold this change and do a lot more validation upgrading to 1.13, which includes upgrading our builder images so our pipelines build with 1.13 and that will impact other repos too.

hferentschik:
For what it's worth, I am using [gvm](https://github.com/moovweb/gvm) to have multiple Go version in parallel on my machine. Works like a charm.

hferentschik:
I am actually going to close this PR. I think we should move to Golang 1.13 in the near future, but I also think we combine this with a general review of the dependencies and maybe trying to upgrade them.

Issue #5463, opened 2019-09-12
Use gitcredential helpers for git authentication


hferentschik:
Following up on issue #5341 which used a different approach for fixing the problem around `ForkAndPullRepo` in `helpers.go`.

The idea is to find a more holistic approach of handling git authentication locally as well as in the pipeline. 

To avoid writing anything to disk, git's [credential helpers](https://git-scm.com/docs/gitcredentials), in particular the [cache](https://git-scm.com/docs/git-credential-cache) helper, could be used. The cache helper stores the credentials for a specified time in memory. The credentials never hit the disk. The idea is to add something along this line to `Gitter`:

```go
// CacheCredendials caches the credentials for the specified user, host and path for the specified number of seconds.
CacheCredendials(host string, path string, username string, password string, time int) error

// Clears potentially cached credentials for the specified host, path and user, if they exist
ClearCredentials(host string, path string, username string) error
```
This way, when dealing with repositories which need authentication one can call `CacheCredendials` once with an appropriate timeout and then call other methods of the `Gitter` interface without having to think about the credentials. 

As a further step, one can then imagine providing our own credential helper. There is an [API](https://git-scm.com/docs/api-credentials) for that. This custom credential helper could be a standalone program or a `jx` command. It would then retrieve the credentials from the cluster.

This mechanism would also solve the issue around `jx import` adding the API token to the git remote. 

See also https://git-scm.com/book/en/v2/Git-Tools-Credential-Storage

hferentschik:
There is second option which might be a bit easier to implement. We could leverage the _GIT_ASKPASS_ env variable. We could set it either where needed or as early as a `jx` call gets initialized (in the Cobra setup). 

We could set _GIT_ASKPASS_ to `jx` itself using a command like `jx step git credential`. 

This approach would solve authentication for all cases where `jx` is involved and where we make git calls from within `jx`.

The disadvantage of this approach is that it won't work for users using the git CLI directly. There a dedicated k8s-cred-helper is superior, since once configured it will work regardless of whether git is called as part of a `jx` command or standalone.

hferentschik:
See discussion around git authentication on this PR as well https://github.com/jenkins-x/jx/pull/5850#issuecomment-543137745

Issue #5114, opened 2019-08-17
jx boot mixes up secrets from vault with k8s secrets


ccojocar:
### Summary 

It seems that `jx boot` mixes up the secrets from vault with k8s secrets e.g. for GitHub pipeline token. 

For instance when disabling the `gh-secret` from boot templates when vault is active, the installation process starts to fail when the environments are created because the `auth config` is not properly set in vault. 

```
creating environment production git repository for URL: https://github.com/ccojocar/environment-cosmin-dev-production.git to namespace jx-production
error: no auth server found for git server https://github.com from gitURL https://github.com/ccojocar/environment-cosmin-dev-production.git
error: failed to interpret pipeline file jenkins-x.yml: failed to run '/bin/sh -c jx step verify env' command in directory '.', output: ''
```
This means that the create environment is using the duplicated github token from k8s secret instead of using the one from vault :disappointed: . Unfortunately the secrets are quite mixed up by boot. 

We need to properly configure the auth in a boot step before creating the environments (verify-env step).

Also the GitHub token seems to be duplicated in 5 places during the installation:

![image](https://user-images.githubusercontent.com/591127/63213618-b876b100-c10e-11e9-902c-5c24cdabf62d.png)

This will definitely make secrets rotation very difficult. 

### Expected behavior

jx should use in the pipeline the secrets from vault for git provider authentication. 

### Actual behavior

jx is using the secrets duplicated in k8s secret `jx-pipeline-git-github-github` when vault is configured.

### Jx version

The output of `jx version` is:

```
COPY OUTPUT HERE
```

### Jenkins type

- [ x] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

### Operating system / Environment

ACs:
- When vault is enabled when using boot the value of the pipeline git secret should only be in vault and not in jx-pipeline-github-github
- When vault is disabled the pipeline git secret should be available in jx-pipeline-github-github

ccojocar:
Block until https://github.com/jenkins-x/jx/issues/5190 is done.

pmuir:
@ccojocar is this unblocked now?

ccojocar:
Yes, it still blocked by https://github.com/cloudbees/cloudbees-jenkins-x-distro/issues/15 which needs https://github.com/jenkins-x/jx/issues/5310.

ccojocar:
[...]

PR #5850, opened 2019-10-17
[wip] feat: Handling git credentials via GIT_ASKPASS

### LINKED ISSUE #5772: Unable to re-run 'jx boot' from the initial boot config directory
### Summary 

When using `jx boot` to bootstrap a cluster, one should be allowed to re-run `jx boot` from within the initial boot config directory, however, this currently results in errors of the form:

```
Attempting to cherry pick commits that were on master but not yet pushed
WARNING: Unable to cherry-pick 1c384abadb71921d01971922930decfecc03e6f3: git output: error: could not apply 1c384ab... initial config based of hf-bee-bot/environment-hardy-bdd-dev-test with ref v1.0.25
hint: after resolving the conflicts, mark the corrected paths
hint: with 'git add <paths>' or 'git rm <paths>'
hint: and commit the result with 'git commit': failed to run 'git cherry-pick 1c384abadb71921d01971922930decfecc03e6f3' command in directory '/Users/hardy/tmp/dummy-boot/jenkins-x-boot-config', output: 'error: could not apply 1c384ab... initial config based of hf-bee-bot/environment-hardy-bdd-dev-test with ref v1.0.25
hint: after resolving the conflicts, mark the corrected paths
hint: with 'git add <paths>' or 'git rm <paths>'
hint: and commit the result with 'git commit''
```

Install Jenkins-X on a new cluster via `jx boot`, then re-run `jx boot` from within the initial boot config directory.

### Expected behavior

The second `jx boot` succeeds as well, resulting in no changes to the cluster. 

### Actual behaviour

See above.

The problem is how we try to merge / cherry-pick local changes during verify environment. Effectively we get the latest master and then cherry-pick any local changes which are not in the remotes on top of it. This happens in `helpers.ForkAndPullRepo` or the more particular in `gitter.GetCommitsNotOnAnyRemote`. Under the hood this calls `git log master  --not --remotes`. This will fail for us due to the way we are pushing to the remote. To avoid having to store credentials for the remote environment repository, we internally (in code) construct an authenticated push URL to which we push. This pushes the code to the remote correctly, but it does not update the remote ref in _.git/refs/remotes/origin_. This means `git log master  --not --remotes` will report the local commits which are actually already on the remote as well missing and we will attempt to cherry pick them leading to a merge conflict.

The problematic code is:

```golang
	userDetails := provider.UserAuth()
	duplicatePushURL, err := gitter.CreateAuthenticatedURL(duplicateInfo.CloneURL, &userDetails)
	if err != nil {
		return nil, errors.Wrapf(err, "failed to create push URL for %s", duplicateInfo.CloneURL)
	}
	err = gitter.Push(dir, duplicatePushURL, true, fmt.Sprintf("%s:%s", "HEAD", toBranch))
	if err != nil {
		return nil, errors.Wrapf(err, "failed to push HEAD to %s", toBranch)
	}
```

hferentschik:
fixes #5772

This is a possible solution for #5772 allowing you to re-run `jx boot` from the same directory multiple times.

At the same time, it could also serve as an alternative solution for #5463, replacing the need for a credential helper. The only drawback is that users might get prompted for a password/token when executing git commands directly. All `jx` executions would be covered.

We could also set `git config core.askpass ~/.jx/bin/git-askpass` directly into the git config as part of cloning / setting a remote. This way `jx` command executions as well as users direct git commands would be covered. We also would not have to set _GIT_ASKPASS_ on each `jx` command execution. 

Thoughts?

Note, this is just a POC. There are several TODOs left in the code. I'd like to determine whether this is a viable path forward.

### DIFF
--- a/jx/bdd/boot-vault/cluster.yaml
+++ b/jx/bdd/boot-vault/cluster.yaml
@@ -15,4 +15,3 @@ clusters:
       - command: jx
         args:
           - boot
-          - -b
--- a/pkg/cmd/cmd.go
+++ b/pkg/cmd/cmd.go
@@ -18,6 +18,10 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/jenkins-x/jx/pkg/util"
 
 	"github.com/jenkins-x/jx/pkg/cmd/profile"
 	"github.com/spf13/viper"
@@ -69,6 +73,10 @@ import (
 	"gopkg.in/AlecAivazis/survey.v1/terminal"
 )
 
+const (
+	askPassBin = "git-askpass"
+)
+
 // NewJXCommand creates the `jx` command and its nested children.
 // args used to determine binary plugin to run can be overridden (does not affect compiled in commands).
 func NewJXCommand(f clients.Factory, in terminal.FileReader, out terminal.FileWriter,
@@ -78,7 +86,7 @@ func NewJXCommand(f clients.Factory, in terminal.FileReader, out terminal.FileWr
 	rootCommand := &cobra.Command{
 		Use:              "jx",
 		Short:            "jx is a command line tool for working with Jenkins X",
-		PersistentPreRun: setLoggingLevel,
+		PersistentPreRun: persistentPreRun,
 		Run:              runHelp,
 	}
 
@@ -301,7 +309,12 @@ func fullPath(command *cobra.Command) string {
 	return name
 }
 
-func setLoggingLevel(cmd *cobra.Command, args []string) {
+func persistentPreRun(cmd *cobra.Command, args []string) {
+	setLoggingLevel(cmd)
+	configureGit()
+}
+
+func setLoggingLevel(cmd *cobra.Command) {
 	verbose, err := strconv.ParseBool(cmd.Flag(opts.OptionVerbose).Value.String())
 	if err != nil {
 		log.Logger().Errorf("Unable to determine log level")
@@ -318,6 +331,37 @@ func setLoggingLevel(cmd *cobra.Command, args []string) {
 			log.Logger().Errorf("Unable to set log level to info")
 		}
 	}
+}
+
+func configureGit() {
+	// TODO issue-5772 add a batch file for windows
+	// TODO issue-5772 is will run on each command execution adding a bit of overhead. Is this ok?
+	// TODO issue-5772 should it be able to skip this via a global flag? are there cases where we don't want this?
+	configDir, err := util.ConfigDir()
+	if err != nil {
+		log.Logger().Errorf("unable to determine current X_HOME: %s", err.Error())
+	}
+
+	askPassBin := filepath.Join(configDir, "bin", askPassBin)
+	exists, err := util.FileExists(askPassBin)
+	if err != nil {
+		log.Logger().Errorf("unable to determine file stats for %s: %s", askPassBin, err.Error())
+	}
+
+	if !exists {
+		data := []byte("jx step git credentials --ask-pass")
+		err := ioutil.WriteFile(askPassBin, data, 0750)
+		if err != nil {
+			log.Logger().Errorf("unable to write %s: %s", askPassBin, err.Error())
+		}
+	}
+
+	err = os.Setenv("GIT_ASKPASS", askPassBin)
+	if err != nil {
+		log.Logger().Warnf("unable to set GIT_ASKPASS: %s", err.Error())
+	}
+
+	log.Logger().Tracef("GIT_ASKPASS successfully set to %s", askPassBin)
 }
 
 func runHelp(cmd *cobra.Command, args []string) {
--- a/pkg/cmd/step/git/step_git_credentials.go
+++ b/pkg/cmd/step/git/step_git_credentials.go
@@ -25,11 +25,18 @@ const (
 	optionOutputFile = "output"
 )
 
+type credentials struct {
+	user       string
+	password   string
+	serviceURL string
+}
+
 // StepGitCredentialsOptions contains the command line flags
 type StepGitCredentialsOptions struct {
 	step.StepOptions
 
 	OutputFile string
+	AskPass    bool
 }
 
 var (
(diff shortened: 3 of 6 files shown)


ccojocar:
@hferentschik Have you also seen the https://github.com/jenkins-x/jx/issues/5114? One issue with boot right now when we are using vault is that the gitAuth.yaml is not configured at all in vault but that's slightly related. 

Some things to consider:
- the commands run locally and in the pipeline - in the `pipeline` is shouldn't prompt for credentials, which use case are we trying to solve?
- also in the pipeline, Tekton is configuring the git credentials as an initialisation step, `git` should pick up those credentials which are stored during the installation in `knative-git-user-pass` secret 
- also we usually store the same credentials in another secret `jx-pipeline-git-github-github` which is looked up by jx when loading the pipeline secrets - also this secret needs to be removed when enabling vault. In that case jx should fetch the secrets from vault not k8s secrets - See the #5114
- sometimes when loading pipeline secretes we performed a kind of merged which tried to lookup first the `~/.jx/gtiAuth.yaml` and the checks the secrets in the cluster
- normally during the installation we store the bot token in `jx-pipeline-git-github-github` so  this means the gitask will always use the bot token. See #https://github.com/jenkins-x/jx/issues/3593 - basically we should split the bot token from local token. The bot token should be used only by the pipeline when running in the cluster and not by jx on the local workstations. The usual use case is when there are a few developers who wants to use their own git token when running jx locally but the bot token when the builds are running in the pipeline.


I think, it makes sense to define the scenarios where the git is executed by jx:
1) on local workstation (boot, create quickstart, import, etc.)
2) pipeline 

There are the types of git tokens/ssh keys:
1) `bot token` - must be exclusively used by the pipeline when executing various jx commands or directly git etc.  
2) `developer token` - should be always used locally by developers
3) `developer ssh key` - setup by developer on its workstation (ssh-agent handles the authentication)

These tokens can be stored:
1) `bot token` -  `vault` or `k8s secrets` 
2) `developer token` - only on developer workstation, it should never be updated in k8s secrets or vault  - currently there is a big issue when a developer can override the bot token in vault 

Problems we should solve:
1) clearly separate the tokens 
2) bot token should be either stored in vault or in k8s secret
3) locally always use the developer token 
4) reuse the git token also a git provider token (e.g. for GItHub API) - this token is stored in the gitAuth.yaml (locally, k8s secret, vault)

I think the GIT_ASKPASS  or git credentials is great for `developer token` - this should be active when running jx locally on developer workstation. This should address problem 3) but we need to make sure we consider the other cases and the overall setup.

hferentschik:
> Have you also seen the #5114? 

Kind of. I am aware of the issue, but don't really understand all the implications in code.

> the commands run locally and in the pipeline - in the pipeline is shouldn't prompt for credentials, which use case are we trying to solve?

Mainly the local development case, but I think this should work in pipeline as well.

> also in the pipeline, Tekton is configuring the git credentials as an initialisation step,

You mean `git step git credentials`? 

> git should pick up those credentials which are stored during the installation in knative-git-user-pass secret

As said, I think the _GIT_ASKPASS_ approach could be made to work locally as well as in pipeline. 
_GIT_ASKPASS_ only is about how git asks for the password. Internally we could retrieve and returns different tokens, depending on the environment. 

This might even allow us to unify the process and have a single way of doing things.

> also we usually store the same credentials in another secret jx-pipeline-git-github-github which is looked up by jx when loading the pipeline secrets - also this secret needs to be removed when enabling vault. In that case jx should fetch the secrets from vault not k8s secrets - See the #5114

Yeah, this is confusing. Maybe we could sit together and define which secrets there should be and for what use case they should be retrieved and which secrets should eventually be removed. I hooked into `jx step git credentials` because that was the one thing known to me and because it was trivial to so. 

So instead of extending `jx step git credentials` we need another/new command, something like `jx git askpass` which depending on some context retrieves and returns the right token.

> sometimes when loading pipeline secretes we performed a kind of merged which tried to lookup first the ~/.jx/gtiAuth.yaml and the checks the secrets in the cluster

Why is that, really? Why do we even need a `gitAuth.yaml`? If all were well we would not need it, right?

> normally during the installation we store the bot token in jx-pipeline-git-github-github so this means the gitask will always use the bot token. 

At the moment yes. But it really up to us how we produce the token. For example, I am using something like `git config  url."https://token@github.com/".insteadOf "git@github.com:"` to make sure git is using the specified command for _GIT_ASKPASS_. _token_ is arbitrary. We could use _local_ vs _pipeline_ in order to retrieve and return different tokens. I guess _pipeline_ would then also be paired with a "in cluster" check.

> See ##3593 - basically we should split the bot token from local token. 

But how do we get this token into Jenkins X? Right now there is no such token at all, right? So would the user have to specify another token during boot? Or have a command to add a developer token? 

> I think, it makes sense to define the scenarios where the git is executed by jx:

> on local workstation (boot, create quickstart, import, etc.)
> pipeline

agreed

> developer token - only on developer workstation, it should never be updated in k8s secrets or vault

If the token is not anywhere in the cluster, how it is then provided to git? Is this all up to the developer to sort out using whatever git config needed? Is that not a bit problematic?
Do we have any means in Jenkins X to differentiate between different developers within a team?

> clearly separate the tokens

+1

> bot token should be either stored in vault or in k8s secret

+1

> locally always use the developer token

I am a bit unclear atm how this local git token would work, especially where it comes from and who/what is configuring it.

> I think the GIT_ASKPASS or git credentials is great for developer token - this should be active when running jx locally on developer workstation. This should address problem 3) but we need to make sure we consider the other cases and the overall setup.

Ok.

hferentschik:
Ok, this was really good feedback. Obviously there are many factors to consider and we need to solve multiple problems. 

Some questions I am having right now:

* Can we split these problems up so that we can solve them one at a time instead of trying to address all at one and ending up with this mega refactoring
* Can the _GIT_ASKPASS_ approach be extended to work in all our scenarios
* If not, should we try to use _GIT_ASKPASS_ for the local development use case only?
* How and where is the developer token created and stored?

hferentschik:
BTW, this is worth reading - https://coolaj86.com/articles/vanilla-devops-git-credentials-cheatsheet/

hferentschik:
Closing this for now. Will open a new PR asap.

PR #5879, opened 2019-10-18
feat(git): refactor AddCommit method

wbrefvem:
#### Submitter checklist

#### Description

This is a readability enhancement that changes the `AddCommit` method of the git CLI interface to `CommitAllChangesToTrackedFiles`, which is a more accurate description of what the method does. The current name can give (and has given) the impression that all files, tracked and untracked, will be committed. The new name should disabuse users of the method of that misconception.


pkg/gits/git_cli.go
```diff
@@ -506,7 +506,7 @@ func (g *GitCLI) CommitDir(dir string, message string) error {
 }
 
 // AddCommit perform an add and commit of the changes from the repository at the given directory with the given messages
-func (g *GitCLI) AddCommit(dir string, msg string) error {
+func (g *GitCLI) CommitAllChangesToTrackedFiles(dir string, msg string) error {
```
hferentschik: The godoc is out of sync here.

That said, why do you want to rename the function? Is `CommitAllChangesToTrackedFiles` not a bit superfluous.  `AddCommit` (or even better `Commit`)  does already imply that, no?

Personally, the new name confuses me more instead of clarifying anything, since I am wondering at that stage what the semantic difference to a 'commit' would be.

wbrefvem: @hferentschik To me, `AddCommit` is ambiguous because it could mean that tracked & untracked files are added, or it could mean simply `git commit -a` which only adds tracked files, the latter of which is the behavior of `AddCommit`. I admit that it's a wordy name! But it tells you exactly what the method is doing.

hferentschik: Ok. Maybe `CommitChangedTrackedFiles` is a bit friendlier?

PR #5891, opened 2019-10-21
fix(pipeline): Use standard version stream cloning logic in metapipeline

### LINKED ISSUE #5075: Installing CJXD with an existing jenkinx-x-versions repo in jenkins home doesn't override it with the correct one
### Summary 
When executing `jx boot` using the CJXD Release 1 binary while having an existing `jenkins-x-versions` repository in `.jx` results in boot using the existing repository without checking if it's a CJXD repository.

I had this happening when creating a brand new CJXD cluster without deleting `.jx`, resulting in CJXD using the versions stream master branch of `jenkins-x-versions` instead of `cloudbees-jenkins-x-versions`.

I found out because metapipelines were enabled by default. 

Create a cluster using `jx boot` with the `oss` profile enabled, it will download `jenkins-x-versions`. Then enable the `cloudbees` profile and attempt to do the same.

### Expected behavior
It should use the versions defined in `cloudbees-jenkins-x-versions`

### Actual behavior
Uses whatever versions are defined in the local `jenkins-x-versions` repository, without checking whether it's the correct one or not.

### Jx version
CJXD Release 1:

```
NAME               VERSION
jx                 2.0.521
Kubernetes cluster v1.12.8-gke.10
kubectl            v1.15.1
helm client        Client: v2.13.1+g618447c
git                git version 2.20.1 (Apple Git-117)
Operating System   Mac OS X 10.14.4 build 18E226
```
### Jenkins type

- [ ] Classic Jenkins

abayer:
#### Submitter checklist

#### Description

Now that the standard version stream cloning logic has merged, let's use that in metapipeline rather than its own logic.

/assign @pmuir 
/assign @hferentschik 

fixes #5075

### DIFF
--- a/pkg/cmd/controller/pipeline/pipelinerunner_cmd.go
+++ b/pkg/cmd/controller/pipeline/pipelinerunner_cmd.go
@@ -105,7 +105,7 @@ func (o *PipelineRunnerOptions) Run() error {
 		return err
 	}
 
-	metapipelineClient, err := metapipeline.NewMetaPipelineClient()
+	metapipelineClient, err := metapipeline.NewMetaPipelineClient(o.Git(), o.In, o.Out, o.Err)
 	if err != nil {
 		return err
 	}
--- a/pkg/tekton/metapipeline/clientfactory.go
+++ b/pkg/tekton/metapipeline/clientfactory.go
@@ -2,11 +2,11 @@ package metapipeline
 
 import (
 	"fmt"
-	"io/ioutil"
+	"io"
 	"os"
-	"strings"
 	"time"
 
+	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/apps"
 	"github.com/jenkins-x/jx/pkg/client/clientset/versioned"
 	jxclient "github.com/jenkins-x/jx/pkg/client/clientset/versioned"
@@ -16,10 +16,11 @@ import (
 	"github.com/jenkins-x/jx/pkg/kube"
 	"github.com/jenkins-x/jx/pkg/log"
 	"github.com/jenkins-x/jx/pkg/tekton"
-	"github.com/jenkins-x/jx/pkg/util"
+	"github.com/jenkins-x/jx/pkg/versionstream/versionstreamrepo"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	tektonclient "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
+	"gopkg.in/AlecAivazis/survey.v1/terminal"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 	kubeclient "k8s.io/client-go/kubernetes"
@@ -41,6 +42,11 @@ type clientFactory struct {
 	kubeClient   kubernetes.Interface
 	ns           string
 
+	git    gits.Gitter
+	stderr io.Writer
+	stdin  terminal.FileReader
+	stdout terminal.FileWriter
+
 	versionDir       string
 	versionStreamURL string
 	versionStreamRef string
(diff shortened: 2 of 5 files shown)


hferentschik:
@abayer how does this fix #5075?

abayer:
@hferentschik I'm kinda guessing. =) I'll add a test to verify, but I believe that the now-standard cloning approach handles that scenario.

hferentschik:
>  I'll add a test to verify,

+1

pkg/cmd/controller/pipeline/pipelinerunner_cmd.go
```diff
@@ -105,7 +105,7 @@ func (o *PipelineRunnerOptions) Run() error {
 		return err
 	}
 
-	metapipelineClient, err := metapipeline.NewMetaPipelineClient()
+	metapipelineClient, err := metapipeline.NewMetaPipelineClient(o.Git(), o.In, o.Out, o.Err)
```
hferentschik: I've seen this now at several places that we pass all three standard file handles explicitly. I've been thinking it would make sense to wrap this into some utility struct like `IOFileHandles`. Definitely would help to reduce the number of function parameters and I think there is potential for some shared behaviour in this struct as well then (maybe not in this case).

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -41,6 +42,12 @@ type clientFactory struct {
 	kubeClient   kubernetes.Interface
 	ns           string
 
+	git    gits.Gitter
+	stderr io.Writer
```
hferentschik: so `stderr` is not a terminal.FileWriter?

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -41,6 +42,12 @@ type clientFactory struct {
...
+	stdin  terminal.FileReader
+	stdout terminal.FileWriter
+
+	teamSettings     *v1.TeamSettings
```
hferentschik: Do we really want to have a pointer? I'd prefer a value instance if possible.

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -75,6 +82,11 @@ func NewMetaPipelineClientWithClientsAndNamespace(jxClient versioned.Interface,
 		tektonClient:     tektonClient,
 		kubeClient:       kubeClient,
 		ns:               ns,
+		git:              gitter,
+		stdin:            stdin,
+		stdout:           stdout,
+		stderr:           stderr,
+		teamSettings:     teamSettings,
```
hferentschik: I might be mistaken, but do we really need _teamSettings_ as an instance variable?

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -216,18 +228,21 @@ func versionStreamURLAndRef(jxClient versioned.Interface, ns string) (string, st
...
+	teamSettings, url, ref, err := versionStreamURLAndRef(c.jxClient, c.ns)
 	if err != nil {
 		return err
 	}
 
 	if c.versionStreamURL != url || c.versionStreamRef != ref {
 		oldVersionStreamDir := c.versionDir
-		c.versionDir, err = cloneVersionStream(url, ref)
+		c.versionDir, _, err = versionstreamrepo.CloneJXVersionsRepo(url, ref, teamSettings, c.git, true, false, c.stdin, c.stdout, c.stderr)
+		c.teamSettings = teamSettings
```
hferentschik: You keep updating the instance `teamSetting`, but I cannot see where it is used.

pkg/tekton/metapipeline/clientfactory.go
```diff
@@ -237,68 +252,6 @@ func (c *clientFactory) cloneVersionStreamIfNeeded() error {
 	return nil
 }
 
-func cloneVersionStream(url string, ref string) (string, error) {
```
hferentschik: 👍

pkg/tekton/metapipeline/clientfactory.go
abayer: Ah, good point. I did need it at one point but then refactored.

pkg/tekton/metapipeline/clientfactory.go
abayer: Nope.

abayer:
@hferentschik Removed the team settings field since it was in fact unused, added a test for #5075 - verified that it does now remove the existing `jenkins-x-versions` clone and replace it.

pkg/tekton/metapipeline/clientfactory.go
hferentschik: if it were me, I'd have a wrapper which wraps all three streams as their underlying io.[Reader|Writer] with a function to wrap then as a terminal xyz.

pkg/tekton/metapipeline/clientfactory.go
hferentschik: let's get rid of it then :-)

hferentschik:
/test boot-vault

@abayer, I've seen several failed boot-vault context over the last couple of days. Any idea why this context fails so often?

abayer:
@hferentschik From what I can tell, it looks like Google Cloud load balancer provisioning has been taking abnormally long today, so...sometimes `boot-vault` times out trying to get the ingress domain. Is annoying.

pkg/tekton/metapipeline/clientfactory.go
abayer: Yeah, I'm gonna open an issue for general refactoring around the streams - starting with the struct idea you had.

PR #5892, opened 2019-10-21
chore: Switch distro-specific strings to overrideable vars, not constants

### LINKED ISSUE #5968: Switch from hardcoded constants for boot + version stream URLs to specifying at build time
This will greatly simplify things for distros of jx, since they won't need to have hardcoded constants in the standard jx binary, and they won't have to mess around with profiles, etc.

See https://github.com/jenkins-x/jx/pull/5892 for a prototype.

abayer:
#### Submitter checklist

#### Description

This will allow for setting these values for a distribution's binary of `jx` without needing to change source code.

NOTE: This is just an experiment right now.

/assign @garethjevans 

fixes #5968


abayer:
`TestStepValidate` is failing because the version is now getting set to the `0.0.0-SNAPSHOT...` version in tests, not just when building the binary. I tried changing the version it's testing against to `0.0.0`, but it still failed, so I need to think about this more.

PR #5894, opened 2019-10-21
fix: GitHubProvider#asGitHubCommit should extract author details specified in the commit

### LINKED ISSUE #5726: GitHubProvider#asGitHubCommit should extract author name and email from commit
### Summary 

```go
	message := ""
	if commit.Commit != nil {
	} else {
	}
	var author *GitUser
	if commit.Author != nil {
		author = &GitUser{
		}
	} else {
	}
	return GitCommit{
		Message: message,
		URL:     commit.GetURL(),
		SHA:     commit.GetSHA(),
		Author:  author,
	}

}
```

```
``` 

khansaAmrouni:
fixes #5726 

## Description

`GitHubProvider#GitHubCommit` uses ` commit.Author` to extract the author of a commit, `commit.Author` contains GitHub user profile, and since users can choose to have a public email address on GuitHub, it happens that sometimes we get email empty. the idea is to use commit.Commit.Author to get author details specified in the commit.

### DIFF
--- a/pkg/gits/github.go
+++ b/pkg/gits/github.go
@@ -782,13 +782,13 @@ func (p *GitHubProvider) asGitHubCommit(commit *github.RepositoryCommit) GitComm
 		log.Logger().Warnf("No Commit object for for commit: %s", commit.GetSHA())
 	}
 	var author *GitUser
-	if commit.Author != nil {
+	if commit.Commit.Author != nil {
 		author = &GitUser{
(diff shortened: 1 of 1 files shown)


khansaAmrouni:
PS: i am an outreachy applicant, it is a pleasure if you accept to review my PR.
thanks

khansaAmrouni:
@hferentschik ok, deal

PR #5899, opened 2019-10-22
fix: ensure helm is using requirements version stream

macox:
#### Description

- Installation of the vault operator was not using version stream specified in the requirements for `jx boot`. It was not set in the helm InstallChartOptions and was ignoring what was set in teamsettings and then defaulted to the default version stream.  This fix passes the values through in InstallChartOptions.

- Similarly in `jx step helm apply` this PR now passes the version steam config from requirements through in helm InstallChartOptions. 

- `jx step verify preinstall` was updating the team settings prior to the above steps. This wasn't taken effect however, as the team settings is being ignored. Removing this functionality as the above means it's no longer required, and this also prevents team settings being updated in multiple places. It should reflect what is in the dev-env template in a `jx boot` gitops cluster.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "^pkg/jenkins/test_data/update_center.json.*$|^.secrets.baseline$|^.*test.*$",
     "lines": null
   },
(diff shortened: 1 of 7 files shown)


abayer:
woo!

PR #5902, opened 2019-10-22
chore: Switch calls passing "o.In, o.Out, o.Err" to using a struct instead

### LINKED ISSUE #5901: Switch calls passing "o.In, o.Out, o.Err" to using a struct instead
Coming out of discussion on #5891 - we have a _lot_ of calls passing `o.In, o.Out, o.Err` (from `CommonOptions`), almost always all three of them together. Switching that to a utility struct will make a lot of functions a little cleaner/simpler.

abayer:
#### Submitter checklist

#### Description

Calls to `survey.WithStdio(o.In, o.Out, o.Err)` stay the same, since that's from a dependency. I left the setting of `o.In`, `o.Out`, and `o.Err` as is - I just added `CommonOptions.GetIOFileHandles()` and switched any other uses of `o.In, o.Out, o.Err` to use `o.GetIOFileHandles()` instead.

/assign @hferentschik 

fixes #5901

### DIFF
--- a/pkg/util/files.go
+++ b/pkg/util/files.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/jenkins-x/jx/pkg/log"
+	"gopkg.in/AlecAivazis/survey.v1/terminal"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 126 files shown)


pkg/util/files.go
```diff
@@ -26,6 +27,14 @@ const (
 	MaximumNewDirectoryAttempts = 1000
 )
 
+// IOFileHandles is a struct for holding CommonOptions' In, Out, and Err I/O handles, to simplify function calls.
+type IOFileHandles struct {
```
hferentschik: :+1 sweet. Not sure whether I'd put it into a dedicated io.go or here, but luckily Go does not care :-)

PR #5910, opened 2019-10-23
feat(pipeline): Allow activity for next build number and branch as revision

### LINKED ISSUE #5909: feat request: optionally avoid creating Source Repository resources during meta pipeline create
### Summary 

There are times where we would like to programatically run the metapipeline adhoc in a cluster so that it generates the Tekton resources from a jenkins-x.yaml needed to process the pipeline.  

ATM this will also create the Source Repository resource which in turn creates prow config so that Tide now watches the repos that contain the jenkins-x.yml.  This is not required as another cluster is responsible for for Tide merging and responding to webhook requests.

Chatting to @abayer and @pmuir it seems the `SR` is created when determining the next build number.  It's been suggested we could pass in a flag or something to avoid the behaviour we are seeing or other clusters having Prow config created that watch repos when we only want an pipeline executed adhoc.

/cc @abayer

abayer:
#### Submitter checklist

#### Description

With the new `UseActivityForNextBuildNumber` and `UseBranchAsRevision` boolean parameters for `PipelineCreateParam`, you can force the metapipeline to:

* Determine the next build number by looking at the build numbers on `PipelineActivity`s for this org/repo/branch combination, rather than using (and if need be creating) the `SourceRepository`.
* Pass the new `--branch-as-revision` flag to `jx step create task`, which will override the default behavior of using the new version tag to checkout for release pipelines, instead just using the given branch.

/assign @hferentschik 
/assign @pmuir 
/assign @rawlingsj 

fixes #5909

### DIFF
--- a/pkg/cmd/step/create/step_create_task.go
+++ b/pkg/cmd/step/create/step_create_task.go
@@ -79,45 +79,46 @@ var (
 type StepCreateTaskOptions struct {
 	step.StepOptions
 
-	Pack              string
-	BuildPackURL      string
-	BuildPackRef      string
-	PipelineKind      string
-	Context           string
-	CustomLabels      []string
-	CustomEnvs        []string
-	NoApply           *bool
-	DryRun            bool
-	InterpretMode     bool
-	DisableConcurrent bool
-	StartStep         string
-	EndStep           string
-	Trigger           string
-	TargetPath        string
-	SourceName        string
-	CustomImage       string
-	DefaultImage      string
-	CloneGitURL       string
-	Branch            string
-	Revision          string
-	PullRequestNumber string
-	DeleteTempDir     bool
-	ViewSteps         bool
-	EffectivePipeline bool
-	NoReleasePrepare  bool
-	Duration          time.Duration
-	FromRepo          bool
-	NoKaniko          bool
-	SemanticRelease   bool
-	KanikoImage       string
-	KanikoSecretMount string
-	KanikoSecret      string
-	KanikoSecretKey   string
-	ProjectID         string
-	DockerRegistry    string
-	DockerRegistryOrg string
-	AdditionalEnvVars map[string]string
-	PodTemplates      map[string]*corev1.Pod
+	Pack                string
+	BuildPackURL        string
+	BuildPackRef        string
+	PipelineKind        string
+	Context             string
+	CustomLabels        []string
+	CustomEnvs          []string
+	NoApply             *bool
+	DryRun              bool
+	InterpretMode       bool
+	DisableConcurrent   bool
+	StartStep           string
+	EndStep             string
+	Trigger             string
+	TargetPath          string
+	SourceName          string
+	CustomImage         string
+	DefaultImage        string
+	CloneGitURL         string
+	Branch              string
+	Revision            string
+	PullRequestNumber   string
+	DeleteTempDir       bool
+	ViewSteps           bool
+	EffectivePipeline   bool
+	NoReleasePrepare    bool
+	Duration            time.Duration
+	FromRepo            bool
+	NoKaniko            bool
+	SemanticRelease     bool
+	KanikoImage         string
+	KanikoSecretMount   string
+	KanikoSecret        string
+	KanikoSecretKey     string
+	ProjectID           string
+	DockerRegistry      string
+	DockerRegistryOrg   string
+	AdditionalEnvVars   map[string]string
+	PodTemplates        map[string]*corev1.Pod
+	UseBranchAsRevision bool
 
 	GitInfo              *gits.GitRepository
 	BuildNumber          string
@@ -176,6 +177,7 @@ func NewCmdStepCreateTaskAndOption(commonOpts *opts.CommonOptions) (*cobra.Comma
 	cmd.Flags().BoolVarP(&options.ViewSteps, "view", "", false, "Just view the steps that would be created")
 	cmd.Flags().BoolVarP(&options.EffectivePipeline, "effective-pipeline", "", false, "Just view the effective pipeline definition that would be created")
 	cmd.Flags().BoolVarP(&options.SemanticRelease, "semantic-release", "", false, "Enable semantic releases")
+	cmd.Flags().BoolVarP(&options.UseBranchAsRevision, "branch-as-revision", "", false, "Use the provided branch as the revision for release pipelines, not the version tag")
 
 	options.AddCommonFlags(cmd)
 	options.setupViper(cmd)
@@ -514,7 +516,7 @@ func (o *StepCreateTaskOptions) createEffectiveProjectConfigFromOptions(tektonCl
 			o.BuildNumber = "1"
 		} else {
 			log.Logger().Debugf("generating build number...")
-			o.BuildNumber, err = tekton.GenerateNextBuildNumber(tektonClient, jxClient, ns, o.GitInfo, o.Branch, o.Duration, o.Context)
+			o.BuildNumber, err = tekton.GenerateNextBuildNumber(tektonClient, jxClient, ns, o.GitInfo, o.Branch, o.Duration, o.Context, false)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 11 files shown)


abayer:
Tests still needed - just wanted to make sure that this didn't break any existing behavior and that it feels generally right.

abayer:
Ignore the first round of `tekton`, `boot-vault`, `bdd` failures - that's just due to the new clusters for the second run spinning up and the first run's clusters getting GC'd.

abayer:
@hferentschik - tests added. I think this is good to go.

pkg/cmd/step/create/step_create_task.go
```diff
@@ -176,6 +177,7 @@ func NewCmdStepCreateTaskAndOption(commonOpts *opts.CommonOptions) (*cobra.Comma
 	cmd.Flags().BoolVarP(&options.ViewSteps, "view", "", false, "Just view the steps that would be created")
 	cmd.Flags().BoolVarP(&options.EffectivePipeline, "effective-pipeline", "", false, "Just view the effective pipeline definition that would be created")
 	cmd.Flags().BoolVarP(&options.SemanticRelease, "semantic-release", "", false, "Enable semantic releases")
+	cmd.Flags().BoolVarP(&options.UseBranchAsRevision, "branch-as-revision", "", false, "Use the provided branch as the revision for release pipelines, not the version tag")
```
hferentschik: Not quite sure the intent of the flag becomes clear here. 

So why is this even a release pipeline if it is not supposed to be doing release stuff? Does this not come down to the fact that we are missing something conceptually, namely the ability to run anything else than a release or pr pipeline?

pkg/cmd/step/create/step_create_task.go
```diff
@@ -514,7 +516,7 @@ func (o *StepCreateTaskOptions) createEffectiveProjectConfigFromOptions(tektonCl
 			o.BuildNumber = "1"
 		} else {
 			log.Logger().Debugf("generating build number...")
-			o.BuildNumber, err = tekton.GenerateNextBuildNumber(tektonClient, jxClient, ns, o.GitInfo, o.Branch, o.Duration, o.Context)
+			o.BuildNumber, err = tekton.GenerateNextBuildNumber(tektonClient, jxClient, ns, o.GitInfo, o.Branch, o.Duration, o.Context, false)
```
hferentschik: so here it is always false?

pkg/cmd/step/create/step_create_task.go
```diff
@@ -1254,7 +1256,11 @@ func (o *StepCreateTaskOptions) setBuildVersion(pipelineConfig *jenkinsfile.Pipe
 			version = "0.0.1"
 		}
 		o.version = version
-		o.Revision = "v" + version
+		if o.UseBranchAsRevision {
```
hferentschik: Extract this into a function? Seems to be duplicated below.

pkg/tekton/pipelines.go
```diff
@@ -118,13 +120,56 @@ func CreateOrUpdateTask(tektonClient tektonclient.Interface, ns string, created
...
-func GenerateNextBuildNumber(tektonClient tektonclient.Interface, jxClient jxClient.Interface, ns string, gitInfo *gits.GitRepository, branch string, duration time.Duration, context string) (string, error) {
+func GenerateNextBuildNumber(tektonClient tektonclient.Interface, jxClient jxClient.Interface, ns string, gitInfo *gits.GitRepository, branch string, duration time.Duration, context string, useActivity bool) (string, error) {
 	nextBuildNumber := ""
 	resourceInterface := jxClient.JenkinsV1().SourceRepositories(ns)
+	activityInterface := jxClient.JenkinsV1().PipelineActivities(ns)
 	// TODO: How does SourceRepository handle name overlap?
 	sourceRepoName := naming.ToValidName(gitInfo.Organisation + "-" + gitInfo.Name)
 
 	f := func() error {
+		if useActivity {
```
hferentschik: What's about splitting this function? Given that the two code paths are quite different I am wondering whether it even makes sense to have a BuildNumberGenerator interface for which you instantiate the right type depending on the flag. This would need some more thought probably. The idea would be to have a better separation of concerns and less entangled code.

pkg/cmd/step/create/step_create_task.go
abayer: Yup, that's the situation.

pkg/cmd/step/create/step_create_task.go
abayer: Correct - I decided not to support this for non-metapipeline.

pkg/tekton/pipelines.go
abayer: I can split this at the least - let's table the interface approach for later thought.

hferentschik: ok

pkg/cmd/step/create/step_create_task.go
```diff
@@ -1293,7 +1299,11 @@ func (o *StepCreateTaskOptions) setBuildVersion(pipelineConfig *jenkinsfile.Pipe
...
 		}
-		o.Revision = "v" + version
+		if o.UseBranchAsRevision {
+			o.Revision = o.Branch
+		} else {
+			o.Revision = "v" + version
+		}
 	} else {
 		// lets use the branch name if we can find it for the version number
 		branch := o.Branch
```
hferentschik: So here we use the branch already, right? Which is the pr pipelinetype I guess. So changing the pipeline type is not an option?

abayer: Yeah, the use case still wants to have "normal" `pullrequest` pipeline builds, so we're a little stuck. I considered just creating a new pipeline type, but that's a pretty complex thing. We can revisit later, though.

pkg/tekton/pipelines.go
```diff
@@ -117,54 +120,115 @@ func CreateOrUpdateTask(tektonClient tektonclient.Interface, ns string, created
...
-				// which could be due to name clashes
-				continue
+// GenerateNextBuildNumber generates a new build number for the given project.
+func GenerateNextBuildNumber(tektonClient tektonclient.Interface, jxClient jxClient.Interface, ns string, gitInfo *gits.GitRepository, branch string, duration time.Duration, context string, useActivity bool) (string, error) {
+	nextBuildNumber := ""
+	activityInterface := jxClient.JenkinsV1().PipelineActivities(ns)
+
+	f := func() error {
+		if useActivity {
+			bn, err := nextBuildNumberFromActivity(activityInterface, gitInfo, branch)
```
hferentschik: +1

pkg/cmd/step/create/step_create_task.go
hferentschik: > Yeah, the use case still wants to have "normal" pullrequest pipeline builds, so we're a little stuck. I considered just creating a new pipeline type, but that's a pretty complex thing. We can revisit later, though.

that is my thinking as well. Is that really so complex. My fear is that we keep digging us deeper and deeper in.

abayer: So my thinking is that if there are any more tweaks like this needed in the future, we say "yeah, this needs a new pipeline type" - but I feel like just this one flag is small enough to not merit that work yet.

hferentschik: Where is the complexity in creating a new type? Because of the jenskins-x.yaml syntax and the fact that the release kind is "hard coded"?

What I mean is:
```yaml
  pipelines:
    release:
      pipeline:
```

Here we have a fixed "release" pipeline, which would be opposed to something more flexible like:

```yaml
  pipelines:
      pipeline:
     kind: "release"
```

Is this the issue? Could we not have hard coded "release" and "pullrequest", but then have "free form" pipelines of the type above? 

From a metapipline point of view a new pipeline type does not seem to be hard, right?

abayer: It means going through everywhere that we refer to or use pipeline types and adding the new one, adding support for it in metapipeline/`step_create_task`, and most importantly, deciding what exactly will distinguish the new pipeline type from the others. If we're going to add a new pipeline type, I want to know exactly _why_ we're adding it, how it differs from the others, what the use case is for it, etc. i.e., I don't want to create another `feature`, given that literally no one knows what you would use _that_ for. =)

abayer: ...which is why I feel that adding these two flags to metapipeline, which can only be used by something that's calling metapipeline programmatically, is less intrusive/complex/etc. Basically, I'm fine with a small addition of complexity for "internal" APIs for a specific use case, but less so for anything that involves changing the _userfacing_ APIs - i.e., what you can do in a `jenkins-x.yml` and what that means behaviorally.

hferentschik:
@abayer yet another rebase needed ;-)

abayer:
Rebased!

PR #5912, opened 2019-10-23
feat: added mergo to handle the differences between versions of jx-requirements.yml when upgrading a cluster

### LINKED ISSUE #5896: jx upgrade boot wipes user requirements
### Summary 

When running `jx upgrade boot` some user modifications to jx-requirements.yml can be wiped. This is due to the cherry picking commits from the boot config repository in to the dev env repository that defaults to accept theirs when a conflict is found. This can be seen below

`// CherryPickTheirs does a git cherry-pick of commit
func (g *GitCLI) CherryPickTheirs(dir string, commitish string) error {
	return g.gitCmd(dir, "cherry-pick", commitish, "--strategy=recursive", "-X", "theirs")
}`

We should exclude jx-requirements.yml from the cherry pick and find a way to correctly merge the changes.

### Expected behavior

After running `jx upgrade boot` the jx-requirements.yml contains both user changed settings and updates from the upstream boot repo.

### Actual behavior

After running `jx upgrade boot` the jx-requirements.yml is missing user changed settings.

dgozalo:
#### Submitter checklist

#### Description
Created a new function to merge a given requirements struct with the calling one's data and save the output in the given path.

We are using `mergo` to do so. It will work for any struct and any number of nested structs within the requirements file.

A Transformer however was needed to handle the slice of EnvironmentConfig structs as we want to modify any existing ones with the ones provided in the given requirements file and also append to the slice if there's a new one.

Mergo will always take the calling requirements as base and replace any property with the given requirement's data if and only if the property from the given requirements file is non-zero valued.

Also removed the function that obtained the VersionsStreamConfig from the requirements file because it basically checked that the file existed and returned the versions stream ref struct. These checks are already done when loading the requirements file so it wasn't needed.

fixes #5896


abayer:
/lgtm

It'd be nice to have a test for how it handles new fields in the "upstream" `jx-requirements.yaml` that didn't exist in the current `jx-requirements.yaml`, since I know that's one of the things prompting this, but I'm not sure how one would do that effectively, so hey. =)

PR #5915, opened 2019-10-23
fix: BDD test reports don't stash after using a test cluster

abayer:
#### Submitter checklist

#### Description

So let's delete `~/.kube/config`, which just contains the test cluster in it, before we stash anything.

/assign @dgozalo 

n/a

### DIFF
--- a/jenkins-x-bdd.yml
+++ b/jenkins-x-bdd.yml
@@ -152,6 +152,10 @@ pipelineConfig:
                   - --suite-name
                   - "Static_BDD_Tests"
 
+              - name: clear-kubeconfig
+                image: gcr.io/jenkinsxio/builder-go:${inputs.params.version}
+                command: rm ~/.kube/config
+
               - name: stash-html-report
                 image: gcr.io/jenkinsxio/builder-go:${inputs.params.version}
                 command: jx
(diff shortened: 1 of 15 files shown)


abayer:
/hold

Turns out that wasn't sufficient - it's still stashing to, e.g., `gs://jx-prod-logs/jenkins-x/static-e2e-tests/jenkins-x/jx/gh-pages/16/workspace/source/build/reports/junit.html`.

abayer:
/hold cancel

Ok, I got annoyed at the bucket paths being wrong so went and fixed it while I was here. =)

jenkins-x-bdd.yml
```diff
@@ -152,6 +152,10 @@ pipelineConfig:
                   - --suite-name
                   - "Static_BDD_Tests"
 
+              - name: clear-kubeconfig
```
hferentschik: what's that good for?

hferentschik:
@abayer rebase needed :-)

jenkins-x-bdd.yml
abayer: It's making sure we're no longer using the test cluster context when we subsequently run `jx step stash`.

abayer:
Rebased, after an initial oopsie where my editor got confused and I inadvertently committed the merge conflict itself. Fixed now. =)

PR #5916, opened 2019-10-23
fix(pipeline): Add GIT_* env vars to non-pod template images

### LINKED ISSUE #5615: Regression on `jx step create pr`
### Summary 

`jx step create pr chart` used inside a pipeline cause an error regarding GIT USER and GIT EMAIL

Upgrade to latest platform

### Expected behavior

The PR should be sent

### Actual behavior

The PR failed to be sent

### Jx version

```
NAME               VERSION
jx                 2.0.757
jenkins x platform 2.0.1296
Kubernetes cluster v1.13.7-gke.8
kubectl            v1.15.2
git                2.20.1 (Apple Git-117)
Operating System   Mac OS X 10.14.6 build 18G95
```

### Jenkins type

- [ x] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

Workaround shown in here: 

https://github.com/zeebe-io/zeebe-cluster-helm/blob/master/jenkins-x.yml
Which is adding to the pipeline configuration these env variables: 
```
buildPack: charts 
pipelineConfig:
  env:
  - name: GIT_COMMITTER_EMAIL
    value: jenkins-x@googlegroups.com
  - name: GIT_AUTHOR_EMAIL
    value: jenkins-x@googlegroups.com
  - name: GIT_AUTHOR_NAME
    value: jenkins-x-bot
  - name: GIT_COMMITTER_NAME
    value: jenkins-x-bot
```

abayer:
#### Submitter checklist

#### Description

If a step uses an image that isn't a pod template alias, like `go` or `maven`, and instead uses an image name, with or without tag, such as `kaniko/executor` or `gcr.io/jenkinsxio/builder-go:2.0.902-250`, we haven't been adding all the environment variables that pod templates automatically include, meaning those env vars aren't set unless they're explicitly specified in the `jenkins-x.yml` or build pack. So far as I can tell, the only ones we're missing right now are `GIT_AUTHOR_NAME` and friends. So this PR adds them.

Figuring out exactly the right value to use is a challenge. The pod templates all have `jenkins-x-bot` and `jenkins-x@googlegroups.com` as the name/email, but it feels like it'd be better to use `PipelineUsername` and `PipelineUserEmail` from team settings...if they're set. Since `PipelineUserEmail` was just added a month and a half ago, we can't assume it's going to be set in existing installs, even after updating, so we'll use those values if they're not empty, and fall back on the defaults if needed.

I also added constants for the default values, because why not.

I've updated the existing `step_create_task_test` test cases - in most cases, it's just adding the env vars to containers using non-pod-template images, but I also configured `kaniko_entrypoint` to set the user/email in team settings, `default-in-buildpack` to set just the email, and `pipeline-timeout` to set just the name, so that we have verification of the use of team settings values when they're set.

/assign @pmuir 
/assign @hferentschik 

fixes #5615


pkg/util/constants.go
```diff
@@ -15,4 +15,10 @@ const (
 
 	// PlaceHolderDockerRegistryOrg placeholder for docker registry
 	PlaceHolderDockerRegistryOrg = PlaceHolderPrefix + "_DOCKER_REGISTRY_ORG"
+
+	// DefaultGitUserName default value to use for git "user.name"
+	DefaultGitUserName = "jenkins-x-bot"
```
hferentschik: +1 for collecting this in one place.

PR #5919, opened 2019-10-24
fix: add a deprecation warning to `jx install`/`jx upgrade platform`

jstrachan:
to give folks a hint they should switch to `jx boot`

### DIFF
--- a/pkg/cmd/opts/deprecations.go
+++ b/pkg/cmd/opts/deprecations.go
@@ -1,0 +1,14 @@
+package opts
+
+import (
+	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/util"
+)
+
+// LogInstallDeprecated logs a warning message about deprecation
+func LogInstallDeprecated() {
+	log.Logger().Warn("this command is now deprecated.")
+	log.Logger().Warnf("We now highly recommend you use %s instead:", util.ColorInfo("jx boot"))
+	log.Logger().Warnf("for more information see: %s\n\n", util.ColorStatus("https://jenkins-x.io/docs/getting-started/setup/boot/"))
+
+}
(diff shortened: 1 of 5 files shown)


garethjevans:
@jstrachan should we add this to `jx create cluster <xxx>` too? (maybe if the --skip-installation flag is not set)

jstrachan:
@garethjevans good call! I did wonder about doing that - but I wondered if `jx create cluster` without the `--skip-installation` would invoke install and so get the warning anyway. I've not tested that though ;)

pkg/cmd/opts/deprecations.go
```diff
@@ -0,0 +1,14 @@
+package opts
+
+import (
+	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/util"
+)
+
+// LogInstallDeprecated logs a warning message about deprecation
+func LogInstallDeprecated() {
```
ccojocar: I would make this function more generic, that other commands can use it. e.g. provide the `recommended command` as an argument and also the link to the docs.

ccojocar: also `Cobra Command` seems to have a `Deprecated` flag which can be set on the command.

```
type Command struct {
	// Deprecated defines, if this command is deprecated and should print this string when used.
	Deprecated string
```

which prints this message before the command is executed:

```Go
func (c *Command) execute(a []string) (err error) {
	if c == nil {
		return fmt.Errorf("Called Execute() on a nil Command")
	}

	if len(c.Deprecated) > 0 {
		c.Printf("Command %q is deprecated, %s\n", c.Name(), c.Deprecated)
	}
```

I am wondering if we can leverage this instead.

ccojocar:
@jstrachan We can close this PR. I implemented a more generic mechanics to mark commands as deprecated. Please have a look over https://github.com/jenkins-x/jx/pull/6063. Thanks!

jstrachan:
@ccojocar awesome thanks!

PR #5924, opened 2019-10-24
fix: Remove defunct CloudBees addon

### LINKED ISSUE #5923: Remove defunct CloudBees addon
This isn't a thing any more, so it should be removed.

abayer:
#### Submitter checklist

fixes #5923


hferentschik:
/lgtm

These ones you can bringing on :-)

pkg/cmd/cloudbees/cloudbees.go
```diff
@@ -1,224 +0,0 @@
...
-func NewCmdCloudBees(commonOpts *opts.CommonOptions) *cobra.Command {
-	options := &CloudBeesOptions{
-		CommonOptions: commonOpts,
-	}
-	cmd := &cobra.Command{
-		Use:     "cloudbees",
-		Short:   "Opens the CloudBees app for Kubernetes for visualising CI/CD and your environments",
-		Long:    core_long,
-		Example: core_example,
-		Aliases: []string{"cloudbee", "cb", "ui", "jxui"},
```
hferentschik: On closer inspection this should not be removed ;-)

This is the `jx ui` command which does the port forward to open the UI on localhost in case the UI is not installed with SSO. 

I think it is unfortunate that the commands main name is 'cloudbees'. I guess the description is also a bit off as well. 

Given that the way we document this in the distribution docs is `jx ui`, what we could do is:

* make the main name of this command "ui"
* rename file and package to get rid of 'cloudbees'
* add 'cloudbees' as alias for good measure in case someone is actually using it

BTW,  that seems wrong as well:

> 		# Print the Jenkins X console URL but do not open a browser
>		jx console -u`)

Should probably be `jx ui -u`

abayer: whoops! I'll clean this up on Monday.

ccojocar:
@abayer I have already removed some parts (the addon command for this) but forgot the login. Thanks  for removing it.

abayer:
/hold

As Hardy pointed out, I kinda overshot here. I’ll fix it up today, but holding it for the moment.

abayer:
/hold cancel

Ok, updated, removing hold.

PR #5925, opened 2019-10-24
feat: set the logging level via JX_LOG_LEVEL

### LINKED ISSUE #5080: Use JX_LOG_LEVEL env variable to configure log level
### Summary 

The '--log-level' CLI option got removed in favour of a single `--verbose` flag. The `--verbose` only increases default info to debug log level. There is no other way to set a log level which. 

The idea is to similar to the `JX_LOG_FORMAT` env variable use a `JX_LOG_LEVEL` variable to control the log level aside of the more user facing `--verbose`.

This issue is tightly related to #4581. IMO --verbose is not the same as changing the log level. It increases the default output to the user, making it more verbose. This is not the same as setting a log level. Logging and in- and output of a command are separate things which we merged together, but treating the info logger as default writer for user stdout.

hferentschik:
fixes #5080 

@ilpianista, here is the new pull request. I added a test as well. Sorry about messing with your original PR #5717.

### DIFF
--- a/charts/jx/values.yaml
+++ b/charts/jx/values.yaml
@@ -107,6 +107,9 @@ env:
   # lets use JSON output in logs
   JX_LOG_FORMAT: json
 
+  # lets set the logging level
+  JX_LOG_LEVEL: info
+
   # lets mark this pod as being a CI/CD pipeline job so reuse the pipeline git user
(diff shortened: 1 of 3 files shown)


ilpianista:
Thanks for implementing the test!

PR #5926, opened 2019-10-24
feat: Completion for namespace and environment

### LINKED ISSUE #6168: Bash completion for namespace and environment commands
Configure bash completion for the environment and namespace commands. See pull request #5926.

msvticket:
This is a first stab at adding completion for command and flag values.

Fixes #6168


hferentschik:
@msvticket, thanks for the PR. 

> This is a first stab at adding completion for command and flag values.

Would you mind creating an issue for this and outlining what your goal is here?
It probably also would help to give some context on the issue or here on how to use this.

pkg/cmd/cmd.go
```diff
@@ -75,11 +76,41 @@ func NewJXCommand(f clients.Factory, in terminal.FileReader, out terminal.FileWr
 	err io.Writer, args []string) *cobra.Command {
 
 	configureViper()
+
+	// It is likely that the user has the completions for kubectl loaded, so reusing function from there if they exist
+	bashCompletionFunction := `
```
ccojocar: There is a completion command in `pkg/cmd/completion.go` which generates the completions for whole commands. I am wondering if you could move this into that command.

msvticket: OK

msvticket:
> @msvticket - so I've done:
> 
> ```shell
> >  jx completion bash > ~/.jx/bash
> > source ~/.jx/bash
> ```
> 
> but I don't think this additional completion functions are exported.
> 
> ```shell
> $ jx completion bash | grep __jx_get_env
> ```
> 
> does not show anything either yet. Am I doing anything wrong?
> 
> Can you explain on how to use/test this?

Cosmins request made me break the functionality. Now I've fixed it again, so it works the way you expected.

The new functionality is completion on `jx namespace`, `jx environment` and `jx get environments --promote`.

hferentschik:
@msvticket , nice. Now it works. Thanks for the PR.

/lgtm

PR #5928, opened 2019-10-24
fix: Don't patch unnecessarily

### LINKED ISSUE #5937: PatchUpdate in the Kube client expansion should not call Patch for empty patches
The PatchUpdate function in the various Kube client expansions calls Patch of the Kube client, even for empty diffs. In this case, an unnecessary call is made which needs to go over the wire.

In the case the diff is empty we should just return the original object.

msvticket:
This PR prevetns patch calls if the patch actually is empty.

Specifically this fixes problems when DevEnvAndTeamSettings in team_settings modifies the environment unnecessarily, thus mitigating problems for user without update permission.

fixes #5937

### DIFF
--- a/pkg/client/clientset/versioned/typed/jenkins.io/v1/app_expansion.go
+++ b/pkg/client/clientset/versioned/typed/jenkins.io/v1/app_expansion.go
@@ -1,6 +1,8 @@
 package v1
 
 import (
+	"bytes"
+
 	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
 	util "github.com/jenkins-x/jx/pkg/util/json"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -27,6 +29,9 @@ func (c *apps) PatchUpdate(app *v1.App) (*v1.App, error) {
 	patch, err := util.CreatePatch(orig, app)
 	if err != nil {
 		return nil, err
+	}
+	if bytes.Equal(patch, []byte("[]")) {
+		return orig, nil
 	}
 	patchedApp, err := c.Patch(resourceName, types.JSONPatchType, patch)
 	if err != nil {
--- a/pkg/client/clientset/versioned/typed/jenkins.io/v1/buildpack_expansion.go
+++ b/pkg/client/clientset/versioned/typed/jenkins.io/v1/buildpack_expansion.go
@@ -1,6 +1,8 @@
 package v1
 
 import (
+	"bytes"
+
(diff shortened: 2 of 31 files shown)


hferentschik:
Awesome, thanks for the PR @msvticket. Would you mind creating an issue as well and referencing it in this PR. Would be nice to even reference the issue in the commit message as well.

abayer:
FYI, here's the test failure:
```
=== FAIL: pkg/util/json TestCreatePatchNoDiff (0.00s)
=== PAUSE TestCreatePatchNoDiff
=== CONT  TestCreatePatchNoDiff
    patch_test.go:48: 
        	Error Trace:	patch_test.go:48
        	Error:      	Not equal: 
        	            	expected: <nil>(<nil>)
        	            	actual  : []uint8([]byte(nil))
        	Test:       	TestCreatePatchNoDiff
        	Messages:   	the patch should have been empty
```

pkg/client/clientset/versioned/typed/jenkins.io/v1/app_expansion.go
```diff
@@ -28,6 +28,9 @@ func (c *apps) PatchUpdate(app *v1.App) (*v1.App, error) {
 	if err != nil {
 		return nil, err
 	}
+	if patch == nil {
```
hferentschik: Have you verified this? I don't think the patch will ever be nil. If the two versions are identical '[]' is returned. S
 something like `if string(patch) == "[]" ` would do. Could you verify this?

Other than that I am all up for saving some unnecessary calls.

pkg/util/json/patch.go
```diff
@@ -30,6 +30,9 @@ func CreatePatch(before, after interface{}) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	if len(patch) == 0 {
```
hferentschik: ahh, you modified the code here as well to return nil. Given that this is supposed to be a generic function, I am not sure whether it makes more sense to handle the empty JSON diff "[]" higher up.

pkg/client/clientset/versioned/typed/jenkins.io/v1/app_expansion.go
msvticket: Well, I changed util.CreatePatch to return nil instead to make these if statements simpler. Or so I thought. I had missed the fact that if you return nil as an array it actually becomes an empty array and not nil. So I I'll revert that change in util.CreatePatch and change the if statements. I think it is uglier, but I can't think of a prettier solution right now.

PR #5930, opened 2019-10-24
fix: Reducing and structuring the log output of `jx boot`

### LINKED ISSUE #5927: Reduce the verbosity of `jx boot` execution
The output of `jx boot` is very verbose with a lot of messages in particular from Helm templating. It would be nice to reduce the verbosity of this.

hferentschik:
fixes #5927

You get a more verbose output which contains most of the original output of `jx boot` via:

```
$ jx --verbose boot
# or
$ JX_LOG_LEVEL=debug jx boot
```

Still room for improvement, but imo the default output is a bit saner.

Would love to separate default UI output (to stdout) from logging. It really is unfortunate that we use the logger for both.

### DIFF
--- a/pkg/cmd/boot/boot.go
+++ b/pkg/cmd/boot/boot.go
@@ -245,7 +245,7 @@ func (o *BootOptions) Run() error {
 		return err
 	}
 	if !isBootClone {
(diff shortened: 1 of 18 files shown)


abayer:
yaaaaaay!

PR #5931, opened 2019-10-24
fix: GitHubProvider#asGitHubCommit should extract author name and email from commit

### LINKED ISSUE #5726: GitHubProvider#asGitHubCommit should extract author name and email from commit

khansaAmrouni:
fixes #5726

asGithubCommit should extract author name and email from commit using commit.Commit.Author to get author name and email specified in the commit instead of getting them from commit.Author.

### DIFF
--- a/pkg/gits/github.go
+++ b/pkg/gits/github.go
@@ -782,13 +782,13 @@ func (p *GitHubProvider) asGitHubCommit(commit *github.RepositoryCommit) GitComm
 		log.Logger().Warnf("No Commit object for for commit: %s", commit.GetSHA())
 	}
 	var author *GitUser
-	if commit.Commit.Author != nil {
+	if commit.Author != nil {
 		author = &GitUser{
-			Login:     commit.Commit.Author.GetLogin(),
+			Login:     commit.Author.GetLogin(),
 			Email:     commit.Commit.Author.GetEmail(),
 			Name:      commit.Commit.Author.GetName(),
-			URL:       commit.Commit.Author.GetURL(),
-			AvatarURL: commit.Commit.Author.GetAvatarURL(),
+			URL:       commit.Author.GetURL(),
+			AvatarURL: commit.Author.GetAvatarURL(),
 		}
 	} else {
 		log.Logger().Warnf("No author for commit: %s", commit.GetSHA())
--- a/pkg/gits/github.go
+++ b/pkg/gits/github.go
@@ -782,13 +782,13 @@ func (p *GitHubProvider) asGitHubCommit(commit *github.RepositoryCommit) GitComm
 		log.Logger().Warnf("No Commit object for for commit: %s", commit.GetSHA())
 	}
 	var author *GitUser
-	if commit.Author != nil {
+	if commit.Commit.Author != nil {
 		author = &GitUser{
-			Login:     commit.Author.GetLogin(),
-			Email:     commit.Author.GetEmail(),
-			Name:      commit.Author.GetName(),
-			URL:       commit.Author.GetURL(),
-			AvatarURL: commit.Author.GetAvatarURL(),
+			Login:     commit.Commit.Author.GetLogin(),
+			Email:     commit.Commit.Author.GetEmail(),
+			Name:      commit.Commit.Author.GetName(),
+			URL:       commit.Commit.Author.GetURL(),
+			AvatarURL: commit.Commit.Author.GetAvatarURL(),
 		}
 	} else {
 		log.Logger().Warnf("No author for commit: %s", commit.GetSHA())


pkg/gits/github.go
```diff
@@ -785,8 +785,8 @@ func (p *GitHubProvider) asGitHubCommit(commit *github.RepositoryCommit) GitComm
 	if commit.Author != nil {
 		author = &GitUser{
 			Login:     commit.Author.GetLogin(),
-			Email:     commit.Author.GetEmail(),
-			Name:      commit.Author.GetName(),
+			Email:     commit.Commit.Author.GetEmail(),
+			Name:      commit.Commit.Author.GetName(),
```
hferentschik: As I mentioned on the previous pull request. I think it makes sense to verify that the commit author and the PR author match. They don't have to be the same. A pull request might contain commits from multiple users. 

In this approach, you could mix up the data from two different users into a single GitUser.

Really some testing is needed here.

khansaAmrouni: huum right! as i read the documentation, there are both author and committer.
according to the  [FREE online Pro Git book](https://git-scm.com/book/en/v2/Git-Basics-Viewing-the-Commit-History)

> The author is the person who originally wrote the patch, whereas the committer is the person who last applied the patch

so maybe we can mix both to a single GitUser?

hferentschik: > huum right! as i read the documentation, there are both author and committer.

There is even more than that. A given PR can contain commits from other authors altogehter. I, for example, could cherry-pick this commit and create my own pull request from it, in which case I would be the author/creator of the PR, but not the author of the containing commits. 

> so maybe we can mix both to a single GitUser?

No. A single GitUser needs to be consistent. We cannot mix data from two different users into it. They need to match.

hferentschik: I recommend you try the API yourself a bit. If you have your Git Hub token, you can do this:

```
$ curl --user "token:$GH_TOKEN" https://api.github.com/repos/jenkins-x/jx/pulls/5939/commits
```

The above is a test PR I created and it shows that a PR can contain commits for which there is no commit.Author data at all.

You can try the above curl command against a few pull requests to see what they return.

Regarding the code, we know that the commit.Commit.Author always has an email and name since git won't let you commit without this information. That means you can always create a GitUser with this information. commit.Authorcan be null though (see above example). So at the very least you have to create the GitUser outside of the `if commit.Author != nil` statement and then within the if statement amend the other fields if commit.Author != nil.

Also, if you look at https://developer.github.com/v3/pulls/#list-commits-on-a-pull-request, you see that there is no email and name for the commit.Author, that's why these two fields will always be empty atm, forcing us to go to the file system to get the information, like here: https://github.com/jenkins-x/jx/blob/593d0aabe9ffd02c2dc12dd1f8328dace0d5207d/pkg/gits/github.go#L818

```go
	for _, commit := range commits {
		summary := p.asGitHubCommit(commit)
		if commit.Author != nil {
			if summary.Author.Email == "" {
				log.Logger().Infof("Commit author email is empty for: %s", commit.GetSHA())
				dir, err := os.Getwd()
				if err != nil {
					return answer, err
				}
				gitDir, _, err := p.Git.FindGitConfigDir(dir)
				if err != nil {
					return answer, err
				}
				log.Logger().Infof("Looking for commits in: %s", gitDir)
				email, err := p.Git.GetAuthorEmailForCommit(gitDir, commit.GetSHA())
				if err != nil {
					log.Logger().Warnf("Commit not found: %s", commit.GetSHA())
					continue
				}
				summary.Author.Email = email
			}
		}
		answer = append(answer, &summary)
	}
```

hferentschik:
BTW, it would be great if you could make sure that your commit messages use the conventional commits format.  See also https://jenkins-x.io/docs/contributing/code/#the-commit-message

khansaAmrouni:
Thank you very much for your feedback. It was very helpful!
so, I would suggest this modification in the function code. 

```
	message := ""
        var author *GitUser
	if commit.Commit != nil {
                author = &GitUser{
			Login:     commit.Commit.Author.GetName(),
			Email:     commit.Commit.Author.GetEmail(),
                      }
	} else {
		log.Logger().Warnf("No Commit object for commit: %s", commit.GetSHA())
	}

        if commit.Author != nil {
                  author = &GitUser{
                  Name:      commit.Author.GetLogin(),
		}
	} else {
	}
	return GitCommit{
		Message: message,
		URL:     commit.GetURL(),
		SHA:     commit.GetSHA(),
		Author:  author,
	}

}
```
hope it is correct

hferentschik:
> Thank you very much for your feedback. It was very helpful!
so, I would suggest this modification in the function code.

I think your code is still not quite correct. You seem to mix up `GitUser.Login` and `GitUser.Name`. It is `GitUser.Name` you should populate from `commit.Commit.Author` not `GitUser.Name`.`

What's, even more, you are creating new instances of `GitUser` in each branch (`if commit.Commit != nil` and ` if commit.Author != nil`). This means if  `commit.Author` is not nil, you don't get the email you extracted from the commit itself. 

It is a step in the right direction, but still not quite right ;-)

hferentschik:
I am going to close this PR for now, since it needs not correct anyways. Once you have something more to review just open a new pull request.

PR #5943, opened 2019-10-25
fix: Correct command name in output

### LINKED ISSUE #5724: Istio addon suggests wrong flags to lock version
### Summary 
During execution of `jx create addon istio` the two charts print that you should lock down the istio version using: jx step create version pr -k charts -n install/kubernetes/helm/istio-init

This command fails due to -k not being an acceptable flag:
`Error: unknown shorthand flag: 'k' in -k`

Add istio through the command above and observe output.

### Expected behavior
Command that is provided should work with my environment. 

### Jx version

2.0.821

### Jenkins type

- [X] Classic Jenkins

### Kubernetes cluster

GKE

OSX

msvticket:
#### Description

Fix for faulty commands being output:

`jx step create version pr` -> jx step create pr versions`

fixes #5724


abayer:
The `boot-vault` failure is due to a change in the BDD tests that...looks like it's not working right with boot, so don't worry about it, it's being looked into. And the `integration` failure is a sporadically flaky one, so I'm rerunning it.

msvticket:
I can't see how the changes in this PR could affect a lint test. It only changes terminal output.

PR #5950, opened 2019-10-25
fix: fixing the gke cluster deletion to make it easier

DMEvanCT:
#### Submitter checklist

#### Description

Cluster deletion was confusing when using an arg instead of an option. Changed arg to option to make it clear. 

This should be done with all other cloud providers to come. Change coming after this is approved. 

fixes #


DMEvanCT:
@hferentschik I fixed a few things and moved it to a new pull request (Related to git). This seemed to be a bit cleaner. Same change

DMEvanCT:
@cagiti

PR #5952, opened 2019-10-26
fix: gke cluser deletion with option

DMEvanCT:
Signed-off-by: Evan Haston <ehaston@darkmatterit.io>

#### Submitter checklist

#### Description

Switched gke delete to use option instead of arg. It's easier to understand 

We may want to do this with eks too 

fixes #

### DIFF
--- a/pkg/cmd/deletecmd/delete_gke.go
+++ b/pkg/cmd/deletecmd/delete_gke.go
@@ -13,11 +13,14 @@ import (
 	"github.com/spf13/cobra"
 )
 
+
 // DeleteGkeOptions Options for deleting a GKE cluster
 type DeleteGkeOptions struct {
 	get.GetOptions
 	Region    string
 	ProjectID string
+	ClusterName string
+
 }
 
 var (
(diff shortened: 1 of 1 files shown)


DMEvanCT:
@hferentschik I updated the pull request to make it less confusing. Here is the correct one with the tests passing.

DMEvanCT:
@hferentschik  any word on this?

DMEvanCT:
@hferentschik  I can't see why the linter failed...

hferentschik:
@DMEvanCT, thanks for the pull request. Could you try to motivate this change a bit more? Why do you think it is easier to understand?

The issue I am having with this is that technically it is breaking backwards compatibility of the CLI. People might, for example, be using the old form of the command and scripts and would get problems when upgrading `jx`.

I understand the intention to make the `jx` CLI more consistent and may be easier to use, but I am not sure how we should best go about it. Having a lot of these seemingly small changes over time might be frustrating to users, so maybe a better strategy would be to do a more general CLI overhaul and let this go hand in hand with a new major version of the CLI. WDYT?

@abayer, any thoughts on this?

DMEvanCT:
@hferentschik  I agree. I can take a look at the overhaul to fix this kind of thing and make the changes.

hferentschik:
For what it's worth, the linting issues seem legit.

DMEvanCT:
> For what it's worth, the linting issues seem legit.

What seemed to be the issue?

hferentschik:
```
Gofmt errors in files:
./pkg/cmd/deletecmd/delete_gke.go
diff -u ./pkg/cmd/deletecmd/delete_gke.go.orig ./pkg/cmd/deletecmd/delete_gke.go
--- ./pkg/cmd/deletecmd/delete_gke.go.orig	2019-11-06 15:28:24.299680792 +0000
+++ ./pkg/cmd/deletecmd/delete_gke.go	2019-11-06 15:28:24.299680792 +0000
@@ -13,14 +13,12 @@
 	"github.com/spf13/cobra"
 )
 
-
 // DeleteGkeOptions Options for deleting a GKE cluster
 type DeleteGkeOptions struct {
 	get.GetOptions
-	Region    string
-	ProjectID string
+	Region      string
+	ProjectID   string
 	ClusterName string
-
 }
 
 var (
@@ -52,9 +50,6 @@
 			err := options.Run()
 			helper.CheckErr(err)
 
-
-
-
 		},
 	}
 	cmd.Flags().StringVarP(&options.Region, "region", "", "europe-west1-c", "GKE region to use. Default: europe-west1-c")
@@ -62,7 +57,6 @@
 	cmd.Flags().StringVarP(&options.ClusterName, "cluster-name", "c", "", "Google Cluster name to delete")
 	options.AddGetFlags(cmd)
 
-
```

You want to run `make fmt` on your feature branch. That will adjust your formatting. I think there are some superfluous empty lines.

hferentschik:
I think a good first step into the right direction would be to create an issue where we can collect the issue(s) we want to address. Then we can see if and how many we want to take on.

Regarding making the cluster to delete an option, I am wondering why you think it should be an option in the first place. It is pretty much a required parameter, is it not? In the case of `cluster create` the name is optional since if you don't specify a name will be generated. However,  at deletion time the name is required. Maybe also something to discuss on an issue.

For now, I am going to close this issue. I hope that's ok with you.

DMEvanCT:
It was better to see the option. People are not automatically aware that it is an arg

Issue #5669, opened 2019-10-01
Diagnose and fix sporadic hangs on environment promotes in boot-vault and boot-local versions BDD tests


abayer:
`boot-local` and `boot-vault` have a tendency to hang like this:

```
STEP: manually promoting app to production environment
      RUNNER:About to execute ** promote --env production --version 0.0.1 bdd-spring-1569932401 in /tmp/bdd-190803862 with timeout 1h0m0s expecting exit code 0
WARNING: prow based install so skip waiting for the merge of Pull Requests to go green as currently there is an issue with gettingstatuses from the PR, see https://github.com/jenkins-x/**/issues/2410
Namespace **-production created 
Promoting app bdd-spring-1569932401 version 0.0.1 to namespace **-production
WARNING: Could not discover the Git repository info failed to parse Git URL  due to Invalid path  could not determine organisation and repository name
Created Pull Request: https://github.com/*************-test/environment-pr-552-4-bdd-boot-v-production/pull/1
WARNING: Could not discover the Git repository info failed to parse Git URL  due to Invalid path  could not determine organisation and repository name
```

I've poked around the clusters when this happens, and what appears to be happening is that the `production` repo PR build goes fine, but the PR never gets merged. Since the repo is private now, I can't tell _why_ it never gets merged, but it definitely doesn't - the status gets updated properly (I see that happen in the `crier` logs), and I see `tide` continuing to sync, but nothing merges.

This is not a consistent thing - but it does seem to happen ~50+% of the time, which is very annoying.

abayer:
Of course, since I opened this, the hang right has dropped dramatically, so I'm deprioritizing this.

garethjevans:
@abayer this is happening on arcalos around 70% of the time - its really annoying

abayer:
That's using Lighthouse, right? Then one real possibility is that it's hitting the `pipeline not found` or `task not found` issues - our Prow setup automatically reruns pipelines that hit that, but LH doesn't have that logic.

garethjevans:
Well, we're currently using Prow.  Maybe switching to lighthouse would help?

abayer:
Oh, then no, switching to LH would make it worse. =) There's got to be some other factor causing the hang than the `pipeline not found` or `task not found` thing - Prow automatically retries in those cases.

abayer:
[...]

PR #5964, opened 2019-10-29
fix: use tide based merging for promotion

pmuir:

#### Submitter checklist

fixes #


abayer:
Looks good, but feels like we should do the same thing in `jx delete application` too.

abayer:
There's a very real possibility this may fix #5669, so once this is in, I'll watch very carefully for any future hangs.

PR #5966, opened 2019-10-29
feat: add a new auth config saver which can load the auth config from a configmap and resolve the secret URIs from vault

ccojocar:
#### Submitter checklist

part of #5114

### DIFF
--- a/pkg/auth/configmap_vault_config_saver.go
+++ b/pkg/auth/configmap_vault_config_saver.go
@@ -15,7 +15,7 @@ const (
 	labelAuthConfigValue = "auth"
 )
 
-// LoadConfig loads the auth config from a ConfigMap witch stores in its
+// LoadConfig loads the auth config from a ConfigMap which stores in its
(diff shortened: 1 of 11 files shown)


pkg/auth/configmap_vault_config_saver.go
```diff
@@ -0,0 +1,96 @@
...
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	labelAuthConfig      = "jenkins.io/config-type"
+	labelAuthConfigValue = "auth"
+)
+
+// LoadConfig loads the auth config from a ConfigMap witch stores in its
```
abayer: `s/witch/which` unless we've got an infestation of magic-using `ConfigMap`s flying around on broomsticks. =)

ccojocar: done

PR #5977, opened 2019-10-29
fix(boot): 'jx upgrade boot' needs to fetch upgrade boot config SHA before cherrypicking

abayer:
#### Submitter checklist

- [x] Change is code complete and matches issue description.
- [x] Change is covered by existing or new tests.

#### Description

Without doing that, the cherry-pick will fail due to the commit from the release tag not being present in the local boot repo, since it's not on master.

#### Special notes for the reviewer(s)

/assign @pmuir 
/assign @macox 

#### Which issue this PR fixes

fixes #5974

PR #5981, opened 2019-10-30
fix(versions): Delete and reclone version stream significantly less.

### LINKED ISSUE #5976: Version stream resolver always deletes and reclones for any version ref but a branch
This seems to be due to https://github.com/jenkins-x/jx/blob/bb422634a084f80aa2775682ca1034e5a966a3b9/pkg/versionstream/versionstreamrepo/gitrepo.go#L87-L89 - `refs/head/SOME_TAG` (or a commit-ish) is always going to fail, so far as I can tell, so `deleteAndReclone` will keep getting called every time you try to fetch anything but a branch.

abayer:
#### Submitter checklist

#### Description

By switching to an approach using our `Gitter` APIs rather than the `go-git` for handling an existing clone, we should drastically decrease the number of delete and reclones.

I verified that with this, I can get rid of the sleep I had to insert in #5977 to get the new test there to pass, and the test still passes.

/assign @pmuir 

fixes #5976


pkg/gits/git_cli.go
```diff
@@ -869,6 +869,11 @@ func (g *GitCLI) FetchTags(dir string) error {
 	return g.gitCmd(dir, "fetch", "--tags")
 }
 
+// FetchRemoteTags fetches all the tags from a remote repository
+func (g *GitCLI) FetchRemoteTags(dir string, repo string) error {
```
pmuir: Might be better to extend the command above, and check if repo is empty?

pkg/versionstream/versionstreamrepo/gitrepo.go
```diff
@@ -131,27 +79,51 @@ func cloneJXVersionsRepo(versionRepository string, versionRef string, settings *
...
+			err = gitter.FetchBranch(wrkDir, versionRepository, versionRef)
 			if err != nil {
 				dir, err := deleteAndReClone(wrkDir, versionRepository, versionRef, gitter)
 				if err != nil {
 					return "", "", errors.WithStack(err)
 				}
 				return dir, versionRef, nil
 			}
+			// TODO: It might be good to do the same behavior for any branch - just the checkout approach works for
+			// tags and commit-ishs, but not branches. I'm not sure how we can tell if it's a branch vs a tag, though.
```
pmuir: The only way I know how to do this is to use the FilterTags method to check if the tag exists or not

pkg/gits/git_cli.go
abayer: Yeah, I think I'm gonna do a dive through `git_cli` at some point soon and clean up a bunch of stuff.

PR #5984, opened 2019-10-30
fix: lets generate a PR on the dev environment when we import a project

jstrachan:
so that if folks re-create the cluster and re-run boot the SourceRepository
gets recreated

### DIFF
--- a/pkg/cmd/importcmd/import.go
+++ b/pkg/cmd/importcmd/import.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/jenkins-x/jx/pkg/cmd/step/create/pr"
 	"github.com/jenkins-x/jx/pkg/maven"
 
 	"github.com/cenkalti/backoff"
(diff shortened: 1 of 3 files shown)


abayer:
@jstrachan - ok, might be worth adding something in the BDD tests to verify the PR creation/merge, maybe?

jstrachan:
lemme add a bdd test that after the tests have completed (if gitops is enabled) we delete all the `SourceRepository` resources, re-run the dev pipeline and verify the SR's are all back again

abayer:
Thanks!

abayer:
Probably should add a `test-single-import` to `jx/bdd/boot-vault/ci.sh`, like this from versions:

https://github.com/jenkins-x/jenkins-x-versions/blob/master/jx/bdd/boot-lh/ci.sh#L47-L48

https://github.com/jenkins-x/jenkins-x-versions/blob/master/jx/bdd/boot-lh/ci.sh#L65

PR #5987, opened 2019-10-30
feat: use a yaml template to create resources for Vault through CloudFormation and use a provided user

### LINKED ISSUE #5434: Modify the EKS Vault installation to avoid creating an IAM User
### Summary 
Right now, we are attempting to create Vault with a CloudFormation stack.

This stack describes every necessary resource for Vault to work.

We are assuming that every user will be allowed to create IAM Users, which is not always the case. In fact, we ourselves can't do this in our AWS accounts as we have a restriction on user creation.

We should find an alternative to the User creation or add a way to provide an existing User with enough privileges to complete the stack creation.

The error is: `API: iam:CreateUser User: arn:aws:iam::<snip>:user/<snip> is not authorized to perform: iam:CreateUser on resource: arn:aws:iam::<snip>:user/vault_<region>-18f431c6` 

Attempt to create a cluster with Vault enabled.

### LINKED ISSUE #5435: Vault needs env variable AWS_REGION set to attempt too create the CloudFormation stack
### Summary 
The Vault stack creation is failing due to `Missing Region` when we attempt to launch it with `aws-sdk-go` in: https://github.com/jenkins-x/jx/blob/ad6b52fb30230b95917aff9c1faeb13db63abd94/pkg/cmd/create/create_vault.go#L351-L355

Ideally, we'd either pass this env variable together with the env variables that we define for Boot or we pass the region when creating the `awsSession` in `aws-sdk-go`.

The `autoCreate` flag also needs to be set to true to make it create any necessary resources that we are not passing through with boot:

https://github.com/jenkins-x/jx/blob/ad6b52fb30230b95917aff9c1faeb13db63abd94/pkg/cmd/create/create_vault.go#L340

dgozalo:
feat: we can't use IRSA yet so let's use IAM Users

feat: let users provide an user to create Vault resources

#### Submitter checklist

#### Description
Move the existing programmatic way to create Vault resources to a CloudFormation template located in the boot-config repository while maintaining the old way for non-boot clusters.

Also take an existing IAMUser to create secret keys and use them to authenticate the Vault pod.

Also added a new struct to `VaultConfig` in `jx-requirements.yml` so existing AWS resources can be added.

This is lacking some tests but we didn't really have any for Vault,  I'll try to get some but it's a pretty hard part to test in isolation so I think this should be tested by a BDD test.

fixes #5434
fixes #5435


pkg/cloud/amazon/vault/vault_resources.go
```diff
@@ -370,29 +449,68 @@ func runCloudformationTemplate(templateBody *string, stackName *string) error {
...
+				vsp.S3BucketARN = v.OutputValue
+			case "AWSKMSKey":
+				vsp.KMSKeyARN = v.OutputValue
+			case "AWSDynamoDBTable":
+				vsp.DynamoDBTableARN = v.OutputValue
+			default:
+				log.Logger().Warnf("CloudFormation parameter %s not expected", *v.OutputKey)
+			}
+		}
+	}
```
ccojocar: I would be nice to validate at this point that none of the expected `vsp` fields is empty. For instance, an output value gets missed or is empty.

pkg/cloud/amazon/vault/vault_resources.go
```diff
@@ -405,3 +523,54 @@ func createAccessKey(region string, username string) (*string, *string, error) {
...
+	case cloudformation.ChangeSetStatusCreateInProgress, cloudformation.ResourceStatusDeleteInProgress, cloudformation.ResourceStatusUpdateInProgress:
+		return util.ColorWarning(status)
+	default:
+		return util.ColorAnswer(status)
+	}
+}
+
+func runStackEventsReporting(donec <-chan bool, svc *cloudformation.CloudFormation, stackName *string) {
+	lastLen := 0
+	fmt.Println()
```
ccojocar: I think this line needs to be removed.

pkg/cloud/amazon/vault/vault_resources.go
```diff
@@ -405,3 +523,54 @@ func createAccessKey(region string, username string) (*string, *string, error) {
...
+			output, _ := svc.DescribeStackEvents(&cloudformation.DescribeStackEventsInput{
+				StackName: stackName,
+			})
+
+			for i, j := 0, len(output.StackEvents)-1; i < j; i, j = i+1, j-1 {
+				output.StackEvents[i], output.StackEvents[j] = output.StackEvents[j], output.StackEvents[i]
+			}
+
+			if lastLen > 0 {
+				fmt.Printf("\033[%dA", lastLen)
```
ccojocar: Maybe consider using a logger instead of printing directly to stdout.

pkg/cloud/amazon/vault/vault_resources.go
```diff
@@ -405,3 +523,54 @@ func createAccessKey(region string, username string) (*string, *string, error) {
...
+			}
+			lastLen = len(output.StackEvents)
+			for _, v := range output.StackEvents {
+				str := fmt.Sprintf("\r\033[K  %-25s | %-60s | %s", trimString(*v.ResourceType, 25),
+					trimString(*v.PhysicalResourceId, 60), trimString(*v.ResourceStatus, 25))
+				fmt.Println(defineStatus(str, v))
```
ccojocar: ditto

pkg/cloud/amazon/vault/vault_resources.go
```diff
@@ -370,29 +449,68 @@ func runCloudformationTemplate(templateBody *string, stackName *string) error {
...
 	_, err := svc.CreateStack(input)
 	if err != nil {
-		return errors.Wrapf(err, "Unable to create vault preresiquite resources")
+		return errors.Wrapf(err, "Unable to create vault prerequisite resources")
 	}
-
+	donec := make(chan bool)
+	go runStackEventsReporting(donec, svc, stackName)
 	// Wait until stack is created
 	err = svc.WaitUntilStackCreateComplete(desInput)
```
ccojocar: Does this wait has any timeout in case the stack creation gets stuck? It would be nice to bail out after a while.

pkg/cmd/create/create_vault.go
```diff
@@ -362,34 +370,64 @@ func (o *CreateVaultOptions) createVaultAWS(vaultOperatorClient versioned.Interf
...
-			TableName: autoCreateTableName,
-		})
+		bucketName := o.BucketName
+		if bucketName == "" {
+			bucketName = "vault-unseal." + o.S3Region + "." + domain
+		}
+
+		valueUUID, err := uuid.NewV4()
+		if err != nil {
+			return errors.Wrapf(err, "Generating UUID failed")
```
ccojocar: capitalised error

pkg/cmd/create/install.go
```diff
@@ -2308,10 +2308,10 @@ func (options *InstallOptions) createSystemVault(client kubernetes.Interface, na
 			CreateOptions: CreateOptions{
 				CommonOptions: options.CommonOptions,
 			},
-			IngressConfig:       *ic,
```
ccojocar: Why is ingress config and namespace removed from the arguments? I think in the installed was just used to expose vault with the ingress configuration.

pkg/cmd/step/boot/step_boot_vault.go
```diff
@@ -132,8 +136,43 @@ func (o *StepBootVaultOptions) Run() error {
...
+			DynamoDBRegion:      awsConfig.DynamoDBRegion,
+			AccessKeyID:         awsConfig.AccessKeyID,
+			SecretAccessKey:     awsConfig.SecretAccessKey,
+			ProvidedIAMUsername: awsConfig.ProvidedIAMUsername,
+		}
+		cvo.AWSTemplatesDir = filepath.Join(o.Dir, o.ProviderValuesDir, cloud.EKS, "templates")
+		defer func() {
+			// we don't want to save secret keys so we have to wipe them before saving the file
+			log.Logger().Info("Deleting AccessKeyID and SecretAccessKey from the jx-requirements.yml file")
+			requirements.Vault.AWSConfig.AccessKeyID = ""
```
ccojocar: This is a bit dangerous. Are the requirements saved on disk with the access key until this function gets executed to clear them up?

You can also ignore the fields from serialisation using the tag:
```
json:"-"`
```

pkg/config/install_requirements.go
```diff
@@ -293,16 +293,39 @@ type ClusterConfig struct {
...
+}
+
+// VaultAWSConfig contains all the Vault configuration needed by Vault to be deployed in AWS
+type VaultAWSConfig struct {
+	VaultAWSUnsealConfig
+	AutoCreate          bool   `json:"autoCreate,omitempty"`
+	DynamoDBTable       string `json:"dynamoDBTable,omitempty"`
+	DynamoDBRegion      string `json:"dynamoDBRegion,omitempty"`
+	AccessKeyID         string `json:"accessKeyId,omitempty"`
+	SecretAccessKey     string `json:"secretAccessKey,omitempty"`
```
ccojocar: This and the above filed should be ignored from sterilisation with tag
```
json:"-"`
```

pkg/config/install_requirements.go
```diff
@@ -293,16 +293,39 @@ type ClusterConfig struct {
...
 	Bucket         string `json:"bucket,omitempty"`
 	Keyring        string `json:"keyring,omitempty"`
 	Key            string `json:"key,omitempty"`
 	ServiceAccount string `json:"serviceAccount,omitempty"`
 	RecreateBucket bool   `json:"recreateBucket,omitempty"`
+	AWSIAMUserARN  string `json:"awsIAMUserARN,omitempty"`
 	// Optionally allow us to override the default lookup of the Vault URL, could be incluster service or external ingress
-	DisableURLDiscovery bool `json:"disableURLDiscovery,omitempty"`
+	DisableURLDiscovery bool           `json:"disableURLDiscovery,omitempty"`
+	AWSConfig           VaultAWSConfig `json:",omitempty"`
```
ccojocar: This config seems to be inconsistent and confusing. We have `Bucket` in the top structure but also `AWSConfig.S3Bucket` in a substructure. Also the `AWSIAMUserARN` should be rather moved into `AWSConfig` section.

We should move the GCP related config into something like GCPConfig, otherwise the users will be confused.

dgozalo: You are right, AWSIAMUserARN should be removed because it's already in AWSConfig, I forgot to remove it.

Shouldn't the change of GCP configuration be done in a different PR? The problem now when working with some provider is that you also have to modify a bunch of things for the other provider and it takes ages to close something.

ccojocar: We should streamline the configuration for each provider since we are still in the early stage. Otherwise we introduce more stuff which it will be quite difficult to deprecate later.

pkg/cmd/step/boot/step_boot_vault.go
dgozalo: They are never saved on disk at this point, that's why I want to run this and save the file after they've been used.
This is just a way to let users provide the access keys instead of an user.

pkg/cloud/amazon/vault/vault_resources.go
dgozalo: What I'm doing with this is rewriting the console at certain points, so I think it wouldn't work with a logger.

pkg/cloud/amazon/vault/vault_resources.go
dgozalo: I'm sure the CloudFormation stack would fail if it can't produce the expected outputs, because it would mean that the creation of the resource that it's getting the reference from would've failed.

pkg/cloud/amazon/vault/vault_resources.go
dgozalo: The default is 120 attempts with a 30 seconds of delay, this results in a total timeout of 1h.

Both attempts and delay can be modified. What would you think it's a good time to wait?

pkg/cloud/amazon/vault/vault_resources.go
```diff
@@ -65,45 +88,101 @@ func CreateVaultResources(vaultParams ResourceCreationOpts) (*string, *string, *
...
+	}
+	templateBytes, err := ioutil.ReadFile(templatePath)
+	if err != nil {
+		return nil, nil, nil, nil, nil, err
+	}
+	stackName := stackNamePrefix + vaultParams.UniqueSuffix
+	err = runCloudFormationTemplate(aws.String(string(templateBytes)), aws.String(stackName), []*cloudformation.Parameter{
+		{
+			ParameterKey:   aws.String(resourceSuffixParamName),
+			ParameterValue: aws.String(vaultParams.UniqueSuffix),
```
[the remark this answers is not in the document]
dgozalo: If it gets to this point either:

a) There's no vault installed previously
b) No existing resources have been provided

The suffix is indeed creating a new set of independent resources and existing resources can be provided so we don't have to create new ones.

pkg/cmd/create/install.go
dgozalo: I removed them because there were basically not being used at all.

pkg/cloud/amazon/vault/vault_resources.go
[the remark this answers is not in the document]
ccojocar: Thanks for clarifications. 

What will happen in the following scenario:
Fist run of boot:
- resources are created for a cluster 
- vault is created with these resources

vault gets somehow deleted. The boot is run second time with the same cluster name. Does the stack creation is executed?

I believe that we save the resource names in the requirements file and they get automatically reused by the second run, and the CloudFormation stack is not applied again. 

It might be that one of these resources doesn't exist anymore in AWS (e.g. was deleted or something happed). I think, currently we don't validate this by re-running the stack. This is why I was asking if is idempotent. 

We can handle this in a follow up. Maybe we can open an issue for it.

pkg/cmd/create/install.go
ccojocar: It makes sense then. Thanks!

pkg/cmd/step/boot/step_boot_vault.go
ccojocar: I think this might be dangerous because if the jx-requirements file gets somehow saved in git, these credentials might get leaked. 

Why not to use the env variables for them? They don't really belong to the requirements since they should never be saved, and are somehow ephemeral.  

I would strongly suggest to remove them and to use env variables instead, and maybe command line arguments but I think the standard way is to provide them in env variables.

PR #6021, opened 2019-11-04
fix: ask user to validate discovered app name

### LINKED ISSUE #5883: running jx promote in a brand new install without specifying app name breaks environment builds
### Summary 

running jx promote in a brand new install without specifying app name or dir breaks environment builds. This provides a bad first time experience as you end up with you install in a state that needs manual intervention to recover. Once any application has been deployed to an environment, you don't hit this problem (and get a valid error instead). 

1. Create a brand new cluster with NO APPS IN IT
2. run a quick start - check it is running in staging
3. run `jx promote --env production` outside of the application dir (do NOT run it in the dir, or it will work). 

### Expected behavior

Should return an error that it doesn't know what application you are promoting (if you run it in an app dir it will be fine, has to be outside), and no other harm done. 

### Actual behavior

Will result in a failed PR to the production environment and you wont' be able to promote to it. 

Error detail here: https://gist.github.com/mic-jx-bot/afc9b51414dc68fa71abaec2f8c46703

No other apps can be promoted and it appears your production environment never works.

If you try to reproduce this once you successfully have an application in production, you will not hit this problem, this is strictly first time problem. 

### Jx version

```
jx version
NAME               VERSION
jx                 2.0.721
Kubernetes cluster v1.13.10-gke.0
kubectl            v1.16.2
git                2.20.1
```

### Jenkins type

- [ ] Classic Jenkins

### Kubernetes cluster

GKE (created via jx create cluster)

Macos

mauromorales:
This PR introduces the function `EnsureApplicationNameIsDefined()` which is used to make sure that the application name is set when running the `jx promote` command. This functionality was extracted from the `Run()` function and additionally added a check for the user to confirm whether the name of the app is correct when we guessed it.

Some things to consider:

- I decided to export it, to make it easier to test without having to do a lot of test preparation.
- Because the process to figure out the app name is tightly coupled to to other function in the promote and helm packages I decided to pass those functions as parameters so it's easier to provide fake functions in the tests. I think this should be ok because the functionality being tested is not the one from figuring out the application name itself.

fixes #5883

### DIFF
--- a/pkg/cmd/opts/helm.go
+++ b/pkg/cmd/opts/helm.go
@@ -523,8 +523,6 @@ func (o *CommonOptions) DiscoverAppName() (string, error) {
 	}
 	answer = gitInfo.Name
 
-	if answer == "" {
-	}
 	return answer, nil
 }
 
--- a/pkg/cmd/promote/promote.go
+++ b/pkg/cmd/promote/promote.go
@@ -143,46 +143,115 @@ func NewCmdPromote(commonOpts *opts.CommonOptions) *cobra.Command {
 	return cmd
 }
 
-func (options *PromoteOptions) AddPromoteOptions(cmd *cobra.Command) {
-	cmd.Flags().StringVarP(&options.Application, opts.OptionApplication, "a", "", "The Application to promote")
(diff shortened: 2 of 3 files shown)


abayer:
/hold

Holding for a bit to deal with batching.

PR #6027, opened 2019-11-04
fix: only use the base file path when loading the auth config from vault

ccojocar:
#### Submitter checklist

fixes #


abayer:
/hold

Holding for a bit to deal with batching.

Issue #6031, opened 2019-11-05
jx commands should detect git/credentials files created by tekton


jstrachan:
the tekton approach to secrets generates a `git/credentials` file from the available secrets on startup of a pipeline: https://github.com/tektoncd/pipeline/blob/master/docs/auth.md#git-basic-auth

it would be great to have `jx` respect the same git credentials file by default for user/pwd values. Then we can avoid explicitly looking at the `Secrets` and instead just reuse the `ServiceAccount => Secret` bindings we reuse in tekton pipelines.

So lets also detect `git/credentials` files and enrich whatever is found at `~/.jx/gitAuth.yaml` (which will be empty inside pipelines)

ccojocar:
This has more implications:
- This Tekton's credentials are set by us in boot, see https://github.com/jenkins-x/jenkins-x-boot-config/blob/7d088ae938eae89fb976d763e1fe47d65038a8f6/env/tekton/values.tmpl.yaml#L14
- They are the same like the pipeline user/token 
- Also they are duplicated in a lot of places see #5514 
- We are trying to improve this and only use the secrets to lookup this credentials when vault is not enabled, otherwise they should be fetched by jx from vault, which will allow users to easily rotated them in vault

It's true that we can read them from the git/credentials file when `jx` is executed in the tekton created Pod, but hat only works with Pods created by Tekton. In case, jx is executed in other pods (e.g. job, build controller etc), this credentials file is not available. It would be rather cleaner to have a uniform approach which fetches them directly from vault when is enabled.

PR #6032, opened 2019-11-05
fix: lets detect git/credentials provided by tekton

jstrachan:
fixes https://github.com/jenkins-x/jx/issues/6031


pkg/auth/gitcreds/gitcredentials_auth.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/jenkins-x/jx/pkg/auth"
+	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/util"
+	"github.com/pkg/errors"
+)
+
+// LoadGitCredentialsAuth loads the git credentials from the `git/credentials` file
+// in `$XDG_CONFIG_HOME/git/credentials` or in the `~/git/credentials` directory
+func LoadGitCredentialsAuth() (*auth.AuthConfig, error) {
```
ccojocar: @jstrachan Have you seen https://github.com/jenkins-x/jx/pull/6034/files? We should abstract this behind a `ConfigHandler` interface which is used by the `ConfigService` to load the config.

ccojocar:
@jstrachan I am wondering what's the advantage to use this approach since we configured the git credentials in boot https://github.com/jenkins-x/jenkins-x-boot-config/blob/7d088ae938eae89fb976d763e1fe47d65038a8f6/env/tekton/values.tmpl.yaml#L14 and are the same with pipeline user/token from vault.

ccojocar:
Also we are working to an overall improvement for sects management. See https://github.com/cloudbees/jenkins-x-security/issues/12. 

The list of issues assigned to this topic https://codetree.com/projects/StX1/issues/10427820?type=epic. Especially, #6036 and  #5114.

jstrachan:
@ccojocar when using github apps there is not a single user + token for github.com; but one per owner (organisation/user). So we need to be able to generate the git credentials dynamically based on the owner of a github repository.

It's easy to patch tekton to do this; then the generated `git/credentials` file is specific to a github owner. 

However unless we have this patch to be able to load `git/credentials` we've other easy way to align the credentials that tekton creates to the ones used by `jx` steps if using github apps.

Also since all steps in a tekton pipeline have access to the generated git credentials; we could consider removing all the custom logic in `jx` to load `Secrets` if inside a pipeline; and rely on getting secrets from the local `git/credentials` - whether running locally or inside a tekton pipeline; further aligning and simplifying things

ccojocar:
Thanks for clarifications!

>However unless we have this patch to be able to load git/credentials we've no easy way to align the credentials that tekton creates to the ones used by jx steps if using github apps (which loads them from ~/.jx/gitAuth.yaml or Secrets directly).

This is not always the case. The gitAuth.yaml is stored in `vault` when vault is enabled. I am wondering why we are not storing those GitHub app generated credentials in vault.

ccojocar:
I am working on moving all these auth config handlers behind the `AuthConfigService`. It would be nice at least to implement the `ConfigHandler` interface, if you want to load them form that git credentials file. Just have a look how is done in `file_config_handler.go`, and all other `*_handler.go` sources.  Then in the factory, we go over all these handler and load the credentials either based on the configured sources, or lookup them in different places. 

This is to avoid having commands which loads directly the pipeline secrets from kubernetes secrets as of today. The plan is to have this behind this service abstraction.

jstrachan:
@ccojocar yeah, should be able to just merge this logic into the `file_config_handler.go`.

If this approach works (using the `git/credentials` from tekton) we could look at disabling the logic of looking at the secrets at all if folks are using tekton.

ccojocar:
It would be great if you could merge them in the file handler.

jstrachan:
@ccojocar actually on reviewing the change - we'll need this logic for file, configmap and secret based auth implementations; so the position of this change looks correct to me

ccojocar:
We partially use this logic to create the auth config service (in some commands):https://github.com/jenkins-x/jx/blob/4a89356b28368bb9724d55fbf44c1c9882f66d2f/pkg/cmd/clients/factory.go#L368, while other commands loads the auth directly from kubernetes secrets see `LoadPipelineSecrets` - which is not quite right and leads to quite some issues (e.g. when vault is enabled). 

The plan is to move everything under this factory method (abstraction) and based on secrets location config and context, we will create the git auth config service in one place. 

It think you can modify the file handler such that:
- first try to load the gitAuth.yaml file form ~/.jx
- second tries to parse the credentials form `.git/credentials` file

jstrachan:
@ccojocar thanks Cosmin, I'll do that.

BTW on further testing of the use of GitHub App tokens and tekton; it turns out that loading git/credentials is great for working with git & promotes consistency with tekton and avoids us needing `jx step git credentials`) - we do still need to load tokens from Secrets. e.g. `jx promote` can often end up using a different git owner and token than the application we are releasing/promoting. So I don't think we can zap the logic to load secrets from Secrets (or vault) any time soon; but we can hopefully make `jx step git credentials` become a no-op when using tekton.

jstrachan:
@ccojocar I moved the git credentials loading inside the file config handler: https://github.com/jenkins-x/jx/blob/a550d97bd1695103022ccd0c360a9af56216159e/pkg/auth/file_config_handler.go#L59

PR #6047, opened 2019-11-06
fix: extract author info from commit

### LINKED ISSUE #5726: GitHubProvider#asGitHubCommit should extract author name and email from commit

mauromorales:
This PR adds unexported function `extractRepositoryCommitAuthor()` which will extract the author information from a GitHub commit. It will give preference to the commit's author above the github user when available.

fixes #5726

### DIFF
--- a/pkg/gits/github.go
+++ b/pkg/gits/github.go
@@ -791,6 +791,23 @@ func (p *GitHubProvider) ListOpenPullRequests(owner string, repo string) ([]*Git
 	return answer, nil
 }
 
+func extractRepositoryCommitAuthor(rc *github.RepositoryCommit) (gu *GitUser) {
+	gu = &GitUser{}
+
+	if rc.Commit.Author != nil {
+		gu.Email = rc.Commit.Author.GetEmail()
+		gu.Name = rc.Commit.Author.GetName()
+
+		if rc.Author != nil {
+			gu.Login = rc.Author.GetLogin()
+			gu.URL = rc.Author.GetURL()
+			gu.AvatarURL = rc.Author.GetAvatarURL()
+		}
+	}
+
+	return
+}
+
 func (p *GitHubProvider) asGitHubCommit(commit *github.RepositoryCommit) GitCommit {
 	message := ""
 	if commit.Commit != nil {
@@ -798,18 +815,8 @@ func (p *GitHubProvider) asGitHubCommit(commit *github.RepositoryCommit) GitComm
 	} else {
 		log.Logger().Warnf("No Commit object for for commit: %s", commit.GetSHA())
 	}
-	var author *GitUser
-	if commit.Author != nil {
-		author = &GitUser{
-			Login:     commit.Author.GetLogin(),
-			Email:     commit.Author.GetEmail(),
-			Name:      commit.Author.GetName(),
-			URL:       commit.Author.GetURL(),
-			AvatarURL: commit.Author.GetAvatarURL(),
-		}
-	} else {
-		log.Logger().Warnf("No author for commit: %s", commit.GetSHA())
-	}
+	author := extractRepositoryCommitAuthor(commit)
+
 	return GitCommit{
 		Message: message,
 		URL:     commit.GetURL(),
@@ -831,26 +838,6 @@ func (p *GitHubProvider) GetPullRequestCommits(owner string, repository *GitRepo
 
 	for _, commit := range commits {
 		summary := p.asGitHubCommit(commit)
-		if commit.Author != nil {
-			if summary.Author.Email == "" {
-				log.Logger().Infof("Commit author email is empty for: %s", commit.GetSHA())
-				dir, err := os.Getwd()
-				if err != nil {
-					return answer, err
-				}
-				gitDir, _, err := p.Git.FindGitConfigDir(dir)
-				if err != nil {
-					return answer, err
-				}
-				log.Logger().Infof("Looking for commits in: %s", gitDir)
-				email, err := p.Git.GetAuthorEmailForCommit(gitDir, commit.GetSHA())
-				if err != nil {
-					log.Logger().Warnf("Commit not found: %s", commit.GetSHA())
-					continue
-				}
-				summary.Author.Email = email
-			}
-		}
 		answer = append(answer, &summary)
 	}
 	return answer, nil
--- a/pkg/gits/github_test.go
+++ b/pkg/gits/github_test.go
@@ -3,6 +3,8 @@ package gits
 import (
 	"testing"
 
+	"github.com/google/go-github/github"
+
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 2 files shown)


hferentschik:
@mauromorales, great that you picked this up ;-) I am at a conference atm, but will have a closer look asap.

pkg/gits/github.go
```diff
@@ -791,25 +791,51 @@ func (p *GitHubProvider) ListOpenPullRequests(owner string, repo string) ([]*Git
...
+}
+
 func (p *GitHubProvider) asGitHubCommit(commit *github.RepositoryCommit) GitCommit {
 	message := ""
 	if commit.Commit != nil {
 		message = util.DereferenceString(commit.Commit.Message)
 	} else {
 		log.Logger().Warnf("No Commit object for for commit: %s", commit.GetSHA())
 	}
-	var author *GitUser
```
hferentschik: 👍 for extracting this into a function.

hferentschik: BTW, there was another pull request for this just a short while back where I left this [comment](https://github.com/jenkins-x/jx/pull/5931#discussion_r338925448)

pkg/gits/github.go
```diff
@@ -791,25 +791,51 @@ func (p *GitHubProvider) ListOpenPullRequests(owner string, repo string) ([]*Git
 	return answer, nil
 }
 
+func extractRepositoryCommitAuthor(rc *github.RepositoryCommit) (gu *GitUser) {
+	commitAuthorPresent := rc.Commit.Author != nil
```
hferentschik: For what it's worth, technically I think there will always be a `rc.Commit.Author`, since git does not let you commit anything without author and email set. So in my world, we can just create the GitUser with this name and email and then just amend any additional information like URL and login in case the `rc.Author` is not nil which basically means that GitHub was able match the commit author against a GitHub user (which might not need to be the case). Does that make sense?

pkg/gits/github.go
```diff
@@ -791,25 +791,51 @@ func (p *GitHubProvider) ListOpenPullRequests(owner string, repo string) ([]*Git
...
+	repositoryAuthorPresent := rc.Author != nil
+	repositoryAuthorOnly := !commitAuthorPresent && repositoryAuthorPresent
+	commitAndRepositoryAuthorMatch := commitAuthorPresent && repositoryAuthorPresent && rc.Commit.Author.GetLogin() == rc.Author.GetLogin()
```
hferentschik: Not sure if we can rely on `rc.Commit.Author.GetLogin` in this case. Looking at the code of the github lib, there is:

```
	// The following fields are only populated by Webhook events.
	Login *string `json:"username,omitempty"` // Renamed for go-github consistency.
```

I think login will in our case never be set. 

Best to run some example requests using for example `curl` to look at the actual JSON returned.

pkg/gits/github.go
mauromorales: yup I think it's clear, let me do the changes and I'll request your feedback again

pkg/gits/github.go
mauromorales: @hferentschik yes, you are right login at the commit author level is unreliable. I've switched to use email.

mauromorales:
Thanks for the feedback @hferentschik, this simplifies the logic a lot.

pkg/gits/github.go
```diff
@@ -791,25 +791,33 @@ func (p *GitHubProvider) ListOpenPullRequests(owner string, repo string) ([]*Git
 	return answer, nil
 }
 
+func extractRepositoryCommitAuthor(rc *github.RepositoryCommit) (gu *GitUser) {
```
hferentschik: 👍

mauromorales:
@hferentschik I'm not sure I follow regarding the risks that brings the removal of the code, could you give me some ideas, I'll be happy to implement them

pkg/gits/github.go
```diff
@@ -791,25 +791,33 @@ func (p *GitHubProvider) ListOpenPullRequests(owner string, repo string) ([]*Git
...
+	gu = &GitUser{}
+
+	if rc.Commit.Author != nil {
+		gu.Email = rc.Commit.Author.GetEmail()
+		gu.Name = rc.Commit.Author.GetName()
+
+		repositoryAuthorPresent := rc.Author != nil
+		if repositoryAuthorPresent && gu.Email == rc.Author.GetEmail() {
```
hferentschik: On second thought. I think this is not quite correct since the origin of this issue was to begin with that `rc.Author.GetEmail()` is quite often empty. AFAIU it will depend on whether you have a public email as part of your profile or not. Many choose not to expose a public email which was leading to the fact that in we went to the file system and the actual git repo to get the email. See `GetPullRequestCommits`. 

With your change, we are in a better position now, since we will always get name and email. But I think it will also lead to the fact that we won't populate these additional fields even though we could/should.

I am wondering whether it would not just be better to just add login, URL and avatar to the `GitUser` provided  `rc.Author != nil`. We basically rely on GitHub to give us a matching Github when possible.  If GitHub cannot match the email it will return a nil Author (the second commit of `curl --user "token:$GH_TOKEN" https://api.github.com/repos/jenkins-x/jx/pulls/5939/commits`, note also that my committer profile does not contain an email, since I don't have a public exposed email).

hferentschik:
>  I'm not sure I follow regarding the risks that brings the removal of the code, could you give me some ideas, I'll be happy to implement them

You are basically removing the fallback. The reason I noticed this issue was that I kept seeing "Commit author email is empty for: ..." in the build logs. So as it stands now, we quite often get into this code branch. I think what happened is that due to the fact `asGitHubCommit` would return empty email addresses, this code was added to remedy it, instead of trying to find the underlying issue. We think your code fixes the problem, but there is a chance we are missing something. So if you delete this if branch in `GetPullRequestCommits` there is a risk that something breaks there. I am pretty sure it won't, but I have not tested it. That's all I'm saying.

In terms of logic, it makes no sense at all to me that in our `GitHubProvider` implementation when we want to retrieve commit information from GitHub assume that the current working directory from which this code is called is a git repository and we try to match commit information from there. That's wrong on so many levels.

pkg/gits/github.go
mauromorales: @hferentschik i see what you are saying but I think this could be a problem because at least from the example in that address, we would be adding as email: foo@bar.com and as login and avatar, invalid-email-address and some url respectively, which don't really belong together. What is this information used for later on? because I could see a scenario where login "invalid-email-address" would have lot's of commits related to it, even if they are actually different users.

mauromorales:
@hferentschik I'm not entirely sure if we will mitigate any risk. Correct me if I'm wrong but I understand that the information in the commit's author is exactly the same as in the local file. So even in the case where data would be missing from the commit, it would be missing in both, places, wouldn't it?

pkg/gits/github.go
hferentschik: Fair point. I think foo@bar.com is a bit of edge case and I am not sure whether there is actually somehow a "user" with this email. If you really choose random non-existent emails you indeed get empty authors.

I think the most important thing we are after is the name as email as per commit. After that, we need to decide on which side we prefer to error. In your case, we will know that if the `rc.Author.GetEmail` exists and matches, the data will be correct. On the other hand, for all the cases where people decide now to make their email public in their GitHub profile, we won't enhance the `GitUser` at all, even though the data would be there.

In what I am suggesting, we will always enhance the GitUser provided GitHub can make a match, but we might have edge cases like foor@bar.com where you get "invalid" or at the very least questionable data. 

My reasoning is that there will be a proper match in almost all cases. To end up with these border cases someone would have to intentionally use these email addresses. Why would you do that?

Also look at:

```bash
curl --user "token:$GH_TOKEN" https://api.github.com/search/users?q=foo@bar.com
```

I think that there are literally users in the GitHub database using this email address and hence you get this strange matches.

If you use something like:
curl --user "token:$GH_TOKEN" https://api.github.com/search/users?q=john.doe@acme.com
```bash

you get an empty result.

Comment (mauromorales) [reply pkg/gits/github.go]: @hferentschik sounds good to me, I've applied the changes

File: pkg/gits/github.go [hunk 6]
```diff
@@ -831,26 +838,6 @@ func (p *GitHubProvider) GetPullRequestCommits(owner string, repository *GitRepo
 
 	for _, commit := range commits {
 		summary := p.asGitHubCommit(commit)
-		if commit.Author != nil {
```
hferentschik: 👍 :-)

PR #6054, opened 2019-11-07
fix: More consistent use of `gitRef` rather than `o.GitRef` in boot

abayer:
#### Submitter checklist

#### Description

We just hit a bug which resulted in `o.GitRef` being unset, and so everything failed to check out etc. We should be passing `gitRef` around, not `o.GitRef`, since `o.GitRef` has legitimate cases for being unset.

n/a


abayer:
/assign @garethjevans 

Since you added this logic last week. =) I'm assuming there wasn't a specific reason for using `o.GitRef` when the existing code used `gitRef`, but just to make sure...

PR #6062, opened 2019-11-08
fix: Don't include pipeline runs that failed/were retried by Prow in logs

### LINKED ISSUE #6061: "jx get build logs" can include failed-but-auto-retried-by-Prow pipelines
There are a couple race conditions in Tekton that can result in a `PipelineRun` failing for no good reason. We've tweaked Prow to automatically retry in those scenarios, but you can still get errors running `jx get build logs` in batch mode with a report that there are multiple builds matching the filter. We should just exclude the fail-right-away-and-retry cases from `jx get build logs`.

abayer:
#### Submitter checklist

#### Description

The immediate impetus here is that BDD tests can fail in this scenario, because more than one pipeline is found for the filter - the one that failed, and the one that was retried. But this should just clean things up a bit anyway.

~~I'll come back and add tests for this later today.~~
EDIT: Test added.

/assign @dgozalo 

fixes #6061

### DIFF
--- a/pkg/logs/tekton_logging.go
+++ b/pkg/logs/tekton_logging.go
@@ -103,6 +103,13 @@ func (t TektonLogger) GetTektonPipelinesWithActivePipelineActivity(filters []str
 	prMap := make(map[string][]*tektonapis.PipelineRun)
 	for _, pr := range tektonPRs.Items {
 		p := pr
+		prStatus := p.Status.GetCondition(knativeapis.ConditionSucceeded)
+		// Don't include any pipeline runs that failed due to Tekton race conditions and were auto-restarted by Prow
+		if prStatus != nil && prStatus.IsFalse() &&
+			(strings.Contains(prStatus.Message, "can't be found:pipeline.tekton.dev") ||
+				strings.Contains(prStatus.Message, "it contains Tasks that don't exist")) {
+			continue
+		}
 		prBuildNumber := p.Labels[v1.LabelBuild]
 		if prBuildNumber == "" {
 			prBuildNumber = findLegacyPipelineRunBuildNumber(&p)
--- a/pkg/logs/tekton_logging_test.go
+++ b/pkg/logs/tekton_logging_test.go
@@ -10,7 +10,6 @@ import (
 	"testing"
 
 	"github.com/acarl005/stripansi"
-
 	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/client/clientset/versioned"
 	jxfake "github.com/jenkins-x/jx/pkg/client/clientset/versioned/fake"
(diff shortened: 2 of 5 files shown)


pkg/logs/tekton_logging_test.go
```diff
@@ -803,6 +805,133 @@ func TestGetRunningBuildLogsMetapipelineAndPendingGenerated(t *testing.T) {
...
+		Namespace:    ns,
+		LogWriter: &TestWriter{
+			StreamLinesLogged: make([]string, 0),
+			SingleLinesLogged: make([]string, 0),
+		},
+		LogsRetrieverFunc: LogsProvider,
+	}
+
+	_, err := jxClient.JenkinsV1().PipelineActivities(ns).Create(&v1.PipelineActivity{
+		ObjectMeta: v12.ObjectMeta{
```
hferentschik: I'd try to separate the test setup out a bit. It is hard to narrow down on the actual test(assertion).
The whole set up of the fake client could, for example, happen in a _setUp_ function and the resources itself could be extracted and appropriately named, e.g. `succeededPipelineRun`, `activePipelineRun`.

pkg/logs/tekton_logging.go
```diff
@@ -103,6 +103,13 @@ func (t TektonLogger) GetTektonPipelinesWithActivePipelineActivity(filters []str
 	prMap := make(map[string][]*tektonapis.PipelineRun)
 	for _, pr := range tektonPRs.Items {
 		p := pr
+		prStatus := p.Status.GetCondition(knativeapis.ConditionSucceeded)
+		// Don't include any pipeline runs that failed due to Tekton race conditions and were auto-restarted by Prow
+		if prStatus != nil && prStatus.IsFalse() &&
+			(strings.Contains(prStatus.Message, "can't be found:pipeline.tekton.dev") ||
```
hferentschik: That seems brittle. Depending on status message text/format.Is this the best we can do?

abayer: Sadly, yes.

pkg/logs/tekton_logging_test.go
abayer: Ok, will do tomorrow.

pkg/logs/tekton_logging.go
hferentschik: :-(

pkg/logs/tekton_logging_test.go
abayer: So the `PipelineRun`s still differ too much to go that route. I did tease out the `PA1` `PipelineActivity` creation, but the cases where we create runs rather than just load from the disk are a hassle - if I have a chance today, I'm going to tweak those to load from disk instead.

abayer: Ok, moved the runs to be loaded from disk. I left the activities as is.

PR #6063, opened 2019-11-08
feat: implement a generic mechanism to deprecate commands and marked a few commands as deprecated

ccojocar:
#### Submitter checklist

#### Description

This generates a message such this before the command execution:
```
$x upgrade ingress
Command "ingress" is deprecated, it will be removed on 01-02-2020. We now highly recommend you use boot instead. Please check https://jenkins-x.io/docs/getting-started/setup/boot/ for more details.


```

fixes #

### DIFF
--- a/pkg/cmd/cmd.go
+++ b/pkg/cmd/cmd.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"fmt"
 
+	"github.com/jenkins-x/jx/pkg/cmd/deprecation"
 	"github.com/jenkins-x/jx/pkg/cmd/profile"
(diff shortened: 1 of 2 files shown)


ccojocar:
> longer term I wonder if we should add the option to show/hide deprecated commands based on some setting (maybe a flag in a config file in ~/.jx somewhere?) so that jx help would show/ hide deprecated commands?

@jstrachan I agree with this. I had the same thoughts when implementing this, but I think is better in the beginning to keep this built into the binary to avoid any configuration confusions. I don't expect that the list of deprecated commands will change too often, or more flexibility is required in terms of configuration (e.g. jx oss vs cjxd vs other setup).

Issue #5671, opened 2019-10-01
jx delete application needs an environment selector


tdcox:
### Summary 
Currently, `jx delete application` removes the selected applications from all environments. Typically, it would be necessary to be able to remove applications selectively from some production or staging environments but leave them running in others so it would be helpful to both warn the user what is going to happen and to provide a filter to permit selected environments to be included/excluded.

Ideally, we would have a review and sign-off process around this activity as well.

### Jx version

The output of `jx version` is:

```
2.0.775
```

### Jenkins type

- [x] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

mauromorales:
Implementation will depend on resolution of #6069

PR #6069, opened 2019-11-11
fix: filter environments from delete application

### LINKED ISSUE #5668: jx delete application can delete environments!
### Summary 
Issuing `jx delete application` shows a default list of applications to delete that includes live environments as selectable items. This makes it far too easy to accidentally delete an environment when trying to remove an application.

### Jx version

```
2.0.775
```

### Jenkins type

- [ ] Classic Jenkins

mauromorales:
This PR removes any environments from the list of applications that can be deleted.

Before

```
$ jx delete application
? Pick Applications to remove from Prow:  [Use arrows to move, space to select, type to filter]
> [ ]  mauromorales/environment-mm-ghi-5668-dev
  [ ]  mauromorales/environment-mm-ghi-5668-production
  [ ]  mauromorales/environment-mm-ghi-5668-staging
  [ ]  mauromorales/jx-qs-test-delete-apps
```

After

```
$ jx delete application
? Pick Applications to remove from Prow:  [Use arrows to move, space to select, type to filter]
> [ ]  mauromorales/jx-qs-test-delete-apps
```

fixes #5668

### DIFF
--- a/pkg/cmd/deletecmd/delete_application.go
+++ b/pkg/cmd/deletecmd/delete_application.go
@@ -139,6 +139,28 @@ func (o *DeleteApplicationOptions) Run() error {
 	return nil
 }
 
+func matchesWithAnyEnvironmentURL(s string, environments map[string]*v1.Environment) bool {
+	for _, environment := range environments {
+		if strings.Contains(environment.Spec.Source.URL, s) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func removeEnvironments(jobs []string, environments map[string]*v1.Environment) []string {
+	var applications []string
+
+	for _, job := range jobs {
+		if !matchesWithAnyEnvironmentURL(job, environments) {
+			applications = append(applications, job)
+		}
+	}
+
+	return applications
+}
+
 func (o *DeleteApplicationOptions) deleteProwApplication(repoService jenkinsv1.SourceRepositoryInterface) (deletedApplications []string, err error) {
 	jxClient, _, err := o.JXClient()
 	if err != nil {
@@ -160,11 +182,12 @@ func (o *DeleteApplicationOptions) deleteProwApplication(repoService jenkinsv1.S
 		NS:           ns,
 		IgnoreBranch: true,
 	}
-	names, err := prowOptions.GetReleaseJobs()
+	jobs, err := prowOptions.GetReleaseJobs()
 	if err != nil {
 		return deletedApplications, fmt.Errorf("Failed to get ProwJobs")
 	}
 
+	names := removeEnvironments(jobs, envMap)
 	if len(names) == 0 {
 		return deletedApplications, fmt.Errorf("There are no Applications in Jenkins")
 	}
--- a/pkg/cmd/deletecmd/delete_application_test.go
+++ b/pkg/cmd/deletecmd/delete_application_test.go
@@ -1,12 +1,13 @@
-package deletecmd_test
+package deletecmd
 
 import (
 	"fmt"
 	"testing"
 
+	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+
 	gojenkins "github.com/jenkins-x/golang-jenkins"
 	clients_test "github.com/jenkins-x/jx/pkg/cmd/clients/mocks"
-	"github.com/jenkins-x/jx/pkg/cmd/deletecmd"
 	"github.com/jenkins-x/jx/pkg/cmd/opts"
 	"github.com/jenkins-x/jx/pkg/cmd/testhelpers"
 	"github.com/jenkins-x/jx/pkg/gits"
@@ -59,7 +60,7 @@ func TestDeleteApplicationInJenkins(t *testing.T) {
 	pegomock.When(jenkinsClient.GetJobs()).ThenReturn(pegomock.ReturnValue(jobs), pegomock.ReturnValue(nil))
 	pegomock.When(jenkinsClient.GetJob(pegomock.EqString(testRepoName))).ThenReturn(pegomock.ReturnValue(job), pegomock.ReturnValue(nil))
 
-	o := &deletecmd.DeleteApplicationOptions{
+	o := &DeleteApplicationOptions{
 		CommonOptions: &commonOpts,
 	}
 	o.Args = []string{testRepoName}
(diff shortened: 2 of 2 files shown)


hferentschik:
@mauromorales, would you mind pasting a before and after example of the output? Either on the pull request or on the issue.

pkg/cmd/deletecmd/delete_application.go
```diff
@@ -139,6 +139,23 @@ func (o *DeleteApplicationOptions) Run() error {
 	return nil
 }
 
+func removeEnvironments(jobs []string, environments map[string]*v1.Environment) (applications []string) {
```
hferentschik: I am not a friend of naked return values. If anything they should only be used in short functions (which this one arguably is). The problem is that functions can change and easily cross the threshold where naked returns become more of a liability than anything else. Also for consistency reasons, I'd prefer to stay clean of them. That's my 0.02$

pkg/cmd/deletecmd/delete_application.go
```diff
@@ -139,6 +139,23 @@ func (o *DeleteApplicationOptions) Run() error {
...
+	var keep bool
+	for _, job := range jobs {
+		keep = true
+		for _, environment := range environments {
+			if strings.Contains(environment.Spec.Source.URL, job) {
+				keep = false
```
hferentschik: Why the detour via the `keep`? If you invert the if statement you can do everything in one line, right?

hferentschik:
> ```
> $ jx delete application
> ? Pick Applications to remove from Prow:  [Use arrows to move, space to select, type to filter] 
> [ ]  mauromorales/environment-mm-ghi-5668-dev
> [ ]  mauromorales/environment-mm-ghi-5668-production
> [ ]  mauromorales/environment-mm-ghi-5668-staging
> [ ]  mauromorales/jx-qs-test-delete-apps
> ```

I guess what gets me here is why the env repos would be "apps". Is there no better way to get apps instead of going via `prowOptions.GetReleaseJobs()`?

pkg/cmd/deletecmd/delete_application.go
mauromorales: @hferentschik I don't think this would work because if it's inverted we'd add records every time we don't match the environment, and that would be the case for every existing environment, so we'd end up with a bunch of duplicates

pkg/cmd/deletecmd/delete_application.go
mauromorales: I've switched the function to explicitly returns the list of applications

pkg/cmd/deletecmd/delete_application.go
```diff
@@ -160,11 +179,12 @@ func (o *DeleteApplicationOptions) deleteProwApplication(repoService jenkinsv1.S
 		NS:           ns,
 		IgnoreBranch: true,
 	}
-	names, err := prowOptions.GetReleaseJobs()
+	jobs, err := prowOptions.GetReleaseJobs()
```
hferentschik: So maybe 'GetReleaseJobs' is really the best we have in the case of _applications_. Feels wrong though. IMO there should be a clear separation between applications/projects and environments. At some stage, we need to try unravelling this instead of monkey patching. 

If we go for filtering, we might want to consider whether it makes sense to push the filtering into `GetReleaseJobs` instead of just applying only in _delete_application_. This sounds like something which could be useful at other places as well. Just thinking out loud atm. @ccojocar WDYT?

pkg/cmd/deletecmd/delete_application.go
hferentschik: ahh, I see what you are saying.

hferentschik: One could split the two for loops, just iterating either the jobs or the envs and then making the "contains" check in a helper function. But that's just nitpick.

pkg/cmd/deletecmd/delete_application.go
hferentschik: Another disadvantage here is that this is tying us t Prow which is probably not really future proof (thinking Lighthouse here).

I am wondering whether there is some sort of label we could base a selection on which differs between a project/application created via jx [import|create] and say the env pipelines.

Or whether it is possible to first retrieve application namespaces (staging|production|...) and then find applications in this namespace. In this model the dev environment would not count as application namespace.

hferentschik: Also, how does this compare with `jx get application`? Does that have the same problem and if so, do we need to align between them?

mauromorales: @hferentschik good point. We seem to be getting a deployments list on `jx get application` which does return only the applications and not the environments.

https://github.com/jenkins-x/jx/blob/a2908f49f3b03b1374fc3e04ab57a93b1bee963c/pkg/cmd/get/get_applications.go#L281

What I wasn't able to figure out is how to get the repository owner out of the deployment. Do you know if there's a way?

Also, as you mention if we were to follow this approach then we'd only need to request the applications from a given environment to solve #5671

pkg/cmd/deletecmd/delete_application.go
mauromorales: @hferentschik I've extracted it into a function now ... but I guess this will depend on what we decide about using a different mechanism to get the applications

pkg/cmd/deletecmd/delete_application.go
```diff
@@ -139,6 +139,28 @@ func (o *DeleteApplicationOptions) Run() error {
 	return nil
 }
 
+func matchesWithAnyEnvironmentURL(s string, environments map[string]*v1.Environment) bool {
```
hferentschik: This looks good to me now in terms of using the "filtering" approach. Still trying to figure out whether there is no better way, aka aligning with `get application`.  Funny enough `GetDeployments` in `get application` takes a completely different approach to what we do here. `GetDeployments` goes via Kubernetes deployments while here we look at the Prow config. IMO this should definitely be aligned. Also, the code in `get applications` is explicitly skipping the dev environment which makes sense as well. So I am wondering whether we could not share somehow the code around `GetApplicationsOptions#getAppData`. Does this make sense?

pkg/cmd/deletecmd/delete_application.go
hferentschik: See https://github.com/jenkins-x/jx/pull/6069/files#r345255299

hferentschik: > What I wasn't able to figure out is how to get the repository owner out of the deployment. Do you know if there's a way?

Not sure I follow. What do you mean?

mauromorales: @hferentschik right now the generated output with the delete command is:

```
[ ]  mauromorales/jx-qs-test-delete-apps
```

however from `GetDeployments` I only seem to be able to get the name of the application`jx-qs-test-delete-apps` and I'm not aware how to relate it to the repo url/owner

pkg/cmd/deletecmd/delete_application.go
mauromorales: @hferentschik yup I know what you mean, I'd be ok using the approach mentioned in the other comment but we should have it clear how much backwards compatibility this should have. If it isn't necessary to show the owner for example then we could also just do the switch.

hferentschik: In the light of https://codetree.com/projects/StX1/board/10487910?type=epic and the recent discussion around applications, I think applications should be retrieved via iterating SourceRepository CRDs. Initially, we can do this without any modifications to the CRD itself, we can just apply the same filtering for env repos as we use here. Another benefit of using the CRDs is that you have the repo information. 

I leave it up to you. I guess we could merge this now and address everything else as part of the "refactor application commands" or we take a first initial step and use SourceRepository CRDs. WDYT?

mauromorales: @hferentschik because of all the changes and back and forth that might happen from the refactoring I'd prefer to just merge/close this PR and address in https://github.com/jenkins-x/jx/issues/6101

hferentschik: We took the refactoring of the application commands into the new sprint, if you want to continue there. The first step would be to let `get` and `delete` application use the same data source (SourceRepository).

Issue #6089, opened 2019-11-13
jx create cluster aws fails w/ 'parsing the current Kubernetes context'


ascheman:
### Summary

jx create cluster aws fails w/ 'parsing the current Kubernetes context'

### Steps to reproduce the behavior

`jx create cluster aws --region us-west-2` on a clean environment (only `~/.aws/credentials` provided)

### Expected behavior

No error

### Actual behavior

```bash
? Missing required dependencies, deselect to avoid auto installing: kops, kubectl, helm
Installing kops
Downloading https://github.com/kubernetes/kops/releases/download/1.14.1/kops-linux-amd64 to /home/vagrant/.jx_aws-191111-2229/bin/kops.tmp...
Downloaded /home/vagrant/.jx_aws-191111-2229/bin/kops.tmp
Installing kubectl
Downloading https://storage.googleapis.com/kubernetes-release/release/v1.13.2/bin/linux/amd64/kubectl to /home/vagrant/.jx_aws-191111-2229/bin/kubectl...
Downloaded /home/vagrant/.jx_aws-191111-2229/bin/kubectl
Installing helm
Downloading https://storage.googleapis.com/kubernetes-helm/helm-v2.16.0-linux-amd64.tar.gz to /home/vagrant/.jx_aws-191111-2229/bin/helm.tgz...
Downloaded /home/vagrant/.jx_aws-191111-2229/bin/helm.tgz
Installing helm secrets plugin
WARNING: Failed to retrieve team settings: failed to create the jx client: invalid configuration: no configuration has been provided - falling back to default settings...
? Pick Availability Zone:  us-west-2b
error: parsing the current Kubernetes context : unable to parse  as <cluster_name>.<region>.* or arn:aws:<region>:account-id:cluster/<cluster_name>
```

### Jx version

The output of `jx --version` is:

```
2.0.974
```

### Jenkins type

N/A

### Kubernetes cluster

N/A

### Operating system / Environment

Happens on 
* OSX (Darwin)
* Ubuntu (Linux)

EnziinSystem:
Me too.
```
jx create cluster aws
Command "aws" is deprecated, it will be removed on Feb 1 2020. 
? nodes 3
? Pick Availability Zone:  us-east-1a
error: parsing the current Kubernetes context : unable to parse  as <cluster_name>.<region>.* or arn:aws:<region>:account-id:cluster/<cluster_name
```

```
jx --version
2.0.1093
```
Ubuntu 19.10

PR #6090, opened 2019-11-13
Respect provided region for create cluster aws

ascheman:
#### Submitter checklist

#### Description

If a region is provided (`--region XXX`) there is no need to look it up
in a cluster configuration for AWS (in particular if a new cluster should
be created and there has never be a cluster configuration up to then).

Cf. https://kubernetes.slack.com/archives/C9MBGQJRH/p1573511916278100

fixes jenkins-x/jx#6089

### DIFF
--- a/pkg/cloud/amazon/ecr.go
+++ b/pkg/cloud/amazon/ecr.go
@@ -19,11 +19,14 @@ import (
 // GetAccountIDAndRegion returns the current account ID and region
 func GetAccountIDAndRegion(profile string, region string) (string, string, error) {
 	sess, err := NewAwsSession(profile, region)
-	// We nee to get the region from the connected cluster instead of the one configured for the calling user
-	// as it might not be found and it would then use the default (us-west-2)
-	_, region, err = GetCurrentlyConnectedRegionAndClusterName()
-	if err != nil {
-		return "", "", err
+	if (region == "") {
+		// TODO for the original author (dgozalo): Is this right? We are setting up a new cluster - so we do not have a current cluster!?
+		// We need to get the region from the connected cluster instead of the one configured for the calling user
+		// as it might not be found and it would then use the default (us-west-2)
+		_, region, err = GetCurrentlyConnectedRegionAndClusterName()
+		if err != nil {
+			return "", "", err
+		}
 	}
 	svc := sts.New(sess)
 
--- a/go.mod
+++ b/go.mod
@@ -132,7 +132,7 @@ require (
 	golang.org/x/oauth2 v0.0.0-20190604053449-0f29369cfe45
 	golang.org/x/sync v0.0.0-20190423024810-112230192c58
 	golang.org/x/sys v0.0.0-20190616124812-15dcb6c0061f
-	golang.org/x/xerrors v0.0.0-20190513163551-3ee3066db522 // indirect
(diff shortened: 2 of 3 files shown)


ascheman:
Unfortunately I cannot check the failed tests. Could you enable me @dgozalo or @jstrachan ?

hferentschik:
> Unfortunately I cannot check the failed tests. Could you enable me @dgozalo or @jstrachan ?

Unfortunately, we don't have a public view for this status checks atm. Let me see what I can do.

hferentschik:
Regarding the failing DCO check - you will need to sign-off your commits.

See:

hferentschik:
Regarding the _lint_ check. There is a Go sytle violation:

```
 // GetAccountIDAndRegion returns the current account ID and region
 func GetAccountIDAndRegion(profile string, region string) (string, string, error) {
 	sess, err := NewAwsSession(profile, region)
-	if (region == "") {
+	if region == "" {
 		// TODO for the original author (dgozalo): Is this right? We are setting up a new cluster - so we do not have a current cluster!?
 		// We need to get the region from the connected cluster instead of the one configured for the calling user
 		// as it might not be found and it would then use the default (us-west-2)
```

I recommend you run `make fmt` on your changes. That will reformat your code according to the Golang standard. You can then amend your commit with reformatted code.

ascheman:
Thanks @hferentschik, I obviously did too much Java programming instead of GoLang lately ;-)
Great that you use/support GoLang formatting in the mean time!

pkg/cloud/amazon/ecr.go
```diff
@@ -19,11 +19,14 @@ import (
 // GetAccountIDAndRegion returns the current account ID and region
 func GetAccountIDAndRegion(profile string, region string) (string, string, error) {
 	sess, err := NewAwsSession(profile, region)
-	// We nee to get the region from the connected cluster instead of the one configured for the calling user
-	// as it might not be found and it would then use the default (us-west-2)
-	_, region, err = GetCurrentlyConnectedRegionAndClusterName()
-	if err != nil {
-		return "", "", err
+	if (region == "") {
+		// TODO for the original author (dgozalo): Is this right? We are setting up a new cluster - so we do not have a current cluster!?
```
dgozalo: You are right. I think this function has been tweaked twice but only in the context of already being connected to a cluster as it was needed for things like ECR where it made sense to use the cluster region. 

It's right to use the provided region when it's not empty and use the cluster one when it is. We may have to make it more clear in the future, but for now we need this to let people create clusters.

dgozalo: It's also true that, if you have your region configured in your local AWS credentials / configuration it should be taken as the region to create the cluster.

hferentschik:
@ascheman @dgozalo - just touching base on this PR. What needs to be done?

@ascheman are you planning to push an update?

ascheman:
Sorry, didn't find the time to set up the GPG things up to now ...

hferentschik:
> Sorry, didn't find the time to set up the GPG things up to now ...

no worries. Let us know if you need more feedback/help. Looking forward to an update of this PR.

hferentschik:
@dgozalo @ascheman - what the status on this pull request? It needs another update, right?

hferentschik:
@ascheman, I am closing this pull request for now. If you get around updating it feel free to open a new pull request. That would be awesome.

PR #6095, opened 2019-11-13
feat: allow to install on PKS

dgozalo:
#### Submitter checklist

#### Description
Some changes needed to get a Jenkins-x running in PKS (and any k8s cluster not in a Cloud really).

We are now going to be defaulting to dockerhub for clusters other than GKE, EKS and AKS, we needed to handle creating the secret correctly and assigning the right push url to dockerhub.

Also, the dev environment master pipeline was failing when using local secrets. While using Vault should be the default and highly encouraged, we need to make it easy for people to try Jenkins-x in a standalone k8s cluster.

Local secrets are going to be added to a cluster secret and mounted as an optional volume in the build pods at `/builder/home/.jx/localSecrets/currentCluster`.

The `FileSystemSecretsClient` has been updated to fall back to this path if it doesn't find the secrets at `/builder/home/.jx/localSecrets/<cluster-name>`.

Also, removed the restriction to need Vault to add apps.

This doesn't have an issue because it's unplanned work for PKS.

### DIFF
--- a/pkg/cmd/add/add_app.go
+++ b/pkg/cmd/add/add_app.go
@@ -163,9 +163,6 @@ func (o *AddAppOptions) Run() error {
 			return util.InvalidOptionf(optionValues, o.SetValues,
 				"no more than one --%s can be specified when using GitOps for your dev environment", optionValues)
 		}
-		if o.GetSecretsLocation() != secrets.VaultLocationKind {
-			return fmt.Errorf("cannot install apps without a vault when using GitOps for your dev environment")
-		}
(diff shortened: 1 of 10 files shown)


hferentschik:
@dgozalo the integration tests panics!? That might be a legit issue.

dgozalo:
> @dgozalo the integration tests panics!? That might be a legit issue.

Yeah, I was trying to get the logs but I couldn't do it from the UI so I ran it again, I'll check what actual error this time.

PR #6108, opened 2019-11-13
feat: allow project name to be overridden for tenant service call

garethjevans:
allow boot to override the project name with the consumer project name when asking
the tenant service for a domain.


garethjevans:
Field could be called `DomainPrefixOverride` instead, it may make more sense and make it reusable

garethjevans:
There may be a better way of doing this without adding to OSS

garethjevans:
This logic has been moved into the jx-tenant-service

PR #6109, opened 2019-11-13
fix: don’t create default factory in tekton logging code

### LINKED ISSUE #6337: tekton_logging.go#L584 instantiates a new factory rather than using the current factory
### Summary 

See https://github.com/jenkins-x/jx/blob/master/pkg/logs/tekton_logging.go#L584 - this is instantiating a new, default factory rather than using the one that called the method

pmuir:
This was a bug we discovered when trying to reuse this code. I copied it and fixed it where I was using it for now but this fixes it here too

fixes #6337

### DIFF
--- a/pkg/cmd/step/step_unstash.go
+++ b/pkg/cmd/step/step_unstash.go
@@ -73,12 +73,19 @@ func NewCmdStepUnstash(commonOpts *opts.CommonOptions) *cobra.Command {
 
 // Run runs the command
 func (o *StepUnstashOptions) Run() error {
-	u := o.URL
+	authSvc, err := o.GitAuthConfigService()
+	if err != nil {
+		return err
+	}
+	return Unstash(o.URL, o.OutDir, o.Timeout, authSvc)
+}
+
+func Unstash(u string, outDir string, timeout time.Duration, authSvc auth.ConfigService) error {
 	if u == "" {
 		// TODO lets guess from the project etc...
 		return util.MissingOption("url")
 	}
-	file := o.OutDir
+	file := outDir
 	if file != "" {
 		isDir, err := util.DirExists(file)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


abayer:
`boot-vault` failure was a flaky (I opened https://github.com/jenkins-x/jx/issues/6114) but `integration` was a legit test compilation failure.

hferentschik:
@pmuir, any chance you can rebase and update this pull request? An issue would be nice as well providing some more context.

pkg/cmd/step/step_unstash.go
```diff
@@ -121,14 +123,16 @@ func (o *StepUnstashOptions) Run() error {
...
 	return func(urlText string) (string, error) {
-		token, err := GetTokenForGitURL(authSvc, urlText)
-		if err != nil {
-			log.Logger().Warnf("Could not find the git token to access urlText %s due to: %s", urlText, err)
-		} else if token != "" {
-			idx := strings.Index(urlText, "://")
-			if idx > 0 {
-				idx += 3
-				urlText = urlText[0:idx] + token + "@" + urlText[idx:]
+		if authSvc != nil {
```
hferentschik: so not passing an auth service is not an error condition?

pkg/logs/tekton_logging.go
```diff
@@ -580,10 +576,22 @@ func downloadLogFile(logsURL string, o *opts.CommonOptions) ([]byte, error) {
 	return logBytes, nil
 }
 
-func performProviderDownload(logsURL string) (*bufio.Scanner, error) {
-	provider, err := factory.NewBucketProviderFromTeamSettingsConfiguration(clients.NewFactory())
+func performProviderDownload(logsURL string, jxClient versioned.Interface, ns string) (*bufio.Scanner, error) {
+	provider, err := NewBucketProviderFromTeamSettingsConfiguration(jxClient, ns)
```
hferentschik: Is the problem that jxClient is not passed or that the namespace is not passed. This is for sure better, just in the end`clients.NewFactory()` works as well, right? Unless you switch context I guess. Trying to understand how the bug manifests itself.

pmuir: the problem is that the factories (i.e. the `jxClient` and `ns` in this case are created with default values, not the ones that the command was called with. This happens if you programatically reuse this code and create a custom jxClient (e.g. with impersonation)

pkg/cmd/step/step_unstash.go
pmuir: not sure - i probably didn't need to change this.

PR #6117, opened 2019-11-14
feat: Add support for specifying node tolerations in pipeline syntax and specifying arbitrary labels for pipeline pods

abayer:
#### Submitter checklist

#### Description

If said arbitrary label(s) are specified and `distributeParallelAcrossNodes` is true, then those label(s) will be used for the anti-affinity rather than the default of the `tekton.dev/pipelineRun` label+value. This will allow multiple runs of one or more pipelines to be on unique nodes, not just multiple pods of a single pipeline run.

Also switched the `distributeParallelAcrossNodes` behavior from `PreferredDuringSchedulingIgnoredDuringExecution` to `RequiredDuringSchedulingIgnoredDuringExecution`.

n/a

### DIFF
--- a/pkg/cmd/step/create/step_create_task.go
+++ b/pkg/cmd/step/create/step_create_task.go
@@ -665,7 +665,8 @@ func (o *StepCreateTaskOptions) generateTektonCRDs(effectiveProjectConfig *confi
 			return nil, errors.Wrapf(err, "parsing of pipeline timeout failed")
 		}
 	}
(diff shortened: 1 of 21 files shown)


abayer:
/hold

Testing this out by hand and I'm not sure it's right...

abayer:
Ok, my last change made things "work", but autoscaling doesn't seem to be triggered by something that can't be scheduled due to anti-affinity. I'll do some more experiments.

abayer:
/hold cancel

Ha, overwrote my own fix of the generated code. Good job, me.

I've now verified that autoscaling will happen, so removing the hold.

PR #5816, opened 2019-10-15
Wrap Exposecontroller UrlTemplate config in quotes

lhr0909:
#### Submitter checklist

- [x] Change is code complete and matches issue description.
- [x] Change is covered by existing or new tests.

#### Description

This change is to react to https://github.com/jenkins-x/exposecontroller/pull/191 this change in exposecontroller template. Need to add quote around the template generation code in `jx` as well.

#### Which issue this PR fixes

fixes #5800
fixes https://github.com/jenkins-x/exposecontroller/issues/192

PR #6130, opened 2019-11-15
fix: wrap Exposecontroller UrlTemplate config in quotes

### LINKED ISSUE #5800: jx upgrade ingress lost all ingress rules
### Summary 

`jx upgrade ingress` lost all ingress rules

`jx upgrade ingress ` and set custom template to `"{{.Service}}.{{.Domain}}"`

### Expected behavior

update success

### Actual behavior

the console shows ingress rules created. but that's not true, I cannot find any ingress rules after that.

### Jx version

```
NAME               VERSION
jx                 2.0.780
Kubernetes cluster v1.13.10-eks-5ac0f1
kubectl            v1.16.0
git                2.23.0
```

### Jenkins type

- [x] Classic Jenkins

I'm using multi-cluster, this happens on the production cluster.

hferentschik:
fixes #5800

supersedes PR  #5816

this PR adds the quotes around the URL template as well (as did #5816), but only if the template is not empty. I think for non-empty URL templates the quotes are indeed needs.

That said, jx upgrade ingress seems to work with an empty URL template, but not with a custom one.

### DIFF
--- a/pkg/expose/exposecontroller.go
+++ b/pkg/expose/exposecontroller.go
@@ -86,12 +86,21 @@ func RunExposecontroller(devNamespace, targetNamespace string, ic kube.IngressCo
 
 	CleanExposecontrollerReources(kubeClient, targetNamespace)
 
-	exValues := []string{
-		"config.exposer=" + ic.Exposer,
-		"config.domain=" + ic.Domain,
-		"config.tlsacme=" + strconv.FormatBool(ic.TLS),
-		"config.urltemplate=" + ic.UrlTemplate,
+	exValues := []string{}
+
+	if ic.Exposer != "" {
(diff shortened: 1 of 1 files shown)


hferentschik:
@lhr0909, any chance you could test the `jx` binary once this is merged and available in the releases `jx` binary?

It applies the quotes to non-empty URL templates, resulting in the same net result as your PR. I am not sure, however, whether this is all that's needed here. Would love to get your feedback on this.

zhaochunqi:
@hferentschik I upgraded and tested and it still not work.
for file `part0-configmap.yaml` , url template is `"\"{{.Service}}.{{.Domain}}"\"`, had to remove it manually like I did in https://github.com/jenkins-x/jx/issues/5800

PR #6140, opened 2019-11-18
chore: Refactor applications commands

mauromorales:
- Application command returns user applications and not K8s deployments
  (fixes jenkins-x#6099)
- Extract business logic outside of the get command into its own package
- Passing the --preview flag will return a message telling the user to
  use a different command in order to separate concerns (fixes jenkins-x#6271)

### DIFF
--- a/pkg/applications/applications.go
+++ b/pkg/applications/applications.go
@@ -1,0 +1,165 @@
+package applications
+
+import (
+	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	"github.com/jenkins-x/jx/pkg/cmd/clients"
+	"github.com/jenkins-x/jx/pkg/flagger"
+	"github.com/jenkins-x/jx/pkg/kube"
+	"github.com/jenkins-x/jx/pkg/kube/services"
+	"github.com/jenkins-x/jx/pkg/util"
+	"github.com/pkg/errors"
+	"k8s.io/api/apps/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// Deployment represents an application deployment in a single environment
+type Deployment struct {
+	*v1beta1.Deployment
+}
+
+// Environment represents an environment in which an application has been
+// deployed
+type Environment struct {
+	v1.Environment
+	Deployments []Deployment
+}
+
+// Application represents an application in jx
+type Application struct {
+	*v1.SourceRepository
+	Environments map[string]Environment
+}
+
+// List is a collection of applications
+type List struct {
+	Items []Application
+}
+
+// Environments loops through all applications in a list and returns a map with
+// all the unique environments
+func (al List) Environments() map[string]v1.Environment {
+	envs := make(map[string]v1.Environment)
+
+	for _, a := range al.Items {
+		for name, env := range a.Environments {
+			if _, ok := envs[name]; !ok {
+				envs[name] = env.Environment
+			}
+		}
+	}
+
+	return envs
+}
+
+// Name returns the application name
+func (a Application) Name() string {
+	return a.SourceRepository.Spec.Repo
+}
+
+// IsPreview returns true if the environment is a preview environment
+func (e Environment) IsPreview() bool {
+	return e.Environment.Spec.Kind == v1.EnvironmentKindTypePreview
+}
+
+// Version returns the deployment version
+func (d Deployment) Version() string {
+	return kube.GetVersion(&d.Deployment.ObjectMeta)
+}
+
+// Pods returns the ratio of pods that are ready/replicas
+func (d Deployment) Pods() string {
+	pods := ""
+	ready := d.Deployment.Status.ReadyReplicas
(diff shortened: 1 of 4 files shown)


pkg/applications/applications.go
```diff
@@ -0,0 +1,27 @@
+package applications
```
hferentschik: +1 for placing shared functionality into dedicated package away from the commands. One problem we are facing is that we have too strong coupling between commands (the Cobra parts of it) and the actual business logic.

We need to get into a state where the command just deals with the options and validation of things and creates the necessary clients but then calls into other packages containing the business logic. 

Commands calling commands is another anti-patterns. Sharing of code should be via defined interfaces in the business logic interfaces. 

Anyways, just some background. Hope it makes sense.

pkg/cmd/get/get_applications.go
```diff
@@ -261,6 +268,14 @@ func (o *GetApplicationsOptions) getAppData(kubeClient kubernetes.Interface) (na
 	if err != nil {
 		return nil, nil, nil, nil, errors.Wrap(err, "listing environments")
 	}
+
+	// Fetch applications
+	envMap, _, err := kube.GetOrderedEnvironments(client, ns)
```
hferentschik: See comment above. I think dealing with the environment should be part of `GetApplications`

pkg/applications/applications.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	"github.com/jenkins-x/jx/pkg/client/clientset/versioned"
+	"github.com/jenkins-x/jx/pkg/kube"
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// GetApplications returns a list of SourceRepositories
+func GetApplications(jxClient versioned.Interface, ns string, envMap map[string]*v1.Environment) ([]v1.SourceRepository, error) {
```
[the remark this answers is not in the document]
mauromorales: Thanks @hferentschik this is very useful feedback! Yes I agree, it makes more sense to do it the way you describe. I was a bit hesitant to move all that logic into applications pkg because I felt we were probably still going to use some of it in the cobra command but it also felt wrong to have to give it all these params. I guess we should make this first step and then refactor whatever is missing in the cobra command in the future.

pkg/applications/applications.go
```diff
@@ -0,0 +1,117 @@
...
+	Deployment *v1beta1.Deployment
+}
+
+type applicationEnvironment struct {
+	Environment v1.Environment
+	Deployments []applicationDeployment
+}
+
+// Application represents an application created by the user
+type Application struct {
```
hferentschik: +1

pkg/applications/applications.go
```diff
@@ -0,0 +1,117 @@
...
+	*v1.SourceRepository
+	Environments map[string]applicationEnvironment
+}
+
+// Name returns the name of an application
+func (a Application) Name() string {
+	return a.SourceRepository.Spec.Repo
```
hferentschik: +1

pkg/applications/applications.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Version returns the version of a deployment
+func (ad applicationDeployment) Version() string {
+	return kube.GetVersion(&ad.Deployment.ObjectMeta)
+}
+
+// Pods returns the ratio of pods that are ready/replicas
+func (ad applicationDeployment) Pods() string {
```
hferentschik: Is that needed somewhere for displaying?

mauromorales:
@hferentschik yes, still work in progress ... I need to actually generate the resulting table and add tests

pkg/applications/applications.go
mauromorales: yes, we display it on the table view unless the user specifies they don't want it

hferentschik:
@mauromorales - how is it going? Let me know when you need another review.

mauromorales:
@hferentschik I've now pushed a change with all the functionality in place. There are no tests atm as I was pretty much experimenting on each step, but I'm planning to work on that now. I'm still a bit torn as to how this should be tested because dependency on some client and external pkgs is always coupled. Any recommendations?

mauromorales:
@hferentschik I've removed the wip label, please have another look at this

pkg/applications/applications_test.go
```diff
@@ -0,0 +1,159 @@
...
+import (
+	"testing"
+
+	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/api/apps/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestAppendMatchingDeployments(t *testing.T) {
```
hferentschik: For what it's worth, I prefer Gingko tests. I think they can be written in a more concise way than Golang table-driven tests. Also, they offer me things like focused tests and better test reports. Probably a matter of taste.

PR #6146, opened 2019-11-19
refactor: use the auth config service to load the config from Kubernetes secrets

### LINKED ISSUE #6036: Use the auth config service to load the config from k8s secrets
### Summary 

It seems that right now some commands are loading directly the secrets from kubernetes secrets instead of relying on the factory abstraction for auth config service. This abstraction should handled when to load the config from kubernetes secrets, vault or local file. 

The code require some refactoring to move the kubernetes secrets loading into the auth config service.

### Jx version

```
COPY OUTPUT HERE
```

### Jenkins type

- [ ] Classic Jenkins

ccojocar:
* Replace the direct load of auth config from Kubernetes secrets with auth config service through the code
* Adapt various test to use the auth config service instead of the Kubernetes secrets.
* Remove the methods from factory interface which loads the auth cofing directly from secrets, or creates the git providers from secrets

#### Submitter checklist

fixes #6036


pkg/auth/auth_config_test.go
```diff
@@ -94,7 +85,7 @@ type ConfigTest struct {
 
 func (c *ConfigTest) Load() *auth.AuthConfig {
 	config, err := c.svc.LoadConfig()
-	c.AssertNoError(err)
+	require.NoError(c.t, err)
```
hferentschik: OOI, what's the difference between `assert` and `require`?

hferentschik:
What would be the core change/file to look at?

pkg/auth/auth_config_test.go
ccojocar: `require` stops immediately the tests without executing the follow up tests while `aseert` still keeps the test execution till to the end.

ccojocar:
> What would be the core change/file to look at?

Please look at all files :). Thanks!

garethjevans:
looks good, I can't spot anything obvious - keen to see how this will work when we don't have a vault ingress.

ccojocar:
> looks good, I can't spot anything obvious - keen to see how this will work when we don't have a vault ingress.

At the moment the factory will look up for secrets sources in this order (and will not fail anymore if e.g. secrets are not found in vault):
1) if vault configured tries to load the secret from vault
2) load secrets from k8s secrets when the step 2) fails or vault is not configured
3) load the secrets from files either from `~/.jx` or `git/credentials` when step 2) fails 

We would most likely tweak the loading sequence but for the initial refactoring, this will ensure that everything keeps working as before.

PR #6148, opened 2019-11-19
feat: checkout to the ref instead of FETCH_HEAD if ref is a branch

dgozalo:

#### Submitter checklist

#### Description
We need to be able to use a branch as a versions stream ref in order to get our automated release system to work, if the ref is a branch it will look for it in the remote branches and then checkout to it if it is.

fixes #

### DIFF
--- a/pkg/versionstream/versionstreamrepo/gitrepo.go
+++ b/pkg/versionstream/versionstreamrepo/gitrepo.go
@@ -95,9 +95,13 @@ func cloneJXVersionsRepo(versionRepository string, versionRef string, settings *
 				}
 				return dir, versionRef, nil
 			}
-			// TODO: It might be good to do the same behavior for any branch - just the checkout approach works for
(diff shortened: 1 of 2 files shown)


pkg/versionstream/versionstreamrepo/gitrepo.go
```diff
@@ -238,3 +253,16 @@ func resolveRefToTag(dir string, commitish string, gitter gits.Gitter) (string,
 	}
 	return resolved, nil
 }
+
+func isBranch(dir string, ref string, gitter gits.Gitter) (bool, error) {
```
abayer: This is useful enough that I think it would make sense to go in the git helpers.

abayer: https://github.com/jenkins-x/jx/blob/master/pkg/gits/helpers.go specifically =)

dgozalo: Done

PR #6161, opened 2019-11-20
feat: boot - attempt restore from existing dev repository

warrenbailey:
Adding a parameter to jx boot that allows you to boot from an existing dev repository

Signed-off-by: warrenbailey <warren@warrenbailey.net>

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/cmd/boot/boot.go
+++ b/pkg/cmd/boot/boot.go
@@ -47,6 +47,8 @@ type BootOptions struct {
 
 	// RequirementsFile provided by the user to override the default requirements file from repository
 	RequirementsFile string
+
+	EnvironmentRepo string
 }
 
 var (
@@ -98,6 +100,7 @@ func NewCmdBoot(commonOpts *opts.CommonOptions) *cobra.Command {
 	cmd.Flags().StringVarP(&options.EndStep, "end-step", "e", "", "the step in the pipeline to end at")
 	cmd.Flags().StringVarP(&options.HelmLogLevel, "helm-log", "v", "", "sets the helm logging level from 0 to 9. Passed into the helm CLI via the '-v' argument. Useful to diagnose helm related issues")
 	cmd.Flags().StringVarP(&options.RequirementsFile, "requirements", "r", "", "requirements file which will overwrite the default requirements file")
+	cmd.Flags().StringVarP(&options.EnvironmentRepo, "environment-repository-url", "n", "", "the  dev environment url repository to boot from")
 
 	return cmd
 }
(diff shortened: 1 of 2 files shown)


abayer:
This seems like it's relevant to #5772.

pkg/cmd/boot/boot_test.go
```diff
@@ -67,6 +67,42 @@ func TestDetermineGitRef_GitURLNotInVersionStream(t *testing.T) {
 	assert.Equal(t, "master", gitRef, "determineGitRef")
 }
 
+func TestConeDevEnvironment(t *testing.T) {
```
cagiti: small type in clone

macox:
> How is this different to `jx boot --git-url`?

I thought `jx boot --git-url` was for specifying the parent boot config, not the existing dev env repo as is the intent of this change?

pkg/cmd/boot/boot.go
```diff
@@ -98,6 +100,7 @@ func NewCmdBoot(commonOpts *opts.CommonOptions) *cobra.Command {
 	cmd.Flags().StringVarP(&options.EndStep, "end-step", "e", "", "the step in the pipeline to end at")
 	cmd.Flags().StringVarP(&options.HelmLogLevel, "helm-log", "v", "", "sets the helm logging level from 0 to 9. Passed into the helm CLI via the '-v' argument. Useful to diagnose helm related issues")
 	cmd.Flags().StringVarP(&options.RequirementsFile, "requirements", "r", "", "requirements file which will overwrite the default requirements file")
+	cmd.Flags().StringVarP(&options.EnvironmentRepo, "environment-repository-url", "n", "", "the  dev environment url repository to boot from")
```
hferentschik: '-n' as shortcut? That seems a bit odd.

Also "the  dev environment url repository to boot from" -> "the URL of the dev environment repository to boot from"

I am wondering though how this is supposed to work. So `--environment-repository-url` allows you to specify a repo URL and it clones and boot from there, ignoring `--git-url` and `--git-ref` which are about specifying a env repo to fork from. Are these options not exclusive to each other? I seem to be missing how this is handled in code. 

It would be nice to have an issue as well which describes the context of this change and what it tries to achieve.

warrenbailey: I misunderstood what @garethjevans wanted for this PR, so I'm reworking it now. The idea is to boot from a previously created dev repository and not from boot-config. This will feed into being able to repair, re-install or upgrade an existing cluster.

There is an issue but not on jx, I'll add one with some context.

I agree "-n" is a bit odd, unfortunately the obvious ones ("e", "d", "r") . I'll just remove the shorthand.

Marking as WIP for now

warrenbailey:
> How is this different to `jx boot --git-url`?

I'll add a ticket with some more context.

PR #6172, opened 2019-11-23
fix: Communicate errors when resolve function can not operate

RafalKorepta:
#### Submitter checklist

#### Description

The function resolve should handle edge cases like niled pointer receiver and niled user argument by providing information to the user.

The messages that resolve function returns are not very informative for the user of jx command-line tool, but developers could quickly identify a problem when the messages appears.

### DIFF
--- a/pkg/users/git.go
+++ b/pkg/users/git.go
@@ -60,8 +60,11 @@ func (r *GitUserResolver) GitUserSliceAsUserDetailsSlice(users []gits.GitUser) (
 // * making a call to the gitProvider
 // as often user info is not complete in a git response
 func (r *GitUserResolver) Resolve(user *gits.GitUser) (*jenkinsv1.User, error) {
-	if r == nil || user == nil {
-		return nil, nil
+	if r == nil {
+		return nil, errors.New("unable to resolve user on nil receiver")
+	}
+	if user == nil {
+		return nil, errors.New("unable to resolve when user is nil")
 	}
 	selectUsers := func(id string, users []jenkinsv1.User) (string, []jenkinsv1.User,
 		*jenkinsv1.User, error) {
--- a/pkg/users/git_test.go
+++ b/pkg/users/git_test.go
@@ -279,6 +279,23 @@ func TestFindUserWithNoEmailButSameGitLogin(t *testing.T) {
 	assert.Len(t, users.Items, 2)
 }
 
+func TestResolve(t *testing.T) {
+	t.Run("Unable to resolve on niled pointer receiver", func(t *testing.T) {
+		var resolver *users.GitUserResolver
+
+		_, err := resolver.Resolve(&gits.GitUser{})
+		assert.Error(t, err)
+	})
+
+	t.Run("Unable to resolve when user is niled", func(t *testing.T) {
+		resolver, _, err := prepare(t)
+		assert.NoError(t, err)
+
+		_, err = resolver.Resolve(nil)
+		assert.Error(t, err)
+	})
+}
+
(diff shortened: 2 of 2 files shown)


hferentschik:
@RafalKorepta, thanks for the pull request. Any chance you could also create a corresponding issue? Some more context would be ideal in terms when these edge cases appeared and what the exception was in this case.

pkg/cmd/preview/preview.go
```diff
@@ -286,24 +286,21 @@ func (o *PreviewOptions) Run() error {
...
-						log.Logger().Warnf("Unable to update user %s from %s because %v", prAuthor.Name, o.PullRequestName, err)
-					}
-					if author != nil {
-						user = &v1.UserSpec{
-							Username: author.Spec.Login,
-							Name:     author.Spec.Name,
-							ImageURL: author.Spec.AvatarURL,
-							LinkURL:  author.Spec.URL,
-						}
+				author, err := resolver.Resolve(pullRequest.Author)
```
hferentschik: Maybe the intention was to ignore a non-set author, as opposed to failing the command. 
Was there another error occurring with the preview command when the nil author was nil?

pkg/users/git.go
```diff
@@ -60,8 +60,11 @@ func (r *GitUserResolver) GitUserSliceAsUserDetailsSlice(users []gits.GitUser) (
 // * making a call to the gitProvider
 // as often user info is not complete in a git response
 func (r *GitUserResolver) Resolve(user *gits.GitUser) (*jenkinsv1.User, error) {
-	if r == nil || user == nil {
-		return nil, nil
+	if r == nil {
+		return nil, errors.New("unable to resolve user on niled receiver")
+	}
+	if user == nil {
+		return nil, errors.New("unable to resolve when user is niled")
```
hferentschik: I think it should be "unable to resolve when user is nil".

pkg/users/git.go
```diff
@@ -60,8 +60,11 @@ func (r *GitUserResolver) GitUserSliceAsUserDetailsSlice(users []gits.GitUser) (
 // * making a call to the gitProvider
 // as often user info is not complete in a git response
 func (r *GitUserResolver) Resolve(user *gits.GitUser) (*jenkinsv1.User, error) {
-	if r == nil || user == nil {
-		return nil, nil
+	if r == nil {
+		return nil, errors.New("unable to resolve user on niled receiver")
```
hferentschik: I think it should be "unable to resolve user on nil receiver". I don't think "niled" is a thing. At least I have not seen it used anywhere before.

RafalKorepta: You are right.

pkg/users/git.go
RafalKorepta: Agree.

pkg/cmd/preview/preview.go
RafalKorepta: I didn't think of such use case. In that case I should revert the preview.go change entirely.

RafalKorepta:
@hferentschik I didn't find any problem when I used preview command. On my debugging session, I realise that I could improve unit test for `GitUserResolve.Resolve()`. I will leave you to decide if this PR is worth merging.

hferentschik:
> I will leave you to decide if this PR is worth merging.

@RafalKorepta, thanks again for the PR. I will close it for now. It is not only _preview_ which relies on this _Resolve_, but also changelog generation. I'd also wish that the code would be a bit better, especially in terms of what it reports to the user, but by merging this change we basically change the semantics of the code. For example, the changelog generation should not error out if a single user cannot be resolved.

Eventually, I'd like to revisit this code, but I think it needs to go hand in hand with a bigger refactoring.

PR #6176, opened 2019-11-25
fix: lets avoid caching environment git repos

### LINKED ISSUE #5609: jx add app uses same cache directory for multiple clusters
In gitops mode `jx add app` uses the directory _$JX_HOME/environment/dev_ as a cache directory for creating the pull requests against the development repository.

This causes issues when the user has more than two Jenkins X cluster he connects to. 

Since the same directory is used for different environment repositories you end up with errors of the form:

```bash
$ jx add app foo
error: adding app fooversion  with alias  using gitops: creating pr for foo: pulling environment repo https://github.com/hf-bee-bot/environment-hardy-dev-dev.git into /Users/hardy/.jx/environments: fetching origin add-app-foo-0.0.46: git output: From https://github.com/hf-bee-bot/environment-hardy-dev-dev
 ! [rejected]        add-app-foo-0.0.46 -> origin/add-app-foo-0.0.46  (non-fast-forward): failed to run 'git fetch '...'
```

The cache directory should be specific to the cluster. Other parts of the code already use a cache directory which is keyed by cluster name, apps does not yet do that.

jstrachan:
as it causes issues for multiple-clusters or for stale git clones

fixes #5609

### DIFF
--- a/pkg/util/dirs.go
+++ b/pkg/util/dirs.go
@@ -1,11 +1,13 @@
 package util
 
 import (
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
 	"github.com/jenkins-x/jx/cmd/codegen/util"
 	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/pkg/errors"
 )
 
 func HomeDir() string {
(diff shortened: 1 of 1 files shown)


pkg/util/dirs.go
```diff
@@ -104,17 +106,14 @@ func CacheDir() (string, error) {
 	return path, nil
 }
 
+// EnvironmentsDir to more easily handle multiple clusters and minimise issues around
+// using old clones of environment git repositories lets use temporary directories for environment dirs
 func EnvironmentsDir() (string, error) {
-	h, err := ConfigDir()
+	dir, err := ioutil.TempDir("", "jx-env-")
```
hferentschik: And how are all these temp directories get deleted?

jstrachan: good point - probably best to wait  until we've time for a more comprehensive solution to this issue...

PR #6180, opened 2019-11-25
fix: lookup only the ingress when use ingress is setup

ccojocar:
Stop looking up the vault service, it is always available even when the vault ingress is not created.

#### Submitter checklist

fixes #


garethjevans:
tested this out locally on a remote cluster without vault ingress and it seems to work nicely

PR #6186, opened 2019-11-25
feat: allow boot upgrade to latest version stream release tag

### LINKED ISSUE #6214: allow upgrade boot to upgrade to latest version stream release tag
### Summary 

`jx upgrade boot` will currently upgrade the version stream to the master SHA by default.  It should be able to upgrade the latest release tag if the current version is a release tag.

macox:
#### Submitter checklist

#### Description

Allow upgrading boot to the latest version stream release tag, rather than just master or a supplied Ref

fixes #6214

### DIFF
--- a/pkg/gits/git_cli.go
+++ b/pkg/gits/git_cli.go
@@ -520,8 +520,13 @@ func (g *GitCLI) AddCommit(dir string, msg string) error {
 
 // AddCommitFiles perform an add and commit selected files from the repository at the given directory with the given messages
 func (g *GitCLI) AddCommitFiles(dir string, msg string, files []string) error {
-	fileString := strings.Trim(fmt.Sprintf("%v", files), "[]")
-	return g.gitCmd(dir, "commit", "-m", msg, "--", fileString)
+	for _, file := range files {
+		err := g.Add(dir, file)
+		if err != nil {
+			return err
+		}
+	}
+	return g.gitCmd(dir, "commit", "-m", msg)
 }
 
 func (g *GitCLI) gitCmd(dir string, args ...string) error {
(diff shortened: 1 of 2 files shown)


abayer:
Hmm, not sure how this will work with a repackaged distro that uses a branch per release train, rather than just master for everything in the versions repo. Getting the latest tag may not be the desired result there. Cc @daveconde

pkg/gits/git_cli.go
```diff
@@ -868,6 +873,12 @@ func (g *GitCLI) GetLatestCommitMessage(dir string) (string, error) {
 	return g.gitCmdWithOutput(dir, "log", "-1", "--pretty=%B")
 }
 
+// GetLatestTag returns the latest git commit message
+func (g *GitCLI) GetLatestTag(dir string) (string, error) {
```
hferentschik: How does this differ from `NthTag(<dir>, 1)`? I am a bit concerned about how `GitCLI` is developing. I am not a big fan of `NthTag` either, but since it's there we might as well use it. @macox WDYT?

pkg/gits/git_cli.go
```diff
@@ -520,8 +520,13 @@ func (g *GitCLI) AddCommit(dir string, msg string) error {
 
 // AddCommitFiles perform an add and commit selected files from the repository at the given directory with the given messages
 func (g *GitCLI) AddCommitFiles(dir string, msg string, files []string) error {
```
hferentschik: Why not use `Add` followed by `CommitDir`? We are adding all these very specific functions making it really hard to reuse any of this and also making it hard to understand/read the code making use of it, since we are not using/mirroring standard git CLI workflows.

macox: This is a change to an existing function. It's aim is to allow adding of specific files to a commit, doing an add may include unintended files to the commit.

PR #6187, opened 2019-11-25
fix: only commit requirement upgrade files if changed

### LINKED ISSUE #6213: upgrade boot fails when no changes to requirements
### Summary 

`jx upgrade` boot is failing as it tries to commit changes to the requirement file even of it contains no changes.

macox:

#### Submitter checklist

#### Description

This was trying to commit the requirements file even when there was no changes, and then failing.

fixes #6213

### DIFF
--- a/pkg/cmd/upgrade/upgrade_boot.go
+++ b/pkg/cmd/upgrade/upgrade_boot.go
@@ -163,10 +163,16 @@ func (o *UpgradeBootOptions) Run() error {
 }
 
 func (o UpgradeBootOptions) createCommitForRequirements(requirementsFileName string) error {
-	err := o.Git().AddCommitFiles(o.Dir, "Merge jx-requirements.yml", []string{requirementsFileName})
+	changedFiles, err := o.Git().ListChangedFilesFromBranch(o.Dir, requirementsFileName)
 	if err != nil {
-		return errors.Wrapf(err, "error creating a commit with the merged jx-requirements.yml file from dir %s",
-			requirementsFileName)
+		return errors.Wrap(err, "failed to list changed files")
+	}
+	if strings.Contains(changedFiles, requirementsFileName) {
+		err := o.Git().AddCommitFiles(o.Dir, "Merge jx-requirements.yml", []string{requirementsFileName})
+		if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/cmd/upgrade/upgrade_boot.go
```diff
@@ -161,10 +161,14 @@ func (o *UpgradeBootOptions) Run() error {
 }
 
 func (o UpgradeBootOptions) createCommitForRequirements(requirementsFileName string) error {
-	err := o.Git().AddCommitFiles(o.Dir, "Merge jx-requirements.yml", []string{requirementsFileName})
+	err := o.Git().AddCommitFiles(o.Dir, fmt.Sprintf("Merge %s", requirementsFileName), []string{requirementsFileName})
 	if err != nil {
-		return errors.Wrapf(err, "error creating a commit with the merged jx-requirements.yml file from dir %s",
-			requirementsFileName)
+		if strings.Contains(err.Error(), "nothing to commit") {
```
hferentschik: I don't think this is a good idea. We should not depend on checking the error message. I don't think the error message is part of the official git "API" and it also is dependent on the user's locale setting. We only had this case recently where code would fail on a computer with Swedish locale. ATM, we are enforcing `LC_ALL=ALL` when executing `git` from the command line. This works as a workaround, but the aim is to get away from matching error messages. 

It should be possible to do this by for example checking first whether there is something to commit.

hferentschik:
Would love to see an issue as well :-)

pkg/gits/git_cli.go
```diff
@@ -456,6 +456,16 @@ func (g *GitCLI) HasChanges(dir string) (bool, error) {
 	return len(text) > 0, nil
 }
 
+// HasFileChanged indicates if there are any changes to a file in the repository from the given directory
+func (g *GitCLI) HasFileChanged(dir string, fileName string) (bool, error) {
+	text, err := g.gitCmdWithOutput(dir, "status", "-s", fileName)
```
hferentschik: I think `--porcelain` would be the safer option, since that is guaranteed to be fixed and can even be pinned by specifying a version. I guess `-s` will work for now though.

PR #6188, opened 2019-11-25
feat: allow selection of chart in next-version

### LINKED ISSUE #6025: `jx step next-version` can not find correct chart
### Summary 

Seeing the same behavior in #2661 which the chart.yaml file does not get updated with the correct version, and therefore the helm release command fails.

I believe it is because we have three charts in our server project - `initdb-test`, `preview`, and `server`

The command `jx step next-version` incorrectly updates the version in `initdb-test` instead of `server` even when the `--dir ./charts/server` flag is provided.

Create a third chart in the charts folder that comes alphabetically before the preview and actual chart, then run `jx step next-version`

jx just bumps the first folder found instead of the folder that needs to be released

### Expected behavior

The correct chart.yaml file is updated so the chart can be released

### Actual behavior

the first folder found is updated instead of the one that needs to be

### Jx version

```
~ jx version       
NAME               VERSION
jx                 2.0.923
Kubernetes cluster v1.13.9-gke.3
kubectl            v1.15.2
git                2.21.0
Operating System   "Fedora release 30 (Thirty)"


verifying packages
```

### Jenkins type

- [ ] Classic Jenkins

### Kubernetes cluster

terraform + jx boot

happens locally on fedora, on mac, and in build pod

### LINKED ISSUE #6076: `jx step next-version` increments the version of an unrelated helm chart in the repository
### Summary 

It seems that the meta-pipeline `step-create-tekton-crds` increase the version of an application unrelated helm chart from a project repository when running the release pipeline. This chart is only used for testing purposes and is not the application related chart. 

It is even located in `test-resources/helm/` folder not `charts`.

See the logs:

```
Showing logs for build jenkins-x/jx-release-version/master #2 release stage meta-pipeline and container step-create-tekton-crds
? A local Jenkins X versions repository already exists, pulling the latest: Yes
running command: jx step next-version --use-git-tag-only --tag
created new version: 1.0.19 and written to file: ./VERSION
Updating chart version in test-resources/helm/Chart.yaml to 1.0.19
Tag v1.0.19 created and pushed to remote origin
```

This chart should be skipped from version updates. `jx` should only increase the version of the charts placed into the `./charts` folder.  

### Jx version

```
COPY OUTPUT HERE
```

### Jenkins type

- [ x] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

shteou:
#### Description

Allow the selection of a specific helm chart in `jx step next-version`

Added a parameter, `charts-dir`, to` jx step next-version` which allows
the user to specify which chart directory should be tagged with the
latest version

When invoked with the tag flag `jx step next-version` delegates 
to `jx step tag` without specifying `charts-dir`, resulting in a default
behaviour which attempts to search for the correct helm chart. In the
case where the repository contains multiple charts this can fail to
select the correct one

Fixes #6025

### DIFF
--- a/pkg/cmd/step/step_next_version_integration_test.go
+++ b/pkg/cmd/step/step_next_version_integration_test.go
@@ -112,3 +112,59 @@ func TestSetVersionChart(t *testing.T) {
 
 	assert.Equal(t, string(testFile), string(updatedFile), "replaced version")
 }
+
+func TestRunChartsDir(t *testing.T) {
+	originalJxHome, tempJxHome, err := testhelpers.CreateTestJxHomeDir()
+	assert.NoError(t, err)
+	defer func() {
+		err := testhelpers.CleanupTestJxHomeDir(originalJxHome, tempJxHome)
+		assert.NoError(t, err)
+	}()
+	originalKubeCfg, tempKubeCfg, err := testhelpers.CreateTestKubeConfigDir()
+	assert.NoError(t, err)
+	defer func() {
+		err := testhelpers.CleanupTestKubeConfigDir(originalKubeCfg, tempKubeCfg)
+		assert.NoError(t, err)
+	}()
+
+	f, err := ioutil.TempDir("", "test-run-chartsdir")
+	assert.NoError(t, err)
+
+	testData := path.Join("test_data", "next_version", "charts")
+	_, err = os.Stat(testData)
+	assert.NoError(t, err)
+
+	err = util.CopyDir(testData, f, true)
+	assert.NoError(t, err)
+
+	git := gits.NewGitCLI()
+	err = git.Init(f)
+	assert.NoError(t, err)
+
+	o := step.StepNextVersionOptions{
+		StepOptions: step2.StepOptions{
+			CommonOptions: &opts.CommonOptions{},
+		},
+	}
+	o.Out = tests.Output()
+	o.Dir = f
+	o.NewVersion = "1.2.3"
+	o.Tag = true
+	o.ChartsDir = path.Join(f, "charts", "foo")
+	o.SetGit(&gits.GitFake{})
+
+	defer os.Remove("VERSION")
+
+	err = o.Run()
+	assert.NoError(t, err)
+
+	// Check explicit chart has been updated
+	updatedChart, err := util.LoadBytes(o.Dir, "charts/foo/Chart.yaml")
+	expectedUpdatedChart, err := util.LoadBytes(testData, "charts/foo/expected_Chart.yaml")
+	assert.Equal(t, string(updatedChart), string(expectedUpdatedChart), "replaced version")
+
+	// And check no other chart was updated
+	bystanderChart, err := util.LoadBytes(o.Dir, "charts/bar/Chart.yaml")
+	expectedBystanderChart, err := util.LoadBytes(testData, "charts/bar/expected_Chart.yaml")
+	assert.Equal(t, string(bystanderChart), string(expectedBystanderChart), "no version replaced")
+}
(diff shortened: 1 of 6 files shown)


shteou:
I've added this PR to allow for the case where the charts structure looks like the following:

```
charts/
  an-unrelated-chart/
  my-service/
  preview/
```

Depending on the ordering of these charts folders, the command will potentially update the unrelated chart. However, I'm not sure if it's a pre-condition of the command that only the two charts (my-service and preview) exist.

@ccojocar Could the pipeline in #6076 be updated to specify the chart directory?
The glob in [step tag](https://github.com/jenkins-x/jx/blob/bb59278c2707e0e99b3c24be926745c324824388/pkg/cmd/step/step_tag.go#L274)  (called by next-version) looks like it could be tightened to the charts directory as well. I could raise a separate PR if that's useful and backward compatible.

abayer:
/ok-to-test

Since this looks like it’s fixing #6076 and/or #6025, could you add `fixes #...` to the PR description?

abayer:
You can fix the lint failure by running `make fmt` and pushing the changes it makes, and you can validate lint in general by running `make lint`, though that takes a bit due to doing all the codegen.

abayer:
The integration test failure looks like a flaky one. Rerunning.

shteou:
@abayer sorted the linting issue, cheers!
I've added fixes 6025 to the PR description.

Is it worth waiting for feedback on 6076? I think it may only be tangentially related.

abayer:
@shteou - Yeah, that's perfectly fine. =) Thanks!

abayer:
Sorry for not bringing this up earlier - I wasn't awake enough yet. =) Could you add a test, probably in `step_next_version_integration_test.go`, similar to `TestSetVersionChart`, using the new option? Thanks!

shteou:
> Sorry for not bringing this up earlier - I wasn't awake enough yet. =) Could you add a test, probably in `step_next_version_integration_test.go`, similar to `TestSetVersionChart`, using the new option? Thanks!

Sure thing, I'll give it a stab. I may prod you guys in Slack if I run into issues, if that's okay 👍

abayer:
Sounds good!

pkg/cmd/step/step_next_version_integration_test.go
```diff
@@ -112,3 +112,59 @@ func TestSetVersionChart(t *testing.T) {
...
+	o.Dir = f
+	o.NewVersion = "1.2.3"
+	o.Tag = true
+	o.ChartsDir = path.Join(f, "charts", "foo")
+	o.SetGit(&gits.GitFake{})
+
+	err = o.Run()
+	assert.NoError(t, err)
+
+	os.Remove("VERSION")
```
shteou: fyi, StepNextVersionOptions.Run has the side-effect of writing out a VERSION file, which ends up in the working directory (so `pkg/cmd/step` in this instance). Any better way to do this? Can we change the working directory for the scope of the test to the temporary directory returned by `ioutil.Tempdir("", "test-run-chartsdir")` ?

hferentschik: > fyi, StepNextVersionOptions.Run has the side-effect of writing out a VERSION file, which ends up in the working directory (so pkg/cmd/step in this instance)

:-( 

> Any better way to do this? Can we change the working directory for the scope of the test to the temporary directory returned by ioutil.Tempdir("", "test-run-chartsdir")

> Any better way to do this? Can we change the working directory for the scope of the test to the temporary directory returned by ioutil.Tempdir("", "test-run-chartsdir")

+1, I think this is the best option for now. Also, the cleanup should happen in a `defer` so that it really occurs, even in the case of an unexpected error

shteou:
Thanks for the help @hferentschik, I've addressed your comments and build otherwise looks good 👍

PR #6198, opened 2019-11-26
fix: regression in git auth from secrets

jstrachan:


pkg/auth/kube_config_handler.go
```diff
@@ -47,7 +47,7 @@ func (k *KubeAuthConfigHandler) LoadConfig() (*AuthConfig, error) {
 		return nil, errors.Wrap(err, "retrieving config from k8s secrets")
 	}
 	if secrets == nil {
-		return nil, fmt.Errorf("no secrets found for server kind %q and service kind %q", k.serverKind, k.serviceKind)
+		return nil, fmt.Errorf("no secrets found for server kind %q and service kind %q", k.serviceKind, k.kind)
```
abayer: Feels like the error message should change given that the relevant fields in the log are changing.

jstrachan: great catch

jstrachan: fixed

PR #6199, opened 2019-11-26
feat: fixes for GitHub Apps and also only use local gitAuth.yaml when importing and creating quickstarts

### LINKED ISSUE #6209: jx import + jx create quickstart should only use the users local git token
to create a new git repo + push to git rather than currently its using the pipeline token.

Currently if using vault and a user does `jx create quickstart` the users local git secrets in `git/credentials` or `~/.jx/gitAuth.yaml` are ignored and the user is prompted for their git user name + token each time

rawlingsj:
Main motivation for this is to support GitHub Apps and also ensure we only ever use users local git auth credentials when creating or importing applications.  We don't want to be using an in cluster git secret.

fixes #6209

### DIFF
--- a/pkg/cmd/clients/factory.go
+++ b/pkg/cmd/clients/factory.go
@@ -386,6 +386,17 @@ func (f *factory) CreateAuthConfigService(fileName string, namespace string,
 	return nil, fmt.Errorf("no auth config found for secret %q", fileName)
 }
 
+// CreateLocalGitAuthConfigService creates a new service which loads/saves the auth config from/to a local file.
+func (f *factory) CreateLocalGitAuthConfigService() (auth.ConfigService, error) {
+
+	if authService, err := f.createAuthConfigServiceFile(auth.GitAuthConfigFile, kube.ValueKindGit); err == nil {
+		return authService, nil
+	}
+	log.Logger().Debugf("No auth config found in file %s", auth.GitAuthConfigFile)
+
+	return nil, fmt.Errorf("no auth config found for secret %q", auth.GitAuthConfigFile)
+}
+
 // SecretsLocation indicates the location where the secrets are stored
 func (f *factory) SecretsLocation() secrets.SecretsLocationKind {
 	client, namespace, err := f.CreateKubeClient()
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "^pkg/jenkins/test_data/update_center.json.*$|^.secrets.baseline$|^.*test.*$",
     "lines": null
   },
-  "generated_at": "2019-11-25T11:30:02Z",
+  "generated_at": "2019-11-27T13:41:11Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -663,7 +663,7 @@
       {
         "hashed_secret": "2ee489d9ee5e2ab410b713e784e474d98cf08c54",
         "is_secret": false,
-        "line_number": 538,
+        "line_number": 539,
         "type": "Secret Keyword"
       }
     ],
--- a/jx/bdd/boot-vault/ci.sh
+++ b/jx/bdd/boot-vault/ci.sh
@@ -15,15 +15,20 @@ export BUILD_NUMBER="$BUILD_ID"
 JX_HOME="/tmp/jxhome"
 KUBECONFIG="/tmp/jxhome/config"
 
-mkdir -p $JX_HOME
+# lets avoid the git/credentials causing confusion during the test
+export XDG_CONFIG_HOME=$JX_HOME
+
+mkdir -p $JX_HOME/git
 
 jx --version
-jx step git credentials
+
+# replace the credentials file with a single user entry
+echo "https://$GH_USERNAME:$GH_ACCESS_TOKEN@github.com" > $JX_HOME/git/credentials
 
 gcloud auth activate-service-account --key-file $GKE_SA
 
 # lets setup git 
-git config --global --add user.name JenkinsXBot
+git config --global --add user.name jenkins-x-bot-test
 git config --global --add user.email jenkins-x@googlegroups.com
 
 echo "running the BDD tests with JX_HOME = $JX_HOME"
@@ -32,7 +37,7 @@ sed -e s/\$VERSION/${VERSION_PREFIX}${VERSION}/g -e s/\$CODECOV_TOKEN/${CODECOV_
 sed -e s/\$VERSION/${VERSION_PREFIX}${VERSION}/g -e s/\$CODECOV_TOKEN/${CODECOV_TOKEN}/g boot-vault.prow.yaml.template > boot-vault.prow.yaml
 
 # setup jx boot parameters
-export JX_VALUE_ADMINUSER_PASSWORD="$JENKINS_PASSWORD"
+export JX_VALUE_ADMINUSER_PASSWORD="$JENKINS_PASSWORD" # pragma: allowlist secret
 export JX_VALUE_PIPELINEUSER_USERNAME="$GH_USERNAME"
 export JX_VALUE_PIPELINEUSER_EMAIL="$GH_EMAIL"
 export JX_VALUE_PIPELINEUSER_TOKEN="$GH_ACCESS_TOKEN"
--- a/pkg/cmd/clients/fake/fake_factory.go
+++ b/pkg/cmd/clients/fake/fake_factory.go
@@ -378,3 +378,9 @@ func (f *FakeFactory) CreateHelm(verbose bool,
 func (f *FakeFactory) CreateCertManagerClient() (certmngclient.Interface, error) {
 	return fake_certmngclient.NewSimpleClientset(), nil
 }
+
+
+// CreateLocalGitAuthConfigService creates a new service which loads/saves the auth config from/to a local file.
+func (f *FakeFactory) CreateLocalGitAuthConfigService() (auth.ConfigService, error) {
+	return  f.GetDelegateFactory().CreateLocalGitAuthConfigService()
+}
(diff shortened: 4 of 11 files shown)


abayer:
@rawlingsj - `boot-vault` failed with 
```
error: failed to create private repository *************-test/bdd-gh-1574778062 due to: POST https://api.github.com/orgs/*************-test/repos: 404 Not Found []
```
Looks like it's trying to use the org API instead of the user API?

And `integration` is failing with:
```
pkg/cmd/clients/fake/fake_factory.go:67: cannot use (*FakeFactory)(nil) (type *FakeFactory) as type clients.Factory in assignment:
	*FakeFactory does not implement clients.Factory (missing CreateLocalGitAuthConfigService method)
pkg/cmd/clients/fake/fake_factory.go:75: cannot use f (type *FakeFactory) as type clients.Factory in return argument:
	*FakeFactory does not implement clients.Factory (missing CreateLocalGitAuthConfigService method)
pkg/cmd/clients/fake/fake_factory.go:124: cannot use &copy (type *FakeFactory) as type clients.Factory in return argument:
	*FakeFactory does not implement clients.Factory (missing CreateLocalGitAuthConfigService method)
pkg/cmd/clients/fake/fake_factory.go:131: cannot use &copy (type *FakeFactory) as type clients.Factory in return argument:
	*FakeFactory does not implement clients.Factory (missing CreateLocalGitAuthConfigService method)
```
Which confused me at first since lint passed and it smelled like a generate was needed...until I remembered that it was `FakeFactory`, not a mock. =)

abayer:
The `boot-vault` failure I put in https://github.com/jenkins-x/jx/pull/6199#issuecomment-558660549 is still there, fyi.

abayer:
@rawlingsj - so I think https://github.com/jenkins-x/jx/blob/ed6a23f9111b6760c7878b4c1c9922449cc1d182/pkg/gits/github.go#L300-L303 is not behaving the same way with this PR - the authenticated user isn't `jenkins-x-bot-test`, I'm assuming, so it's deciding `jenkins-x-bot-test` is an org.

abayer:
@warrenbailey will we need to make that change everywhere that uses boot in bdd tests?

abayer:
Well, it still failed in the same way, so I guess I don't need to worry about the `ci.sh` changes. =) Mind rolling them back?

rawlingsj:
> Well, it still failed in the same way, so I guess I don't need to worry about the ci.sh changes. =) Mind rolling them back?

@abayer @warrenbailey FWIW I just forced pushed to that branch to roll the last commit back and also rebased to sort the conflict with HEAD, am still expecting boot-vault to fail ATM though.

> Looks like it's trying to use the org API instead of the user API?

I'm not sure that is the issue yet as this repo doesn't exist, I just logged in and checked the test bot private repos https://github.com/jenkins-x-bot-test/bdd-gh-1574778062 and it doesn't exist, unless it was cleaned up?

I'm wondering if its the way we use git creds in this BDD test as the change from this PR forces create quickstart and import to use a local `gitAuth.yaml` file which the BDD test will need to create.  Will check in the morning.

pkg/cmd/clients/factory.go
```diff
@@ -386,6 +386,17 @@ func (f *factory) CreateAuthConfigService(fileName string, namespace string,
 	return nil, fmt.Errorf("no auth config found for secret %q", fileName)
 }
 
+// CreateLocalGitAuthConfigService creates a new service which loads/saves the auth config from/to a local file.
+func (f *factory) CreateLocalGitAuthConfigService() (auth.ConfigService, error) {
```
hferentschik: Me and @ccojocar have recently been talking about git and the various authentication needs around it. Personally I am not a fan using `gitAuth.yaml` and writing the credentials to disk. 

I was more envisioning a custom git credential helper or the use of `GIT_ASKPASS` and the ability to call into `jx` itself to get the appropriate credentials. Here is a PR I created a while back using `GIT_ASKPASS` - https://github.com/jenkins-x/jx/pull/5850/. We did not merge this since @ccojocar was working on his auth refactoring. I am pretty sure `GIT_ASKPASS` would work in this scenario as well w/o the need to write anything to disk.

hferentschik:
I would love to have a matching issue for this change. If nothing else than for proper changelog generation. Also, we have a whole bunch of git auth related issues. Being able to cross-link these issues would be nice.

jstrachan:
@hferentschik here you go: https://github.com/jenkins-x/jx/issues/6209

once we get things working with `~/.jx/gitAuth.yaml` we can easily add solutions which load git tokens from alternative preferred places first as you mention which should remove the need for git auth - but also still support those users who've been using jx for a couple of years and have a valid `~/.jx/gitAuth.yaml`

jstrachan:
@hferentschik not sure what you are not a fan of - but in general there are operations in the `jx` client which should never really even try to look at vault / pipeline secrets as they are clearly personal user operations like creating repos, pushing code, raising issues, creating personal PRs

jstrachan:
@hferentschik btw there's currently a massive UX fail; if using vault creating a quickstart you are often promoted multiple times to enter your github user + pwd which is a very bad UX

jstrachan:
@hferentschik btw I've already added support for discovering local git tokens from `git/credentials` files https://github.com/jenkins-x/jx/blob/2cd1ac87f2a3fd7d72ec5835b982601675efd145/pkg/auth/gitcredentials_auth.go#L15 and I welcome more efforts to automatically discover local git tokens via other means for local operations to minimise the need for prompting the user for tokens or even using `~/.jx/gitAuth.yaml`

However this issue is not about improving local detection, its about fixing the code so that local secrets are not *completely ignored* as they are right now and instead the user is prompted multiple times for them if using vault + no ingress

warrenbailey:
> Well, it still failed in the same way, so I guess I don't need to worry about the `ci.sh` changes. =) Mind rolling them back?

No problem, I didn't think it would work but was hoping as it's a bit more complicated to fix

jstrachan:
@ccojocar how is a generic function to create the auth config going to know the difference between being creating an auth service designed only for local secrets (e.g. for `jx create quickstart / jx import / jx create issue / jx create pr`) versus one which discovers vault v secrets v local stuff?

We definitely need a new function for when we definitely don't want to ever look at the pipeline secrets at all. This isn't about modifying the generic ordering really.

i.e. we need to add more functions to describe what kind of auth we want (local only, localOrPipeline, pipelineOnly etc) so we can use better logic to figure things out. 

I'm all for cleaning up the abstractions and layering - but I think we need a few different entry points at the top level so its easier to use the right logic/ordering for the right occassion. For other stuff like `jx preview / jx promote` where we can run it locally or we can run it in a pipeline then sure.

jstrachan:
@ccojocar i like your idea in the more medium term - though there's extra complications. when running BDD tests to create quickstarts we are inside the cluster but want to pretend we are local so its gonna take a reasonable amount of work to go through all the various code paths for all operations to get things aligned nicely to just use a single function which magically does the right thing in all contexts & intensions

ccojocar:
> how is a generic function to create the auth config going to know the difference between being creating an auth service designed only for local secrets (e.g. for jx create quickstart / jx import / jx create issue / jx create pr) versus one which discovers vault v secrets v local stuff?

Unfortunately until now was a hell of a mess. Commands loading the secrets directly from k8s secrets (load pipeline secrets), merging secretes from vault with k8s secrets which are duplicated etc. This was due to a lot of short-sighted hacks without clear design and so on.

> i.e. we need to add more functions to describe what kind of auth we want (local only, localOrPipeline, pipelineOnly etc) so we can use better logic to figure things out.

Why? 

I just see 2 use cases:
1) executed in the pipeline with pipeline token either from (vault, k8s secrets, git/credentials)
2) executed locally with a user token (pipeline vs use token needs to be clearly defined from UX perspective that the user understand that is his own token)

The sequence should be something like:

if executed locally:
- load token from file (gitAuth.yaml or git credentials), not found pop  up to user for token with clear description that it is his token and store that token locally in a file

else if executed in the pipeline:
- is vault configured tried to load from vault 
- not found in vault tried to load form k8s secrets
- not found in k8s secrets tried to load from files (git/credentials or gitAuth.yaml) - note that chart-museum auth is also using this so is not only git

ccojocar:
>  i like your idea in the more medium term - though there's extra complications. when running BDD tests to create quickstarts we are inside the cluster but want to pretend we are local so its gonna take a reasonable amount of work to go through all the various code paths for all operations to get things aligned nicely to just use a single function which magically does the right thing in all contexts & intensions

This is a clear sign that the use of static assignment per command to load the secrets is not going to work. This needs to be decoupled form commands execution because every command is potentially able to run either locally or in the cluster. They should get the auth config based on the context where they are being executed.

jstrachan:
@ccojocar I agree; we need a way to pass in the intention/context (or have separate functions). In this case we've added a new function for now (we can refactor later when we do a bigger refactor) to allow context/intention to affect the ordering of discovery of auth

jstrachan:
so `jx create quickstart / jx import / jx create issue / jx create pr` should *never* look in the pipeline secrets whether they are run from inside a pipeline or not.

however `jx preview / jx promote` should default to pipeline secrets if they are run in pipelines.

@ccojocar @hferentschik can you recommend a good quick fix for this while we figure out a nicer longer term refactor to do this?

ccojocar:
You can do whatever you want and merge what ever you what. I am just trying to improve the things.

jstrachan:
btw `jx preview / jx promote` were created before tekton which now creates the `git/credentials` file inside pipelines; so we may be able to get those 2 commands (when using tekton - to avoid breaking jenkins) to ignore pipeline secrets too whether inside or outside of a pod. So we may be able to move more `jx` commands over to a pure local discovery of secrets?

Its worth a spike at that in a couple of weeks to see how many `jx` commands can be converted to just use local secrets only + never look at the pipeline secrets?

jstrachan:
imagine we get most uses of the git auth using a new only-local git secrets function (which never looks in vault/secrets) - it'd be interesting to review the remaining uses in the codebase to see if they are still even valid; particularly since boot has come along too. 

At least then we'd be able to query the code and see how many `jx` commands really do need/use secrets/vault for git tokens which would help in refactoring/spring cleaning

PR #6211, opened 2019-11-27
fix: Apply labels as part of pull req creation, not just after

### LINKED ISSUE #6210: Nil pointer error when creating a PR with no changes
See #6203 for an example - I'm creating a separate issue for just this part of it, since that issue's real problem is that `jx import` tries to add the imported repo to the dev env repo, but `jx delete application` doesn't remove it, so if you try to import something that's already been imported/deleted, things go weird. But the nil pointer error is a separate problem in `pull_request_op` that could get hit by any situation where a PR isn't actually created.

abayer:
#### Submitter checklist

#### Description

Analogous to #6244 - gets us away from the nil pointer issue and cuts down on API calls a bit.

fixes #6210

### DIFF
--- a/pkg/gits/operations/pull_request_op.go
+++ b/pkg/gits/operations/pull_request_op.go
@@ -88,13 +88,10 @@ func (o *PullRequestOperation) CreatePullRequest(kind string, update ChangeFiles
 		filter := &gits.PullRequestFilter{
 			Labels: labels,
 		}
+		details.Labels = labels
 		result, err = gits.PushRepoAndCreatePullRequest(dir, upstreamInfo, forkInfo, o.Base, details, filter, !o.SkipCommit, commitMessage, true, o.DryRun, o.Git(), provider)
 		if err != nil {
 			return nil, errors.Wrapf(err, "failed to create PR for base %s and head branch %s from temp dir %s", o.Base, details.BranchName, dir)
-		}
-		err = gits.AddLabelsToPullRequest(result, labels)
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to add labels %+v to PR %s", labels, result.PullRequest.URL)
 		}
 	}
 	return result, nil
--- a/pkg/gits/operations/pull_request_op_test.go
+++ b/pkg/gits/operations/pull_request_op_test.go
@@ -26,7 +26,6 @@ import (
 
 	"github.com/jenkins-x/jx/pkg/gits/operations"
 
-	"github.com/acarl005/stripansi"
 	"github.com/ghodss/yaml"
 	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/client/clientset/versioned"
@@ -37,7 +36,6 @@ import (
 	"github.com/jenkins-x/jx/pkg/gits"
 	gits_test "github.com/jenkins-x/jx/pkg/gits/mocks"
 	resources_test "github.com/jenkins-x/jx/pkg/kube/resources/mocks"
-	"github.com/jenkins-x/jx/pkg/log"
 	"github.com/petergtz/pegomock"
 	"github.com/stretchr/testify/assert"
 	tektonclient "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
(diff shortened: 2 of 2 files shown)


pkg/gits/operations/pull_request_op.go
```diff
@@ -92,9 +92,11 @@ func (o *PullRequestOperation) CreatePullRequest(kind string, update ChangeFiles
 		if err != nil {
 			return nil, errors.Wrapf(err, "failed to create PR for base %s and head branch %s from temp dir %s", o.Base, details.BranchName, dir)
 		}
-		err = gits.AddLabelsToPullRequest(result, labels)
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to add labels %+v to PR %s", labels, result.PullRequest.URL)
+		if result != nil {
```
hferentschik: So what are we saying here? That `gits.PushRepoAndCreatePullRequest` might return nil and no error? Is there then not something semantically wrong in that function? In which cases do we get `nil` and no error?

ccojocar: I think is this case inside of the function:
```
		changed, err := gitter.HasChanges(dir)
		if err != nil {
			return nil, errors.WithStack(err)
		}
		if !changed {
			log.Logger().Warnf("No changes made to the source code in %s. Code must be up to date!", dir)
			return nil, nil
		}
```

Maybe it  makes sense to explicitly check if there are any changes before calling the function and skip everything afterwords. The logic would look more clear to me. This doesn't look like an error case.

abayer: @hferentschik We get `nil` and no error if there are no changes to create a PR with. And @ccojocar, which function do you mean?

ccojocar: `changed, err := gitter.HasChanges(dir)`

abayer: I'll tackle that in a followup - I'm gonna rework this like #6244 to apply the labels at creation time instead of immediately after to deal with the nil pointer specifically.

pkg/gits/operations/pull_request_op.go
```diff
@@ -88,14 +88,11 @@ func (o *PullRequestOperation) CreatePullRequest(kind string, update ChangeFiles
 		filter := &gits.PullRequestFilter{
 			Labels: labels,
 		}
+		details.Labels = labels
```
hferentschik: +1, that seems ways more sensible.

Any idea on why the labels were initially added in a separate call?  Seems to make no sense.

hferentschik:
@abayer, there seems to be a legit issue with the integration tests:

```
=== Failed
=== FAIL: pkg/gits/operations TestCreatePullRequest (0.04s)
WARNING: Failed to retrieve team settings: failed to create the ** client: namespaces "**" not found - falling back to default settings...
    pull_request_op_test.go:93: 
        	Error Trace:	pull_request_op_test.go:93
        	Error:      	"Created Pull Request: https://fake.git/testowner/testrepo/pulls/1
        	            	" does not contain "Added label updatebot to Pull Request https://fake.git/testowner/testrepo/pulls/1"
        	Test:       	TestCreatePullRequest
        	Messages:   	Updatebot label should be added to the PR
=== FAIL: pkg/gits/operations TestCreatePullRequestWithMatrixUpdatePaths (0.03s)
WARNING: Failed to retrieve team settings: failed to create the ** client: namespaces "**" not found - falling back to default settings...
    pull_request_op_test.go:214: 
        	Error Trace:	pull_request_op_test.go:214
        	Error:      	"Created Pull Request: https://fake.git/testowner/testrepo/pulls/1
        	            	" does not contain "Added label updatebot to Pull Request https://fake.git/testowner/testrepo/pulls/1"
        	Test:       	TestCreatePullRequestWithMatrixUpdatePaths
        	Messages:   	Updatebot label should be added to the PR
```

abayer:
@hferentschik Yup, definitely legit. I’ll try to look into it today.

abayer:
So it wasn't an actual problem with the code, it's that the tests were looking for logging in something we're not using any more, so I changed them to instead verify that the labels were on the `PullRequestArguments`. Should be good now.

PR #6222, opened 2019-11-27
chore: disable the initial build trigger if github app enabled as the…

rawlingsj:
… webhooks will already be setup


warrenbailey:
I've seen this behaviour as well, it triggers two master builds after creating a quickstart

rawlingsj:
fixes: #6222

PR #6224, opened 2019-11-28
fix: making sure client factory is faked in promote_test and step_create_pr_versions_test

### LINKED ISSUE #6223: promote_test and step_create_pr_versions_test connect to cluster
_promote_test_ and _step_create_pr_versions_test_ open a cluster connection atm which effectively make them integration tests.

Turns out, however, that the connection is not needed. So these tests can stay "unit" test by using a mock/fake factory.

hferentschik:
* Making sure unit tests don't connect to cluster
* Moving helper script into hack directory

fixes #6223


jx/bdd/boot-vault/cluster.yaml
```diff
@@ -15,4 +15,3 @@ clusters:
       - command: jx
         args:
           - boot
-          - -b
```
ccojocar: Why did you remove this? I think this allows to run boot in batch mode, otherwise it might ask for interactive input. I am not sure if it makes any difference.

hferentschik: ahh, my wrong. You are right. Let me remove this.

hferentschik: I was thinking only of -b as a option to boot which does not exist. But in this case it is a global option `jx -b boot`.

hferentschik: Fixed. Good catch. That said, I don't think it makes a difference in our case.

PR #6229, opened 2019-11-28
fix: make sure the cluster zone is stored in jx-install-config

### LINKED ISSUE #6225: 'jx boot' does not record the cluster zone in 'jx-install-config' config map
When using a cloud provider, `jx boot` need to store the cluster zone into the _jx-install-config_ config map. The config map is the only reliable way to determine the right cluster zone when trying to make other cloud API calls, e.g. trying to create cloud resources like buckets.

hferentschik:
fixes #6225


abayer:
/retest

Athens burped.

abayer:
/hold

PR #6231, opened 2019-11-28
chore: add git file change tests

macox:
#### Description

Some tests for git HasChanges and HasFileChanged functions


pkg/gits/git_cli_test.go
```diff
@@ -583,6 +583,55 @@ var _ = Describe("Git CLI", func() {
 			})
 		})
 	})
+
+	Describe("#HasChanges", func() {
```
hferentschik: :-) Yeah

abayer:
/hold

PR #6241, opened 2019-11-28
fix: check helm compatibility to avoid using helm v3

### LINKED ISSUE #6181: Improve the error message when using helm v3
### Summary 
Helm v3 came out recently and I updated my helm version to v3.
When using jx create cluster .... I received the following 
WARNING: retrying after error:failed to run 'helm init --client-only' command in directory '', output: 'Error: unknown flag: --client-only'
WARNING:
retrying after error:failed to run 'helm init --client-only' command in directory '', output: 'Error: unknown flag: --client-only'
FATAL: helm init failed: after 3 attempts, last error: failed to run 'helm init --client-only' command in directory '', output: 'Error: unknown flag: --client-only'

On macOS brew install helm

### Expected behavior
An error message that helm v3 is not supported yet and that one need to install helm v2.

### Actual behavior
WARNING: retrying after error:failed to run 'helm init --client-only' command in directory '', output: 'Error: unknown flag: --client-only'
WARNING:
retrying after error:failed to run 'helm init --client-only' command in directory '', output: 'Error: unknown flag: --client-only'
FATAL: helm init failed: after 3 attempts, last error: failed to run 'helm init --client-only' command in directory '', output: 'Error: unknown flag: --client-only'

### Jx version

~ $ jx version
NAME               VERSION
jx                 2.0.881
jenkins x platform 2.0.1642
kubectl            v1.16.3
git                2.24.0
Operating System   Mac OS X 10.15.1 build 19B88

verifying packages

WARNING: jx version 2.0.1011 is available in the version stream. We highly recommend you upgrade to it.

### Jenkins type

- [ X] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

### Kubernetes cluster

GCP

macOS

ccojocar:
#### Submitter checklist

fixes #6181

### DIFF
--- a/pkg/cmd/clients/factory.go
+++ b/pkg/cmd/clients/factory.go
@@ -763,7 +763,7 @@ func checkHelmCompatibility(helm *helm.HelmCLI) {
 		return
 	}
 	if v.Major > 2 {
-		log.Logger().Fatalf("You current helm version v3 is not supported. Please downgrade to helm v2.")
+		log.Logger().Fatalf("Your current helm version v3 is not supported. Please downgrade to helm v2.")
 	}
 }
 
--- a/pkg/cmd/clients/factory.go
+++ b/pkg/cmd/clients/factory.go
@@ -5,9 +5,7 @@ import (
 	"io"
 	"net/url"
 	"os"
-	"strings"
 
-	"github.com/blang/semver"
 	"github.com/jenkins-x/jx/pkg/config"
 
 	"github.com/jenkins-x/jx/pkg/kube/cluster"
@@ -734,8 +732,7 @@ func (f *factory) CreateHelm(verbose bool,
 	if verbose {
 		log.Logger().Debugf("Using helmBinary %s with feature flag: %s", util.ColorInfo(helmBinary), util.ColorInfo(featureFlag))
 	}
-	helmCLI := helm.NewHelmCLI(helmBinary, helm.V2, "", verbose)
-	checkHelmCompatibility(helmCLI)
+	helmCLI := helm.NewHelmCLIWithCompatibilityCheck(helmBinary, helm.V2, "", verbose)
 	var h helm.Helmer = helmCLI
 	if helmTemplate {
 		kubeClient, ns, _ := f.CreateKubeClient()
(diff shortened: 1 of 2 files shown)


pkg/cmd/clients/factory.go
```diff
@@ -747,6 +750,23 @@ func (f *factory) CreateHelm(verbose bool,
...
+		log.Logger().Warnf("Unable to detect the current helm version due to: %s", err)
+		return
+	}
+	v, err := semver.ParseTolerant(version)
+	if err != nil {
+		log.Logger().Warnf("Unable to parse the current helm version: %s", version)
+		return
+	}
+	if v.Major > 2 {
+		log.Logger().Fatalf("Your current helm version v3 is not supported. Please downgrade to helm v2.")
```
abayer: Probably worth change to `Your current helm version v%d` with `v.Major`, just in case something goes truly insane and they release v4 while we're not looking. =)

ccojocar: Fixed.

pkg/helm/helm_cli.go
```diff
@@ -66,6 +65,35 @@ func NewHelmCLI(binary string, version Version, cwd string, debug bool, args ...
...
+		log.Logger().Warnf("Unable to detect the current helm version due to: %s", err)
+		return
+	}
+	v, err := semver.ParseTolerant(version)
+	if err != nil {
+		log.Logger().Warnf("Unable to parse the current helm version: %s", version)
+		return
+	}
+	if v.Major > 2 {
+		log.Logger().Fatalf("Your current helm version v%d is not supported. Please downgrade to helm v2.", v.Major)
```
hferentschik: I am not a big fan of using `Fatalf`, but I guess w/o further changes that's the best we can do.

One could, however, return an error from `checkCompatibility` or a `bool` and a message `string`.  One can then use `Fatalf` in `NewHelmCLIWithCompatibilityCheck` based on the returned value. The benefit is that ` checkCompatibility`. Up to you. We can also merge as is.

ccojocar: I though about about returning an error and let the caller to handle it but if it doesn't treat that error as fatal the execution might continue with an invalid helm version.

ccojocar: and the `CreateHelm` doesn't propagate further any error, so we could log it there but keeping  this check encapsulated in the helm client wrapper seems more appropriate to me because we can have in future a different wrapper implementation for helm 3

ccojocar: We can merge as it is if you are fine with it.

hferentschik: Fair enough. Just wanted to hear your reasoning :-)

Issue #6246, opened 2019-11-29
controller build hangs


jstrachan:
just seen a controllerbuild hang and stop updating `PipelineActivity` records when using lighthouse...

```

goroutine 0 [idle]:
runtime.epollwait(0x4, 0x7ffdcf62cdf0, 0xffffffff00000080, 0x0, 0xffffffff03f4e06b, 0x22500000000, 0x0, 0x0, 0x0, 0x0, ...)
	/usr/local/go/src/runtime/sys_linux_amd64.s:675 +0x20
runtime.netpoll(0xc000060501, 0x1)
	/usr/local/go/src/runtime/netpoll_epoll.go:71 +0x126
runtime.findrunnable(0xc00005e000, 0x0)
	/usr/local/go/src/runtime/proc.go:2380 +0x51f
runtime.schedule()
	/usr/local/go/src/runtime/proc.go:2525 +0x21c
runtime.goexit0(0xc000d2af00)
	/usr/local/go/src/runtime/proc.go:2722 +0x1dd
runtime.mcall(0x0)
	/usr/local/go/src/runtime/asm_amd64.s:299 +0x5b

goroutine 1 [select (no cases), 123 minutes]:
github.com/jenkins-x/jx/pkg/cmd/controller.(*ControllerBuildOptions).Run(0xc000358690, 0x0, 0x0)
	/workspace/source/pkg/cmd/controller/controller_build.go:248 +0x11fb
github.com/jenkins-x/jx/pkg/cmd/controller.NewCmdControllerBuild.func1(0xc000648000, 0xc0008ed940, 0x0, 0x4)
	/workspace/source/pkg/cmd/controller/controller_build.go:118 +0x6f
github.com/spf13/cobra.(*Command).execute(0xc000648000, 0xc0008ed900, 0x4, 0x4, 0xc000648000, 0xc0008ed900)
	/workspace/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:830 +0x2ae
github.com/spf13/cobra.(*Command).ExecuteC(0xc000673400, 0xc0007a4e40, 0x5905f20, 0xc00000e010)
	/workspace/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914 +0x2fc
github.com/spf13/cobra.(*Command).Execute(...)
	/workspace/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
[...]

PR #6250, opened 2019-11-29
fix: lets use the gcloud API to write to buckets

jstrachan:
so that we can use write timeouts

### DIFF
--- a/pkg/cloud/buckets/buckets.go
+++ b/pkg/cloud/buckets/buckets.go
@@ -103,6 +103,28 @@ func ReadBucketURL(u *url.URL, timeout time.Duration) ([]byte, error) {
 	return data, nil
 }
 
+// WriteBucketURL writes the data to a bucket URL of the for 's3://bucketName/foo/bar/whatnot.txt?param=123'
+// with the given timeout
+func WriteBucketURL(u *url.URL, data []byte, timeout time.Duration) error {
+	bucketURL, key := SplitBucketURL(u)
+	return WriteBucket(bucketURL, key, data, timeout)
+}
+
+// WriteBucket writes the data to a bucket URL and key of the for 's3://bucketName' and key 'foo/bar/whatnot.txt'
+// with the given timeout
+func WriteBucket(bucketURL string, key string, data []byte, timeout time.Duration) error {
+	ctx, _ := context.WithTimeout(context.Background(), timeout)
+	bucket, err := blob.Open(ctx, bucketURL)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/cloud/gke/storage/bucket_provider.go
```diff
@@ -74,12 +79,15 @@ func (b *GKEBucketProvider) EnsureBucketIsCreated(bucketURL string) error {
...
-func (b *GKEBucketProvider) UploadFileToBucket(reader io.Reader, outputName string, bucketURL string) (string, error) {
-	bytes, err := ioutil.ReadAll(reader)
+func (b *GKEBucketProvider) UploadFileToBucket(reader io.Reader, key string, bucketURL string) (string, error) {
+	data, err := ioutil.ReadAll(reader)
 	if err != nil {
 		return "", err
 	}
-	return gke.UploadFileToBucket(bytes, outputName, bucketURL)
+
+	log.Logger().Debugf("Uploading data %s to bucket %s with key %s", string(data), bucketURL, key)
```
abayer: I tweaked this in #6252 to not include the data in the debug log, since, well, that's gonna be a a lot of text. Probably worth doing here too - I don't think there's really any value in seeing the data we're about to write, and I have a feeling that trying to log that much at once may be causing this log line not to show up at all.

jstrachan: ooh lemme do the same! :) was a copy/paste

abayer: Yup. =) Thanks!

pkg/cloud/buckets/buckets.go
```diff
@@ -103,6 +103,28 @@ func ReadBucketURL(u *url.URL, timeout time.Duration) ([]byte, error) {
 	return data, nil
 }
 
+// WriteBucketURL writes the data to a bucket URL of the for 's3://bucketName/foo/bar/whatnot.txt?param=123'
```
hferentschik: The godoc seem a bit messed up. There are some parts missing, right?

pkg/cloud/buckets/buckets.go
```diff
@@ -103,6 +103,28 @@ func ReadBucketURL(u *url.URL, timeout time.Duration) ([]byte, error) {
...
+// with the given timeout
+func WriteBucketURL(u *url.URL, data []byte, timeout time.Duration) error {
```
hferentschik: Would `WriteToBucketURL` be a better name? For me this makes more sense given the semantics of a bucket.

hferentschik:
relates to #6246

PR #6261, opened 2019-11-29
[wip] chore: introducing 'unit' build flag for non integration tests

hferentschik:
Splitting between a unit and integration check.


Makefile
```diff
@@ -135,30 +134,24 @@ tidy-deps: ## Cleans up dependencies
...
 
-test: make-reports-dir ## Run ONLY the tests that have no build flags (and example of a build tag is the "integration" build tag)
-	CGO_ENABLED=$(CGO_ENABLED) $(GOTEST) $(COVERFLAGS) -failfast -short ./... $(TEST_BUILDFLAGS)
+test: ## Run tests with the "unit" build tag
+	KUBECONFIG=/cluster/connections/not/allowed CGO_ENABLED=$(CGO_ENABLED) $(GOTEST) --tags=unit -failfast -short ./... $(TEST_BUILDFLAGS)
 
-test-cloud-gke: make-reports-dir ## Run ONLY the tests that have no build flags (and example of a build tag is the "integration" build tag)
-	CGO_ENABLED=$(CGO_ENABLED) $(GOTEST) -coverprofile=$(COVER_OUT) --covermode=count --coverpkg=./pkg/cloud/gke -failfast -short ./pkg/cloud/gke $(TEST_BUILDFLAGS)
+test-coverage : make-reports-dir ## Run tests and coverage for all tests with the "unit" build tag
+	CGO_ENABLED=$(CGO_ENABLED) $(GOTEST) --tags=unit $(COVERFLAGS) -failfast -short ./... $(TEST_BUILDFLAGS)
```
ccojocar: This will exclude the integration tests from the coverage. I think it make sense to have both since right now some integration tests are exclusive to unit tests.

hack/linter.sh
```diff
@@ -14,22 +14,21 @@ if ! [ -x "$(command -v golangci-lint)" ]; then
 	${DIR}/install_golint.sh -b $GOPATH/bin v1.20.0
 fi
 
-golangci-lint run \
+GO111MODULE=on golangci-lint run \
 	--no-config \
     --disable-all \
 	-E misspell \
 	-E unconvert \
-    -E deadcode \
```
ccojocar: Why is this checker removed?

hack/linter.sh
```diff
@@ -14,22 +14,21 @@ if ! [ -x "$(command -v golangci-lint)" ]; then
...
     -E unconvert \
-    -E gosec \
```
ccojocar: :D also this one

jenkins-x-unit.yml
```diff
@@ -0,0 +1,52 @@
...
+                    cpu: 4
+                    memory: 8Gi
+                  requests:
+                    cpu: 3
+                    memory: 3072Mi
+            steps:
+              - name: init-jx
+                image: gcr.io/jenkinsxio/builder-jx:0.1.639
+                command: jx
+                args: ['create','git','token','-n','fake','-t','1234','-u','https://github.com','fake']
```
ccojocar: I think this command is not stateless when is executed in the cluster. This might stored this config in k8s secret or vault. I would avoid running it.

Makefile
```diff
@@ -186,9 +179,6 @@ test-slow-integration-report: make-reports-dir get-test-deps test-slow-integrati
 test-slow-integration-report-html: make-reports-dir get-test-deps test-slow-integration
```
ccojocar: It makes sense to rename this to `test-intergration-report-html`

jenkins-x-unit.yml
```diff
@@ -0,0 +1,52 @@
+buildPack: none
```
ccojocar: Do you plan to addd a new context for this one?

jenkins-x-unit.yml
```diff
@@ -0,0 +1,52 @@
...
+                value: jenkins-x-bot
+              - name: GIT_COMMITTER_NAME
+                value: jenkins-x-bot
+              - name: BASE_WORKSPACE
+                value: /workspace/source
+              - name: GOPROXY
+                value: http://jenkins-x-athens-proxy:80
+              - name: JUNIT_REPORT
+                value: "true"
+              - name: PARALLEL_BUILDS
```
ccojocar: Maybe we can get rid of this fan out limit for parallel unit tests.

jenkins-x-integration.yml
```diff
@@ -53,7 +53,7 @@ pipelineConfig:
                 command: make
                 args: ['build']
 
-              - name: test-integration
```
ccojocar: I think makes sense to rename the `test-slow-integration` to just `test-integration`

jenkins-x-unit.yml
hferentschik: yes

Makefile
hferentschik: will do

hack/linter.sh
hferentschik: this is temporary. I needed to remove this, since it would not work for me locally. Need to double check on the reason for deadcode, but the gosec issue was an issue with random number generator which was a false positive for all I can tell.

hferentschik: See also https://github.com/jenkins-x/jx/pull/6286/commits/ba9c23b4d7d7c55edd705b8ad44da488b2992d7e

hferentschik:
@ccojocar thanks for the review. This is very much still work in progress. It is a little side thing I try to get ready. I'll incorporate your feedback.

Regarding `linter.sh`, I am seeing some issues when trying to run it locally. For exmaple, go modules should be enforced in the script and then I keep getting a false positive in the gosec mule.

hferentschik:
Closing this PR for now

PR #6266, opened 2019-12-01
WIP: feat: Add i18n support

LinuxSuRen:
#### Submitter checklist

fixes #


LinuxSuRen:
`pkg/i18n/bindata.go` is auto-generated. I'd prefer set it as git ignore. What do you think?

abayer:
Hey @LinuxSuRen ! Great to see you doing your localization magic for jx too! =)

Our apologies for the build logs not being viewable yet - we’re still nailing down some edge cases that can result in secrets exposed in the logs before we can share access to them publicly. But for now, I can share what’s failing with you.

`lint` is failing due to a go format error. I would suggest you run `make fmt`.

The `bdd`, `tekton`, and `boot-vault` tests are failing because they in many places explicitly parse the output from `jx` commands, and in runs with this PR, the output always starts with `Asset jx/zh_CN/LC_MESSAGES/jx.po not found`. While the BDD tests are probably a little too fragile when it comes to the exact format and content of `jx` command output, this in particular is obviously something we don’t want to have showing up whenever you run `jx whatever ...`.

I’m going to mark this PR as WIP (work in progress) for now so that it won’t accidentally get merged before you’re happy with it, and once it has been merged, we can try to update the BDD tests to support multiple locales, and add a new test run (probably not on every jx PR) to verify that the tests can pass using a Chinese locale. I think there may be some other problems inside `jx` itself where we parse the output of other commands, such as git, and expect the output to be using English locale, so some deeper changes may be needed.

LinuxSuRen:
@abayer Thanks for your response. But I made a mistake that I create this PR with the master branch.

PR #6286, opened 2019-12-03
fix: 'step boot vault' updates the vault CRD when there are changes

### LINKED ISSUE #5867: 'step boot vault' should update the vault CRD when there are changes
### Summary 

The vault CRD is programatically created by `step boot vault` but when a vault exists the command just skips and exists. It that case, it should check if there are any changes to the vault CRD (e.g. the docker image version update or some default settings enabled) and update the CRD.  

There is currently an issue during the boot update process when the vault version gets updated. This new version doesn't get propagated into the vault CRD.

### Expected behavior

The vault CRD should be updated when there are any changes.

hferentschik:
fixes #5867


hferentschik:
@dgozalo, @ccojocar, @daveconde - I think I got things working again :-)

I tested the following against GKE:

- `jx [create|delete|get] vault`
- `jx step boot vault` w/ and w/o pre-existing Vault instance. In the latter case, the Vault CRD gets now updated
- `jx create ...` with Vault enabled

For all, I can tell it's working. Not sure about AWS. I shifted around a lot of variables, so there is a good chance that I messed up something. Needs to be verified.

Not sure whether we want to merge as is and address AWS issue in a follow-up or whether we want to try sorting things out in this PR.

pkg/vault/create/create_integration_test.go
```diff
@@ -0,0 +1,251 @@
...
+			}
+			Expect(errorMessages).Should(ContainElement("the Vault name needs to be provided"))
+			Expect(errorMessages).Should(ContainElement("the namespace to create the Vault instance into needs to be provided"))
+			Expect(errorMessages).Should(ContainElement("a kube client needs to be provided"))
+			Expect(errorMessages).Should(ContainElement("a vault operator client needs to be provided"))
+			Expect(errorMessages).Should(ContainElement("a kube/cloud provider needs be provided"))
+		})
+
+		It("successfully installs vault", func() {
+			if cluster.IsInCluster() {
```
hferentschik: Unfortunately, this test only works locally. The pipeline service account does not have enough permissions to run this test. Not sure what's the best way forward here. Would be nice to have these type of tests running in the cluster as well, but elevating the pipeline user's privileges might be a security issue. @ccojocar @daveconde

hferentschik:
Should be good to go now. @dgozalo did a sanity check against AWS as well.

PR #6288, opened 2019-12-04
fix: `jx step helm build --recursive` 

### LINKED ISSUE #6267: `jx step helm build --recursive` is broken: *.tgz extraction does not create subfolders
### Summary 

`jx step helm build --recursive` is currently broken because when extracting the dependencies `.tgz`, sub-folders are not created (especially the `template` sub-folder). Thus, sub-charts have no impact what so ever when applying the chart.

This raises another question: shouldn't be `--recursive` the default behavior ? Currently, the preview chart does not use the dependencies of the main chart for these reasons. At least, all quickstarts `Makefile` should be changed: https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes/blob/master/packs/go/preview/Makefile#L18

This is the minimal fix to make it work: https://github.com/olli-ai/jx/commit/73d2969648a912e88c9877905de9724e1d156e60

However, before applying such a fix, we should:
- Write a test
- Check that it won't break other calls to `util.UnTargz`
- Remove `.tgz` files, to avoid `helm` to load 2 times each chart (and potentially choose the wrong one)

I'll work on a proper PR.

- create a project
```bash
jx create quickstart
```
- call `jx step helm build --recursive` or preview
```bash
cd charts/preview
echo "  version: 0.1.0-SNAPSHOT" >> requirements.yaml
jx step helm build --recursive
```
- acknowledge that there is no template folder in the dependencies
```bash
ls charts/my-project/
```
- and that helm fails to load them too
```bash
helm template .
```

### Expected behavior

sub-charts should be properly extracted

### Actual behavior

sub-charts have no template folder

### Jx version

current head: 6e47e73fc2909aba33fe8ed7f6721cf67975a2e1

### Jenkins type

- [ ] Classic Jenkins

aure-olli:
#### Submitter checklist

#### Description

Prior to this PR, `jx step helm build --recursive` was borken. When extracting dependencies, sub-directories were not created, including the precious `templates` directory. This, importation was useless.

The problems were the following:
- `opts.HelmInitRecursiveDependencyBuild` was calling `util.UnTargz` (which does not create sub-directories) instead of `util.UnTargzAll`.
- `util.UnTargzAll` failed to extract properly archives created by `helm`, because the `tgz` archive does not contain directories, only files. Directories needs to be created manually before trying to create each file.
- leaving the archive next to the extracted folder may confuse `helm`, which will read both to parse the sub-charts. The directory was fortunately read before the archive, but better safe than sorry.

2 tests are added:
- `Test_UnTargzAll_NestedNoDir` tests that `util.UnTargzAll` can properly extract a `tgz` archive with no directory.
- `Test_HelmInitRecursiveDependencyBuild_extraction` test that `opts.HelmInitRecursiveDependencyBuild` properly extracts all the `helm` archives and cleans them up.

I think `jx step helm build --recursive` should become the default behavior of build packs, or else the dependencies of the chart are not installed by the preview: https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes/blob/master/packs/go/preview/Makefile#L18

fixes #6267


hferentschik:
@aure-olli, thanks for a great pull request. Not only in terms of the actual fix provided but also the level of detail provided in the issue and pull request description.

abayer:
The `integration` failure was our one known flaky test there, `pkg/cmd/step/scheduler TestStepSchedulerConfigMigrateNonGitopsAdvanced`, which I really need to nail down one of these days. The `boot-vault` failure looks to have been a problem installing the UI, which I hadn’t seen before. Either way, I’ve relauched the failed tests.

aure-olli:
I've used `helm` for one of my test. I hope it's not related.

abayer:
Nah, you should be fine. The most recent failure is completely unrelated and seems to be due to something awry in our test environment - if it’s not fixed soon, I’ll just override the bdd context.

abayer:
We’re working on fixing the problem, so I’m gonna keep retrying periodically. =)

abayer:
Yaaaay, things are fixed, so we should be good. The test should pass in 15-20 minutes and then this’ll be merged. =)

aure-olli:
Thanks :)

PR #6289, opened 2019-12-04
fix: Correct link to import documentation

msvticket:
Fixing that documentation has moved on the website making following links in command line help and also reference on website give 404.


hferentschik:
thanks for the pull request @msvticket :-)

PR #6293, opened 2019-12-05
feat: Bring history for step split monorepo

### LINKED ISSUE #1486: jx step split monorepo failing
### Summary 
The `jx step split monorepo` command creates the remote repo but fails committing the files. 
Using the `angular-io-quickstart` shipped with Jenkins X as example source code for splitting code.
Replace myorg below with appropriate values.

```
mkdir /tmp/jxstep
cd /tmp/jxstep
git clone git@github.com:myorg/angular-io-quickstart.git
cd ang*
jx step split monorepo --glob "src*" --no-git=false --organisation myorg  

```
### Jx version

```
1.3.167
```

### Kubernetes cluster
N/A. Running locally.

MacOs 10.13.6
git version 2.18.0
### Expected behavior
Repo `src` created on Github with files from angular-io-quickstart/src committed.

### Actual behavior
Repo `src` created but no files committed. The output from first run:

> Created git repository to https://github.com/myorg/src
> 
> Pushed git repository to https://github.com/myorg/src

Furter investigation... something wrong with security? Trying this works fine:

```
# https://help.github.com/articles/splitting-a-subfolder-out-into-a-new-repository/
mkdir /tmp/gitfilter
cd /tmp/gitfilter
git clone git@github.com:myorg/angular-io-quickstart.git
cd ang*
branch=$(git branch | sed -n -e 's/^\* \(.*\)/\1/p')
git filter-branch --prune-empty --subdirectory-filter src $branch
# manual step, creating the myorg/mysrcproject in github
git remote set-url origin  https://github.com/myorg/mysrcproject.git
git push -u origin $branch
```

With the 'src' repo in place @Github (created by jx above) I try this:

```
sudo rm -r /tmp/jxstep
mkdir /tmp/jxstep
cd /tmp/jxstep
git clone git@github.com:myorg/angular-io-quickstart.git
cd ang*
jx step split monorepo --glob "src*" --no-git=false --organisation myorg  
```

Output from the jx command:
```
Cloning https://github.com/myorg/src.git into directory generated/src
error: failed to run 'git commit -m generated by: jx step split monorepo' command in directory 'generated/src', output: 'On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
	app/
	assets/
	browserslist
	environments/
	favicon.ico
	index.html
	karma.conf.js
	main.ts
	polyfills.ts
	styles.css
	test.ts
	tsconfig.app.json
	tsconfig.spec.json
	tslint.json

nothing added to commit but untracked files present': exit status 1
```

msvticket:
#### Submitter checklist

#### Description
Various features and fixes for the command `jx step split monorepo` has been made:

Made git history come along to new repo. This made it tricky to support existing remote repos for the split out repos, so I removed that.
Removed some code related to earlier handling copying files and commit message.
Removed test that didn't test what it was supposed to.
When 

I think these fixes warrants removing the command from the deprecation list.
There are more things that should be done, but I didn't want to make this PR too big. Supporting other git provider than github and fetching git provider and organisation from configuration to let user choose from comes to mind.

Actually fixes #1486 which was closed due to being rotten

### DIFF
--- a/pkg/cmd/step/step_split_monorepo.go
+++ b/pkg/cmd/step/step_split_monorepo.go
@@ -33,17 +33,17 @@ var (
 
 	stepSplitMonorepoExample = templates.Examples(`
 		# Split the current folder up into separate Git repositories 
-		jx step split monorepo -o mygithuborg
+		jx step split -o mygithuborg
 			`)
 )
 
 // StepSplitMonorepoOptions contains the command line flags
 type StepSplitMonorepoOptions struct {
 	step.StepOptions
+	gits.GitRepositoryOptions
 
 	Glob          string
 	Organisation  string
-	RepoName      string
 	Dir           string
 	OutputDir     string
 	KubernetesDir string
@@ -59,7 +59,7 @@ func NewCmdStepSplitMonorepo(commonOpts *opts.CommonOptions) *cobra.Command {
 	}
 
 	cmd := &cobra.Command{
-		Use:     "split monorepo",
+		Use:     "split",
 		Short:   "Mirrors the code from a monorepo into separate microservice style Git repositories so its easier to do finer grained releases",
 		Long:    stepSplitMonorepoLong,
 		Example: stepSplitMonorepoExample,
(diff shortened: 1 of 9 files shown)


msvticket:
When testing the code I realised that it always created the repositories public. Not good... So I had to handle some git options since even though defaulting to private is better, public needs to be an option. So while at it I did add support for other git provider and other git options.

hferentschik:
@msvticket, interesting pull request. I am actually surprised to get feedback on the deprecation of `step split monorepo`. Are you actually using this regularly? Don't you find it limiting in what it does? 

Regarding the deprecation, the idea is to reduce the surface area of `jx` in terms of code and the need for maintenance and documentation.  `step split monorepo` is really just a helper command with no direct coupling to Jenkins X. So the idea is to remove these commands from core `jx`. 

We are working on a binary extension approach similar to kubectl and docker where one can add new commands to `jx`, but that is still work in process. I could see `step split monorepo` being such an extension.

I am interested to hear your take on this.

hferentschik:
For what it's worth, there seems to be a formatting issue:

```
+++ ./pkg/cmd/step/step_split_monorepo_test.go	2019-12-05 15:07:25.147385985 +0000
@@ -33,7 +33,7 @@
 		OutputDir:    tempDir,
 		NoGit:        true,
 	}
-	testhelpers.ConfigureTestOptions(options.CommonOptions, nil, nil);
+	testhelpers.ConfigureTestOptions(options.CommonOptions, nil, nil)
 
 	err = options.Run()
 	assert.NoError(t, err, "Failed to run split monorepo on source %s output %s", testData, tempDir)
@@ -48,4 +48,3 @@
 		filepath.Join(tempDir, "bar", "charts", "bar", "Chart.yaml"),
 		filepath.Join(tempDir, "bar", "charts", "bar", "templates", "deployment.yaml"))
 }
```

`make fmt` should help.

msvticket:
Well, if I had known there was a plan do deprecate the command I might not have fixed it. The state of the command was such that deprecating it would not have been much of a loss. But if this PR would be merged the command would become useful and thus I don’t think it should be deprecated. After the extension mechanism  is in place this could of course be re-evaluated.

hferentschik:
> Well, if I had known there was a plan do deprecate the command I might not have fixed it. 

Yeah, deprecation is centralised in deprecated.go - https://github.com/jenkins-x/jx/blob/84e411625fbe33925351892b3b09c6b616cc2cd0/pkg/cmd/deprecation/deprecated.go#L198 which makes it hard to see when working on the monorepo command itself. 

> The state of the command was such that deprecating it would not have been much of a loss.

Right. Also, tbh I have a hard time seeing such a command working generically. There are so many factors to consider that I don't think you get around to plan and execute the split in a customized manner. 

> But if this PR would be merged the command would become useful and thus I don’t think it should be deprecated. 

If the pull request passes all tests I think we go ahead and merge it. I cannot make a call alone whether this also would harbour removal of the deprecation. Personally I'd say no. My main reason for the deprecation is the fact that I believe that the core `jx` binary needs to become more specialised, focusing on the core Jenkins X. Everything else, like this helper commands should go into a binary extension App.

I could imagine deprecating the command by initially hiding it. This way the code is still in `jx` and even executable. However, it would not be visible in the docs or the help message. This way the code would still be there and could easily moved to an App later.

> After the extension mechanism is in place this could of course be re-evaluated.

+1

pkg/cmd/step/step_split_monorepo.go
```diff
@@ -33,17 +33,17 @@ var (
 
 	stepSplitMonorepoExample = templates.Examples(`
 		# Split the current folder up into separate Git repositories
```
hferentschik: Maybe we could expand a bit here what the command actually does. IMO that is not obvious.

hferentschik:
@msvticket, I leave this up to you. You can just close this pull request or if you like get it up to date and we merge it (no guarantee in terms of how we handle deprecation though)

hferentschik:
@msvticket, I am going to close this pull request. If you get around fixing the remaining issues it had you can just open a new one. Happy to merge, but no guarantee in terms of the future of the command.

PR #6295, opened 2019-12-05
feat: display the github app name during quickstarts

cagiti:

If a cluster is using github apps to provide auth for the bot user then it would be useful to display the name of the github app during the initial setup phase when creating a quickstart.


cagiti:
still have some updates to this change, found a bug

PR #6296, opened 2019-12-06
[wip] feat: add i18n support

LinuxSuRen:
Replace #6266

#### Submitter checklist

#### Description

Add i18n support for the command line description. Currently, I just put one entry of Chinese.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ get-test-deps: ## Install test dependencies
 print-version: ## Print version
 	@echo $(VERSION)
 
-build: $(GO_DEPENDENCIES) ## Build jx binary for current OS
+build: $(GO_DEPENDENCIES) gen-data ## Build jx binary for current OS
 	CGO_ENABLED=$(CGO_ENABLED) $(GO) $(BUILD_TARGET) $(BUILDFLAGS) -o build/$(NAME) $(MAIN_SRC_FILE)
 
 build-all: $(GO_DEPENDENCIES) build make-reports-dir ## Build all files - runtime, all tests etc.
@@ -267,6 +267,12 @@ fmt: ## Format the code
       	else \
       		echo "Fixed formatting for: $(FORMATTED)"; \
       	fi
+JX_FILES="pkg/cmd/*.go"
+gettext:
+	go-xgettext -k=i18n.T "${JX_FILES}" > pkg/i18n/jx.pot
+
+gen-data:
+	cd pkg/i18n && go-bindata -o bindata.go -pkg i18n jx/zh_CN/LC_MESSAGES/
 
 .PHONY: importfmt
 importfmt:
--- a/pkg/cmd/cmd.go
+++ b/pkg/cmd/cmd.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/jenkins-x/jx/pkg/i18n"
 
 	"github.com/jenkins-x/jx/pkg/cmd/deprecation"
 	"github.com/jenkins-x/jx/pkg/cmd/profile"
@@ -78,7 +79,7 @@ func NewJXCommand(f clients.Factory, in terminal.FileReader, out terminal.FileWr
 	configureViper()
 	rootCommand := &cobra.Command{
 		Use:              "jx",
-		Short:            "jx is a command line tool for working with Jenkins X",
+		Short:            i18n.T("jx is a command line tool for working with Jenkins X"),
 		PersistentPreRun: setLoggingLevel,
 		Run:              runHelp,
 	}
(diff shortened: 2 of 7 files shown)


LinuxSuRen:
Here's the screenshot of the manual test.
image

LinuxSuRen:
What's the reason why this build was failed? I tested it locally.

abayer:
Lemme rerun and get you the log.

abayer:
Here’s the error:
```
cd pkg/i18n && go-bindata -o bindata.go -pkg i18n **/zh_CN/LC_MESSAGES/
/bin/bash: go-bindata: command not found
```

abayer:
I’m guessing you need to add a target like `get-test-deps` to do https://github.com/go-bindata/go-bindata#installation and then make that a dependency of the target that calls go-bindata.

Makefile
```diff
@@ -267,6 +267,12 @@ fmt: ## Format the code
       	else \
       		echo "Fixed formatting for: $(FORMATTED)"; \
       	fi
+JX_FILES="pkg/cmd/*.go"
+gettext:
+	go-xgettext -k=i18n.T "${JX_FILES}" > pkg/i18n/jx.pot
+
+gen-data:
+	cd pkg/i18n && go-bindata -o bindata.go -pkg i18n jx/zh_CN/LC_MESSAGES/
```
hferentschik: So you are using go-bindata to include the translations into the binary, right? I used go-bindata in another project and we got big problems when it suddenly disappeared. Seems the owner brought at least the repo back, but it is archived. Not sure I like the idea of depending on go-bindata. Are there any alternatives?

hferentschik: BTW, in terms of the `Makefile`, we would have to make sure that `go-bindata` as well as `go-xgettext` are getting installed.

LinuxSuRen: Yes, I met the `go-bindata` issue too. I given a temporary solution in [another project](https://github.com/jenkins-zh/jenkins-cli/blob/master/Makefile#L60). I don't have an alternative one for now.

pkg/cmd/cmd.go
```diff
@@ -78,7 +79,7 @@ func NewJXCommand(f clients.Factory, in terminal.FileReader, out terminal.FileWr
 	configureViper()
 	rootCommand := &cobra.Command{
 		Use:              "jx",
-		Short:            "jx is a command line tool for working with Jenkins X",
+		Short:            i18n.T("jx is a command line tool for working with Jenkins X"),
```
hferentschik: Would the plan be to internationalise just the help texts or also error messages? I am in particular concerned about error messages. We heavily rely on grepping for error messages when users report issues. This might get much harder in the case error messages are internationalised. What's your take on that?

Another question which comes to my mind - it seems you are suggesting to use the English text as message key, right. The benefit, of course, is that this was the default text is still directly in the code. On the other hand, the actual text is duplicated between the code and the message bundle where it serves as id. What happens if one fixes a typo in the code. That will make the translation fail for this message, right? Does the tooling (eg go-xgettext) handle this in some form? Can one easily detect when message IDs are out of sync?

hferentschik:
@LinuxSuRen, thanks for the pull request. A few thoughts:

> Add i18n support for the command line description. Currently, I just put one entry of Chinese.

What would be your plan if we decide to go ahead with internationalisation? Do a full sweep of the codebase and update this PR? Are you planning to just internationalise just the help messages or even error messages? The latter concerns me a bit since it potentially will make it harder to track down issues. 

Personally, I am not sure whether it is good timing to start with internationalization? I think we will see a fair bit of movement in the codebase over the coming months and I am concerned about the additional flux internationalization would introduce.

@jenkins-x/core any more thoughts on this?

pkg/cmd/cmd.go
LinuxSuRen: Thanks for your deep thoughts.

In my opinion, international for help text is the primary task. We can generate the help document base on the different languages. Error messages are complicated, I agree with you. But the internationalized error messages are also helpful and useful. We can request the users provide the 
 error message as English or other language bases on the main communication language.

About the second question, I think that we can figure a way to detect the changes in the message keys. But I don't the solution now. I mean it should be possible to detect them.

LinuxSuRen:
This is my plan:

- let jx support the i18n feature (current pr)
- provide the localization of Chinese for help text part by part
- generate [the document for jx](https://jenkins-x.io/docs/reference/commands/) as Chinese version

I know that the current solution might have some problems.  But it could bring a huge benefit for the Chinese users, and help me to promote `jx` in China.

hferentschik:
I fully understand the benefits of i18n and eventually would like to have it as well, but as outlined in the discussion, there are also technical costs involved with i18n. Right now I feel the timing to start with i18n is not ideal. I'd like to stabilize the codebase first more and apply some of the CLI refactorings we have been discussing within the team before. 

@LinuxSuRen, what do you think about putting this on hold for a few months and then reevaluate on where things stand.

LinuxSuRen:
hi @hferentschik, thanks for your response. I'm ok with your proposal. I'll try to attend the office hour to discuss more.

hferentschik:
Closing this pull request for now. Hopefully, we can use this as a foundation further down the road.

Issue #6158, opened 2019-11-20
jx boot doesn't recognize gitKind: bitbucketserver as a Bitbucket git server


SmartassSkeleton:
### Summary 

When trying to install Jenkins-X on an existing EKS cluster using `jx boot` the process fails on creating a repository for the `dev` environment. From the error message it seems like the git server API the `jx` CLI is trying to use is GitHub API.   

### Steps to reproduce the behavior

Using `jx boot` with `jx-requirements.yml` file cloned from the [jenkins-x-boot-config](https://github.com/jenkins-x/jenkins-x-boot-config/) repository:

```
autoUpdate:
  enabled: false
  schedule: ""
bootConfigURL: https://github.com/jenkins-x/jenkins-x-boot-config
cluster:
  clusterName: my-cluster
  environmentGitOwner: cicd
  gitKind: bitbucketserver
  gitName: bs
  gitServer: https://bitbucket.example.com
  namespace: jx
  provider: eks
  region: eu-west-1
  registry: xxxxxxxxxxxx.dkr.ecr.eu-west-1.amazonaws.com
  zone: eu-west-1
environments:
- ingress:
    domain: staging.example.com
    externalDNS: false
    namespaceSubDomain: -jx.
    tls:
      email: user@example.com
      enabled: true
      production: true
  key: dev
- ingress:
    domain: ""
    externalDNS: false
    namespaceSubDomain: ""
    tls:
      email: ""
      enabled: false
      production: false
  key: staging
- ingress:
    domain: ""
    externalDNS: false
    namespaceSubDomain: ""
    tls:
      email: ""
      enabled: false
      production: false
  key: production
gitops: true
ingress:
  domain: staging.example.com
  externalDNS: false
  namespaceSubDomain: -jx.
  tls:
    email: user@example.com
    enabled: true
    production: true
kaniko: true
repository: nexus
secretStorage: local
storage:
  backup:
    enabled: true
    url: s3://bucket
  logs:
    enabled: true
    url: s3://bucket
  reports:
    enabled: true
    url: s3://bucket
  repository:
    enabled: true
    url: s3://bucket
vault: {}
velero: {}
versionStream:
  ref: v1.0.201
  url: https://github.com/jenkins-x/jenkins-x-versions.git
webhook: lighthouse
```

### Expected behavior

I expected `jx boot` to successfully deploy Jenkins X on my Kubernetes cluster and create the appropriate git repositories and hooks. 

### Actual behavior

[...]

PR #6297, opened 2019-12-06
fix: Update go-gitlab to latest version to resolve project creation

TonyNoble:
fix: Remove hard-coding of github as target for duplication in verify_environment step

Signed-off-by: Tony Noble <tony.noble@gmail.com>

#### Submitter checklist

#### Description
Further fixes to allow jx boot to work cleanly with GitLab (on-prem or .com).  
* Updated to use the latest version of the go-gitlab module, as this includes updates for required fields in API calls to latest versions of GitLab.
* When creating environment config repos (or checking for their existence), the GitKind field was hard-coded to GitHub, meaning the wrong client and API was used.  Given that the the GitKind field is available in the jx-requirements config for the repo in use, it makes sense to use that but default to GitHub if not specified

fixes #
This should provide a partial fix for #6158 as well.

### DIFF
--- a/pkg/gits/gitlab.go
+++ b/pkg/gits/gitlab.go
@@ -172,7 +172,7 @@ func (g *GitlabProvider) GetRepository(org, name string) (*GitRepository, error)
 	if err != nil {
 		return nil, err
 	}
-	project, response, err := g.Client.Projects.GetProject(pid)
+	project, response, err := g.Client.Projects.GetProject(pid, nil, nil)
 	if err != nil {
 		return nil, fmt.Errorf("request: %s failed due to: %s", response.Request.URL, err)
 	}
@@ -224,7 +224,7 @@ func (g *GitlabProvider) ForkRepository(originalOrg, name, destinationOrg string
 	if err != nil {
 		return nil, err
 	}
-	project, _, err := g.Client.Projects.ForkProject(pid)
+	project, _, err := g.Client.Projects.ForkProject(pid, nil, nil)
 	if err != nil {
 		return nil, err
 	}
@@ -352,7 +352,7 @@ func (g *GitlabProvider) UpdatePullRequestStatus(pr *GitPullRequest) error {
 	if err != nil {
 		return err
 	}
-	mr, _, err := g.Client.MergeRequests.GetMergeRequest(pid, *pr.Number)
+	mr, _, err := g.Client.MergeRequests.GetMergeRequest(pid, *pr.Number, nil, nil)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


TonyNoble:
Added as WIP because while the GitKind change works for me, I don't have the ability to check with other setups and wanted to check that the config location I'm taking the value from is sane.

pkg/gits/gitlab.go
```diff
@@ -172,7 +172,7 @@ func (g *GitlabProvider) GetRepository(org, name string) (*GitRepository, error)
 	if err != nil {
 		return nil, err
 	}
-	project, response, err := g.Client.Projects.GetProject(pid)
+	project, response, err := g.Client.Projects.GetProject(pid, nil, nil)
```
hferentschik: these _nil_ values are needed now due to the upgrade of the GitLab client?

hferentschik:
@TonyNoble thanks for the pull request. How far does this change get you to using GitLab with Jenkins X? Can you successfully install Jenkins X via `jx boot` now?

pkg/gits/gitlab.go
TonyNoble: The method signature for GetProject (plus a few others) changed after version 0.6.0 to GetProject(pid interface{}, opt *GetProjectOptions, options ...OptionFunc)  - neither the second or third parameters seemed to be required (and are nil-checked as such), do anything specifically useful in the context of use by jx currently.

TonyNoble:
@hferentschik - I'm getting there.  The duplication of the environment config repos now works, as does jx import on existing repos though I'm still slightly hamstrung by the inability to use gitlab instances that use a self-signed certificate.

There also seems to be an inherent confusion when gitlab work is not being performed in the user's personal space - the 'environmentGitOwner' value is being used for two different purposes at various times, as far as I can tell.  I've worked round that with a certain amount of manual intervention, but I think that will require a separate PR.

pkg/gits/gitlab.go
hferentschik: thanks

hferentschik:
> I'm getting there. The duplication of the environment config repos now works, as does jx import on existing repos 

Cool. TBH, I am impressed you get so long.

> though I'm still slightly hamstrung by the inability to use gitlab instances that use a self-signed certificate.

I see.

> There also seems to be an inherent confusion when gitlab work is not being performed in the user's personal space - the 'environmentGitOwner' value is being used for two different purposes at various times, as far as I can tell. 

Might be worth creating an issue for that and outlining the problem. Maybe someone can then chime in with some suggestions. 

> I've worked round that with a certain amount of manual intervention, but I think that will require a separate PR.

Sounds good to me.

TonyNoble:
> > There also seems to be an inherent confusion when gitlab work is not being performed in the user's personal space - the 'environmentGitOwner' value is being used for two different purposes at various times, as far as I can tell.
> 
> Might be worth creating an issue for that and outlining the problem. Maybe someone can then chime in with some suggestions.
> 
I'm hoping that once these fixes go in, I can try again from the start and narrow-down the issue a little further.
> > I've worked round that with a certain amount of manual intervention, but I think that will require a separate PR.
> 
> Sounds good to me.
>
The main thing is that I need this to be a 'push-button' repeatable process for our needs.  So every time I get something fixed, I wipe my test cluster and start again.  As you can imagine, it's not a fast process :)

Issue #6097, opened 2019-11-13
Refactor applications commands


hferentschik:
Moving forward "Applications" is the terms for what the user develops. 

A user creates applications by creating a quickstart `jx quickstart` or by creating an application from existing source `jx create application`. 

The user is the able to list configured applications `jx get applications`, filtering for example by environment.

The user can also delete an application via `jx delete application` which will remove any associated resources within Jenkins X. This means the application gets removed from the [Prow|Lighthouse] configuration, as well as from environments in which it is installed. The latter needs to be paired with appropriate user confirmation to avoid for example deleting running applications from production environments. Should even the repo be deleted?

Notes:
An application is a project once it is present in the Prow config.

hferentschik:
See also epic #6096 which tries to resolve further inconsistencies between Apps vs AddOns vs Applications

PR #6307, opened 2019-12-09
fix: do not display undeployed applications

mauromorales:
How and when to display these applications needs to be defined

## Context

We are currently in the process of refactoring all application commands to deal with user applications and not deployments in general (relates to #6097). However there are some additional cases that might need to be addressed, one of them is **undeployed applications** e.g. you create a quickstart but for some reason deployment fails, you are left with one source repository that has no deployments. It would be good to have a way to display these in the get application command, either by default or through a flag.

I'm reverting this change because atm there's a BDD test that breaks with the following error:

```
reates a spring application and promotes it to staging
       [It]
      /workspace/source/test/suite/spring/jx_create_spring.go:36
      get applications with a URL
      Unexpected error:
          <*errors.fundamental | 0xc000517420>: {
              msg: "must be at least 3 fields in bdd-jx, entire output was APPLICATION           STAGING PODS URL\nbdd-jx                             \nbdd-spring-1575891571 0.0.1   1/1  http://bdd-spring-1575891571-jx-staging.cbjx-vs-pr-200-46arc.staging-jxaas.live",
              stack: [0x17fa1c4, 0x18aba3c, 0x1740360, 0x18abedc, 0xfd5589, 0x18a9504, 0x18a94dc, 0x18ac8ea, 0x1740360, 0x18a6e8c, 0x18acfc9, 0x18acfa6, 0x1740360, 0x18a7af4, 0x18b14a3, 0x1740360, 0x18b2029, 0x172fd2c, 0x172f99f, 0x172ee34, 0x1735e06, 0x17356b4, 0x173b5ff, 0x173b114, 0x173a957, 0x173cfbe, 0x173f997, 0x173f6
dd, 0x18a2782, 0x18b2a50, 0x504fa0, 0x45eaf1],
          }
          must be at least 3 fields in bdd-jx, entire output was APPLICATION           STAGING PODS URL
          bdd-jx
          bdd-spring-1575891571 0.0.1   1/1  http://bdd-spring-1575891571-jx-staging.cbjx-vs-pr-200-46arc.staging-jxaas.live
      occurred
```

Until the proper functionality is defined, we've decided to keep the functionality for undeployed applications untouched.

### DIFF
--- a/pkg/cmd/get/get_applications.go
+++ b/pkg/cmd/get/get_applications.go
@@ -135,40 +135,44 @@ func (o *GetApplicationsOptions) generateTable(kubeClient kubernetes.Interface, 
 		row := []string{}
 		name := a.Name()
 
-		for _, k := range o.sortedKeys(list.Environments()) {
+		if len(a.Environments) > 0 {
 
-			if ae, ok := a.Environments[k]; ok {
-				for _, d := range ae.Deployments {
-					name = kube.GetAppName(d.Deployment.Name, k)
-					if ae.Environment.Spec.Kind == v1.EnvironmentKindTypeEdit {
-						name = kube.GetEditAppName(name)
-					} else if ae.Environment.Spec.Kind == v1.EnvironmentKindTypePreview {
-						name = ae.Environment.Spec.PullRequestURL
+			for _, k := range o.sortedKeys(list.Environments()) {
+
+				if ae, ok := a.Environments[k]; ok {
+					for _, d := range ae.Deployments {
+						name = kube.GetAppName(d.Deployment.Name, k)
+						if ae.Environment.Spec.Kind == v1.EnvironmentKindTypeEdit {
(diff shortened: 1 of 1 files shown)


mauromorales:
fixes #6308

pkg/cmd/get/get_applications.go
```diff
@@ -131,12 +131,17 @@ func (o *GetApplicationsOptions) Run() error {
...
 
+OUTER:
 	for _, a := range list.Items {
 		row := []string{}
 		name := a.Name()
 
 		for _, k := range o.sortedKeys(list.Environments()) {
 
+			if len(a.Environments) == 0 {
+				continue OUTER
```
macox: maybe just use a `break` here to reduce complexity

mauromorales: @macox if I break that only affects the inner loop not the outer one, or am I missing something?

PR #6314, opened 2019-12-10
fix: exclude env vars from build pack if passed in

### LINKED ISSUE #6313: create effective pipeline fails if passing in DOCKER_REGISTRY
if we pass in the env var DOCKER_REGISTRY to creating the effective pipeline we've seen failures where the pod gets created with an env var of the form:

```yaml
    - name: DOCKER_REGISTRY
      value: gcr.io
      valueFrom:
        configMapKeyRef:
          key: docker.registry
          name: jenkins-x-docker-registry
```

which is invalid; either `value` or `valueFrom` should be used.

It looks like this is coming from logic of taking the `DOCKER_REGISTRY` from the pod templates and not detecting its being passed in to the meta pipeline client as a custom environment variable so should be discarded from the build pack. i.e. it should just generate this...

```yaml
    - name: DOCKER_REGISTRY
      value: gcr.io
```

jstrachan:
if we pass in custom env vars when creating an effective pipeline we should remove any inherited env vars of the same name from the build pack to avoid creating invalid combined env var definitions in the pods which tekton + k8s cannot instantiate

fixes #6313

### DIFF
--- a/pkg/cmd/step/syntax/step_syntax_effective.go
+++ b/pkg/cmd/step/syntax/step_syntax_effective.go
@@ -420,6 +420,21 @@ func (o *StepSyntaxEffectiveOptions) createPipelineForKind(kind string, lifecycl
 		if err != nil {
 			return nil, errors.Wrapf(err, "Failed to generate pipeline from build pack")
 		}
+
+		// lets remove any environment variables defined in the container options from build packs/pod templates
+		// which are explicitly passed in to avoid possibly adding a value and valueFrom expression in the
+		// generated pod's container env vars
+		co := parsed.Options.ContainerOptions
+		for _, ev := range pipelineConfig.Env {
+			name := ev.Name
+			for i, e := range co.Env {
+				if e.Name == name {
+					co.Env = append(co.Env[0:i], co.Env[i+1:]...)
+					break
+				}
+			}
+		}
+		co.Env = RemoveDuplicateEnvVars(co.Env)
 	}
 
 	parsed.AddContainerEnvVarsToPipeline(pipelineConfig.Env)
@@ -449,6 +464,25 @@ func (o *StepSyntaxEffectiveOptions) createPipelineForKind(kind string, lifecycl
 	}
 
 	return parsed, nil
+}
+
+func RemoveDuplicateEnvVars(env []corev1.EnvVar) []corev1.EnvVar {
+	counts := map[string]int{}
+	for _, e := range env {
+		counts[e.Name] = counts[e.Name] + 1
+	}
+	for name, c := range counts {
+		for i := c - 1; i > 0; i-- {
+			// lets remove the dupe
+			for i, e := range env {
+				if e.Name == name {
+					env = append(env[0:i], env[i+1:]...)
+					break
+				}
+			}
+		}
+	}
+	return env
 }
 
 func (o *StepSyntaxEffectiveOptions) combineEnvVars(projectConfig *jenkinsfile.PipelineConfig) error {
--- a/pkg/cmd/step/syntax/step_syntax_effective.go
+++ b/pkg/cmd/step/syntax/step_syntax_effective.go
@@ -466,6 +466,7 @@ func (o *StepSyntaxEffectiveOptions) createPipelineForKind(kind string, lifecycl
 	return parsed, nil
 }
 
+// RemoveDuplicateEnvVars removes any duplicate env var definitions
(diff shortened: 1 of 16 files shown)


pkg/cmd/step/syntax/step_syntax_effective.go
```diff
@@ -420,6 +420,21 @@ func (o *StepSyntaxEffectiveOptions) createPipelineForKind(kind string, lifecycl
...
 		}
+
+		// lets remove any environment variables defined in the container options from build packs/pod templates
+		// which are explicitly passed in to avoid possibly adding a value and valueFrom expression in the
+		// generated pod's container env vars
+		co := parsed.Options.ContainerOptions
+		for _, ev := range pipelineConfig.Env {
+			name := ev.Name
+			for i, e := range co.Env {
+				if e.Name == name {
```
hferentschik: So this is the actual removal from container options, right?

pkg/cmd/step/syntax/step_syntax_effective.go
```diff
@@ -420,6 +420,21 @@ func (o *StepSyntaxEffectiveOptions) createPipelineForKind(kind string, lifecycl
...
+					co.Env = append(co.Env[0:i], co.Env[i+1:]...)
+					break
+				}
+			}
+		}
+		co.Env = RemoveDuplicateEnvVars(co.Env)
```
hferentschik: Is this an additional step? This is has nothing to do with the removal between duplicated in container options and pipeline env, right?

pkg/cmd/step/syntax/step_syntax_effective.go
```diff
@@ -451,6 +466,26 @@ func (o *StepSyntaxEffectiveOptions) createPipelineForKind(kind string, lifecycl
 	return parsed, nil
 }
 
+// RemoveDuplicateEnvVars removes any duplicate env var definitions
+func RemoveDuplicateEnvVars(env []corev1.EnvVar) []corev1.EnvVar {
+	counts := map[string]int{}
+	for _, e := range env {
+		counts[e.Name] = counts[e.Name] + 1
+	}
+	for name, c := range counts {
```
hferentschik: We don't care which env variable is left in the end as long as there are no duplicated?

pkg/cmd/step/syntax/step_syntax_effective.go
abayer: Yeah, I'm very confused at this point. =)

hferentschik: Still not sure what the intend of ` RemoveDuplicateEnvVars` is as that stage.

abayer:
@jstrachan - so lemme try to make sure I understand what this is doing - it's meant to have anything specified in `pipelineConfig.Env` override any env vars specified in pod templates or the build pack? It looks to me like it just gets env vars from pod templates, since it's just looking at the container options, not the stage/step env. Is that right?

abayer:
You might want to take a look at `combineEnvVars` in `step_syntax_effective` - it's currently specific to using `o.CustomEnvs` to override `projectConfig.Env`, but its cousin `scopedEnv` in `tekton/syntax/pipeline.go` could get moved to utils somewhere and be used as a general purpose "take this first list of env vars and this existing list of env vars and add the first to the existing, overriding anything that was there" function.

abayer:
(and yeah, any `Env` other than in `ContainerOptions` will be properly overridden at time of translation to Tekton CRDs - `ContainerOptions` is the only place where that doesn't already happen)

pkg/cmd/step/syntax/step_syntax_effective.go
jstrachan: fwiw the existing test cases were generating duplicate name + values which is why I added this code. I could always just leave the duplicates there if you'd rather?

pkg/cmd/step/syntax/step_syntax_effective.go
jstrachan: I'm confused too - removing duplicate values resulted in duplicates in the test case

jstrachan:
@abayer @hferentschik thanks for the feedback - am gonna try a much simpler + way less confusing fix instead https://github.com/jenkins-x/jx/pull/6316 gonna close this one for now

hferentschik:
> thanks for the feedback - am gonna try a much simpler + way less confusing fix instead #6316 gonna close this one for now

awesome thanks

PR #6316, opened 2019-12-10
fix: override env vars from build pack if passed in

### LINKED ISSUE #6313: create effective pipeline fails if passing in DOCKER_REGISTRY

jstrachan:
if we pass in custom env vars when creating an effective pipeline we should override any inherited env vars of the same name from the build pack to avoid creating invalid combined env var definitions in the pods which tekton + k8s cannot instantiate

fixes #6313

### DIFF
--- a/pkg/cmd/step/syntax/step_syntax_effective.go
+++ b/pkg/cmd/step/syntax/step_syntax_effective.go
@@ -449,12 +449,8 @@ func (o *StepSyntaxEffectiveOptions) createPipelineForKind(kind string, lifecycl
 	}
 
 	// lets override any container options env vars from any custom injected env vars from the metapipeline client
-	for _, e := range pipelineConfig.Env {
-		for i, e2 := range parsed.Options.ContainerOptions.Env {
-			if e.Name == e2.Name {
-				parsed.Options.ContainerOptions.Env[i] = e
-			}
-		}
+	if parsed != nil && parsed.Options != nil && parsed.Options.ContainerOptions != nil {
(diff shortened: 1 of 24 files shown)


pkg/cmd/step/syntax/step_syntax_effective.go
```diff
@@ -448,6 +448,14 @@ func (o *StepSyntaxEffectiveOptions) createPipelineForKind(kind string, lifecycl
 		return nil, errors.Wrapf(validateErr, "validation failed for Pipeline")
 	}
 
+	// lets override any container options env vars from any custom injected env vars from the metapipeline client
```
abayer: I'd rather either move `scopedEnv` out of `pkg/tekton/syntax/pipeline.go` or just make it publicly accessible, rather than having similarish behavior here.

jstrachan: sorted - just need to make sure I can get all the tests passing with the reordered env vars

Issue #5727, opened 2019-10-07
users.jenkins.io "<user>" already exists


hferentschik:
### Summary 

Creating a preview can fail with:

```
users.jenkins.io "foo" already exists
```

This happens when there is already a User CRD with the name `foo` and `GitUserResolver.Resolve` fails tries to create a new User CRD with the same name.

hferentschik:
The problem is that GitHub user names are case sensitive, but our CRD is  lowercased, hence this code to generate a unique id does not work:

```go
		new := r.GitUserToUser(gitUser)
		id = gitUser.Login
		// Check if the user id is available, if not append "-<n>" where <n> is some integer
		for i := 0; true; i++ {
			_, err := r.JXClient.JenkinsV1().Users(r.Namespace).Get(id, v1.GetOptions{})
			if k8serrors.IsNotFound(err) {
				break
			}
			id = fmt.Sprintf("%s-%d", gitUser.Login, i)
		}
		new.Name = naming.ToValidName(id)
		return id, possibles, new, nil
```

hferentschik:
It might be as easy as moving `naming.ToValidName(id)` higher up.

khansaAmrouni:
Hi! I am an outreachy applicant, can i take this issue to be my first contribution and work on ?

khansaAmrouni:
or can we use strings.ToLower() and render github username lower case !!

mrageh:
@khansaAmrouni are you still working on this issue? If not I'd be happy to jump on it

khansaAmrouni:
@mrageh Yes ! Go a head 
good luck 👍

PR #6323, opened 2019-12-10
fix: matching jenkins user to git user by name

mrageh:
#### Submitter checklist

#### Description
Git user names are case sensitive whereas JX usernames are all case insensitive. Within our app we lowercase the Git username and sometimes this causes issues if the Git user name contains uppercase characters. 

fixes: #5727


hferentschik:
@mrageh, you will need to sign-off the commit in order to pass the DCO check. 

See:

pkg/users/git.go
```diff
@@ -85,6 +85,8 @@ func (r *GitUserResolver) Resolve(user *gits.GitUser) (*jenkinsv1.User, error) {
 				}
 			}
 		}
+
+		gitUser.Login = naming.ToValidName(gitUser.Login)
```
hferentschik: why the assignment to `gitUser.Login` and then a bit later to `id`? Is that not a bit confusing? Would it not work by just calling:
```
id = naming.ToValidName(gitUser.Login)
```
The is no need to change the `Login` member of `gitUser`, is it?

pkg/users/git_test.go
```diff
@@ -53,6 +53,27 @@ func TestFindUserByLabel(t *testing.T) {
...
 
+func TestFindUserByUpperCaseGitUserName(t *testing.T) {
+	t.Parallel()
+	resolver, _, err := prepare(t)
+	assert.NoError(t, err)
+	assert.NoError(t, err)
+	gituserID := "foo"
+
+	gitUser := gits.GitUser{
+		Login: gituserID,
```
hferentschik: any reason you are not using "foo" directly? For symmetry reason and to make it easier to understand the test, I think it makes sense to put in "foo" directly here.

pkg/users/git_test.go
```diff
@@ -53,6 +53,27 @@ func TestFindUserByLabel(t *testing.T) {
...
+	}
+
+	user, err := resolver.Resolve(&gitUser)
+	assert.NoError(t, err)
+	assert.NotNil(t, user)
+
+	gitUser.Login = "Foo"
```
hferentschik: Instead of just updating `Login` I'd create a new instance. Mainly for symmetry. At least for me, it gets much easier to grasp what's going on.

pkg/users/git.go
mrageh: The code here is very confusing you're right, sometimes we use the `gitUser.Login` and sometimes the `id`. But we always want the `gitUser.Login` to be lowercased before we try to create the CRD. That happens over here and we pass the `gitUser.Login` there. I've refactored the code but I'm not happy with the way things look atm. 

Let me know what you think :)

pkg/users/git.go
```diff
@@ -60,6 +60,9 @@ func (r *GitUserResolver) GitUserSliceAsUserDetailsSlice(users []gits.GitUser) (
 // * making a call to the gitProvider
 // as often user info is not complete in a git response
 func (r *GitUserResolver) Resolve(user *gits.GitUser) (*jenkinsv1.User, error) {
+	user.Login = naming.ToValidValue(user.Login)
+	user.Login = naming.ToValidName(user.Login)
```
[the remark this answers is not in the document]
hferentschik: Also, this change breaks an existing test in git_test.go. Not sure whether this is the right way to do it.
You also risk a panic now. The _nil_ check for user comes below this assignment, so for a _nil_ user you will get a panic.

pkg/users/git_test.go
```diff
@@ -53,6 +53,30 @@ func TestFindUserByLabel(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.NoError(t, err)
+
+	gitUser := gits.GitUser{
+		Login: "foo",
+	}
+
+	user, err := resolver.Resolve(&gitUser)
+	assert.NoError(t, err)
+	assert.NotNil(t, user)
```
hferentschik: what's about adding some assertions for the actual expected state of the user on top of just checking that it is not _nil_?

pkg/users/git.go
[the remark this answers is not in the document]
mrageh: No those are two different things we do the user login, that's always needed. I think if we first sanitized the users login, we could then use it below as we please.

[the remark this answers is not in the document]
hferentschik: Hmm, it still breaks something though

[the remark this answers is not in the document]
hferentschik: I see now that these re two different calls. My mistake. Looks very similar :-(. The panic issue remains and I have not checked whether this is conceptually right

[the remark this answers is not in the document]
hferentschik: actually this is not at all what you want. You pulled the modification which otherwise happens in the _selectUsers_ function into Resolve as well, changing the behaviour completely.

_selectUsers_ is a function declaration and the function gets passed as parameter. Don't ask me why. IMO it does not make any sense at all at this point to use function pointers and it just obfuscates the code.

hferentschik:
@mrageh, you will need to run `make fmt` to make sure the formatting is correct.

mrageh:
Rewrote some of the code, right now it's broken I've got a bit to go to make it work. But this is just a sketch to get some feedback about how to proceed

pkg/users/git.go
```diff
@@ -63,8 +59,7 @@ func (r *GitUserResolver) Resolve(user *gits.GitUser) (*jenkinsv1.User, error) {
 	if r == nil || user == nil {
 		return nil, nil
 	}
-	selectUsers := func(id string, users []jenkinsv1.User) (string, []jenkinsv1.User,
-		*jenkinsv1.User, error) {
+	selectUsers := func(user *gits.GitUser) (*jenkinsv1.User, error) {
```
hferentschik: so this is still an intermediate step, right? The aim is still to get rid of passing the function?

mrageh: Done in the WIP commit below

mrageh: I need to refactor some of the tests now, but it kind of works

hferentschik:
Nicely crafted commit messages :-)

hferentschik:
@mrageh - can you run `make fmt` to make sure the format is right

pkg/users/git.go
```diff
@@ -63,43 +59,38 @@ func (r *GitUserResolver) Resolve(user *gits.GitUser) (*jenkinsv1.User, error) {
 	if r == nil || user == nil {
 		return nil, nil
 	}
-	selectUsers := func(id string, users []jenkinsv1.User) (string, []jenkinsv1.User,
```
hferentschik: 👍 for getting rid of this

mrageh:
@hferentschik thanks for the feedback, the only failing test I see locally is 

```
--- FAIL: TestFindUserByAccountReference (0.00s)
    git_test.go:111: 
        	Error Trace:	git_test.go:111
        	Error:      	Not equal: 
        	            	expected: "2aeb00ea-eb74-4554-9e59-2a11aa6ecfd5"
        	            	actual  : "bb3df-72e7-4c20-9cce-8248aaa91d04"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-2aeb00ea-eb74-4554-9e59-2a11aa6ecfd5
        	            	+bb3df-72e7-4c20-9cce-8248aaa91d04
        	Test:       	TestFindUserByAccountReference
    git_test.go:115: 
        	Error Trace:	git_test.go:115
        	Error:      	Not equal: 
        	            	expected: "001bb3df-72e7-4c20-9cce-8248aaa91d04"
        	            	actual  : "bb3df-72e7-4c20-9cce-8248aaa91d04"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-001bb3df-72e7-4c20-9cce-8248aaa91d04
        	            	+bb3df-72e7-4c20-9cce-8248aaa91d04
        	Test:       	TestFindUserByAccountReference


Expected :001bb3df-72e7-4c20-9cce-8248aaa91d04
Actual   :bb3df-72e7-4c20-9cce-8248aaa91d04
```

mrageh:
I reran the tests and now two tests are failing, I think I've introduced some kind of bug by making all these changes. Not sure on how to move forward now

hferentschik:
> I reran the tests and now two tests are failing, I think I've introduced some kind of bug by making all these changes. Not sure on how to move forward now

@mrageh, there were still a couple of issues with the code. I tried to rectify them in #6365. Have a look and let me know what you think.

hferentschik:
/ hold

We might want to close this PR in favour of #6365

hferentschik:
Closing this PR in favour of #6365

PR #6328, opened 2019-12-11
feat: format the imports when executing the fmt goal

cagiti:


hferentschik:
@cagiti you want to be careful here. `go fmt` will always work, since it is part of the built-in toolchain. `goimports` is not and we really should make sure it gets installed when required. This is not the case atm.

hferentschik:
@cagiti any chance you could address this as well?

PR #6332, opened 2019-12-11
fix: upgrades using branch ref

macox:

#### Submitter checklist

#### Description

When the UpgradeVersionStreamRef is set to a branch the upgradeRef was being set to a tagged release, rather than looking up the commit.

This PR removes some of the recently added code which was trying to determine whether we should upgrade to a tagged release or a SHA. 

The `LatestRelease` will now be required to upgrade to the latest tagged release


hferentschik:
An issue would be nice as well. Maybe next time? ;-)

PR #6333, opened 2019-12-11
feat: make it easier for `jx step pr comment` to be run in a pipeline

garethjevans:
#### Submitter checklist

fixes #

### DIFF
--- a/pkg/cmd/step/pr/step_pr_comment.go
+++ b/pkg/cmd/step/pr/step_pr_comment.go
@@ -1,6 +1,8 @@
 package pr
 
 import (
+	"os"
+
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
 	"github.com/jenkins-x/jx/pkg/cmd/opts/step"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


pkg/cmd/step/pr/step_pr_comment.go
```diff
@@ -59,15 +61,27 @@ func NewCmdStepPRComment(commonOpts *opts.CommonOptions) *cobra.Command {
 
 // Run implements this command
 func (o *StepPRCommentOptions) Run() error {
+	if o.Flags.PR == "" {
+		o.Flags.PR = os.Getenv("PULL_NUMBER")
```
hferentschik: We seem to be more and more setting options via environment variables. Instead of doing this ad hoc I am wondering whether a more generic approach makes more sense using, for example, Cobra's ability to automatically bind environment variables to options. The benefit would be a more consistent approach to option handling.

hferentschik: @garethjevans ^^

PR #6339, opened 2019-12-12
chore: allow escaping for code comments on PRs

garethjevans:
--code can be used to turn on escaping

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/cmd/step/pr/step_pr_comment.go
+++ b/pkg/cmd/step/pr/step_pr_comment.go
@@ -28,6 +28,7 @@ type StepPRCommentFlags struct {
 	Owner      string
 	Repository string
 	PR         string
+	Code       bool
 }
 
 // NewCmdStepPRComment Steps a command object for the "step pr comment" command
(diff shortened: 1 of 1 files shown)


pkg/cmd/step/pr/step_pr_comment.go
```diff
@@ -120,5 +122,12 @@ func (o *StepPRCommentOptions) Run() error {
...
 	}
 
+	if o.Flags.Code {
+		return provider.AddPRComment(&pr, escapeAsCode(o.Flags.Comment))
+	}
 	return provider.AddPRComment(&pr, o.Flags.Comment)
 }
+
+func escapeAsCode(comment string) string {
+	return "```\n" + comment
```
hferentschik: what's about the closing backticks?

garethjevans: you don't seem to need to close them on github, it works without it.

hferentschik: it works without on GitHub, but that does not mean that it is the correct syntax. GitHub is just more lenient. It might not work in other cases. Also, it just works if you treat the whole comment as code.

Or image a tool which tries to download the comments and parse the markdown. It might not be as lenient either.

There is a difference between "it works" and doing the right thing.

hferentschik: 👎

PR #6345, opened 2019-12-13
fix: Allow all crds with hooks to be processed

cagiti:
#### Submitter checklist
- [x] Change is code complete and matches issue description.
- [x] Change is covered by existing or new tests.

#### Description
This change allows the helm apply function to iterate and apply all crd part files

#### Which issue this PR fixes
fixes #6344

PR #6351, opened 2019-12-16
chore: update golang image version to 1.12.4

cagiti:

Since our builder images are using golang 1.12.4, lets update the docker golang images  within our pipelines to 1.12.4 too

Contributes towards #6345

### DIFF
--- a/jenkins-x-bdd.yml
+++ b/jenkins-x-bdd.yml
@@ -47,12 +47,12 @@ pipelineConfig:
                     mountPath: /secrets
             steps:
               - name: build-binary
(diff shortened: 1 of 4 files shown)


abayer:
This PR has managed to fail on boot vault in four different flaky ways (first being the once in like 300 chance of the Tekton PipelineRun race condition, where the build still gets rerun, but the BDD tests expect only one activity and as a result there were two, second being a go mod fetching burp, and the last two being fragility in the UI app test showing up in two different places). So if it fails again, I’m just gonna override it. =)

PR #6357, opened 2019-12-18
feat: podAnnotations on jx pod

msvticket:
#### Submitter checklist

#### Description

Adds support for setting podAnnotations in jx chart.

The reason I needed this was that gcactivities stopped working when I enabled istio for the namespace. With the annotation sidecar.istio.io/inject: "false" on the pod it would have worked. But now there is no way to add this annotation via the chart.

### DIFF
--- a/charts/jx/templates/cronjob.yaml
+++ b/charts/jx/templates/cronjob.yaml
@@ -20,6 +20,10 @@ spec:
           labels:
             app: {{ template "jx.name" . }}
             release: {{ .Release.Name }}
+    {{- if .Values.podAnnotations }}
+          annotations:
+{{ toYaml .Values.podAnnotations | indent 12 }}
+    {{- end }}
         spec:
     {{- if .Values.restartPolicy }}
           restartPolicy: {{ .Values.restartPolicy }}
--- a/charts/jx/templates/daemonset.yaml
+++ b/charts/jx/templates/daemonset.yaml
@@ -20,6 +20,10 @@ spec:
       labels:
         app: {{ template "jx.name" . }}
         release: {{ .Release.Name }}
+{{- if .Values.podAnnotations }}
+      annotations:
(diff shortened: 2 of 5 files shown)


hferentschik:
@msvticket do you mind opening an issue as well and providing a bit more context?

msvticket:
> do you mind opening an issue as well and providing a bit more context?

What more context do you think is is needed?

hferentschik:
> What more context do you think is is needed?

An issue with the information in from this pull request would be a good start. 

I guess what's not clear to me is how the change in these templates files solves your problem. Some more details on how to reproduce the issue could be helpful. So you have a Jenkins X install and you want to enable istio for the jx namespace? Doing so breaks gcactivities? They don't run at all anymore or is there an error? 
So what you want to do is to run the gc pods w/o Istio sidecar, right? Are you saying they cannot be run with Istio or is just a way to work around the issue?

Last but not least, I am not familiar with how changes to these templates would propagate to the appropriate Helm charts for installing the gc jobs. Would have thought that is somewhere in the platform repo. But that's just me not fully following how this all is tied together (and hence my question for more context). 

@abayer, this might be more your cup of tea.

msvticket:
Having the the ability to setting annotations on pods created by a helm chart is a standard functionality. That I need it to work around a problem with istio is just an example of the use. I don't see how more details about that is really relevant to the basic feature the PR is adding.

hferentschik:
Ok, I was missing the part that the `jx` Chart is aliased and templated in the jenkins-x-platform repo - https://github.com/jenkins-x/jenkins-x-platform/blob/master/jenkins-x-platform/values.yaml#L25-L198

> Having the the ability to setting annotations on pods created by a helm chart is a standard functionality. 

Sure. And having this discussion helped (at least me) clarifying this.

Issue #6352, opened 2019-12-16
jx create cluster makes an incompatible cluster on AKS


Vaccano:
### Summary 
jx install fails on AKS with 

>Your current helm version v3 is not supported. Please downgrade to helm v2.

### Steps to reproduce the behavior
1. Browse to https://shell.azure.com/
2. Start an Azure Cloud Shell connected to an Azure account.
3. Run `jx version`.  If asked to upgrade select yes.
3a. If you upgraded, change directories to `/home/your_user_name_here/.jx/bin`
3b. Replace `jx` with `./jx` in all commands below.
3. Run `jx create cluster aks -c KubeJx2`
4. Complete the device login challenge
5. Wait for the cluster to be created.
6. When asked for "Jenkins installation type" select "Serverless Jenkins X Pipelines with Tekton"
RESULT: The following error is given:
> FATAL: Your current helm version v3 is not supported. Please downgrade to helm v2.

NOTE: I also tried these same steps, replacing step 3's command with `jx create cluster aks -c KubeJx -- helm3`.  The result was the same error requesting a downgrade to helm v2.

### Expected behavior
When running `jx create cluster aks` it will create a cluster that is compatible with Jenkins X.

### Actual behavior
`jx create cluster aks` creates a cluster that is incompatible with Jenkins X (uses a version of Helm that it does not support). 

### Jx version

The output of `jx version` is:

```
2.0.1082
```

### Jenkins type

- [X ] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

### Kubernetes cluster

AKS Cluster created using `jx create cluster`

### Operating system / Environment

Azure Cloud Shell

Vaccano:
Is there a work around to this issue?  

Some command line parameter that can be passed into jx to make it force AKS to use the older version of Helm maybe?

(I want to evaluate JenkinsX somehow, but can't make a testing system in Azure (where I have cloud computer time))

bitsofinfo:
same issue here

Labibme:
i have same issue with advanced version
[mini@vlmnk8 ~]$ jx add app jx-app-ui
ERROR: Error loading team settings. the server could not find the requested resource (get environments.jenkins.io)
WARNING: Failed to retrieve team settings: failed to setup the dev environment for namespace 'default': the server could not find the requested resource (post environments.jenkins.io) - falling back to default settings...
FATAL: Your current helm version v3 is not supported. Please downgrade to helm v2.
[...]

PR #6359, opened 2019-12-18
feat: Bump acceptable Helm version and installs

deanesmith:
#### Submitter checklist

#### Description
JX versions now will permit Helm versions < 2.16.2 (bumped from 2.15).  Dependency installs were always grabbing Helm3.  We'll lock in on this version until Helm3 is fully supported.

Helps with https://github.com/jenkins-x/jx/issues/6352, among others.


pkg/cmd/opts/install.go
```diff
@@ -328,7 +328,7 @@ func (o *CommonOptions) InstallHelm() error {
 	binary := "helm"
 
 	if runtime.GOOS == "darwin" && !o.NoBrew {
-		err := o.RunCommand("brew", "install", "kubernetes-helm")
+		err := o.RunCommand("brew", "install", "helm@2")
```
hferentschik: I am not quite sure this is going to cut it. At best this might be a solution for macOS. There is a good chance the author of #6352 is even on Linux which would mean this change is not even going to affect him (in the issue we see _/home/<user>_ opposed to _/Users/<user>_).

I think what needs to happen is:

- Remove the use of brew for installing dependencies (just jx itself is installed via Brew)
- Install dependencies in the desired version using the non-Brew approach for all OS types into ~/.jx/bin
- When calling these dependencies adjust the PATH to call our supported binary

hferentschik:
@deanesmith, thanks for the PR. You will need to sign-off the commit in order to pass the DCO check. 

See:

pkg/cmd/opts/install.go
```diff
@@ -344,10 +344,7 @@ func (o *CommonOptions) InstallHelm() error {
 	if err != nil || !flag {
 		return err
 	}
-	latestVersion, err := util.GetLatestVersionFromGitHub("kubernetes", "helm")
-	if err != nil {
-		return err
-	}
+	latestVersion := "v2.16.1"
```
hferentschik: Ahh, I see. So this is fixing the non-Brew part. I still argue that what I said above holds, but we can take this as a stepping stone. However, the version should not be hardcoded here, but be fetched from the version stream.

deanesmith: I can add an issue for that, as well as getting away from Brew.  These repairs should buy some time.

hferentschik: Can we not at least use the version stream. Hard coding the version here is really not a good idea, especially since the version just got upgraded in the version stream. 

So instead of getting the latest version from GitHub, which is really the crux here, we get the latest acceptable version from the version stream. You are in `CommonOptions`, so you have access to the version stream via `o.GetVersionResolver`. All you need to do is to look up the version.

hferentschik: I am fine with creating a new issue for removing Brew, since this affects really all dependencies we install and it is a bigger change. Loading the right version from the version stream, however, we should be doing here and now.

pmuir: Yeah, this needs to use the version stream

pkg/cmd/opts/install.go
deanesmith: Will create an issue for these recommendations since it increases scope and testing required.

pkg/cmd/opts/install.go
deanesmith: How can this be achieved?

deanesmith: Will do

hferentschik:
Closing this PR. Will be superseded by another one.

PR #6360, opened 2019-12-18
fix: enable GH user names with upper case letters

### LINKED ISSUE #5727: users.jenkins.io "<user>" already exists

MarckK:
#### Submitter checklist

- [x ] Change is code complete and matches issue description.

fixes #5727

### DIFF
--- a/pkg/users/git.go
+++ b/pkg/users/git.go
@@ -2,6 +2,7 @@ package users
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/jenkins-x/jx/pkg/kube/naming"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


abayer:
Could you add a test in `pkg/users/git_test.go` like `TestFindUserByFromGitProvider` to verify this does the right thing? Thanks!

hferentschik:
@MarckK this is the same issue @mrageh is working on in pull request #6323. You guys are working on the same thing :-)

@mrageh is also trying to refactor the code which I think is a good idea. We might want to synchronize what we want to do here.

pkg/users/git.go
```diff
@@ -86,14 +87,15 @@ func (r *GitUserResolver) Resolve(user *gits.GitUser) (*jenkinsv1.User, error) {
 			}
 		}
 		new := r.GitUserToUser(gitUser)
-		id = gitUser.Login
+		lowerCaseGitLogin := strings.ToLower(gitUser.Login)
```
hferentschik: I don't think this is going to cut it, at least not for all cases. Lower casing is part of the issue since we use this login as a Kuberenetes label value, but there are more restrictions which apply. ` naming.ToValidName(gitUser.Login)` probably works. @mrageh has done this already and he also had a test.

Right now @mrageh tries to further improve the code, since passing `selectUsers` as a function parameter is super confusing and fulfils no real purpose really.

hferentschik:
/hold 

@MarckK, @mrageh - lets sync what we want to achieve and how we want to progress.

MarckK:
Thank you @hferentschik @abayer for the feedback. I realised after making the PR that @mrageh had made a recent PR with a similar fix. Absolutely, we should go with his PR, and for me, it was good to see how he's written the test. :) 

Let the CI tests run last night as was curious to see if the change would break anything else. Apologies for the confusion. Will close this now.

hferentschik:
no problem

PR #6365, opened 2019-12-19
fix: making sure case sensitive GitHub user names are handled properly

### LINKED ISSUE #5727: users.jenkins.io "<user>" already exists

hferentschik:
fixes #5727

Taking @mrageh refactoring from pull request #6323 the last mile.


pkg/users/users.go
```diff
@@ -80,113 +83,121 @@ func DeleteUser(jxClient versioned.Interface, ns string, userName string) error
 	return err
 }
 
-// Resolve does the heavy lifting for user resolution.
-// This function is not normally called directly but by a dedicated user resolver (e.g. GitUserResolver)
-// * checking the user custom resources to see if the user is present there
-// * calling selectUsers to try to identify the user
-// as often user info is not complete in a git response
-func Resolve(id string, providerKey string, jxClient versioned.Interface,
```
pmuir: This is a public API in JX and cannot be changed in this way as it will break calling code. Please keep the API signature as it is and if you need to introduce a different signature add a new function.

The signature of this API is purposefully designed to not be tied to the `GitUser` interface so that other systems that handle users (e.g. Slack) can use it - and the slack bot does indeed use it.

pkg/users/users.go
```diff
@@ -80,113 +83,121 @@ func DeleteUser(jxClient versioned.Interface, ns string, userName string) error
...
-				found.Labels = make(map[string]string)
-			}
-			found.Labels[providerKey] = id
-			found, err := jxClient.JenkinsV1().Users(namespace).Update(found)
-			if err != nil {
-				return nil, err
-			}
-			log.Logger().Infof("Adding label %s=%s to user %s in users.jenkins.io", providerKey, id, found.Name)
-			return found, nil
+	matchedUser, err := matchGitUser(user, jxUsers, providerKey)
```
pmuir: See above, this code is intentionally not directly connected to Git.

hferentschik: one could return the list of users and let the caller do the matching. You also don't need to pass the full gits.GitUser, but just the id. But this would still require to adjust the Slack app.

pkg/users/users.go
hferentschik: I think there are several things wrong with the current code, but to move this forward I created #6366 with the minimal required changes.

I think we will have a very hard time to refactor and improve our code base if we treat every exported function as API. It is unfortunate, that we don't have a clear separation between internal code and actual code we want others (e.g. Apps) to use. 

Regarding the function specifically, I have a very hard time with its design. Both functions parameters and return values are very cryptic and passing the function pointer seem very specific.

hferentschik:
closed in favour of #6366

PR #6382, opened 2019-12-26
fix: Upgrade Tekton to 0.8.0, remove Knative Build

### LINKED ISSUE #4230: Remove Knative Build and JFR from Jenkins X installation
### Summary 

The knative build and JFR should be removed from Jenkins X installation, as soon as can be used as a buildpack in JenkinsX/Tekton pipeline. #4229 

The the moment this topology is using a old fork of prow which is unmaintained. 

### Jx version

```
COPY OUTPUT HERE
```

### Jenkins type

- [ ] Next Generation (Tekton + Prow)
- [ ] Classic Jenkins
- [x] Serverless Jenkins (JenkinsFileRunner + Prow)

### LINKED ISSUE #6317: Upgrade Tekton to 0.8.0
Note that this doesn't mean exposing any new functionality from Tekton 0.9.1, just upgrading to use Tekton 0.9.1 instead of our current 0.5.1

abayer:
#### Submitter checklist

#### Description

Note that the original plan was to upgrade to 0.9.x, but that turns
out to be...messy. Tekton 0.9.x depends on k8s 1.13.x at compile time,
which results in a bunch of mock/fake changes being needed on our
side. Going to 0.8.0, with its dependency on k8s 1.12.x, did require
some changes, but not as many.

And more importantly than changes here, going to k8s 1.13.x at compile
time results in breaking changes to compilation for Knative Build and
Serving, Cert Manager, and Bankvault. While some of these can be
solved by upgrading, Knative Build ended at 0.7.0 (which depends on
k8s 1.12.x, and which we're upgrading our dependency to here, even
though we also remove all the Knative Build usage here at the same
time), and even if we remove our `go.mod` dependency on Knative Build,
we then end up pulling in 0.2.0 thanks to our Prow
dependency. So...things need to be reworked in some janky ways to
upgrade all the way to 0.9.2.

This also removes Knative Build - the addon is left around for now,
    but any code actually using it is gone. go.mod still references it,
    because otherwise we'd pick up Knative Build 0.2.0 via depending on
    Prow, which would just mix things up more. I'm working on a Prow
    change to remove Knative Build there which will completely eradicate
    Knative Build.

fixes #6317 
fixes #4230


abayer:
So! Another "fun" problem. There are a number of backwards compatibility breaking changes between Tekton v0.5.1, which is what we're using now, and v0.8.0. I thought I was prepared for a lot of them, but I missed one big one - starting with v0.6.0, `PipelineRun` parameters (and `Task` parameters) are typed, either array or string. But that change was made without a deprecation period, so `PipelineRun`s and `Task`s that are valid for v0.5.1 just aren't valid at all for v0.6.0. The other backwards incompatible changes between v0.5.1 and v0.8.0 are properly handled via deprecation, so, parameters aside, we could in theory upgrade release-by-release gradually, but that isn't even an option at all due to the parameters thing.

I'm currently leaning towards doing a forked v0.8.0 release with the relevant backwards incompatible changes switched back to being deprecated, so that CRDs generated with a jx using Tekton v0.5.1 are still valid. I _think_ the only ones we care about are the parameters one and the switch from `${inputs.params.version}` to `$(inputs.params.version)`, though I haven't yet gotten far enough to verify exactly how a change to input/output behavior works with our pipelines.

abayer:
Also just found a failure on quickstart/import claiming the pipeline didn't get started when, in fact, it did, so trying to figure that out now as well.

abayer:
The `start_pipeline.go`-induced failure was because I deleted too much stuff from that command. Whoops. =)

abayer:
Ok, input/output copying isn't working right, so I'll dig into that.

hferentschik:
>  I'm working on a Prow change to remove Knative Build there which will completely eradicate
Knative Build.

+1 - I take it you are talking about our Prow fork? I am wondering whether it makes sense to start there? First update Prow and get that done which will mean we can properly remove KNative as part of this PR as well. WDYT?

abayer:
@hferentschik See all the insanity in #6413 - there are so many moving pieces involved that I’d say even just Prow alone is part of the follow up.

abayer:
Ignore that `boot-vault` failure - previous run, due to the new run's cluster causing the old one to get deleted.

PR #6393, opened 2019-12-30
fix: GC ProwJobs in gc_activities

### LINKED ISSUE #6094: gcactivities cronjob should also clean up prow jobs
Probably at least as aggressively as it cleans up `PipelineActivity`s, definitely not as aggressively as it cleans up `PipelineRun`s. Maybe we should be adding `prowJobName` to `PipelineActivity` labels, and whenever we're GCing a `PipelineActivity`, we GC its corresponding `ProwJob` as well? On prod, that'd still leave us with ~2500 or so `ProwJob`s floating around much of the time, but that's a lot better than "just grow forever" like we have now. There's probably a simpler way we could do it as well - just always deleting any `ProwJob` older than 3 days or something like that.

abayer:
#### Submitter checklist

#### Description

We're already GCing `PipelineActivity` and `PipelineRun` directly,
which also results in the downstream CRDs
`Pipeline`/`Task`/`TaskRun`/`PipelineStructure` all getting GCed too,
but until now, we're not doing anything about `ProwJob`s, so they can
get out of control and cause performance and OOM issues on the various
Prow pods. So let's GC `ProwJob`s too.

The default is to keep a week of `ProwJob`s - even on prod, I haven't
seen issues caused by excesses of `ProwJob`s until it's been at least
two weeks of regular rates of creation, so a week seems fine.

fixes #6094

### DIFF
--- a/pkg/cmd/gc/gc_activities.go
+++ b/pkg/cmd/gc/gc_activities.go
@@ -8,6 +8,8 @@ import (
 	gojenkins "github.com/jenkins-x/golang-jenkins"
 	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
+	"github.com/pkg/errors"
+	prowjobv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 
 	jv1 "github.com/jenkins-x/jx/pkg/client/clientset/versioned/typed/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/util"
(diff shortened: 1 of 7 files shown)


pkg/cmd/gc/gc_activities.go
```diff
@@ -267,6 +277,45 @@ func (o *GCActivitiesOptions) deletePipelineRun(pipelineRunInterface tektonclien
...
+	now := time.Now()
+
+	for _, pj := range pjList.Items {
+		completionTime := pj.Status.CompletionTime
+		if completionTime != nil && completionTime.Add(o.ProwJobAgeLimit).Before(now) {
+			err = o.deleteProwJob(pjInterface, &pj)
+			if err != nil {
+				return err
+			}
+			continue
```
ccojocar: nit: this `continue` statement is not required

pkg/cmd/gc/gc_activities.go
```diff
@@ -267,6 +277,45 @@ func (o *GCActivitiesOptions) deletePipelineRun(pipelineRunInterface tektonclien
...
+		return nil
+	}
+
+	now := time.Now()
+
+	for _, pj := range pjList.Items {
+		completionTime := pj.Status.CompletionTime
+		if completionTime != nil && completionTime.Add(o.ProwJobAgeLimit).Before(now) {
+			err = o.deleteProwJob(pjInterface, &pj)
+			if err != nil {
```
ccojocar: It would be nice to wrap these errors with some context.

abayer:
@ccojocar - updated!

Issue #6110, opened 2019-11-13
jx create cluster eks fails with 'ERROR: error installing helm: Unable to download'


pguimaraes:
### Summary

Jenkins X version 2.0.979 cannot download helm.

### Steps to reproduce the behavior

jx --version
2.0.979

jx create cluster eks --skip-installation=true --install-dependencies=true --eksctl-log-level=4 --cluster-name="XXX" --region="us-east-1" --zones="us-east-1a,us-east-1b" --node-type="m5.large" --nodes=2 --nodes-min=2 --nodes-max=4 --tags="CreatedBy=JenkinsX" --username="XXX" --verbose=true --batch-mode=true

### Expected behavior

Successful cluster creation.

### Actual behavior

DEBUG: Dependencies to be installed: eksctl, aws-iam-authenticator
Installing eksctl
Downloading https://github.com/weaveworks/eksctl/releases/download/latest_release/eksctl_Linux_amd64.tar.gz to /home/XXX/.jx/bin/eksctl.tar.gz...
Downloaded /home/XXX/.jx/bin/eksctl.tar.gz
Installing aws-iam-authenticator
Downloading https://amazon-eks.s3-us-west-2.amazonaws.com/1.12.7/2019-03-27/bin/linux/amd64/aws-iam-authenticator to /home/XXX/.jx/bin/aws-iam-authenticator...
Downloaded /home/XXX/.jx/bin/aws-iam-authenticator
Installing kubectl
Downloading https://storage.googleapis.com/kubernetes-release/release/v1.13.2/bin/linux/amd64/kubectl to /home/XXX/.jx/bin/kubectl...
Downloaded /home/XXX/.jx/bin/kubectl
Installing helm
Downloading https://storage.googleapis.com/kubernetes-helm/helm-v3.0.0-linux-amd64.tar.gz to /home/XXX/.jx/bin/helm.tgz...
ERROR: error installing helm: Unable to download file /home/XXX/.jx/bin/helm.tgz from https://storage.googleapis.com/kubernetes-helm/helm-v3.0.0-linux-amd64.tar.gz due to: download of https://storage.googleapis.com/kubernetes-helm/helm-v3.0.0-linux-amd64.tar.gz failed with return code 404

Please fix the error or install manually then try again

### Jx version

The output of `jx version` is:

```
$ jx version
error: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
$
[...]

PR #6397, opened 2019-12-31
feat: Updated Helm install URL

deanesmith:
#### Submitter checklist

#### Description
Not only is there now Helm3 (which isn't yet supported by jx) but the binary download locations have changed as per https://helm.sh/blog/get-helm-sh/#where-are-we-now.  

https://github.com/jenkins-x/jx/issues/6110

### DIFF
--- a/pkg/cmd/opts/install.go
+++ b/pkg/cmd/opts/install.go
@@ -355,7 +355,7 @@ func (o *CommonOptions) InstallHelm() error {
 		return err
 	}
 
-	clientURL := fmt.Sprintf("https://storage.googleapis.com/kubernetes-helm/helm-v%s-%s-%s.tar.gz", stableVersion, runtime.GOOS, runtime.GOARCH)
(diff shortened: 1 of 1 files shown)


pkg/cmd/opts/install.go
```diff
@@ -355,7 +355,7 @@ func (o *CommonOptions) InstallHelm() error {
 		return err
 	}
 
-	clientURL := fmt.Sprintf("https://storage.googleapis.com/kubernetes-helm/helm-v%s-%s-%s.tar.gz", stableVersion, runtime.GOOS, runtime.GOARCH)
+	clientURL := fmt.Sprintf("https://get.helm.sh/helm-v%s-%s-%s.tar.gz", stableVersion, runtime.GOOS, runtime.GOARCH)
```
ccojocar: There is a second URL above in the `InstallTiller` command which also needs to be updated. It is probably not very relevant but it is good to be consistent.

deanesmith: Submitting a new PR

PR #6400, opened 2020-01-02
[wip] fix: remove functionality to install jx on cluster with create cluste…

### LINKED ISSUE #5537: 'jx create cluster' should not install Jenkins X
### Summary 

At the moment `jx create cluster` creates a cluster and installs directly Jenkins X. For me, this is a bit unexpected, especially since there is a dedicated `jx install` which is now superseded` by `jx boot`.

I'd rather have to type in two clear commands:

```bash
> jx create ...
> jx install ...
```

Also, with `jx boot` you need to explicitly skip the installation if you want to use boot over install:

```
> jx create cluster --skip-installation ...
```

My understanding is that moving forward `jx install` gets deprecated and `jx boot` becomes the default. With `jx boot` it should also be possible to just bring your own cluster. So `jx create cluster` becomes more of a convenient method.

I recommend removing the direct Jenkins X installation behaviour from `jx create cluster`. Nothing wrong imo to type in two commands if needed.

MarckK:
…r command

Signed-off-by: Kara de la Marck <MarckK@users.noreply.github.com>

#### Submitter checklist

#### Description

 `jx create cluster` creates a cluster and installs Jenkins X by default. 
It is recommended that installation of Jenkins X be done by `jx boot`, therefore this default functionality should be taken out.

fixes #5537

### DIFF
--- a/jx/bdd/static/ci.sh
+++ b/jx/bdd/static/ci.sh
@@ -51,7 +51,7 @@ cd boot-source
 
 # lets trigger the BDD tests in a clusterand git provider
 jx step bdd \
-  --config jx/bdd/static/cluster.yaml \
+  --config ../jx/bdd/static/cluster.yaml \
   --versions-repo https://github.com/jenkins-x/jenkins-x-versions.git \
   --provider=gke \
   --git-provider=ghe \
--- a/jx/bdd/static/ci.sh
+++ b/jx/bdd/static/ci.sh
@@ -46,8 +46,11 @@ git config --global --add user.email jenkins-x@googlegroups.com
 # Disable checking that the PipelineActivity has been updated by the build controller properly, since we're not using the build controller with static masters.
 export BDD_DISABLE_PIPELINEACTIVITY_CHECK=true
 
+git clone https://github.com/jenkins-x/jenkins-x-boot-config.git boot-source
+cd boot-source
+
 # lets trigger the BDD tests in a clusterand git provider
-jx step bdd -b \
+jx step bdd \
   --config jx/bdd/static/cluster.yaml \
   --versions-repo https://github.com/jenkins-x/jenkins-x-versions.git \
   --provider=gke \
--- a/jx/bdd/static/jx-requirements.yml
+++ b/jx/bdd/static/jx-requirements.yml
@@ -1,0 +1,46 @@
+gitops: true
+cluster:
+  clusterName: bdd-boot-lh-ghe
+  environmentGitOwner: dev1
+  gitKind: github
+  gitName: ghe
+  gitServer: https://github.beescloud.com
+  project: jenkins-x-bdd3
+  provider: gke
+  zone: europe-west1-c
+environments:
+  - key: dev
+    owner: ""
+    repository: ""
+  - key: staging
+    owner: ""
+    repository: ""
+  - key: production
+    owner: ""
+    repository: ""
+ingress:
+  domain: ""
+  externalDNS: false
+  tls:
+    email: ""
+    enabled: false
+    production: false
+kaniko: true
+secretStorage: vault
+storage:
+  logs:
+    enabled: false
+    url: ""
+  reports:
+    enabled: false
+    url: ""
+  repository:
+    enabled: false
+    url: ""
+webhook: lighthouse
+vault:
+  disableURLDiscovery: true
+versionStream:
+  ref: "master"
+  url: https://github.com/jenkins-x/jenkins-x-versions.git
+
--- a/pkg/cmd/create/create_cluster.go
+++ b/pkg/cmd/create/create_cluster.go
@@ -6,12 +6,9 @@ import (
 	"github.com/jenkins-x/jx/pkg/cmd/create/options"
 
 	"github.com/jenkins-x/jx/pkg/cmd/initcmd"
-	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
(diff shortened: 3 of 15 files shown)


pkg/cmd/create/create_cluster.go
```diff
@@ -23,10 +19,9 @@ type KubernetesProvider string
 // CreateClusterOptions the flags for running create cluster
 type CreateClusterOptions struct {
 	options.CreateOptions
-	InstallOptions   InstallOptions
-	Flags            initcmd.InitFlags
-	Provider         string
-	SkipInstallation bool `mapstructure:"skip-installation"`
+	InstallOptions InstallOptions
```
ccojocar: I don't think there is a reason to keep the other 2 flags (InstallOptions, init Flags) since we are removing the ability to call install from the `create cluster` command. I would also remove them.

pkg/cmd/create/create_cluster_eks.go
```diff
@@ -227,7 +227,5 @@ cluster provisioning. Cleaning up stack %s and recreating it with eksctl.`,
 	o.InstallOptions.SetInstallValues(map[string]string{
```
ccojocar: There is no reason to set the install values here since the install doesn't get call anymore after create cluster command.

pkg/cmd/create/create_cluster_gke.go
```diff
@@ -592,20 +592,13 @@ func (o *CreateClusterGKEOptions) createClusterGKE() error {
 		return err
 	}
 
-	log.Logger().Info("Initialising cluster ...")
-
 	o.InstallOptions.SetInstallValues(map[string]string{
```
ccojocar: ditto

ccojocar:
@MarckK Do you need any help to follow up on this?

pkg/cmd/create/create_cluster_gke.go
MarckK: OK. Also, removing in /create_cluster_aws.go

hferentschik:
@MarckK, how is it going with this PR?

MarckK:
Thanks, @hferentschik, I've gotten `make build` to work, and so it is now compiling. The e2e tests are now failing and I need some guidance on understanding the failures and how best to fix. Thanks!

hferentschik:
@MarckK - sorry missed your comment. We can see whether we can move this PR forward. I like the idea for sure ;-)

Triggering the tests again to get new logs. 

/test all

hferentschik:
@MarckK, sorry super long feedback loop on this one, but let's try to move forward on this. 
So the obvious reasons why your change is blowing up the BDD tests is that your change conflicts with how the BDD test clusters get created. 
Have a look at _jx/bdd_ directory, it contains the directories _boot-lh-ghe_, _boot-vault_, _static_ and _tekton_. These directories contain the setup scripts and configuration too boot the clusters for the BDD tests. 

If we start with _boot-lh-ghe_ and  _boot-vault_ and look at their _cluster.yaml_, we see:
```
clusters:
  - name: boot-lh-ghe
    args:
      - create
      - cluster
      - gke
      - --project-id=jenkins-x-bdd3
      - -m=n1-standard-2
      - --min-num-nodes=3
      - --max-num-nodes=5
      - -z=europe-west1-c
      - --skip-login
      - --skip-installation
    commands:
      - command: jx
        args:
          - boot
          - -b
```
Your pull request is removing the '--skip installation', so these two test clusters won't even come up, since the setup command will fail saying that the option _--skip-installation_ is missing. One could start by removing _--skip-installation_ in these two contexts and see how it goes.

In _static_ and _tekton_ we have the opposite problem. There the cluster is created by the single `jx create` call. In these cases, we need to split into two setup commands (one for create and one for install).

hferentschik:
Another approach we could take is to initially not remove the option, but just change its default. Let's make `--skip-installation` per default true. This means we get the desired behaviour, but can without actually removing anything. The adjustments needed in _ jx/bdd_ become trivial as well then. 

Either way, complete removal of the flag or switching of the default, we will need a walkthrough through the docs to make sure we still properly describe the create/install procedure.

hferentschik:
Seems this PR needs a rebase now.

jx/bdd/static/ci.sh
```diff
@@ -46,9 +46,12 @@ git config --global --add user.email jenkins-x@googlegroups.com
 # Disable checking that the PipelineActivity has been updated by the build controller properly, since we're not using the build controller with static masters.
 export BDD_DISABLE_PIPELINEACTIVITY_CHECK=true
 
+git clone https://github.com/jenkins-x/jenkins-x-boot-config.git boot-source
```
hferentschik: what is this about?

jx/bdd/static/jx-requirements.yml
```diff
@@ -0,0 +1,46 @@
+gitops: true
```
hferentschik: @MarckK, what are you trying to do? Making a "non Boot" cluster to use Boot?

MarckK: This was just me trying to push things around last weekend. Don't worry, I've already rolled this back locally. Am trying a new approach, will push up this weekend.

hferentschik: No problem at all :-) I was just wondering ;-)

hferentschik:
@MarckK, what is the state on this PR? I am wondering whether it makes to close this particular pull request and start over? WDYT?

hferentschik:
Closing this for now. You can always reopen when you have an update.

PR #6410, opened 2020-01-04
fix: updated comments and docs to use correct `.yml` extension

jstrachan:
for the `jx-requirements.yml` file name


abayer:
`boot-vault` looks like a flaky failure, but the lint failure is legit - openapi needs to be regenerated.

PR #6413, opened 2020-01-05
[wip] Experimenting with dependency updates to compile against Tekton 0.9.2

abayer:
Do not ever merge this - it'll get replaced by something real later, if ever.

This bumps a bunch of go.mod dependencies so that we can actually build against Tekton 0.9.2:
* tektoncd/pipeline -> v0.9.2
* jetstack/cert-manager -> v0.11.0
* banzaicloud/bank-vaults -> b327d7f2b681 (v0.7.0)
* k8s/test-infra -> jenkins-x/test-infra fork with Knative Build removed
* knative/pkg and knative/serving -> knative.dev/pkg and knative.dev/serving 0.10.1
* various k8s.io dependencies -> v1.15.3 - Tekton only needs 1.13.12, but Knative Serving needs 1.15.3, so...yeah.

And various transitive dependencies as a result. It's...ugly. 

There are a couple unit test failures - the one in `pkg/kube/pki TestWaitCertificateIssuedReady` I'm really not sure what the answer is, since the underlying behavior seems to have changed. And it explodes due to the prow fork I've got still building against an earlier k8s code gen, which...well, that's entirely on me, since I'm the one who added a client dependency on ProwJobs last week anyway. Sigh. Will figure that out.

I'm opening this now to see how many things explode horribly in the `boot-vault` test.


abayer:
`jx upgrade ingress` breaks with the cert-manager dependency bump, so we're probably going to want to wait on this until after we remove that command. For giggles, I'm experimenting with cert-manager 0.9.1 right now.

PR #6427, opened 2020-01-07
feat: pass tags and global chart variables when deploying an application

### LINKED ISSUE #6312: feat: Create helm tags and/or global variables for environments
Currently, in charts, the only way to know which environment is being deployed is to read `.Release.Namespace`. `jx-<environment>` for staging or production, and `jx-<ownwe>-<repo>-pr-<n>` for a preview environment. 

This is quite limited for the following reasons:
- It cannot be used in dependencies conditions
- Extracting the PR number is quite empirical, as different owners can post their PR or namespaces could change in the future

If we pass those info as tags, they can easily be used to condition dependencies. And if we pass it as global, they can be easily used anywhere in the templates.

`helm` already makes it easy to pass those. A call to `helm` could be
```bash
helm template charts/my-project \
    --set tags.jx-production=true \
    --set global.jx-env=production \
    --set global.jx-env-production=true
```
or
```bash
helm template charts/preview \
    --set tags.jx-preview=true \
    --set global.jx-env=preview \
    --set global.jx-env-preview=true \
    --set global.jx-preview-owner=my-owner \
    --set global.jx-preview-project=my-project \
    --set global.jx-preview-pr=3
```

Dependencies could then be like
```yaml
dependencies:
- alias: local-test-redis
  name: redis
  repository: https://kubernetes-charts.storage.googleapis.com
  version: 9.3.3
  tags:
  - jx-preview
```
and templates like
```
{{- template "myDatabase" -}}
  {{- if .Values.global.jx-env-preview -}}
    db-preview-{{ .Values.global.jx-preview-pr }}
  {{- else -}}
    db-{{ .Values.global.jx-env }}
  {{- end -}}
{{- end -}}
```

aure-olli:
#### Submitter checklist

#### Description

This PR passes useful tags and global chart variables to helm when deploying an application.

Prior to this PR, the only way to check in which environment a chart was deployed was to read `.Release.Namespace`. It was impossible to use such conditions in chart requirements, and preview environment were hard to detect.

Tags can be used to condition chart requirements.
Global variables can be used in chart templates.

`jx promote` and `jx preview` now pass the following variables to helm:
- `tags.jx-ns-<namespace>=true` (ex: `tags.jx-ns-jx-production=true`)
- `tags.jx-<env-type>=true` (ex: `tags.jx-preview=true`)
- `tags.jx-env-<env>=true` (ex: `tags.jx-env-production=true`)
- `global.jx-ns-<namespace>=true` (ex: `tags.jx-ns-jx-production=true`)
- `global.jx-<env-type>=true` (ex: `tags.jx-preview=true`)
- `global.jx-env-<env>=true` (ex: `tags.jx-env-production=true`)
- `global.jx-ns=<namespace>` (ex: `tags.jx-ns=jx-production`)
- `global.jx-type-env=<env-type>` (ex: `tags.jx-type-env=preview`)
- `global.jx-env=<env>` (ex: `tags.jx-env=production`)
- `global.jx-preview-app=<preview-app>` preview only (ex: `global.jx-preview-app=my-app`)
- `global.jx-preview-pr=<preview-pr>` preview only (ex: `global.jx-preview-pr=6`)

`jx step helm install` and `jx step helm apply` now pass the following variables to helm:
- `tags.jx-ns-<namespace>=true` (ex: `tags.jx-ns-jx-production=true`)
- `global.jx-ns-<namespace>=true` (ex: `tags.jx-ns-jx-production=true`)
- `global.jx-ns=<namespace>` (ex: `tags.jx-ns=jx-production`)

Additionally, `jx step helm install` now accepts the `--set-string` parameter, similarly to `helm install`.

I don't know where such feature can be documented.

Regular environments are currently deployed using `jx step helm apply --namespace <namespace>`, without any mention of which environment is actually being deployed. This unfortunately limits the variables I can pass when deploying a regular environment. Maybe another command or option would be useful.

fixes #6312

### DIFF
--- a/pkg/cmd/add/add_app_test.go
+++ b/pkg/cmd/add/add_app_test.go
@@ -291,6 +291,7 @@ func TestAddAppWithSecrets(t *testing.T) {
 			pegomock.AnyBool(),
 			pegomock.AnyStringSlice(),
 			pegomock.AnyStringSlice(),
+			pegomock.AnyStringSlice(),
 			pegomock.EqString(kube.DefaultChartMuseumURL),
 			pegomock.AnyString(),
 			pegomock.AnyString())).
@@ -348,6 +349,7 @@ func TestAddAppWithSecrets(t *testing.T) {
 				pegomock.AnyInt(),
 				pegomock.AnyBool(),
 				pegomock.AnyBool(),
+				pegomock.AnyStringSlice(),
 				pegomock.AnyStringSlice(),
 				pegomock.AnyStringSlice(),
 				pegomock.EqString(kube.DefaultChartMuseumURL),
@@ -436,6 +438,7 @@ func TestAddAppWithDefaults(t *testing.T) {
 			pegomock.AnyBool(),
 			pegomock.AnyStringSlice(),
 			pegomock.AnyStringSlice(),
+			pegomock.AnyStringSlice(),
 			pegomock.EqString(kube.DefaultChartMuseumURL),
 			pegomock.AnyString(),
 			pegomock.AnyString())).
@@ -478,6 +481,7 @@ func TestAddAppWithDefaults(t *testing.T) {
 				pegomock.AnyInt(),
 				pegomock.AnyBool(),
 				pegomock.AnyBool(),
+				pegomock.AnyStringSlice(),
 				pegomock.AnyStringSlice(),
 				pegomock.AnyStringSlice(),
 				pegomock.EqString(kube.DefaultChartMuseumURL),
@@ -715,6 +719,7 @@ func TestAddApp(t *testing.T) {
 			pegomock.AnyBool(),
 			pegomock.AnyStringSlice(),
 			pegomock.AnyStringSlice(),
+			pegomock.AnyStringSlice(),
 			pegomock.EqString(kube.DefaultChartMuseumURL),
 			pegomock.AnyString(),
 			pegomock.AnyString())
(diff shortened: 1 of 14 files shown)


abayer:
lint failure is two things - a case of import ordering, and some formatting. As a general rule, I’d recommend running `make fmt` before pushing every time (I personally slip up on this sometimes due to having an old and slow laptop as my primary work computer, and `make fmt` takes a few minutes there thanks to the import formatting), and if you’ve got time, `make lint` as well to catch any other potential lint errors.

We’re gradually working on getting the logs scrubbed well enough in the UI to be 99% sure of not exposing unmasked secrets there. Once we’ve (finally!) got that, we’ll be able to give contributors like yourself access to the UI. Sorry. =(

abayer:
And the integration failure is due to a bunch of compilation errors in `pkg/cmd/add/add_app_test.go` - if you don’t have the time for a full `make test`, I’d suggest running `make build-all`, which will compile all the test code but not run any tests.

aure-olli:
Hi !

Thank you for you advice, it's much helpful. I also face problem with my modest computer: I had to reboot several times because I forgot to remove `-p 4`, `make test` takes forever even when I only want one test, and I had to let `make generate` run during the night ^^ Many tests interact with `git` and `kubectl`, which provokes many errors so it's sometimes hard to see the true errors.

Anyway, I hope everything is fixed.

abayer:
Yeah, I always set my k8s context to an existing cluster and set my namespace to jx before running the full tests. That said, I should really fix the lingering non-integration tests that barf if you don’t have a cluster.

Issue #6185, opened 2019-11-25
JX boot incorrectly changes environment ingress tls secret name


patrickleet:
### Summary 

Any time something jx boot related changes - adding an app, deleting an app, etc - it validates environments - part of this process is changing all of the tls secret names in each to be invalid by replacing a hyphen with a dot - which breaks https for the whole environment.

![Add environment configuration](https://user-images.githubusercontent.com/474827/69562765-d8bfd280-0f7d-11ea-9f13-42754c812600.png)

### Steps to reproduce the behavior

`jx boot` with tls wildcard cert configured

### Expected behavior

Leave the correct value alone, or at least calculate the same value

### Actual behavior

hyphen is changed to a dot

### Jx version

The output of `jx version` is:

```
➜  ~ jx version
NAME               VERSION
jx                 2.0.1005
Kubernetes cluster v1.13.11-gke.14
kubectl            v1.15.2
helm client        Client: v2.14.3+g0e7f3b6
git                2.21.0
Operating System   "Fedora release 30 (Thirty)"
```

### Jenkins type

- [x] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

### Kubernetes cluster

terraform on gke + jx boot

### Operating system / Environment

happens via pipeline, using fedora locally though

PR #6464, opened 2020-01-08
fix: issue with generated tlsSecretName expose config containing dots…

### LINKED ISSUE #6463: generated expose config tlsSecretName contains dots which does not match the secret name
Generated config:
```
expose:
  config:
    domain: jx.domain.tld
    exposer: Ingress
    http: "false"
    tlsSecretName: tls-jx.domain.tld-p
    tlsacme: "true"
    urltemplate: '{{.Service}}-{{.Namespace}}.{{.Domain}}'
  production: true
```
However the generated Kubernetes Secret is actually called `tls-jx-domain-tld-p`

/kind bug
/assign rawlingsj

rawlingsj:
… fixes #6463

### DIFF
--- a/pkg/cmd/step/verify/step_verify_environments.go
+++ b/pkg/cmd/step/verify/step_verify_environments.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


sfynx:
Not sure why I did not find it before, but it looks like this one has been reported before and therefor also fixes #6185.

Issue #6459, opened 2020-01-08
Update jx promote to stop updating project PA during manual promotions


romainverduci:
At the moment, when running `jx promote`, some information is added to the master pipeline of the app we want to promote regarding the promotion in progress (link and status to the PR created in the environment repo).
This is fine in the context of an automatic promotion that happens right after a release of the app.
But when it comes to manual promotion, it can happen at any time and it seems weird to revive an old pipeline activity for the app master build and add promote information in it.

Acceptance criteria:
- when `jx promote` is run in a script / pipeline and for environments configured with promotion strategy = automatic, we want to keep the current behaviour
- when `jx promote` is run manually or for a an environment configured with promotion strategy = manual, we don't want to update the application master PA, we just want to create a PR in the environment repo and let the normal jx process to happen.

PR #6466, opened 2020-01-08
[wip] feat: add promotion information to PA for automatic promotion only

romainverduci:
fixes https://github.com/jenkins-x/jx/issues/6459

Current behaviour:
- no change for classic jenkins relying on controller_workflow
- if tekton && jx promote batch mode in an environment configure with promotion strategy auto: keeping the same behaviour (updating the project master PA with promotion information)
- if tekton && jx promote in an environment configured with promotion strategy manual: not updating the PA anymore but still creates the PR in the environment repo

Drawback:
- with the current implementation we lose the the ability to follow the promotion PR until it's merged, this code being tightly coupled with the one updating the PA

Is that an acceptable short term change, knowing that if we want to make it perfect, there is probably a lot of refactoring / redesign involved.

### DIFF
--- a/pkg/cmd/controller/controller_workflow_integration_test.go
+++ b/pkg/cmd/controller/controller_workflow_integration_test.go
@@ -153,11 +153,11 @@ func TestSequentialWorkflow(t *testing.T) {
 
 	// lets make sure we don't create a PR for production as we have not completed the staging PR yet
 	err = o.Run()
-	testhelpers.AssertHasNoPullRequestForEnv(t, activities, a.Name, "production")
+	testhelpers.AssertHasNoPullRequestForEnvInPipelineActivity(t, activities, a.Name, "production")
 
 	// still no PR merged so cannot create a PR for production
 	pollGitStatusAndReactToPipelineChanges(t, o, jxClient, ns)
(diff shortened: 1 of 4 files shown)


abayer:
I may just not be deep enough into the code, but why don’t we just not call `OnPromotePullRequest` if the promotion is manual and we’re Tekton? The function does nothing but update `PipelineActivity`, so in the desired scenario, we really just don’t need or want to call it.

romainverduci:
@abayer that's what I had in mind but I may need some help to get that info in the different places the method is called.

hferentschik:
@romainverduci - hi, what is the state on this PR? Is this still wip or is this PR basically ready to go, but we are undecided on whether this is the right thing to do?

romainverduci:
@hferentschik I think we don't want to go with that as it's more a hack than an improvement.
Ideally we need to refactor more the promote code to make it do what we want and in a clear way.

hferentschik:
>  I think we don't want to go with that as it's more a hack than an improvement.

I see. Shall we close this PR then?

romainverduci:
@hferentschik closing it, I won't delete the branch in case we want to come back on it some day.

PR #6479, opened 2020-01-09
fix: Cloning for environments shouldn't leave strange state in JX_HOME

### LINKED ISSUE #6350: Delete app fails with cherry-pick error
### Summary 
Attempting to delete an App using `jx delete app` fails under some circumstances due to a cherry pick error.

```
➜ jx delete app jx-app-fossa
Attempting to cherry pick commits that were on master but not yet pushed
error: couldn't pull the environment repository from dev: Unable to cherry-pick a155412b72bd8ca8123a8e3b0993a5bd49ca577d: git output: error: commit a155412b72bd8ca8123a8e3b0993a5bd49ca577d is a merge but no -m option was given.
fatal: cherry-pick failed: failed to run 'git cherry-pick a155412b72bd8ca8123a8e3b0993a5bd49ca577d' command in directory '/Users/terry/.jx/environments', output: 'error: commit a155412b72bd8ca8123a8e3b0993a5bd49ca577d is a merge but no -m option was given.
fatal: cherry-pick failed'
```

### Jx version

```
jx                 2.0.1011
kubectl            v1.16.2
git                2.21.0
Operating System   Mac OS X 10.14.6 build 18G1012
```

### Jenkins type

- [ ] Classic Jenkins

abayer:
#### Submitter checklist

#### Description

Until this, `jx delete app`, `jx add app`, and `jx get apps` all did weird things with/under `~/.jx/environments`. All of them used `ForkAndPullRepo` to clone under there - `jx get apps` directly into `~/.jx/environments`, and `jx add app` and `jx delete app` into `~/.jx/environments/dev`. Literally none of this made sense. =)

This reworks not just the app commands, but all commands that create PRs (i.e., gitops) against the dev/prod/staging environments to use temporary directories for the fork/clone/push, with the option, just used in tests, of specifying an existing directory to use instead, while `jx get apps` just clones to a temporary directory as well.

Switching all this to temporary directories also really helps users who manage/work with multiple JX clusters from a single laptop/desktop/host, since namespacing `~/.jx/environments` by cluster is basically impossible so far as I can tell.

Also adds `--auto-merge` to `jx delete app` because I was here.

fixes #6350

### DIFF
--- a/pkg/apps/gitops.go
+++ b/pkg/apps/gitops.go
@@ -167,15 +167,14 @@ func (o *GitOpsOptions) DeleteApp(app string, alias string, autoMerge bool) erro
 
 // GetApps retrieves all the apps information for the given appNames from the repository and / or the CRD API
 func (o *GitOpsOptions) GetApps(appNames map[string]bool, expandFn func([]string) (*v1.AppList, error)) (*v1.AppList, error) {
-	dir := o.EnvironmentCloneDir
-	if dir == "" {
-		tmpDir, err := ioutil.TempDir("", "get-apps-")
-		if err != nil {
-			return nil, err
-		}
-		defer os.RemoveAll(tmpDir)
-		dir = tmpDir
+	// AddApp, DeleteApp, and UpgradeApps delegate selecting/creating the directory to clone in to environments/gitops.go's
+	// Create function, but here we need to create the directory explicitly. since we aren't calling Create, because we're
+	// not creating a pull request.
+	dir, err := ioutil.TempDir("", "get-apps-")
+	if err != nil {
+		return nil, err
 	}
+	defer os.RemoveAll(dir)
 
 	gitInfo, err := gits.ParseGitURL(o.DevEnv.Spec.Source.URL)
 	if err != nil {
--- a/pkg/apps/gitops.go
+++ b/pkg/apps/gitops.go
@@ -42,7 +42,7 @@ func (o *GitOpsOptions) AddApp(app string, dir string, version string, repositor
 		GitProvider: o.GitProvider,
 	}
 
-	info, err := options.Create(o.DevEnv, o.EnvironmentsDir, &details, nil, "", autoMerge)
+	info, err := options.Create(o.DevEnv, o.EnvironmentCloneDir, &details, nil, "", autoMerge)
 	if err != nil {
 		return errors.Wrapf(err, "creating pr for %s", app)
 	}
@@ -104,7 +104,8 @@ func (o *GitOpsOptions) UpgradeApp(app string, version string, repository string
 			o.Helmer, inspectChartFunc, o.Verbose, o.valuesFiles),
 		GitProvider: o.GitProvider,
 	}
-	_, err = options.Create(o.DevEnv, o.EnvironmentsDir, &details, nil, app, autoMerge)
+
+	_, err = options.Create(o.DevEnv, o.EnvironmentCloneDir, &details, nil, app, autoMerge)
 	if err != nil {
 		return err
 	}
@@ -156,7 +157,7 @@ func (o *GitOpsOptions) DeleteApp(app string, alias string, autoMerge bool) erro
 		GitProvider:   o.GitProvider,
 	}
 
-	info, err := options.Create(o.DevEnv, o.EnvironmentsDir, &details, nil, "", autoMerge)
+	info, err := options.Create(o.DevEnv, o.EnvironmentCloneDir, &details, nil, "", autoMerge)
 	if err != nil {
 		return err
 	}
@@ -166,9 +167,38 @@ func (o *GitOpsOptions) DeleteApp(app string, alias string, autoMerge bool) erro
 
 // GetApps retrieves all the apps information for the given appNames from the repository and / or the CRD API
 func (o *GitOpsOptions) GetApps(appNames map[string]bool, expandFn func([]string) (*v1.AppList, error)) (*v1.AppList, error) {
-	dir, _, _, _, err := gits.ForkAndPullRepo(o.DevEnv.Spec.Source.URL, o.EnvironmentsDir, o.DevEnv.Spec.Source.Ref, "master", o.GitProvider, o.Gitter, "")
+	dir := o.EnvironmentCloneDir
+	if dir == "" {
+		tmpDir, err := ioutil.TempDir("", "get-apps-")
+		if err != nil {
+			return nil, err
+		}
+		defer os.RemoveAll(tmpDir)
+		dir = tmpDir
+	}
+
+	gitInfo, err := gits.ParseGitURL(o.DevEnv.Spec.Source.URL)
 	if err != nil {
-		return nil, errors.Wrapf(err, "couldn't pull the environment repository from %s", o.DevEnv.Name)
+		return nil, errors.Wrapf(err, "parsing dev env repo URL %s", o.DevEnv.Spec.Source.URL)
+	}
+
+	providerInfo, err := o.GitProvider.GetRepository(gitInfo.Organisation, gitInfo.Name)
+	if err != nil {
+		return nil, errors.Wrapf(err, "determining git provider information for %s", o.DevEnv.Spec.Source.URL)
+	}
+	cloneUrl := providerInfo.CloneURL
+	userDetails := o.GitProvider.UserAuth()
+	originFetchURL, err := o.Gitter.CreateAuthenticatedURL(cloneUrl, &userDetails)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create authenticated fetch URL for %s", cloneUrl)
+	}
+	err = o.Gitter.Clone(originFetchURL, dir)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to clone %s to dir %s", cloneUrl, dir)
+	}
+	err = o.Gitter.Checkout(dir, o.DevEnv.Spec.Source.Ref)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to checkout %s to dir %s", o.DevEnv.Spec.Source.Ref, dir)
 	}
 
 	envDir := filepath.Join(dir, helm.DefaultEnvironmentChartDir)
--- a/pkg/apps/install.go
+++ b/pkg/apps/install.go
@@ -36,24 +36,24 @@ import (
 
 // InstallOptions are shared options for installing, removing or upgrading apps for either GitOps or HelmOps
 type InstallOptions struct {
-	Helmer          helm.Helmer
-	KubeClient      kubernetes.Interface
-	InstallTimeout  string
-	JxClient        versioned.Interface
-	Namespace       string
-	EnvironmentsDir string
-	GitProvider     gits.GitProvider
-	Gitter          gits.Gitter
-	Verbose         bool
-	DevEnv          *jenkinsv1.Environment
-	BatchMode       bool
-	IOFileHandles   util.IOFileHandles
-	GitOps          bool
-	TeamName        string
-	BasePath        string
-	VaultClient     vault.Client
-	AutoMerge       bool
-	SecretsScheme   string
+	Helmer              helm.Helmer
+	KubeClient          kubernetes.Interface
+	InstallTimeout      string
+	JxClient            versioned.Interface
+	Namespace           string
+	EnvironmentCloneDir string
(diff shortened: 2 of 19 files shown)


abayer:
I'm not shocked at the failures - I just wanted to push up what worked for actual apps usage before I had to run to an appointment. Will fix later today

pkg/apps/gitops.go
```diff
@@ -166,9 +176,18 @@ func (o *GitOpsOptions) DeleteApp(app string, alias string, autoMerge bool) erro
 
 // GetApps retrieves all the apps information for the given appNames from the repository and / or the CRD API
 func (o *GitOpsOptions) GetApps(appNames map[string]bool, expandFn func([]string) (*v1.AppList, error)) (*v1.AppList, error) {
-	dir, _, _, _, err := gits.ForkAndPullRepo(o.DevEnv.Spec.Source.URL, o.EnvironmentsDir, o.DevEnv.Spec.Source.Ref, "master", o.GitProvider, o.Gitter, "")
+	dir, err := ioutil.TempDir("", "create-pr")
```
hferentschik: on a side note, I think we should start using a base dir within JX_HOME, e.g. JX_HOME/tmp for all out temporary directory creations. This probably makes it easier to debug and it is also easier to see when things don't get cleaned up.

hferentschik: and why would 'get app' create a tmp directory with 'create-pr' in it?

hferentschik: So for each 'get apps' we clone the repo? That does not seem terribly effective? I guess the idea of using ~/.jx/environment was to have some sort of a cache. 

Admittedly something which works it better than something which is "more performant" but not working ;-)

pkg/apps/gitops.go
```diff
@@ -166,9 +176,18 @@ func (o *GitOpsOptions) DeleteApp(app string, alias string, autoMerge bool) erro
 
 // GetApps retrieves all the apps information for the given appNames from the repository and / or the CRD API
 func (o *GitOpsOptions) GetApps(appNames map[string]bool, expandFn func([]string) (*v1.AppList, error)) (*v1.AppList, error) {
-	dir, _, _, _, err := gits.ForkAndPullRepo(o.DevEnv.Spec.Source.URL, o.EnvironmentsDir, o.DevEnv.Spec.Source.Ref, "master", o.GitProvider, o.Gitter, "")
```
hferentschik: As much as I also dislike the use of `ForAndPullRepo` here, if for no other reason than the fact that it is a terrible name, I believe this would actually work, if there were not the #5772 bug.

hferentschik: Another issue with ~/.jx/environment is that in the case of app it is not keys against context. So if you have multiple clusters and you switch between them you also get an error, this time a legit merge error. This is due to the fact that this app code came first. Later code which uses JX_HOME to cache checkout is keying against the cluster. I believe there is an issue for this as well.

pkg/apps/gitops.go
abayer: `jx get app` and `jx get delete app` create PRs, so...

And yeah, I decided to ditch the sorta cache in `~/.jx/environments` because (a) it was horribly broken as written, and (b) there currently is no viable way to namespace the clone inside `~/.jx/environments` by cluster, so if you've got multiple clusters, you'd still end up having potential problems.

pkg/apps/gitops.go
abayer: But _why_ use `ForkAndPullRepo` here? It's the forking that drove me off - there's absolutely no reason to _fork_ the dev env as part of running `jx get apps`.

pkg/apps/gitops.go
abayer: wait, my brain fried - this shouldn't be `create-pr-`. duh. =)

pkg/apps/gitops.go
hferentschik: Exactly, the name is terrible. It implies completely different semantics to what is actually happening, depending on the passed parameters. It is not always forking. The docs also says "ForkAndPullRepo pulls the specified gitUrl into dir using gitter, creating a remote fork if needed using the git provider". But that makes the function not less confusing. It is one of these cases where a single methods tries to do everything. 

I agree it completely confuses the reader of the code at this point in the code. I also think there are several things wrong with ForkAndPullRepo, so the less we use it the better. 

Doing a fresh throw-away checkout for each of these app operations will fix it, but with a pretty high cost. It feels like it there should be a better way. But let's go for something which works.

abayer:
Fixed most of the test failures - `controller_workflow_integration_test` is failing but I literally have no idea how this change makes it fail and I've had a looooong day so I'll look into it tomorrow.

pkg/apps/gitops.go
abayer: We already effectively are doing throw away checkouts for `jx add app`, `jx delete app`, `jx upgrade app`, and their friends (namely `jx step scheduler [apply|migrate]`. I just threw in a throwaway checkout for `jx get apps` too. =)

pkg/cmd/add/add_app_test.go
```diff
@@ -80,6 +80,11 @@ func TestAddAppForGitOps(t *testing.T) {
 			DevEnv:     testOptions.DevEnv,
 			HelmUpdate: true, // Flag default when run on CLI
 		}
+		envDir, err := o.CommonOptions.EnvironmentsDir()
```
hferentschik: is this test retrying!?

pkg/apps/gitops.go
```diff
@@ -42,6 +42,16 @@ func (o *GitOpsOptions) AddApp(app string, dir string, version string, repositor
 		GitProvider: o.GitProvider,
 	}
 
+	prDir := o.EnvironmentsDir
```
hferentschik: I find it very confusing that parts of the required input for this function come via function parameters and others by accessing the `GitOpsOptions` instance. For example, `autoMerge` is passed as a parameter,  `o.EnvironmentsDir` not.

pkg/apps/gitops.go
```diff
@@ -42,6 +42,16 @@ func (o *GitOpsOptions) AddApp(app string, dir string, version string, repositor
...
+	if prDir == "" {
```
hferentschik: This confuses me. What's happening here. So first you set the `prDir` to the `EnvironmentsDir` which I believe is `~/.jx/environments`, right? If it does not exist you create a tmp directory which eventually gets deleted. I get that part, but what if `~/.jx/environments` already exists?

hferentschik: hmm, on second thought, I might be looking at something here which is work in progress, since `prDir` is not even used.

pkg/cmd/add/add_app_test.go
abayer: Because the tests want to look at the contents of the resulting PR to make sure it did the right thing, and because of how `FakeProvider` works, that means we've gotta keep around the repo we create the PR in. Hence the `CloneDir` option I've added to all the various commands but not exposed via the CLI - a way to say "actually, don't use a temporary directory, use this one".

pkg/apps/gitops.go
abayer: I've tidied it up a bit, but it's still kinda janky.

pkg/apps/gitops.go
abayer: Fixed that.

abayer:
This will pass again once https://github.com/jenkins-x/bdd-jx/pull/104 merges.

pkg/apps/gitops.go
```diff
@@ -166,9 +167,38 @@ func (o *GitOpsOptions) DeleteApp(app string, alias string, autoMerge bool) erro
...
+		return nil, errors.Wrapf(err, "parsing dev env repo URL %s", o.DevEnv.Spec.Source.URL)
+	}
+
+	providerInfo, err := o.GitProvider.GetRepository(gitInfo.Organisation, gitInfo.Name)
+	if err != nil {
+		return nil, errors.Wrapf(err, "determining git provider information for %s", o.DevEnv.Spec.Source.URL)
+	}
+	cloneUrl := providerInfo.CloneURL
+	userDetails := o.GitProvider.UserAuth()
+	originFetchURL, err := o.Gitter.CreateAuthenticatedURL(cloneUrl, &userDetails)
```
hferentschik: :-( - this is bad. We should not use this approach. Sad to see that this proliferates now outside ForkAndPullRepo

pkg/apps/install.go
```diff
@@ -175,6 +175,12 @@ func (o *InstallOptions) DeleteApp(app string, alias string, releaseName string,
 		Items: make([]string, 0),
 	}
 
+	// Make sure that we use a temporary directory for GetApps even if we're not using a temporary directory for DeleteApps
```
hferentschik: What does that mean? Thought we are going to use tmp directories for all of it now?

pkg/cmd/add/add_app_test.go
hferentschik: So we need to add this just for testing purposes? Seems wrong, especially since it confuses the reader of the production code. I already stumbled across your comment.

Is there no other way to test this?

pkg/apps/gitops.go
abayer: I don't disagree, but I couldn't figure out how to ensure that `ForkAndPullRepo` _doesn't_ fork (it seems to fork if the current user is different from the owner of the repo) and I needed an authenticated clone URL, so...

pkg/cmd/deletecmd/delete_app.go
```diff
@@ -45,6 +45,10 @@ type DeleteAppOptions struct {
 	Namespace   string
 	Purge       bool
 	Alias       string
+	AutoMerge   bool
+
+	// Used for testing
+	CloneDir string
```
hferentschik: :-(

pkg/apps/install.go
abayer: That's an out-of-date comment - I'll remove.

abayer: ...and out of date code. Fixing. =)

pkg/cmd/add/add_app_test.go
abayer: Not so far as I can tell - opening a PR with `FakeProvider` just adds a record on the target repository's struct saying "there's a PR from this other repo", so once the other repo's directory gets deleted, the actual content of the PR is gone. So we either need to do the `CloneDir` stuff or we need to remove checking the content of the PR from the tests, which seems like a bad idea.

pkg/apps/gitops.go
hferentschik: Sure

abayer: Though I'd be fine with adding an argument to `ForkAndPullRepo` to explicitly block forking and just using it here.

pkg/apps/gitops.go
```diff
@@ -166,9 +167,37 @@ func (o *GitOpsOptions) DeleteApp(app string, alias string, autoMerge bool) erro
 
 // GetApps retrieves all the apps information for the given appNames from the repository and / or the CRD API
 func (o *GitOpsOptions) GetApps(appNames map[string]bool, expandFn func([]string) (*v1.AppList, error)) (*v1.AppList, error) {
-	dir, _, _, _, err := gits.ForkAndPullRepo(o.DevEnv.Spec.Source.URL, o.EnvironmentsDir, o.DevEnv.Spec.Source.Ref, "master", o.GitProvider, o.Gitter, "")
+	// AddApp, DeleteApp, and UpgradeApps delegate selecting/creating the directory to clone in to environments/gitops.go's
+	// Create function, but here we need to create the directory explicitly. since we aren't calling Create, because we're
```
hferentschik: 👍

pkg/cmd/add/add_app_test.go
hferentschik: Another approach would be to work against known locations. As I mentioned before, I think we should create all these temporary directories in a common base, aka JX_HOME/tmp. This way we have a place to look/scan for the created directory. 

This combined with creating the tmp directory quite high up in the call hierachy (somewhere close to _Run_), would probably get us closter to where we want to be.

For now that is probably too much to ask for ;-)

hferentschik: > So we either need to do the CloneDir stuff or we need to remove checking the content of the PR from the tests, which seems like a bad idea.

Maybe the tests could be split as well. Personally, I would be very reluctant to have this flag just because I need it for testing. For me it is a code smell.

PR #6488, opened 2020-01-10
feat: add the option to run the bdd tests directly without importing

garethjevans:
This is to attempt to fix https://github.com/cloudbees/arcalos/issues/567, in this configuration we run the bdd tests from inside the cluster, but the user that the tests should run as is a local user (configured via a secret) rather than the pipeline user (the default)

### DIFF
--- a/pkg/tekton/metapipeline/client.go
+++ b/pkg/tekton/metapipeline/client.go
@@ -38,6 +38,9 @@ type PipelineCreateParam struct {
 	// UseBranchAsRevision forces step_create_task to use the branch it's passed as the revision to checkout for release
 	// pipelines, rather than use the version tag
 	UseBranchAsRevision bool
+
+	// NoReleasePrepare do not prepare the release, this passes the --no-release-prepare flag to `jx step create task`
+	NoReleasePrepare bool
 }
 
 // Client defines the interface for meta pipeline creation and application.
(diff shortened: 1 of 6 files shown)


pkg/tekton/metapipeline/client.go
```diff
@@ -38,6 +38,9 @@ type PipelineCreateParam struct {
 	// UseBranchAsRevision forces step_create_task to use the branch it's passed as the revision to checkout for release
 	// pipelines, rather than use the version tag
 	UseBranchAsRevision bool
+
+	// NoReleasePrepare do not prepare the release, this passes the --no-release-prepare flag to `jx step create task`
+	NoReleasePrepare bool
```
abayer: How did we forget to include this originally? I feel silly.

abayer:
I think this looks good, but wanna give it another pass on Monday.

garethjevans:
@abayer its quite a hard one to test. On Arcalos the bdd tests are run from inside the cluster, which makes testing changes to JX really hard cause it reads the versions directly from the version stream

garethjevans:
The meta pipeline client changes seems to work nicely btw

abayer:
@garethjevans Yeah, it’s definitely hairy to test, and it feels right, just wanna sleep on it because I get nervous with anything that touches git creds these days. =)

garethjevans:
Any progress on this @abayer? The new logic shouldn't be invoked unless the new flag/env var is set - i think it should be safe to merge - I can't test all the interactions until this versions exists in the version stream

abayer:
/lgtm

@garethjevans Thanks for the reminder. =)

PR #6491, opened 2020-01-10
chore: misc task to make it easier to debug boot as well as cleanup for use of GIT_ASKPASS

hferentschik:
some chore tasks which I extracted from my current work on #5772


hferentschik:
rebased

Issue #2302, opened 2018-11-22
Add support for kustomize


jstrachan:
it would be awesome to be able to use kustomize YAML in an environment git repository to provide environment specific overrides for apps.

e.g. having a folder structure...
```
kustomize/
  app1/
    mydeployment.yaml
  app2/
    someservice.yaml
```

Then when using [no tiller]() we use helm to generate the YAML and post process it. We could trivially detect the `kustomize` folder and then just run `kustomize` against the YAML before we apply it - i.e. here: https://github.com/jenkins-x/jx/blob/master/pkg/helm/helm_template.go#L284

we'd just need to 

(i) add kustomise to our builder images
(ii) detect the `kustomize` folder in the current folder and apply it before the above line. 

simples!

jstrachan:
we may also want to support kustomise folder in the preview chart in an app if we want to do any customisation to the preview chart `charts/preview/kustomize/*.yaml` e.g. to kustomize a dependent chart (postgres or something)

polothy:
is there a way to support these things with `helm3`?  `no tiller` has its own drawbacks* and we were hoping `helm3` would solve all our problems ;) (I know... it'll make new ones!)

\* Doesn't support deploying to multiple namespaces, misses deleting resources like custom CRDS, etc.

PR #6494, opened 2020-01-11
feat: Ensure Kustomize

nehagup:
Step 2 towards : 
#2302

### DIFF
--- a/pkg/cmd/clients/factory.go
+++ b/pkg/cmd/clients/factory.go
@@ -11,6 +11,8 @@ import (
 	"net/url"
 	"os"
 
+	"github.com/jenkins-x/jx/pkg/kustomize"
+
 	"github.com/jenkins-x/jx/pkg/config"
 
 	"github.com/jenkins-x/jx/pkg/kube/cluster"
@@ -775,4 +777,9 @@ func (f *factory) CreateCertManagerClient() (certmngclient.Interface, error) {
 		return nil, err
 	}
 	return certmngclient.NewForConfig(config)
+}
+
+// CreateKustomizer creates a Kustomize client
+func (f *factory) CreateKustomizer() kustomize.Kustomizer {
+	return kustomize.NewKustomizeCLI()
 }
--- a/pkg/cmd/clients/interface.go
+++ b/pkg/cmd/clients/interface.go
@@ -3,6 +3,8 @@ package clients
 import (
 	"io"
 
+	"github.com/jenkins-x/jx/pkg/kustomize"
+
 	gojenkins "github.com/jenkins-x/golang-jenkins"
 	"github.com/jenkins-x/jx/pkg/io/secrets"
 	"github.com/jenkins-x/jx/pkg/util"
@@ -165,4 +167,7 @@ type Factory interface {
 
 	// ResetSecretsLocation resets the location of the secrets
 	ResetSecretsLocation()
+
+	// CreateKustomizer creates a kustomize client
+	CreateKustomizer() kustomize.Kustomizer
 }
--- a/pkg/cmd/opts/common.go
+++ b/pkg/cmd/opts/common.go
@@ -37,6 +37,7 @@ import (
 	"github.com/jenkins-x/jx/pkg/gits"
 	"github.com/jenkins-x/jx/pkg/helm"
 	"github.com/jenkins-x/jx/pkg/kube"
+	"github.com/jenkins-x/jx/pkg/kustomize"
 	"github.com/jenkins-x/jx/pkg/log"
 	"github.com/jenkins-x/jx/pkg/table"
 	"github.com/jenkins-x/jx/pkg/util"
@@ -146,6 +147,7 @@ type CommonOptions struct {
 	kserveClient        kserve.Interface
 	kubeClient          kubernetes.Interface
 	kuber               kube.Kuber
+	kustomizer          kustomize.Kustomizer
 	prowJobClient       prowjobclient.Interface
 	resourcesInstaller  resources.Installer
 	systemVaultClient   vault.Client
@@ -1287,4 +1289,12 @@ func (o *CommonOptions) IsInsecureSSLWebhooks() (bool, error) {
 		return false, nil
 	}
 	return isStaging, nil
+}
+
+// Kustomize returns or creates the kustomize client
+func (o *CommonOptions) Kustomize() kustomize.Kustomizer {
+	if o.kustomizer == nil {
+		return o.factory.CreateKustomizer()
+	}
+	return o.kustomizer
 }
--- a/pkg/cmd/opts/install.go
+++ b/pkg/cmd/opts/install.go
@@ -227,31 +227,70 @@ func (o *CommonOptions) InstallKustomize() error {
 	if runtime.GOOS == "darwin" && !o.NoBrew {
 		return o.RunCommand("brew", "install", "kustomize")
 	}
+
 	binDir, err := util.JXBinLocation()
 	if err != nil {
-		return err
-	}
-	fileName, flag, err := packages.ShouldInstallBinary("kustomize")
-	if err != nil || !flag {
-		return err
+		return errors.Wrapf(err, "unable to find JXBinLocation")
 	}
 
-	latestVersion, err := util.GetLatestVersionFromGitHub("kubernetes-sigs", "kustomize")
+	fullBinaryPath := filepath.Join(binDir, "kustomize")
+	exists, err := util.FileExists(fullBinaryPath)
 	if err != nil {
-		return fmt.Errorf("unable to get latest version for github.com/%s/%s %v", "kubernetes-sigs", "kustomize", err)
+		return errors.Wrapf(err, "unable to verify if binary exists")
+	}
+	if exists {
+		log.Logger().Debugf("binary %s already exists", fullBinaryPath)
+		return nil
 	}
 
-	clientURL := fmt.Sprintf("https://github.com/kubernetes-sigs/kustomize/releases/download/v%v/kustomize_%s_%s_%s", latestVersion, latestVersion, runtime.GOOS, runtime.GOARCH)
-	fullPath := filepath.Join(binDir, fileName)
-	tmpFile := fullPath + ".tmp"
-	err = packages.DownloadFile(clientURL, tmpFile)
+	// get the stable jx supported version of kustomize to be install
+	versionResolver, err := o.GetVersionResolver()
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "unable to retrieve version resolver")
 	}
-	err = util.RenameFile(tmpFile, fullPath)
+
+	stableVersion, err := versionResolver.StableVersion(versionstream.KindPackage, "kustomize")
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "unable to get stable version from the jenkins-x-versions for github.com/%s/%s %v ", "kubernetes-sigs", "kustomize", err)
 	}
+
+	clientURL := fmt.Sprintf("https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%%2Fv%s/kustomize_v%s_%s_%s.tar.gz", stableVersion.Version, stableVersion.Version, runtime.GOOS, runtime.GOARCH)
+	tmpDir := filepath.Join(binDir, "kustomize.tmp")
+	err = os.MkdirAll(tmpDir, util.DefaultWritePermissions)
+	if err != nil {
+		return errors.Wrapf(err, "failed to create tmp directory")
+	}
+
+	defer func() {
+		err = os.RemoveAll(tmpDir)
+		if err != nil {
+			log.Logger().Warnf("Failed to Remove tmp directory: %v", err)
+		}
+	}()
+
+	fullPath := filepath.Join(binDir, "kustomize")
+	tarFile := filepath.Join(tmpDir, "kustomize.tar.gz")
+	defer func() {
+		err = os.Remove(tarFile)
+		if err != nil {
+			log.Logger().Warnf("failed to Remove tarFile : %v", err)
+		}
+	}()
+
+	err = packages.DownloadFile(clientURL, tarFile)
+	if err != nil {
+		return errors.Wrapf(err, "failed to Download File")
+	}
+	err = util.UnTargz(tarFile, tmpDir, []string{"kustomize"})
+	if err != nil {
+		return errors.Wrapf(err, "failed to Un-tar file")
+	}
+
+	err = os.Rename(filepath.Join(tmpDir, "kustomize"), fullPath)
+	if err != nil {
+		return errors.Wrapf(err, "failed to rename file")
+	}
+
 	return os.Chmod(fullPath, 0755)
 }
 
--- a/pkg/cmd/opts/kustomize.go
+++ b/pkg/cmd/opts/kustomize.go
@@ -1,0 +1,75 @@
+package opts
+
+import (
+	"github.com/blang/semver"
+	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/versionstream"
+	"github.com/pkg/errors"
+)
+
+// EnsureKustomize ensures kustomize is installed
+func (o *CommonOptions) EnsureKustomize() error {
+	version, err := o.Kustomize().Version()
+
+	// if there is an error we assume that the binary is not installed
+	if err != nil {
+		return o.InstallKustomize()
+	}
+
+	stableVersion, err := o.stableKustomizeVersion()
+	if err != nil {
+		return errors.Wrap(err, "failed to install Kustomize")
+	}
+
+	supported, err := isInstalledKustomizeVersionSupported(version, stableVersion)
+	if err != nil {
+		return errors.Wrapf(err, "problem finding if installed version of kustomize is supported")
+	}
+
+	if !supported {
+		err = o.InstallKustomize()
+		if err != nil {
+			return errors.Wrap(err, "failed to install Kustomize")
+		}
+		log.Logger().Info("installed Kustomize")
+	}
+
+	log.Logger().Infof("keeping currently installed Kustomize version: %s", version)
+	return nil
+}
+
+func (o *CommonOptions) stableKustomizeVersion() (*versionstream.StableVersion, error) {
+	versionResolver, err := o.GetVersionResolver()
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get version resolver for jenkins-x-versions")
+	}
+
+	// get the stable jx supported version of kustomize to be install
+	stableVersion, err := versionResolver.StableVersion(versionstream.KindPackage, "kustomize")
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get stable version from the jenkins-x-versions for github.com/%s/%s %v ", "kubernetes-sigs", "kustomize", err)
+	}
+	return stableVersion, nil
+}
+
+func isInstalledKustomizeVersionSupported(version string, stableVersion *versionstream.StableVersion) (bool, error) {
+	currVersion, err := semver.Make(version)
+	if err != nil {
+		log.Logger().Warnf("unable to get currently installed Kustomize sem-version %s", err)
+	}
+	lowerLimit, err := semver.Make(stableVersion.Version)
+	if err != nil {
+		log.Logger().Warnf("unable to get lowest supported stable Kustomize sem-version %s", err)
+	}
+	upperLimit, err := semver.Make(stableVersion.UpperLimit)
+	if err != nil {
+		log.Logger().Warnf("unable to get highest supported stable Kustomize sem-version %s", err)
+	}
+
+	if currVersion.GTE(lowerLimit) && currVersion.LT(upperLimit) {
+		log.Logger().Debugf("kustomize is already installed version")
+		return true, nil
+	}
+
+	return false, errors.Wrapf(err, "unsupported version of Kustomize installed. Install kustomize version above %s or below %s ", lowerLimit, upperLimit)
+}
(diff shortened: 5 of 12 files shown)


pkg/kustomize/kustomize_cli.go
```diff
@@ -0,0 +1,58 @@
+package kustomize
+
+import (
+	"strings"
+
+	"github.com/jenkins-x/jx/pkg/util"
+)
+
+// KustomizeCLI implements common kustomize actions based on kustomize CLI
+type KustomizeCLI struct {
```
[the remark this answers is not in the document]
nehagup: Any suggestions?

nehagup:
/assign @hferentschik 

Can you please review.

pkg/cmd/clients/factory.go
```diff
@@ -773,3 +775,8 @@ func (f *factory) CreateCertManagerClient() (certmngclient.Interface, error) {
 	}
 	return certmngclient.NewForConfig(config)
 }
+
+// NewKustomize creates a new Kustomize client
+func (f *factory) NewKustomize() kustomize.Kustomizer {
```
hferentschik: I think to stick with the other functions in here, it should be `func (f *factory) CreateKustomizer`

pkg/cmd/clients/interface.go
```diff
@@ -166,4 +168,7 @@ type Factory interface {
 
 	// ResetSecretsLocation resets the location of the secrets
 	ResetSecretsLocation()
+
+	// New Kustomize creates a new kustomize client
+	NewKustomize() kustomize.Kustomizer
```
hferentschik: `CreateKustomizer` - needs to align with what we have in _factory.go_.

pkg/cmd/opts/common.go
```diff
@@ -144,6 +146,7 @@ type CommonOptions struct {
 	kserveClient        kserve.Interface
 	kubeClient          kubernetes.Interface
 	kuber               kube.Kuber
+	kustomize           kustomize.Kustomizer
```
hferentschik: I think `kustomizer` would be more idiomatic Go

pkg/cmd/opts/kustomize.go
```diff
@@ -0,0 +1,17 @@
+package opts
+
+import "github.com/pkg/errors"
+
+// EnsureKustomize ensures kustomize is installed
+func (o *CommonOptions) EnsureKustomize() error {
+	_, err := o.Kustomize().Version(false)
```
hferentschik: just `o.Kustomize().Version()` - no need for a tls flag here.

pkg/cmd/opts/kustomize_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"os/exec"
+	"testing"
+
+	"github.com/jenkins-x/jx/pkg/cmd/clients"
+)
+
+// TestCommonOptions_EnsureKustomize tests that Kustomize is
+//  properly installed
+func TestCommonOptions_EnsureKustomize(t *testing.T) {
+	cmd := exec.Command("kustomize", "version")
```
hferentschik: This is bit tricky, since it means depending on your local state the tests might do something or not. Really, you want `o.EnsureKustomize()` always to be run and it should always succeed.

You also want to make sure that your test is not interfering with your local JX_HOME which you use for development. 

So what you can do is:

- make sure a temporary JX_HOME is used - https://github.com/jenkins-x/jx/blob/23f9f49350a8aba8bab8ceaf1ff4a1fee8b288ed/pkg/cmd/testhelpers/cmd_test_helpers.go#L149
- make sure the temporary directory is deleted - https://github.com/jenkins-x/jx/blob/23f9f49350a8aba8bab8ceaf1ff4a1fee8b288ed/pkg/cmd/testhelpers/cmd_test_helpers.go#L186
- set o.NoBrew to _true_ to force an install w/o Brew, but via a download from GitHub. In this case the binary ends up in JX_HOME/bin/ (one can then also check via the file system whether the install worked)

pkg/kustomize/interface.go
```diff
@@ -0,0 +1,8 @@
+package kustomize
+
+// Kustomizer defines common kustomize actions used within Jenkins X
+type Kustomizer interface {
+	Version(tls bool) (string, error)
```
hferentschik: I think the interface should look something like this:

```
type Kustomizer interface {
	Version(extraArgs ...string) (string, error)
        Build(extraArgs ...string) (string, error)
}

File: pkg/kustomize/interface.go [hunk 8]
```diff
@@ -0,0 +1,8 @@
...
+	VersionWithArgs(tls bool, extraArgs ...string) (string, error)
+	runKustomizeWithOutput(args ...string) (string, error)
```
hferentschik: private method in interface does not really make much sense

pkg/kustomize/kustomize_cli.go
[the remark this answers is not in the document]
hferentschik: This one we can ignore.

pkg/kustomize/kustomize_cli.go
```diff
@@ -0,0 +1,42 @@
...
+	Runner util.Commander
+}
+
+// NewKustomizeCLI creates a new KustomizeCLI instance configured to use the provided helm CLI in
+// the given current working directory
+func NewKustomizeCLI(args ...string) *KustomizeCLI {
```
hferentschik: why the vararg here? How would you use this? Is it not just _NewKustomizeCLI_?

pkg/kustomize/kustomize_cli.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/jenkins-x/jx/pkg/util"
+)
+
+// KustomizeCLI implements common kustomize actions based on kustomize CLI
+type KustomizeCLI struct {
+	Runner util.Commander
+}
+
+// NewKustomizeCLI creates a new KustomizeCLI instance configured to use the provided helm CLI in
```
hferentschik: I guess the Godocs are still off here, right? - "configured to use the provided helm CLI"

pkg/cmd/opts/install.go
```diff
@@ -223,25 +223,39 @@ func (o *CommonOptions) InstallGlooctl() error {
 }
 
 // InstallKustomize installs kustomize
-func (o *CommonOptions) InstallKustomize() error {
+func (o *CommonOptions) InstallKustomize(installationDir ...string) (err error) {
```
hferentschik: _installationDir ...string_ - a vararg/array for the _installationDir_? I would think that if we really need to pass the installation dir a single string should do.

pkg/cmd/opts/kustomize.go
```diff
@@ -0,0 +1,28 @@
+package opts
+
+import (
+	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/pkg/errors"
+)
+
+// EnsureKustomize ensures kustomize is installed
+func (o *CommonOptions) EnsureKustomize(installationBinDir ...string) error {
```
hferentschik: _ installationBinDir_ vararg?

hferentschik:
regarding the lint check, `kustomize_integration_test.go` seems to be not properly formatted, `make fmt` should fix that.

```
--- a/pkg/cmd/opts/kustomize_integration_test.go
+++ b/pkg/cmd/opts/kustomize_integration_test.go
@@ -3,12 +3,13 @@
 package opts_test
 
 import (
+	"path/filepath"
+	"testing"
+
 	"github.com/blang/semver"
 	"github.com/jenkins-x/**/pkg/cmd/testhelpers"
 	"github.com/jenkins-x/**/pkg/versionstream"
 	"github.com/stretchr/testify/assert"
-	"path/filepath"
-	"testing"
 
 	"github.com/jenkins-x/**/pkg/cmd/clients"
 	"github.com/jenkins-x/**/pkg/cmd/opts"
```

pkg/cmd/opts/install.go
```diff
@@ -223,35 +223,61 @@ func (o *CommonOptions) InstallGlooctl() error {
 }
 
 // InstallKustomize installs kustomize
-func (o *CommonOptions) InstallKustomize() error {
+func (o *CommonOptions) InstallKustomize() (err error) {
```
hferentschik: Let's not use named return values here. We don't use them otherwise and they are recommended for short functions only anyways - https://tour.golang.org/basics/7

Personally,  I stay completely clear of them

pkg/cmd/opts/install.go
```diff
@@ -223,35 +223,61 @@ func (o *CommonOptions) InstallGlooctl() error {
...
 	}
+
 	fileName, flag, err := packages.ShouldInstallBinary("kustomize")
 	if err != nil || !flag {
 		return err
 	}
 
-	latestVersion, err := util.GetLatestVersionFromGitHub("kubernetes-sigs", "kustomize")
+	// get the stable jx supported version of kustomize to be install
+	versionResolver, err := o.GetVersionResolver()
```
hferentschik: +1

pkg/cmd/opts/install.go
```diff
@@ -223,35 +223,61 @@ func (o *CommonOptions) InstallGlooctl() error {
...
+	}
+	err = os.Remove(tarFile)
+	if err != nil {
+		return err
+	}
+	err = os.Rename(filepath.Join(tmpDir, "kustomize"), fullPath)
 	if err != nil {
 		return err
 	}
+	err = os.RemoveAll(tmpDir)
```
hferentschik: This has to happen in a _defer_ clause which you register after the _MkdirAll_ call. Maybe even before.

pkg/cmd/opts/install.go
```diff
@@ -223,35 +223,61 @@ func (o *CommonOptions) InstallGlooctl() error {
...
+	err = os.MkdirAll(tmpDir, util.DefaultWritePermissions)
 	if err != nil {
 		return err
 	}
-	err = util.RenameFile(tmpFile, fullPath)
+	fullPath := filepath.Join(binDir, "kustomize")
+	tarFile := filepath.Join(tmpDir, fileName+".tar.gz")
+	err = packages.DownloadFile(clientURL, tarFile)
+	if err != nil {
+		return err
```
hferentschik: wrapping all these errors might make sense. If an error occurs it makes locating the point where things go wrong easier.

pkg/cmd/opts/kustomize.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/jenkins-x/jx/pkg/versionstream"
+	"github.com/pkg/errors"
+)
+
+// EnsureKustomize ensures kustomize is installed
+func (o *CommonOptions) EnsureKustomize() error {
+	version, err := o.Kustomize().Version()
+
+	if err == nil && version != "" {
+		// get the stable jx supported version of kustomize to be install
```
hferentschik: maybe we can extract something like a _installedBinaryInSupportedVersionRange_? There might be even already something like this.

pkg/cmd/opts/kustomize.go
```diff
@@ -0,0 +1,54 @@
...
+		if err != nil {
+			return errors.Wrapf(err, "Unable to get version resolver for jenkins-x-versions")
+		}
+
+		stableVersion, err := versionResolver.StableVersion(versionstream.KindPackage, "kustomize")
+		if err != nil {
+			return errors.Wrapf(err, "Unable to get stable version from the jenkins-x-versions for github.com/%s/%s %v ", "kubernetes-sigs", "kustomize", err)
+		}
+
+		currVersion, err := semver.Make(version[12:17])
```
hferentschik: That seems wrong and fragile. where is '12:17' coming from. When I run `kustomize version --short`, I'll get `{3.5.4  2020-01-17T14:28:58+00:00  }`, so if anything it should be something like '1:6', right?

Right now I get test failures due to that.

I think working with `kustomize version` might be more understandable as well. It returns: 
```
{Version:3.5.4 GitCommit:3af514fa9f85430f0c1557c4a0291e62112ab026 BuildDate:2020-01-17T14:28:58+00:00 GoOs:darwin GoArch:amd64}
```
I'd probably use a regexp like `^*Version:(.*) .*` to capture the version and take it from there. It makes it more explicit what you are expecting at this point. If you just take a slice out of string you depend much more on the exact string representation. Also if doing though, I'd add a comment about what you expect the _version_ to be in order for you to use the slice '12:17'.

hferentschik: And should we really just continue if there is an error?

pkg/cmd/opts/kustomize_integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+	}()
+
+	err = o.EnsureKustomize()
+	if err != nil {
+		t.Errorf("EnsureKustomize() error for test case TestCommonOptions_EnsureKustomizeNoBinaryInstalled = %v", err)
+	}
+	// test that tmpJXHome/bin contains kustomize binary
+	version, err = o.Kustomize().Version()
+
+	if !assert.FileExists(t, filepath.Join(testJXHome, "bin", "kustomize")) || err != nil {
```
hferentschik: just `assert.FileExists(t, filepath.Join(testJXHome, "bin", "kustomize"))`

hferentschik: and `assertNoError(t, err)`

pkg/cmd/opts/kustomize_integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		log.Logger().Warnf("Unable to get stable version from the jenkins-x-versions for github.com/%s/%s %v ", "kubernetes-sigs", "kustomize", err)
+	}
+
+	assert.Contains(t, version, stableVersion.Version)
+}
+
+func TestCommonOptions_EnsureKustomizeBinaryExistsWrongVersion(t *testing.T) {
+	o := opts.NewCommonOptionsWithFactory(clients.NewFactory())
+	version, _ := o.Kustomize().Version()
```
hferentschik: I don't think this will be a reliable test. Easiest would be for now to skip it, but there are some things we could try. Let's talk about it.

pkg/cmd/testhelpers/cmd_test_helpers.go
```diff
@@ -168,17 +168,23 @@ func CreateTestJxHomeDir() (string, string, error) {
...
-		}
+func CleanupTestJxHomeDir(originalDir, tempDir string) (err error) {
+	// Don't delete if it's not a temp dir or if it's the original dir
+	if originalDir != "" && strings.HasPrefix(tempDir, os.TempDir()) && originalDir != tempDir {
+		err = os.Setenv("JX_HOME", originalDir)
 	}
+
+	if err != nil || originalDir == "" {
+		os.Unsetenv("JX_HOME")
+		log.Logger().Warnf("Unable to set JX original home directory variable : %s", err)
```
hferentschik: %v instead of %s, otherwise you get a warning if `err == nil`

pkg/cmd/opts/kustomize.go
hferentschik: [0-9]+\.[0-9]+\.[0-9]

hferentschik: [\d]+.[\d]+.[\d]

hferentschik:
@nehagup - have a look at https://github.com/nehagup/jx/pull/5 and integrate it into this PR. Probably best to squash all commits into a single one now.

pkg/kustomize/kustomize_cli.go
```diff
@@ -0,0 +1,34 @@
...
+	cli := &KustomizeCLI{
+		Runner: runner,
+	}
+	return cli
+}
+
+// Version executes the Kustomize version command and returns its output
+func (k *KustomizeCLI) Version(extraArgs ...string) (string, error) {
+	args := []string{"version", "--short"}
+	args = append(args, extraArgs...)
```
hferentschik: I am wondering whether we should place this whole regular expression stuff to extract the atual semantic version into here. This function would eg just return 3.5.4. Otherwise, we have the logic spread across mutliple places and it is not clear why you, for example, use a regular expression later to extract the actual version.

pkg/cmd/opts/kustomize.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "Unable to get version resolver for jenkins-x-versions")
+	}
+
+	stableVersion, err := versionResolver.StableVersion(versionstream.KindPackage, "kustomize")
+	if err != nil {
+		return errors.Wrapf(err, "Unable to get stable version from the jenkins-x-versions for github.com/%s/%s %v ", "kubernetes-sigs", "kustomize", err)
+	}
+
+	regex := regexp.MustCompile(`[0-9]+\.[0-9]+\.[0-9]+`)
```
hferentschik: we probably should do this in the implementation of Version()

nehagup: Done. :) Makes sense.

pkg/cmd/opts/install.go
```diff
@@ -227,31 +227,70 @@ func (o *CommonOptions) InstallKustomize() error {
 	if runtime.GOOS == "darwin" && !o.NoBrew {
 		return o.RunCommand("brew", "install", "kustomize")
 	}
+
 	binDir, err := util.JXBinLocation()
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "unable to find JXBinLocation")
 	}
-	fileName, flag, err := packages.ShouldInstallBinary("kustomize")
```
hferentschik: BTW, _ShouldInstallBinary_ is not working for us. It does a check against PATH and returning _false_ if it finds anything anywhere on the PATH. Pretty pointless in our case. Maybe we can change this in a separate issue.

This part is problematic:

```
	pgmPath, err := exec.LookPath(fileName)
	if err == nil {
		log.Logger().Debugf("%s is already available on your PATH at %s", util.ColorInfo(fileName), util.ColorInfo(pgmPath))
		return
	}
```

pkg/cmd/opts/install.go
```diff
@@ -227,31 +227,70 @@ func (o *CommonOptions) InstallKustomize() error {
...
 	}
 
-	clientURL := fmt.Sprintf("https://github.com/kubernetes-sigs/kustomize/releases/download/v%v/kustomize_%s_%s_%s", latestVersion, latestVersion, runtime.GOOS, runtime.GOARCH)
-	fullPath := filepath.Join(binDir, fileName)
-	tmpFile := fullPath + ".tmp"
-	err = packages.DownloadFile(clientURL, tmpFile)
+	stableVersion, err := versionResolver.StableVersion(versionstream.KindPackage, "kustomize")
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "Unable to get stable version from the jenkins-x-versions for github.com/%s/%s %v ", "kubernetes-sigs", "kustomize", err)
```
hferentschik: error messages should start lower case

pkg/cmd/opts/install.go
```diff
@@ -227,31 +227,70 @@ func (o *CommonOptions) InstallKustomize() error {
...
 	}
-	err = util.RenameFile(tmpFile, fullPath)
+
+	clientURL := fmt.Sprintf("https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%%2Fv%s/kustomize_v%s_%s_%s.tar.gz", stableVersion.Version, stableVersion.Version, runtime.GOOS, runtime.GOARCH)
+	tmpDir := filepath.Join(binDir, "kustomize.tmp")
+	err = os.MkdirAll(tmpDir, util.DefaultWritePermissions)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "Failed to create tmp directory")
```
hferentschik: same as above

pkg/kustomize/kustomize_cli.go
```diff
@@ -0,0 +1,40 @@
...
+	version, err := k.runKustomizeWithOutput(args...)
+	if err != nil {
+		return "", err
+	}
+	regex := regexp.MustCompile(`[0-9]+\.[0-9]+\.[0-9]+`)
```
hferentschik: I'd extract the regular expression stuff into a function like:

`extractSemanticVersion(fullVersionString string) (string, error)` 

The main reason is that you can add some simple tests to ensure that this regular expression handles the version strings we know about properly. It also serves as documentation on what we expect to receive by the version call to `customize`.

pkg/cmd/opts/kustomize_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+var versionTests = []struct {
+	currentVersion string
+	stableVersion  *versionstream.StableVersion
+	supported      bool
+}{
+	{"3.5.1", &versionstream.StableVersion{Version: "3.5.4", UpperLimit: "3.6.0"}, false},
+	{"3.5.4", &versionstream.StableVersion{Version: "3.5.4", UpperLimit: "3.6.0"}, true},
+	{"3.6.0", &versionstream.StableVersion{Version: "3.5.4", UpperLimit: "3.6.0"}, false},
+	{"{3.5.4  2020-01-17T14:28:58+00:00  }", &versionstream.StableVersion{Version: "3.5.4", UpperLimit: "3.6.0"}, true},
```
hferentschik: This test and the one below make not much sense anymore, since we do the version extraction now in _kustomize_cli#Version_ directly.

See comment on that function.

pkg/cmd/opts/install.go
nehagup: This was reason why the other test case I deleted was failing. ShouldInstallBinary was returning false even after adding binary at the place.

pkg/cmd/opts/kustomize_test.go
nehagup: Shall I move this to kustomize_cli_test?

nehagup:
Hi @hferentschik can you please tell me where is it breaking?

pkg/cmd/opts/install.go
hferentschik: Yeah, if you want we can look at this function and maybe fix it.

Issue #6465, opened 2020-01-08
Make devEnvApprovers field required in jx-requirements.yml


abayer:
Followup to #6457 and its initial fix on #6461 - we'll start with the new `devEnvApprovers` field as optional, prompting for it when not run in batch mode, but we'll want to move to it being mandatory. Doing so will require making changes to a bunch of BDD tests and a couple other places, once #6461 has released. Staging the change like this keeps us from running into the hell of a new field in `jx-requirements.yml` that's required from right away, because that would break everything downstream until they updated, but adding the field as optional initially gives us a window after it's released to update everything downstream to have `- devEnvApprovers: []`, and _then_ make it mandatory.

abayer:
Just realized we’re almost certain to eventually hit cherry pick errors in `OWNERS` as soon as the upstream one has changes and someone runs `jx boot upgrade`. We always want to keep the downstream version of `OWNERS` in that scenario. There’s logic in `jx boot upgrade` to exclude upstream changes to `OWNERS`, but only after the cherry picking is complete, so the errors will still show up. I’m thinking about doing something like what’s described in https://stackoverflow.com/a/930495 (with `true` as the driver, or something equivalent for Windows) to make cherry picking just ignore the `OWNERS` file completely. I think we do need to address this before making the field mandatory, and probably before making any changes to upstream boot config’s `OWNERS`.

hferentschik:
I am not sure whether _ devEnvApprovers_ should even be in the requirements file. I think we should revisit this, in particular making it a required field.

PR #6502, opened 2020-01-13
fix: Resolve OWNERS merge conflicts in boot upgrade with existing version

### LINKED ISSUE #6543: jx boot upgrade replaces OWNERS with OWNERS from jenkins-x-boot-config
### Summary

Forked repo's `OWNERS` configuration is replaced with the `OWNERS` configuration from the `jenkins-x-boot-config` repository during an upgrade from `v1.0.69` to `v1.0.71` with the commit: "chore: exclude files from upgrade" (i.e. https://github.com/yields-io/environment-k8s-dev/pull/5).

Upgrade environment using `jx upgrade boot`.

### Expected behavior

`OWNERS` file created with the values from the `jx-requirements.yml` > `devEnvApprovers` during the initial `jx boot` shouldn't be touched during an upgrade.

### Actual behavior

`OWNERS` file is replaced with the `OWNERS` file from the `jenkins-x-boot-config` repository. I'm currently not an approver / reviewer in the `jenkins-x-boot-config` repository hence I can't approve the PR created for the upgrade nor would I want to given the `OWNERS` modifications. Similar to [this issue](https://github.com/jenkins-x/jx/issues/5035): "This results in these jx devs having a high prow privileges on other peoples dev environments."

### Jx version

```
NAME               VERSION
jx                 2.0.1123
Kubernetes cluster v1.15.4-gke.22
kubectl            v1.17.0
helm client        Client: v2.13.1+g618447c
git                2.17.1
Operating System   Ubuntu 18.04.3 LTS
```

### Jenkins type

- [ ] Classic Jenkins

### Kubernetes cluster

GKE: `jx create cluster gke --skip-installation --cluster-name k8s`

Ubuntu 18.04

abayer:
#### Submitter checklist

#### Description

Realized this was an issue while thinking about #6465 - we don't commit changes to `OWNERS` from upstream when running `jx upgrade boot`, but we still will hit merge conflicts when cherry-picking if `OWNERS` has changed in both the dev env repo and upstream, so let's use a custom merge driver to ensure that all conflicts for `OWNERS` are automatically resolved in favor of the existing version in the dev env repo, as described in https://stackoverflow.com/a/930495

I still need to write a unit test for this.

fixes #6543

### DIFF
--- a/pkg/cmd/upgrade/upgrade_boot.go
+++ b/pkg/cmd/upgrade/upgrade_boot.go
@@ -49,11 +49,9 @@ var (
 )
 
 const (
-	builderImage           = "gcr.io/jenkinsxio/builder-go"
-	keepDevEnvKey          = "keepDevEnv"
-	keepDevEnvForGitConfig = "\n[merge \"keepDevEnv\"]\n" +
-		"\tname = Always keep the dev env version during merges and cherry-picks\n" +
-		"\tdriver = true\n"
+	builderImage         = "gcr.io/jenkinsxio/builder-go"
+	keepDevEnvKey        = "keepDevEnv"
+	keepDevEnvDriverName = "Always keep the dev env version during merges and cherry-picks"
 )
 
 // NewCmdUpgradeBoot creates the command
@@ -330,35 +328,17 @@ func (o *UpgradeBootOptions) updateBootConfig(versionStreamURL string, versionSt
 	if err != nil {
 		return errors.Wrap(err, "failed to cherry pick upgrade commits")
 	}
-	err = o.excludeFiles(currentSha)
-	if err != nil {
-		return errors.Wrap(err, "failed to exclude files from commit")
-	}
 	return nil
 }
 
 func (o *UpgradeBootOptions) configureGitMergeExcludes() error {
-	gitConfigFile := filepath.Join(o.Dir, ".git", "config")
-	var existingGitConfig string
-	gcExists, err := util.FileExists(gitConfigFile)
+	err := o.Git().Config(o.Dir, "--local", fmt.Sprintf("merge.%s.name", keepDevEnvKey), keepDevEnvDriverName)
 	if err != nil {
-		return errors.Wrapf(err, "checking if git config for repo at %s exists", gitConfigFile)
+		return errors.Wrapf(err, "configuring name for exclude merge driver in %s", o.Dir)
 	}
-	if gcExists {
-		// Read the existing .git/config
-		gcBytes, err := ioutil.ReadFile(gitConfigFile)
-		if err != nil {
-			return errors.Wrapf(err, "reading existing git config for repo from %s", gitConfigFile)
-		}
-		existingGitConfig = string(gcBytes)
-	}
-
-	// Write the existing .git/config content and our custom driver to .git/config in the repo
-	gitConfigContent := existingGitConfig + keepDevEnvForGitConfig
-
-	err = ioutil.WriteFile(gitConfigFile, []byte(gitConfigContent), util.DefaultFileWritePermissions)
+	err = o.Git().Config(o.Dir, "--local", fmt.Sprintf("merge.%s.driver", keepDevEnvKey), "true")
 	if err != nil {
-		return errors.Wrapf(err, "writing new git config for repo to %s", gitConfigFile)
+		return errors.Wrapf(err, "configuring driver for exclude merge driver in %s", o.Dir)
 	}
 
 	// Read the existing .git/info/attributes if it exists
@@ -461,18 +441,6 @@ func (o *UpgradeBootOptions) setupGitConfig(dir string) error {
 	err = o.Git().SetEmail(dir, email)
 	if err != nil {
 		return errors.Wrapf(err, "failed to set email for %s", email)
-	}
-	return nil
-}
-
-func (o *UpgradeBootOptions) excludeFiles(commit string) error {
-	err := o.Git().CheckoutCommitFiles(o.Dir, commit, filesExcludedFromCherryPick)
-	if err != nil {
-		return errors.Wrap(err, "failed to checkout files")
-	}
-	err = o.Git().AddCommitFiles(o.Dir, "chore: exclude files from upgrade", filesExcludedFromCherryPick)
-	if err != nil && !strings.Contains(err.Error(), "nothing to commit") {
-		return errors.Wrapf(err, "failed to commit excluded files %v", filesExcludedFromCherryPick)
 	}
 	return nil
 }
--- a/pkg/cmd/upgrade/upgrade_boot.go
+++ b/pkg/cmd/upgrade/upgrade_boot.go
@@ -328,9 +328,17 @@ func (o *UpgradeBootOptions) updateBootConfig(versionStreamURL string, versionSt
 	if err != nil {
 		return errors.Wrap(err, "failed to cherry pick upgrade commits")
 	}
+
+	err = o.removeMergeExcludesFromAttributes()
+	if err != nil {
+		return errors.Wrapf(err, "removing special handling of files to exclude from merging from upstream")
+	}
 	return nil
 }
 
+// Add a custom merge driver to .git/config that will always choose the current version when there's a change upstream
+// in selected files, and use that merge driver in .git/info/attributes for those selected files.
+// see https://stackoverflow.com/a/930495 for more details
 func (o *UpgradeBootOptions) configureGitMergeExcludes() error {
 	err := o.Git().Config(o.Dir, "--local", fmt.Sprintf("merge.%s.name", keepDevEnvKey), keepDevEnvDriverName)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/cmd/upgrade/upgrade_boot.go
```diff
@@ -399,14 +466,13 @@ func (o *UpgradeBootOptions) setupGitConfig(dir string) error {
...
 
 func (o *UpgradeBootOptions) excludeFiles(commit string) error {
-	excludedFiles := []string{"OWNERS"}
-	err := o.Git().CheckoutCommitFiles(o.Dir, commit, excludedFiles)
+	err := o.Git().CheckoutCommitFiles(o.Dir, commit, filesExcludedFromCherryPick)
 	if err != nil {
 		return errors.Wrap(err, "failed to checkout files")
 	}
-	err = o.Git().AddCommitFiles(o.Dir, "chore: exclude files from upgrade", excludedFiles)
+	err = o.Git().AddCommitFiles(o.Dir, "chore: exclude files from upgrade", filesExcludedFromCherryPick)
```
hferentschik: I really have a bad feeling about this. Can we not just get rid of cherry-picking altogether?

abayer: I just want to fix this one thing with `OWNERS` here and defer the root-and-branch work til later. =)

pkg/cmd/upgrade/upgrade_boot.go
```diff
@@ -312,14 +320,69 @@ func (o *UpgradeBootOptions) updateBootConfig(versionStreamURL string, versionSt
...
 		return errors.Wrap(err, "failed to cherry pick upgrade commits")
 	}
-	err = o.excludeFiles(currentSha)
+	return nil
+}
+
+func (o *UpgradeBootOptions) configureGitMergeExcludes() error {
+	gitConfigFile := filepath.Join(o.Dir, ".git", "config")
+	var existingGitConfig string
+	gcExists, err := util.FileExists(gitConfigFile)
```
hferentschik: hmm, should that not go through the gitter really using the `git config` command? See also https://github.com/jenkins-x/jx/pull/6572/files#diff-ea63bc4be56a5cd0bf52d8031a36cefcR179

not sure whether it is a good idea to just append this line? Can that never conflict with existing sections?

abayer: Yeah, I'd be happy to change that to use `git config`. Will do!

abayer: I'm waiting on #6572 to get in first, since it seems silly to call `git config` directly when you're about to add a helper function for exactly that.

abayer:
/hold cancel

@hferentschik - updated!

pkg/cmd/upgrade/upgrade_boot.go
```diff
@@ -312,14 +318,55 @@ func (o *UpgradeBootOptions) updateBootConfig(versionStreamURL string, versionSt
...
+		existingGitAttr = string(gaBytes)
+	}
+
+	// Write the existing .git/info/attributes content and marking the selected files as using our custom driver
+	gitAttrContent := existingGitAttr + "\n"
+	for _, excludedFile := range filesExcludedFromCherryPick {
+		gitAttrContent += fmt.Sprintf("%s merge=%s\n", excludedFile, keepDevEnvKey)
+	}
+
+	err = ioutil.WriteFile(gitAttrFile, []byte(gitAttrContent), util.DefaultFileWritePermissions)
```
hferentschik: Don't you have to unset this as well after the upgrade? You only want the custom merge driver during `upgrade boot` right? Afterwards, the normal git workflow should apply, right? Does your change now not mean that merging is off for regular evolution of the dev repo?

pkg/cmd/upgrade/upgrade_boot.go
```diff
@@ -312,14 +318,55 @@ func (o *UpgradeBootOptions) updateBootConfig(versionStreamURL string, versionSt
...
+	err := o.Git().Config(o.Dir, "--local", fmt.Sprintf("merge.%s.name", keepDevEnvKey), keepDevEnvDriverName)
+	if err != nil {
+		return errors.Wrapf(err, "configuring name for exclude merge driver in %s", o.Dir)
+	}
+	err = o.Git().Config(o.Dir, "--local", fmt.Sprintf("merge.%s.driver", keepDevEnvKey), "true")
+	if err != nil {
+		return errors.Wrapf(err, "configuring driver for exclude merge driver in %s", o.Dir)
+	}
+
+	// Read the existing .git/info/attributes if it exists
```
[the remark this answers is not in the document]
abayer: I'd say no for two reasons: this is the only place we're doing it currently, and `gitter` feels to me like it should only be for `git` commands, and this isn't that.

pkg/cmd/upgrade/upgrade_boot.go
```diff
@@ -312,13 +318,101 @@ func (o *UpgradeBootOptions) updateBootConfig(versionStreamURL string, versionSt
...
+
+	err = o.removeMergeExcludesFromAttributes()
+	if err != nil {
+		return errors.Wrapf(err, "removing special handling of files to exclude from merging from upstream")
+	}
+	return nil
+}
+
+// Add a custom merge driver to .git/config that will always choose the current version when there's a change upstream
+// in selected files, and use that merge driver in .git/info/attributes for those selected files.
```
hferentschik: thanks :-)

pkg/cmd/upgrade/upgrade_boot.go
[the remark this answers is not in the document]
hferentschik: > I'd say no for two reasons: this is the only place we're doing it currently,

right, but say someone else needs to do the same, where would he naturally be looking for it?

> gitter feels to me like it should only be for git commands, and this isn't that.

It is already more than pure git command execution. There are functions in there already dealing with .git internas. 
But since you feel so strongly about it, I resy my case.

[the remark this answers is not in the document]
abayer: I also can't think of a good abstraction, but lemme try to bang something out.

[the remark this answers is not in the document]
hferentschik: btw, another argument for moving it would be to write some tests for it in the gitter area.

pkg/cmd/upgrade/upgrade_boot.go
```diff
@@ -312,13 +318,101 @@ func (o *UpgradeBootOptions) updateBootConfig(versionStreamURL string, versionSt
...
+		return errors.Wrapf(err, "configuring files (%s) to always use existing version when conflicting with upstream", strings.Join(filesExcludedFromCherryPick, ", "))
+	}
+
 	err = o.cherryPickCommits(configCloneDir, currentSha, upgradeSha)
 	if err != nil {
 		return errors.Wrap(err, "failed to cherry pick upgrade commits")
 	}
-	err = o.excludeFiles(currentSha)
+
+	err = o.removeMergeExcludesFromAttributes()
```
hferentschik: this needs to happen in a defer, otherwise the cleanup is not happening if a cherry-pick error occurs.

pkg/gits/git_local.go
```diff
@@ -538,3 +538,13 @@ func (g *GitLocal) Describe(dir string, contains bool, commitish string, abbrev
 func (g *GitLocal) IsAncestor(dir string, possibleAncestor string, commitish string) (bool, error) {
 	return g.GitCLI.IsAncestor(dir, possibleAncestor, commitish)
 }
+
+// WriteRepoAttributes writes the given content to .git/info/attributes
+func (g *GitLocal) WriteRepoAttributes(dir string, content string) error {
+	return g.GitCLI.WriteRepoAttributes(dir, content)
```
hferentschik: This git_local impl is really a special beast. IMO we shot ourselves into the foot with this one.

pkg/gits/git_cli.go
```diff
@@ -1263,3 +1263,32 @@ func (g *GitCLI) IsAncestor(dir string, possibleAncestor string, commitish strin
...
+
+	err := ioutil.WriteFile(gitAttrFile, []byte(content), util.DefaultFileWritePermissions)
+	if err != nil {
+		return errors.Wrapf(err, "writing new repo-local git attributes to %s", gitAttrFile)
+	}
+	return nil
+}
+
+// ReadRepoAttributes reads the existing content, if any, in .git/info/attributes
+func (g *GitCLI) ReadRepoAttributes(dir string) (string, error) {
```
hferentschik: 👍

pkg/gits/git_cli.go
```diff
@@ -1263,3 +1263,32 @@ func (g *GitCLI) IsAncestor(dir string, possibleAncestor string, commitish strin
 	// Default case is that this is an ancestor, since there's no error from the merge-base call.
 	return true, nil
 }
+
+// WriteRepoAttributes writes the given content to .git/info/attributes
+func (g *GitCLI) WriteRepoAttributes(dir string, content string) error {
```
hferentschik: 👍

PR #6525, opened 2020-01-15
fix: secure the endpoints exposed by the DevPod

### LINKED ISSUE #6475: Secure the IDE exposed by devpods
### Summary 

The IDE that runs as part of a `jx create devpod` is exposed via a public ingress. Unauthorised access to the IDE should be prevented.

One possible solution to lock down access to the devpod IDE would be to front the IDE URL by http basic auth.

ccojocar:
#### Submitter checklist

#### Description

This contains the following changes:
- Make the `auto-expose` flag to behave consistently and set it by default to false. This results in not exposing any DevPod port by default.
- Check whether or not the TLS is enabled when the `auto-expose` flag is active, and ask for user confirmation in case it is disable in order to proceed with an insecure setup.
- Configure the basic authentication when the services are exposed.
- Configure the expose-controller to only recreate the ingress resources of the service created by the DevPod, before it was recreating all ingress resources from `jx` namespace. This will prevent that the other ingress resources are accidentally recreated.
- Augment the error messages with more context and also handle some errors which previously were ignored.

fixes #6475

### DIFF
--- a/pkg/cmd/create/create_devpod.go
+++ b/pkg/cmd/create/create_devpod.go
@@ -773,32 +773,38 @@ func (o *CreateDevPodOptions) Run() error {
 	o.NotifyProgress(opts.LogInfo, "Pod %s is now ready!\n", util.ColorInfo(pod.Name))
 	log.Logger().Infof("You can open other shells into this DevPod via %s", util.ColorInfo("jx create devpod"))
 
-	if !o.Sync && o.AutoExpose {
-		ideServiceURL, err := services.FindServiceURL(client, curNs, ideServiceName)
-		if err != nil {
-			return errors.Wrapf(err, "finding the URL for service %q", ideServiceName)
-		}
-		if ideServiceURL != "" {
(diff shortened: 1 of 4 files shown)


pkg/cmd/create/create_devpod.go
```diff
@@ -1091,7 +1152,7 @@ func FindDevPodLabelFromJenkinsfile(filename string, labels []string) (string, e
 						return a, nil
 					}
 				}
-				return answer, fmt.Errorf("Cannot find pipeline agent %s in the list of available DevPods: %s", a, strings.Join(labels, ", "))
+				return answer, fmt.Errorf("cannot find pipeline agent %s in the list of available DevPods: %s", a, strings.Join(labels, ", "))
```
hferentschik: geee - over 1100 lines of code.

hferentschik:
Except for the "hound fix" commit nicely crafted PR ;-)

/lgtm

PR #6535, opened 2020-01-15
fix: lets make the import deploy test not use maven

jstrachan:
to avoid issues downloading / running maven as part of the import


abayer:
/lgtm

I tend to fall back on the `golang-http` quickstart as my "need a project for a thing", but nodejs should be fine too, I think?

PR #6540, opened 2020-01-15
fix: Don't use anonymous GitHub API for dev env repo creation and get quickstarts

### LINKED ISSUE #6539: Anonymous GitHub API for creating dev env repo and get quickstarts is very vulnerable to rate limiting
With #6462, we switched to falling back to anonymous github.com API calls for getting information on the `jenkins-x-boot-config` repo and fetching the list of quickstart repos in `jx get quickstarts` when the git provider in use for the cluster is anything but github.com. Which works! Which is great! But is extremely vulnerable to GitHub's very draconian anonymous rate limiting by IP.  I think we probably should rework both of these cases to not use the API at all.

This should be fairly easy for the `jenkins-x-boot-config` case - I almost rewrote it to not use the API at the time, but since `jx get quickstarts` needed to do a repo listing anyway, well, I thought might as well stay with it. But given that I just hit rate limiting on three Lighthouse PRs at once at dev env creation, I think we need a better answer here, despite the fact that the `jx get quickstarts` case is hairier - for the default quickstarts, we can look to the version stream, so that may actually just be the right way to handle it, leaving non-standard quickstart listings to be handled via the configured provider's API.

abayer:
#### Submitter checklist

#### Description

As described in #6539, I forgot just how stingy the anonymous GitHub API rate limit is. So let's refactor both the `jx step verify environments` logic for getting the required info about the upstream `jenkins-x-boot-config` and `LoadQuickStartsModel`'s logic for loading quickstart information work to not use anonymous GitHub API calls after all. The required information should be available in both cases without needing to make those API calls, thanks to `https://github.com/jenkins-x/jenkins-x-boot-config.git` actually having everything that's needed in its own URL, and the default quickstarts being in the version stream.

Starting out on WIP to make sure the non-required `boot-lh-ghe` test...well, doesn't pass, since it'll run into https://github.com/jenkins-x/lighthouse/issues/462, but at least gets past that.

fixes #6539

### DIFF
--- a/pkg/cmd/step/verify/step_verify_environments.go
+++ b/pkg/cmd/step/verify/step_verify_environments.go
@@ -432,27 +432,19 @@ func (o *StepVerifyEnvironmentsOptions) createDevEnvironmentRepository(gitInfo *
 		log.Logger().Debugf("set commitish to '%s'", commitish)
 	}
 
-	var fromProvider gits.GitProvider
+	var fromRepo *gits.GitRepository
 	// If the to URL isn't github.com, and the fromGitURL is the default boot repository, use a non-authenticated github.com provider for the from provider.
 	if !gitInfo.IsGitHub() && isDefaultBootURL {
 		fromGitInfo, err := gits.ParseGitURL(fromGitURL)
 		if err != nil {
 			return nil, errors.Wrapf(err, "failed to parse upstream boot config URL %s", fromGitURL)
 		}
-		if fromGitInfo.IsGitHub() {
-			ghServer := &auth.AuthServer{
-				URL:  "https://github.com",
-				Name: "gh",
-				Kind: "github",
-			}
-			fromProvider, err = gits.NewAnonymousGitHubProvider(ghServer, gitter)
-			if err != nil {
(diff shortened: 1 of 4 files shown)


abayer:
`boot-lh-ghe` is failing because of https://github.com/jenkins-x/lighthouse/issues/465, so I'm just gonna un-WIP this for now.

pkg/cmd/step/verify/step_verify_environments.go
```diff
@@ -432,27 +432,19 @@ func (o *StepVerifyEnvironmentsOptions) createDevEnvironmentRepository(gitInfo *
...
-				Kind: "github",
-			}
-			fromProvider, err = gits.NewAnonymousGitHubProvider(ghServer, gitter)
-			if err != nil {
-				return nil, errors.Wrapf(err, "failed to configure anonymous GitHub provider for upstream boot config URL %s", fromGitURL)
-			}
-		}
+		fromRepo = fromGitInfo
+		fromRepo.CloneURL = fromGitURL
+		fromRepo.HTMLURL = strings.TrimSuffix(fromGitURL, ".git")
```
hferentschik: would be nice if we had this in GitRepository, but hey

PR #6546, opened 2020-01-16
fix: lets make step verify git github app aware

cagiti:
Let's make `jx step verify git` github app aware.


cagiti:
Most likely failing due to the spring http endpoints for nexus

PR #6556, opened 2020-01-17
chore: default spring version to 2.2.3 in `jx create spring`

garethjevans:
Currently defaulting to 2.2.3.BUILD.SNAPSHOT (possibly an initializer project problem)

#### Submitter checklist

fixes #


abayer:
/lgtm

We should have a followup issue for trying to work around this when the spring initializer gets messed up like this in the future.

PR #6569, opened 2020-01-21
fix: cmd_test_helpers - CreateTestJxHomeDir()

### LINKED ISSUE #6577: CreateTestJxHomeDir only creates tmp JX_HOME in case gitAuth.yaml exists
`CreateTestJxHomeDir` in `cmd_test_helpers.go` only creates a tmp JX_HOME directory in case the original/current JX_HOME contains a _gitAuth.yaml_:

```golang
originalDir, err := util.ConfigDir()
	if err != nil {
		return "", "", err
	}
	exists, err := util.FileExists(path.Join(originalDir, "gitAuth.yaml"))
	if err != nil {
		return "", "", err
	}
	if exists {
             // create tmp directory and set JX_HOME
            ...
	}
	return "", "", nil
```

That's really surprising behaviour, in particular since if _gitAuth.yaml_ does not exist, the function silently returns the empty string.

For this function to be useful to isolate test execution from the developer's default _JX_HOME_ directory it needs to create the tmp directory regardless. Also, files should not be copied from the existing/original JX_HOME, since there is no guarantee what exists. If a certain file is needed in the tmp JX_HOME, it needs to be placed there in a more predictable way.

nehagup:
fix: cmd_test_helpers - CreateTestJxHomeDir() returns empty string if gitAuth.yaml is not present.

fixes #6577

### DIFF
--- a/pkg/cmd/testhelpers/cmd_test_helpers.go
+++ b/pkg/cmd/testhelpers/cmd_test_helpers.go
@@ -151,35 +151,28 @@ func CreateTestJxHomeDir() (string, string, error) {
 	if err != nil {
 		return "", "", err
 	}
-	exists, err := util.FileExists(path.Join(originalDir, "gitAuth.yaml"))
+	newDir, err := ioutil.TempDir("", ".jx")
 	if err != nil {
 		return "", "", err
 	}
-	if exists {
-		newDir, err := ioutil.TempDir("", ".jx")
-		if err != nil {
-			return "", "", err
-		}
-		contents, err := ioutil.ReadDir(originalDir)
-		if err != nil {
-			return "", "", err
-		}
-		for _, f := range contents {
-			if strings.HasSuffix(f.Name(), ".yaml") {
-				err = util.CopyFileOrDir(path.Join(originalDir, f.Name()), path.Join(newDir, f.Name()), true)
-				if err != nil {
-					return "", "", err
-				}
+	contents, err := ioutil.ReadDir(originalDir)
+	if err != nil {
+		return "", "", err
+	}
+	for _, f := range contents {
+		if strings.HasSuffix(f.Name(), ".yaml") {
+			err = util.CopyFileOrDir(path.Join(originalDir, f.Name()), path.Join(newDir, f.Name()), true)
+			if err != nil {
+				return "", "", err
 			}
 		}
-		err = os.Setenv("JX_HOME", newDir)
-		if err != nil {
-			os.Unsetenv("JX_HOME")
-			return "", "", err
-		}
-		return originalDir, newDir, nil
 	}
-	return "", "", nil
+	err = os.Setenv("JX_HOME", newDir)
+	if err != nil {
+		os.Unsetenv("JX_HOME")
+		return "", "", err
+	}
+	return originalDir, newDir, nil
 }
 
 // CleanupTestJxHomeDir should be called in a deferred function whenever CreateTestJxHomeDir is called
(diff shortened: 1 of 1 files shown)


hferentschik:
@nehagup could you do me a favour and create an issue for this as well and then reference this issue in the commit message.

hferentschik:
@nehagup, I am also wondering whether you have accidentally changed your global git author setting? You commit was made as 'jx-helloworld'. You might want to rectify that as well.

pkg/cmd/testhelpers/cmd_test_helpers.go
```diff
@@ -151,35 +151,28 @@ func CreateTestJxHomeDir() (string, string, error) {
 	if err != nil {
 		return "", "", err
```
hferentschik: probably good to wrap the error and add a custom error string. Makes it easier to trace in case this error gets ever triggered.

pkg/cmd/testhelpers/cmd_test_helpers.go
```diff
@@ -151,35 +151,28 @@ func CreateTestJxHomeDir() (string, string, error) {
...
-				err = util.CopyFileOrDir(path.Join(originalDir, f.Name()), path.Join(newDir, f.Name()), true)
-				if err != nil {
-					return "", "", err
-				}
+	contents, err := ioutil.ReadDir(originalDir)
+	if err != nil {
+		return "", "", err
+	}
+	for _, f := range contents {
+		if strings.HasSuffix(f.Name(), ".yaml") {
```
hferentschik: hmmm, I am not sure whether we should copy anything from the existing JX_HOME directory. I don't see how we could reliably write tests against this, since everyone might have different content in his JX_HOME. In pipeline execution, I would expect the initial JX_HOME to be empty anyways. 

If a specific file is needed in this tmp JX_HOME test directory is should be part of the test setup to make sure it gets placed there. 

What happens if we remove this whole content handling?

pkg/cmd/testhelpers/cmd_test_helpers.go
```diff
@@ -151,35 +151,28 @@ func CreateTestJxHomeDir() (string, string, error) {
...
-		if err != nil {
-			os.Unsetenv("JX_HOME")
-			return "", "", err
-		}
-		return originalDir, newDir, nil
 	}
-	return "", "", nil
+	err = os.Setenv("JX_HOME", newDir)
+	if err != nil {
+		os.Unsetenv("JX_HOME")
```
hferentschik: Technically it would probably be more accurate to call `os.Getenv`  prior to the call of `os.Setenv` to get the original value of JX_HOME and to reset to this value on error.

hferentschik:
@nehagup - I created the issue - https://github.com/jenkins-x/jx/issues/6577

pkg/cmd/testhelpers/cmd_test_helpers.go
nehagup: Makes sense. I included it as it was already part of the helper, so I was wondering if it's expected to behave this way.

hferentschik: Sure. I understand that this was how it already was, but the aim is to make it better now that we are already making changes ;-)

nehagup:
> @nehagup, I am also wondering whether you have accidentally changed your global git author setting? You commit was made as 'jx-helloworld'. You might want to rectify that as well.

Corrected it! Thanks.

Also, can you please send me the Tekton error.

pkg/cmd/testhelpers/cmd_test_helpers.go
```diff
@@ -149,37 +151,20 @@ func MockFactoryFakeClients(mockFactory *clients_test.MockFactory) {
 func CreateTestJxHomeDir() (string, string, error) {
```
hferentschik: Good change looking good now. Could you do me another favour and adjust the Godoc, getting rid of "copying over any existing config"

nehagup: Sure. Thanks.

PR #6575, opened 2020-01-21
fix: Look up SourceRepositorys by labels, not just name.

### LINKED ISSUE #6545: Duplicated SourceRepository resources
### Summary 

I "manually" created sourcerepositories by adding files in `repositories/templates` in the git repo for the dev env. When the first pipeline ran, it created a duplicated resource for my repository, resulting in an invalid hook configuration.

- add a sourcerepository in the `repositories/templates` directory of the git repo for the dev env. name it after the repo. push/merge/...
- trigger a first pipeline run for your application repo. it will automatically create a new SourceRepository resource in the cluster, named `${owner}-${repo}`, which duplicates the existing one - but with a different name
- this generates an invalid configuration for `hook`, with duplicated entries in the `config` configmap:

```
owner/repo:
      - agent: tekton
        always_run: true
        context: pr-build
        name: pr-build
        rerun_command: /test this
        trigger: (?m)^/test( all| this),?(\s+|$)
      - agent: tekton
        always_run: true
        context: pr-build
        name: pr-build
        rerun_command: /test this
        trigger: (?m)^/test( all| this),?(\s+|$)
```
- hook is full of errors such as
```
{"component":"hook","error":"duplicated presubmit job: pr-build","jobConfig":"","level":"error","msg":"Error loading config.","prowConfig":"/etc/config/config.yaml"...}
```

### Expected behavior

- the second resource for the same repo should not have been created
- the invalid hook config should not have been generated

### Actual behavior

- duplicated SourceRepository resources - which in itself is not really the issue, just confusing
- invalid hook configuration

### Jx version

```
NAME               VERSION
jx                 2.0.1120
Kubernetes cluster v1.15.4-gke.22
kubectl            v1.15.7
git                2.24.0
Operating System   Mac OS X 10.12.6 build 16G2136
```

### Jenkins type

- [ ] Classic Jenkins

### Kubernetes cluster

GKE, using jx boot

macos

### More informations

see this slack thread: https://kubernetes.slack.com/archives/C9MBGQJRH/p1579207114237400

abayer:
#### Submitter checklist

#### Description

If you add a `SourceRepository` to your dev env repo by hand, you can specify an arbitrary name, but then when the pipeline for that repository runs, it will try to look for the repo's `SourceRepository` based on its "standard" naming scheme of `organization-repo`. When it doesn't find a `SourceRepository` with that name, it will create a new one instead with that name, resulting in duplicate `SourceRepository`s for that provider/org/repo combination.

So let's first try to look up by name, and then by labels.

fixes #6545

### DIFF
--- a/pkg/cmd/start/start_pipeline.go
+++ b/pkg/cmd/start/start_pipeline.go
@@ -204,7 +204,7 @@ func (o *StartPipelineOptions) createMetaPipeline(jobName string) error {
 		return errors.Wrap(err, "failed to create JX client")
 	}
 
-	sr, err := kube.FindSourceRepository(jxClient, ns, owner, repo)
+	sr, err := kube.FindSourceRepository(jxClient, ns, owner, repo, "", false)
 	if err != nil {
 		return errors.Wrap(err, "cannot determine git source URL")
 	}
--- a/pkg/cmd/start/start_pipeline.go
+++ b/pkg/cmd/start/start_pipeline.go
@@ -204,9 +204,12 @@ func (o *StartPipelineOptions) createMetaPipeline(jobName string) error {
 		return errors.Wrap(err, "failed to create JX client")
 	}
 
-	sr, err := kube.FindSourceRepository(jxClient, ns, owner, repo, "", false)
+	sr, err := kube.FindSourceRepositoryWithoutProvider(jxClient, ns, owner, repo)
 	if err != nil {
 		return errors.Wrap(err, "cannot determine git source URL")
+	}
+	if sr == nil {
+		return fmt.Errorf("could not find existing SourceRepository for owner %s and repo %s", owner, repo)
 	}
 
 	sourceURL, err := kube.GetRepositoryGitURL(sr)
--- a/pkg/kube/sourcerepositories.go
+++ b/pkg/kube/sourcerepositories.go
@@ -12,6 +12,7 @@ import (
 	"github.com/jenkins-x/jx/pkg/kube/naming"
 	"github.com/jenkins-x/jx/pkg/util"
 	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -39,13 +40,40 @@ func GetRepositoryGitURL(s *v1.SourceRepository) (string, error) {
 	return spec.HTTPCloneURL, nil
 }
 
-// FindSourceRepository returns a SourceRepository for the given namespace, owner and name
-// or returns an error if it cannot be found
-func FindSourceRepository(jxClient versioned.Interface, ns string, owner string, name string) (*v1.SourceRepository, error) {
+// FindSourceRepository returns a SourceRepository for the given namespace, owner, repo name, and (optional) provider name.
+// If no SourceRepository is found, return an error if doNotFailIfNotFound is false, otherwise just return a nil repo.
+func FindSourceRepository(jxClient versioned.Interface, ns string, owner string, name string, providerName string, doNotFailIfNotFound bool) (*v1.SourceRepository, error) {
 	resourceName := naming.ToValidName(owner + "-" + name)
+	var errToReturn error
 	repo, err := jxClient.JenkinsV1().SourceRepositories(ns).Get(resourceName, metav1.GetOptions{})
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to find SourceRepository %s in namespace %s", resourceName, ns)
+		if apierrors.IsNotFound(err) {
+			// Wipe the repo, since Get returns an empty SourceRepository, not a nil one, if not found
+			repo = nil
+
+			labelSelector := fmt.Sprintf("%s=%s,%s=%s", v1.LabelOwner, owner, v1.LabelRepository, name)
+			if providerName != "" {
+				labelSelector += fmt.Sprintf(",%s=%s", v1.LabelProvider, providerName)
+			}
+
+			repos, err := jxClient.JenkinsV1().SourceRepositories(ns).List(metav1.ListOptions{
+				LabelSelector: labelSelector,
+			})
+			if err != nil {
+				errToReturn = err
+			}
+			if repos != nil && len(repos.Items) == 1 {
+				repo = &repos.Items[0]
+			}
+		} else {
+			errToReturn = err
+		}
+	}
+	if errToReturn != nil {
+		return nil, errors.Wrapf(errToReturn, "failed to find SourceRepository %s in namespace %s", resourceName, ns)
+	}
+	if repo == nil && !doNotFailIfNotFound {
+		return nil, fmt.Errorf("couldn't find SourceRepository for owner %s, repository %s, optional provider %s in namespace %s", owner, name, providerName, ns)
 	}
 	return repo, nil
 }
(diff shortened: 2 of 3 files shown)


pkg/cmd/start/start_pipeline.go
```diff
@@ -204,7 +204,7 @@ func (o *StartPipelineOptions) createMetaPipeline(jobName string) error {
 		return errors.Wrap(err, "failed to create JX client")
 	}
 
-	sr, err := kube.FindSourceRepository(jxClient, ns, owner, repo)
+	sr, err := kube.FindSourceRepository(jxClient, ns, owner, repo, "", false)
```
hferentschik: Why the empty string here? Having to pass empty strings is always confusing imo.

pkg/kube/sourcerepositories.go
```diff
@@ -39,13 +40,40 @@ func GetRepositoryGitURL(s *v1.SourceRepository) (string, error) {
 	return spec.HTTPCloneURL, nil
 }
 
-// FindSourceRepository returns a SourceRepository for the given namespace, owner and name
-// or returns an error if it cannot be found
-func FindSourceRepository(jxClient versioned.Interface, ns string, owner string, name string) (*v1.SourceRepository, error) {
+// FindSourceRepository returns a SourceRepository for the given namespace, owner, repo name, and (optional) provider name.
+// If no SourceRepository is found, return an error if doNotFailIfNotFound is false, otherwise just return a nil repo.
+func FindSourceRepository(jxClient versioned.Interface, ns string, owner string, name string, providerName string, doNotFailIfNotFound bool) (*v1.SourceRepository, error) {
```
hferentschik: I would think we had a struct for repo provider + ower + name, but that is maybe not the case. There is `GitRepository` but that's too broad. Maybe we should add such a struct or we split `GitRepositor` in some way.

why the `doNotFailIfNotFound`? What do you try to achieve? I think it makes for a bad API. 
Why can the method not behave like:

* return repo + nil in case the source repository could be found
* return nil + nil if it could not be found (no error)
* return nil + error if any other error than apierrors.IsNotFound is encountered 

Are you trying to be be somewhat backwards compatible? Maybe in that case adding a new method makes more sense? Rename `FindSourceRepository` to `FindSourceRepositoryFailNotFound` then add a new implementation of FindSourceRepository which behaves as discussed above and use it. Eventually, we can then try to switch FindSourceRepositoryFailNotFound over to FindSourceRepository.

This is a general issue which I ran a lot into as well. These type of Find and Get operations we have written, fail to distinguish between not found and any other error. This makes it super tricky  use this methods.

pkg/kube/sourcerepositories.go
```diff
@@ -39,13 +40,40 @@ func GetRepositoryGitURL(s *v1.SourceRepository) (string, error) {
...
 	resourceName := naming.ToValidName(owner + "-" + name)
+	var errToReturn error
 	repo, err := jxClient.JenkinsV1().SourceRepositories(ns).Get(resourceName, metav1.GetOptions{})
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to find SourceRepository %s in namespace %s", resourceName, ns)
+		if apierrors.IsNotFound(err) {
+			// Wipe the repo, since Get returns an empty SourceRepository, not a nil one, if not found
```
hferentschik: why not extract this into a new method `FindSourceRepositoryByLabel`. The function gets shorter and easier to understand and I think a  `FindSourceRepositoryByLabel` makes sense in its own right. Not to mention that you could test `FindSourceRepositoryByLabel` now in isolation.

pkg/kube/sourcerepositories.go
```diff
@@ -60,66 +88,78 @@ func GetOrCreateSourceRepositoryCallback(jxClient versioned.Interface, ns string
...
 
-		if callback != nil {
-			callback(copy)
-		}
-		copy.Sanitize()
-		if reflect.DeepEqual(copy.Spec, sr.Spec) && reflect.DeepEqual(copy.Labels, sr.Labels) {
-			return answer, nil
+	// If we did not find an existing SourceRepository for this org/repo, create one
+	if foundSr == nil {
+		labels := map[string]string{}
```
hferentschik: new function?

pkg/kube/sourcerepositories.go
abayer: Actually, there's just one place that called it expecting an error if not found, so I'll fix that.

pkg/cmd/start/start_pipeline.go
abayer: Changing to have a `FindSourceRepositoryWithoutProvider` function. It's moments like this that I miss method overloading in Java. =)

pkg/kube/sourcerepositories.go
abayer: Okiedokie.

abayer:
@hferentschik - tweaked - let me know what you think.

pkg/kube/sourcerepositories.go
```diff
@@ -39,13 +40,42 @@ func GetRepositoryGitURL(s *v1.SourceRepository) (string, error) {
 	return spec.HTTPCloneURL, nil
 }
 
-// FindSourceRepository returns a SourceRepository for the given namespace, owner and name
-// or returns an error if it cannot be found
-func FindSourceRepository(jxClient versioned.Interface, ns string, owner string, name string) (*v1.SourceRepository, error) {
+// FindSourceRepositoryWithoutProvider returns a SourceRepository for the given namespace, owner and repo name.
```
hferentschik: 👍 I am wondering though whether `FindSourceRepositoryByName` (menaing CRD name) would be a more fitting name. The pair of ``FindSourceRepositoryByName` and `FindSourceRepositoryByLabels` would then make more sense together.

pkg/kube/sourcerepositories.go
```diff
@@ -39,13 +40,42 @@ func GetRepositoryGitURL(s *v1.SourceRepository) (string, error) {
...
+// If no SourceRepository is found, return nil.
+func FindSourceRepositoryWithoutProvider(jxClient versioned.Interface, ns string, owner string, name string) (*v1.SourceRepository, error) {
+	return FindSourceRepository(jxClient, ns, owner, name, "")
+}
+
+// FindSourceRepositoryByLabels returns a SourceRepository matching the given label selector, if it exists.
+func FindSourceRepositoryByLabels(jxClient versioned.Interface, ns string, labelSelector string) (*v1.SourceRepository, error) {
```
hferentschik: I find it funny that in one case we have three pieces of information to identify a repo and in the other two. That seems odd to me. If the provider is needed, why is it not needed for the lookup by name? I guess once we would have multiple providers and we allow for using multiple providers in parallel, we can get ambiguity with `FindSourceRepositoryWithoutProvider`. 

My guess is that we don't add the provider to the CRD name, because that would make it even easier to hist the name length limit. Maybe we should get rid of `FindSourceRepositoryWithoutProvider` and place our bet on label lookup only?

Regarding, _labelSelector_, why putting the burden on how to build the selector on the caller? How does the caller know the right label format? Would you not pass _name_, _owner_ and _provider_ and let _FindSourceRepositoryByLabels_ sort this out?

pkg/kube/sourcerepositories.go
```diff
@@ -39,13 +40,42 @@ func GetRepositoryGitURL(s *v1.SourceRepository) (string, error) {
...
+		return nil, errors.Wrapf(err, "listing SourceRepositorys matching label selector %s in namespace %s", labelSelector, ns)
+	}
+	if repos != nil && len(repos.Items) == 1 {
+		return &repos.Items[0], nil
+	}
+	return nil, nil
+}
+
+// FindSourceRepository returns a SourceRepository for the given namespace, owner, repo name, and (optional) provider name.
+// If no SourceRepository is found, return nil.
```
hferentschik: 👍 Nice one. Now let's fix all our resource find functions ;-)

pkg/kube/sourcerepositories.go
```diff
@@ -56,70 +86,57 @@ func GetOrCreateSourceRepositoryCallback(jxClient versioned.Interface, ns string
...
-			Provider:     providerURL,
-			ProviderName: providerName,
-			Repo:         name,
-		},
+	foundSr, err := FindSourceRepository(jxClient, ns, organisation, name, providerName)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to find existing SourceRepository")
 	}
+
+	// If we did not find an existing SourceRepository for this org/repo, create one
```
hferentschik: It beats me why we would use a callback here? any ideas?

pkg/kube/sourcerepositories.go
abayer: Basically, this is because `jx start pipeline` doesn’t have the provider URL available yet when it calls this stuff - in fact, it uses the `SourceRepository` to find the provider URL! I’m probably overdoing things here with the provider to hedge for future issues, if/when we properly support multiple providers on a single cluster, but the more I think about it, the more I think I’m probably making it more complicated than it needs to be. I’ll remove the provider from the label and change the function to take org and repo as its arguments.

pkg/kube/sourcerepositories.go
abayer: Basically, in most cases, the caller just needs a fairly basic set to be set in the SR, but when import’s `addProwConfig` calls it at https://github.com/jenkins-x/jx/blob/d3105b309c55d150328bfe8dd7f7f0081f566d1c/pkg/cmd/importcmd/import.go#L1036, it needs to set more information that isn’t necessarily available yet in other cases.
 
I thought about changing other callers to just lookup, since at this point, the SR should actually always already exist, but because of manually specified SRs in the dev env repo, we can’t be 100% sure those SRs are properly labeled yet, so those fields are ones we do need to ensure are there and add if they’re not.

pkg/kube/sourcerepositories.go
abayer: Fine by me - I have an unconscious tendency to try retain existing names even when that’s silly.

abayer:
@hferentschik - Actually, I'm going to leave this all as is. `FindSourceRepositoryByName` isn't the right name for what's now `FindSourceRepositoryWithoutProvider`, since the function just calls `FindSourceRepository` with an empty string for provider and then everything else behaves the same as if you'd called `FindSourceRepository` directly. `FindSourceRepositoryByLabel` isn't called anywhere outside of `FindSourceRepository` currently, and is only a separate function because you asked for it to be one for cleanliness - which is fine! Just that it's not a particularly downstream-facing API. In fact, the one change I will make is to make it package-private.

pkg/kube/sourcerepositories.go
hferentschik: you would not need a callback for that, right? anyways

PR #6581, opened 2020-01-22
fix: Use requirements gitServer as default for import, create env

### LINKED ISSUE #6579: Git Authentication fails on GHE
### Summary
I installed Jenkins X with the latest `jx` binary and jx-boot-config
and afterwards `jx import` was failing due to bad credentials,
here is what I found out.

After nuking everything to replicate the error I tried to delete a repository before
even running `jx import` with `jx delete repo`, there I was asked if I want to use the  git bot-user for that, which I previously defined in `jx boot`, or some other user.

Here is the first issue, now that I chose a different user, without even being prompted to enter OAuth token, some other OAuth token is apparently used by default, as I receive this error:

```
? Do you wish to use jx-bot-serviceuser as the Git user name No
? Git user name I519840
? Which organisation do you want to use? I519840
Owner of repo is same as username, using GitHub API for Users
error: GET https://github.tools.sap/api/v3/users/I519840/repos?per_page=100: 401 Bad credentials []
```

The issue really kicks in when running `jx import` though, here I `can` enter the real OAuth token, though it is still not used?

Therefore I receive this error:

```
WARNING: No username defined for the current Git server!
? Do you wish to use I519840 as the Git user name: Yes
trying to use draft pack: none
ERROR: Could not find pack: none going to try detect which pack to use
selected pack: /Users/i519840/.jx/draft/packs/github.com/jenkins-x-buildpacks/jenkins-x-kubernetes/packs/maven
replacing placeholders in directory /Users/i519840/Code/jx_related_repos/jx-test
app name: jx-test, git server: github.com, org: i519840, Docker registry org: sap-cpe-cloudpot-dev
skipping directory "/Users/i519840/Code/jx_related_repos/jx-test/.git"
WARNING: organisation I519840 detected from URL https://github.tools.sap/I519840/jx-test.git. '--org i519840' will be ignored
WARNING: Querying webhooks on I519840/jx-test: GET https://api.github.com/repos/I519840/jx-test/hooks: 401 Bad credentials []
Creating GitHub webhook for I519840/jx-test for url http://hook-jx.35.204.94.235.nip.io/hook
error: unable to create webhook for I519840/jx-test - permission denied: POST https://api.github.com/repos/I519840/jx-test/hooks: 403 API rate limit of 60 still exceeded until 2020-01-22 11:43:03 +0100 CET, not making remote request. [rate reset in 19m35s]
```

I tried to curl the same url with my OAuth token to verify that it's not my token's fault and in fact, the curl was working, which means that the wrong token is being used for authentication.

Run `jx boot` with the latest CLI, afterwards try importing a repository which is created outside of the org, preferably your normal user account.

Alternativly, run `jx delete repo` and choose a different user than the default.

### Jx version

```
jx                 2.0.1142
kubectl            v1.17.1
git                2.25.0
```

### Jenkins type

- [X] Serverless Jenkins X Pipelines (Tekton + Lighthouse)
- [ ] Classic Jenkins

Mac OS X 10.15.2 build 19C57

abayer:
#### Submitter checklist

#### Description

`opts.AddGitRepoOptionsArguments` defaults `--git-provider-url` to `https://github.com`, no matter what the actual default should be for the cluster. This makes sense for some of the cases where it's used (i.e., `jx install`, `jx create cluster`, `jx create terraform`, etc), but not for others (`jx import` (and therefore `jx create quickstart`), `jx edit env`, and `jx create env`). In those cases, we should default it to the `gitServer` in the team settings.

Except we don't actually set that with `jx boot`! It just stays on `https://github.com`. So we also need to set that from the requirements.

fixes #6579

### DIFF
--- a/pkg/cmd/create/install.go
+++ b/pkg/cmd/create/install.go
@@ -398,7 +398,7 @@ func (options *InstallOptions) AddInstallFlags(cmd *cobra.Command, includesInit 
 	cmd.Flags().BoolVar(&options.SkipAuthSecretsMerge, opts.OptionSkipAuthSecMerge, false, "Skips merging the secrets from local files with the secrets from Kubernetes cluster")
 
 	bindInstallConfigToFlags(cmd)
-	opts.AddGitRepoOptionsArguments(cmd, &options.GitRepositoryOptions)
+	opts.AddGitRepoOptionsArgumentsWithGithubDefault(cmd, &options.GitRepositoryOptions)
 	options.HelmValuesConfig.AddExposeControllerValues(cmd, true)
 	options.AdminSecretsService.AddAdminSecretsValues(cmd)
 	options.InitOptions.AddInitFlags(cmd)
--- a/pkg/cmd/opts/git.go
+++ b/pkg/cmd/opts/git.go
@@ -145,9 +145,19 @@ func (o *CommonOptions) DiscoverGitURL(gitConf string) (string, error) {
 	return url, nil
 }
 
-// AddGitRepoOptionsArguments adds common git flags to the given cobra command
+// AddGitRepoOptionsArgumentsWithGithubDefault adds common git flags with a default provider URL of github.com to the given cobra command
+func AddGitRepoOptionsArgumentsWithGithubDefault(cmd *cobra.Command, repositoryOptions *gits.GitRepositoryOptions) {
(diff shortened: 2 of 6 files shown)


abayer:
Marking as WIP because I think I want to add some unit tests.

PR #6590, opened 2020-01-22
fix: Lookup git service URL kind from jx-auth-config as well

### LINKED ISSUE #6589: BitBucket Server with boot/gitops/vault gets error for "No git service found"
```
create spring
 Given valid parameters when running ** create spring 
  creates a spring application and promotes it to staging
  
  /tmp/jenkins-x/bdd-**/test/suite/spring/**_create_spring.go:36
      Using Git provider URL https://bitbucket.beescloud.com
STEP: calling ** create spring -b --org **bdd --artifact bdd-spring-1579611559 --name bdd-spring-1579611559 -d web -d actuator --git-provider-url https://bitbucket.beescloud.com
      RUNNER:About to execute ** create spring -b --org **bdd --artifact bdd-spring-1579611559 --name bdd-spring-1579611559 -d web -d actuator --git-provider-url https://bitbucket.beescloud.com in /tmp/bdd-953564943 with timeout 1h0m0s expecting exit code 0
Command "spring" is deprecated, it will be removed on Feb 1 2020. We now highly recommend you use ** create quickstart instead. Please check https://jenkins-x.io/docs/getting-started/first-project/create-quickstart/ for more details.
Created Spring Boot project at /tmp/bdd-953564943/bdd-spring-1579611559
error: no Git service resource found with URL 'https://bitbucket.beescloud.com' in namespace **
      RUNNER:Execution completed with exit code 1
      FAILED got unexpected error: 
expected exit code 0 but got 1 whilst running command ** create spring -b --org **bdd --artifact bdd-spring-1579611559 --name bdd-spring-1579611559 -d web -d actuator --git-provider-url https://bitbucket.beescloud.com
```

This is another side effect of the move away from secrets to the `jx-auth-config` configmap. When there's no explicit kind for the server during `jx import` and friends, we first check to see if the URL matches a known pattern for providers supported by the SaaS, then we check for `jx-pipeline-git-` secrets, and lastly we check the at-this-point-defunct list of `GitService`s. GitHub, BitBucket Cloud, and gitlab.com all get their kind detected by the SaaS provider check, as does GitHub Enterprise _if_ the service URL starts with `https://github`. Anything else (i.e., any self-hosted Gitlab or BitBucket Server, and any self-hosted GHE that isn't on `https://github(something)`) just won't get detected.

abayer:
#### Submitter checklist

#### Description

We're looking for this in a simple set of patterns, secrets, and `GitServices`, but once the `jx-auth-config` configmap is present (i.e., via a combination of boot/vault/gitops), we won't be able to find the kind for something that doesn't match one of the patterns we check first, since the `jx-pipeline-git-` secrets won't exist any more. So if we're in a cluster with `jx-auth-config`, let's look up there too, before we check secrets or `GitServices`.

fixes #6589

### DIFF
--- a/pkg/cmd/opts/git.go
+++ b/pkg/cmd/opts/git.go
@@ -182,7 +182,19 @@ func (o *CommonOptions) GitServerHostURLKind(hostURL string) (string, error) {
 		return "", err
 	}
 
-	kind, err := kube.GetGitServiceKind(jxClient, kubeClient, devNs, hostURL)
+	var clusterAuthConfig *auth.AuthConfig
+
+	// Pass the in-cluster git auth config to GetGitServiceKind
+	clusterAuthConfigSvc, err := o.GitAuthConfigService()
+	if err != nil {
+		return "", errors.Wrapf(err, "getting cluster-based auth configmap for checking kind of git url %s", hostURL)
+	}
+	if clusterAuthConfigSvc != nil {
+		clusterAuthConfig = clusterAuthConfigSvc.Config()
+	}
+
+	kind, err := kube.GetGitServiceKind(jxClient, kubeClient, devNs, clusterAuthConfig, hostURL)
+
 	if err != nil {
 		return kind, err
 	}
--- a/pkg/kube/git_services.go
+++ b/pkg/kube/git_services.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/jenkins-x/jx/pkg/auth"
 	"github.com/jenkins-x/jx/pkg/kube/naming"
 	"github.com/jenkins-x/jx/pkg/log"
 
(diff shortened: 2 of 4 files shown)


abayer:
/assign @ccojocar 
/assign @cagiti 

@cagiti - thought you should probably give this a look to make sure it doesn't cause issues with the github app mode stuff. I don't think it should - we're not using this for actual cred lookups, just the kind, and it'll only apply if the `jx-auth-config` configmap is present.

pkg/cmd/opts/git.go
```diff
@@ -172,7 +172,20 @@ func (o *CommonOptions) GitServerHostURLKind(hostURL string) (string, error) {
 		return "", err
 	}
 
-	kind, err := kube.GetGitServiceKind(jxClient, kubeClient, devNs, hostURL)
+	var clusterAuthConfig *auth.AuthConfig
+
+	// If the jx-auth-config configmap exists, create its config and pass it to GetGitServiceKind
+	configMapClient := kubeClient.CoreV1().ConfigMaps(devNs)
```
ccojocar: Why is the configmap check required here? 

We should use this logic for all cases:
```
clusterAuthConfigSvc, err := o.GitAuthConfigService()
		if err != nil {
			return "", errors.Wrapf(err, "getting cluster-based auth configmap for checking kind of git url %s", hostURL)
		}
		clusterAuthConfig = clusterAuthConfigSvc.Config()
```
The auth config service is meant to abstract the kubernetes check away. It does check for the existence of the configmap and creates the appropriate auth config service: See `pkg/cmd/clients/factory.go: CreateAuthConfigService`. When the secret is there, it will create the kuberntes auth service which parses the kind from the secret, so we don't need to perform this separately.

abayer: Ok. I was just being super defensive here.

PR #6594, opened 2020-01-23
feat: add a command to get the custom resource counts in a cluster

### LINKED ISSUE #6596: knowing how many resources exist for CRD's is useful when maintaining cluster, a command to list these counts would be helpful

rawlingsj:
There are lots of custom resources that can be created by various applications and it can be hard to keep a handle of how many exist.  This can lead to them growing over time if not managed.  This command allows users to get the counts for each custom resource given their permissions.

Example:
```
➜  ~ jx get crd count
looking for cluster wide custom resource counts and namespaced custom resources in these namespaces:
cert-manager
default
jx
jx-production
jx-staging
kube-node-lease
kube-public
kube-system
this operation may take a while depending on how many custom resources exist
NAME                                         VERSION  COUNT NAMESPACE
apps.jenkins.io                              v1       0     cert-manager
apps.jenkins.io                              v1       0     default
workflows.jenkins.io                         v1       0     kube-public
releases.jenkins.io                          v1       365   jx
pipelineactivities.jenkins.io                v1       1183  jx
prowjobs.prow.k8s.io                         v1       1270  jx
pipelinestructures.jenkins.io                v1       1513  jx
```

Future work could integrate with prometheus for alerting. 

fixes #6596

### DIFF
--- a/pkg/cmd/get/get_crd_counts.go
+++ b/pkg/cmd/get/get_crd_counts.go
@@ -1,0 +1,165 @@
+package get
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+
+	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/util"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/jenkins-x/jx/pkg/cmd/helper"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/jenkins-x/jx/pkg/cmd/opts"
+	"github.com/jenkins-x/jx/pkg/cmd/templates"
+)
+
+// CRDCountOptions the command line options
+type CRDCountOptions struct {
+	*opts.CommonOptions
+}
+
+type tableLine struct {
+	name      string
+	version   string
+	count     int
+	namespace string
+}
+
+var (
+	getCrdCountLong = templates.LongDesc(`
+		Count the number of resources for all custom resources definitions
+
+`)
+
+	getCrdCountExample = templates.Examples(`
+
+		# Count the number of resources for all custom resources definitions
+		jx get crd count
+	`)
+)
+
+// NewCmdGetCRDCount creates the command object
+func NewCmdGetCRDCount(commonOpts *opts.CommonOptions) *cobra.Command {
+	options := &CRDCountOptions{
+		CommonOptions: commonOpts,
+	}
+
+	cmd := &cobra.Command{
+		Use:     "crd count",
+		Short:   "Display resources count for all custom resources",
+		Long:    getCrdCountLong,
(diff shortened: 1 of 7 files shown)


pkg/cmd/get/get_crd_counts.go
```diff
@@ -0,0 +1,165 @@
...
+		table.AddRow(r.name, r.version, strconv.Itoa(r.count), r.namespace)
+	}
+
+	table.Render()
+	return nil
+}
+
+func (o *CRDCountOptions) getCustomResourceCounts() ([]tableLine, error) {
+
+	exClient, err := o.ApiExtensionsClient()
```
hferentschik: What's about extracting this into a package (crd?) outside of cmd structure, making it potentially reusable from other places. It could return a slice of CRDCount or similar which you then render into a table?

go.mod
```diff
@@ -96,6 +96,7 @@ require (
 	github.com/pierrec/lz4 v2.0.5+incompatible // indirect
 	github.com/pkg/browser v0.0.0-20170505125900-c90ca0c84f15
 	github.com/pkg/errors v0.8.1
+	github.com/prometheus/common v0.0.0-20181126121408-4724e9255275
```
hferentschik: where is this coming from?

hferentschik:
I'd squash that into a single commit except for the dynamic client one (maybe)

rawlingsj:
> I'd squash that into a single commit except for the dynamic client one (maybe)

Happy to and are we applying that everywhere?  I've been noticing a LOT of PRs lately with a large number of tiny one word changes per commit.

go.mod
rawlingsj: probably the switch from using the sonaboy dynamic client to the kubernetes one, I can check

pkg/cmd/get/get_crd_counts.go
rawlingsj: Feels ok here to me until there's a time for reuse TBH.  If it concerns you a lot then sure I can do but I prefer not to unless there's real cases of reuse, I dont see any yet mind.

go.mod
rawlingsj: just checked and this can be removed, pushing that now

rawlingsj:
> I'd squash that into a single commit except for the dynamic client one (maybe)

> Happy to and are we applying that everywhere? I've been noticing a LOT of PRs lately with a large number of tiny one word changes per commit.

There were conflicts trying to squash into two so I just squashed into 1 commit.

abayer:
Consider this a bit of a devil’s advocate - why do we need this as a jx command? Is adding another jx command the right way to address the need for CRD counts? Not saying I’m against adding this, just that I’d like to make sure we ask that question before we add any new commands, basically.

Also, issue please? =)

rawlingsj:
> Consider this a bit of a devil’s advocate - why do we need this as a jx command? Is adding another jx command the right way to address the need for CRD counts? Not saying I’m against adding this, just that I’d like to make sure we ask that question before we add any new commands, basically.

folks in the SaaS team have been using a script to do the same which is very useful, this just makes it easier for anyone to keep an eye on these numbers.

abayer:
@rawlingsj - fair enough! I would still very much appreciate an issue, though. =)

rawlingsj:
https://github.com/jenkins-x/jx/issues/6596

abayer:
/lgtm

the `boot-vault` failure was just a flaky one in the app test, should be fine next run.

Issue #6555, opened 2020-01-17
Vault creation for AWS fails with null pointer exception


Painyjames:
### Summary 
Cannot run `jx create vault` on an existing jenkins x installation.

### Steps to reproduce the behavior

```sh
AWS_DEFAULT_REGION=eu-west-2 jx create vault jx-vault --aws-access-key-id $AWS_ACCESS_KEY_ID --aws-secret-access-key $AWS_SECRET_ACCESS_KEY --cluster-name fb -n jx  --aws-s3-region=eu-west-2 --aws-auto-create=true --aws-kms-region=eu-west-2  --aws-dynamodb-region=eu-west-2 --verbose
```

### Expected behavior
Vault being created successfully

### Actual behavior
```
DEBUG: vault jx-vault not found in namespace jx, err is the server could not find the requested resource (get vaults.vault.banzaicloud.com jx-vault)
DEBUG: Current configuration dir: /home/victor/.jx
DEBUG: VersionRepository: https://github.com/jenkins-x/jenkins-x-versions.git git ref: master
DEBUG: ? A local Jenkins X versions repository already exists, pulling the latest: Yes
DEBUG: Created service account 'jx-vault-auth-sa' for Vault authentication
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x68 pc=0x2cc2a77]

goroutine 1 [running]:
github.com/jenkins-x/jx/pkg/vault/create.(*defaultVaultCreator).applyDefaultRegionIfEmpty(0xa5b9d00, 0x0, 0xc000050013, 0x9)
	/workspace/source/pkg/vault/create/create.go:440 +0x37
github.com/jenkins-x/jx/pkg/vault/create.(*defaultVaultCreator).vaultAWSConfig(0xa5b9d00, 0xc0007fd880, 0x7ffe406ce00c, 0x8, 0x7ffe406ce08e, 0x2, 0x7ffe406ce094, 0x2, 0x0, 0x0, ...)
	/workspace/source/pkg/vault/create/create.go:304 +0xa4
github.com/jenkins-x/jx/pkg/vault/create.(*defaultVaultCreator).setCloudProviderSpecificSettings(0xa5b9d00, 0xc0007fd880, 0x7ffe406ce00c, 0x8, 0x7ffe406ce08e, 0x2, 0x7ffe406ce094, 0x2, 0x0, 0x0, ...)
[...]

PR #6602, opened 2020-01-23
fix: validate AWS parameters to prevent panics

RafalKorepta:
Signed-off-by: Rafal Korepta <rafal.korepta@gmail.com>

#### Submitter checklist

#### Description

Issue #6555 shows that cloud provider AWS panic when a user wants to create a vault.

This PR remove panic from the jx program flow and handle gracefully the not valid parameters. The main issue isn't resolved yet.

fixes #

### DIFF
--- a/pkg/vault/create/create.go
+++ b/pkg/vault/create/create.go
@@ -4,11 +4,10 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/jenkins-x/jx/pkg/cloud/amazon/session"
-
 	"github.com/banzaicloud/bank-vaults/operator/pkg/apis/vault/v1alpha1"
 	"github.com/banzaicloud/bank-vaults/operator/pkg/client/clientset/versioned"
 	"github.com/jenkins-x/jx/pkg/cloud"
+	"github.com/jenkins-x/jx/pkg/cloud/amazon/session"
 	awsvault "github.com/jenkins-x/jx/pkg/cloud/amazon/vault"
 	"github.com/jenkins-x/jx/pkg/cloud/gke"
 	gkevault "github.com/jenkins-x/jx/pkg/cloud/gke/vault"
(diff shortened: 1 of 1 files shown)


hferentschik:
@RafalKorepta thanks for the pull request and good catch :-) I am going to add a review.

pkg/vault/create/create.go
```diff
@@ -117,11 +116,23 @@ func (p *VaultCreationParam) validate() error {
 		}
 	}
 
+	if p.KubeProvider == cloud.AWS {
```
hferentschik: 👍

pkg/vault/create/create.go
```diff
@@ -117,11 +116,23 @@ func (p *VaultCreationParam) validate() error {
...
+			validationErrors = append(validationErrors, err)
+		}
+	}
+
 	return util.CombineErrors(validationErrors...)
 }
 
 func (p *GKEParam) validate() error {
 	var validationErrors []error
+	if p == nil {
```
hferentschik: I don't think this is technically possible. _p_ is the instance pointer on which the _validate_ method is called on. If this instance pointer is _nil_, you are getting a panic directly, not in this function.

That said, we could add some unit tests for `GKEParam` and `AWSParam`. WDYT?

pkg/vault/create/create.go
```diff
@@ -131,6 +142,23 @@ func (p *GKEParam) validate() error {
 	return util.CombineErrors(validationErrors...)
 }
 
+func (p *AWSParam) validate() error {
+	var validationErrors []error
+	if p == nil {
```
hferentschik: same as above

pkg/vault/create/create.go
ccojocar: In Go is possible to call a method on a null pointer receiver. It is perfectly allowed. Actually sometimes is a good practice to check the receiver for nil before doing anything into that method. https://medium.com/@mehdy/the-story-of-a-nil-pointer-receiver-b530e00a5f11

pkg/vault/create/create.go
RafalKorepta: same as above

pkg/vault/create/create.go
RafalKorepta: As @ccojocar mentioned the method on the nil pointer receiver can be executed. I will quote Dave Cheney 

> Methods in Go are just syntactic sugar for a function which passes the receiver as the first formal parameter.

Reference: https://dave.cheney.net/tag/receivers

The brief comment from Francesc Campoy Flores https://youtu.be/ynoY2xz-F8s?t=987 watch till 18:36 :) 

Regarding the unit tests, I could write them, but there is already `create_test.go` file which is an integration test with GKE. I could move `create_test.go` to `create_intergration_test.go`, so that `create_test.go` will have new unit tests. 

I'm not sure if CloudBees could have integration tests with AWS environment?

WDYT?  @ccojocar @hferentschik

hferentschik: Look at that. Thanks for filling this gap of mine.

PR #6611, opened 2020-01-24
fix: make PipelineResource names have a random suffix

### LINKED ISSUE #6610: PipelineResource names can be reused, causing collisions
I think this is causing issues with super-long organization names, where if `meta-(organization name)` is longer than 31 characters, the same `PipelineResource` ends up getting used by builds of multiple repositories which were kicked off at about the same time. This results in those multiple builds actually building the same repo/branch, which is obviously wrong. I've seen that symptom show up in some test runs with very long org names, and this would be a plausible reason for it happening.

abayer:
#### Submitter checklist

#### Description

We already do this with the other Tekton CRDs, so let's do it here too and hopefully avoid some edge case collisions where multiple builds of different repositories in an org with a very long name kick off at nearly the exact same moment, and they end up both building the same repo/branch, not different repos.

Also adds owner references to `PipelineResource` so it gets GCed with `PipelineActivity`s.

/assign @hferentschik 

fixes #6610

### DIFF
--- a/pkg/cmd/step/create/step_create_task.go
+++ b/pkg/cmd/step/create/step_create_task.go
@@ -302,9 +302,7 @@ func (o *StepCreateTaskOptions) Run() error {
 		return errors.Wrap(err, "Unable to load pod templates")
 	}
 
-	// resourceName is shared across all builds of a branch, while the pipelineName is unique for each build.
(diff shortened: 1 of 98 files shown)


abayer:
I'm about 95% sure this was the root cause of a weird glitch I've been seeing on Lighthouse tests, which use a 32 character long org, where sometimes for no initially obvious reason, the initial spring build would end up actually running the dev env master build. Which doesn't seem to be happening when using an image with this fix. So...yay?

Issue #6442, opened 2020-01-08
support boot addons for pluggable install of other systems


jstrachan:
we are using a new design process for this issue:
https://github.com/jenkins-x/enhancements/pull/1 details here: https://github.com/jstrachan/enhancements/blob/1/proposals/2/README.md

the user guide for the PoC is here: https://github.com/jstrachan/enhancements/blob/1/proposals/2/docs/README.md
with the boot git config repo here: https://github.com/jenkins-x-labs/boot-helmfile-poc 

we currently manually configure in the boot pipeline the install of multiple systems usually in separate namespaces:

* velero (if enabled)
* nginx
* externaldns
* cert-manager
* acme certs chart

ideally those should be modular and optional. e.g it should be really trivial to disable nginx if folks are using, say, istio - without having to hack a pipeline yaml.

we'd also like to make it easy to add a number of other systems in order in separate namespaces independently of the `Staging / Production` environment namespaces:

* knative https://github.com/jenkins-x/jx/issues/6331
* gloo
* istio
* linkerd https://github.com/jenkins-x/jx/issues/6330
* flagger

so it'd be nice to have a simple addon model where anyone can add any systems/charts to any namespace at any point in the flow before we setup the dev/staging/production environment in the traditional way.

So some kind of yaml file with an ordered list of releases with a list of namespaces and charts would be great that would be easy to PR against like we've done with the `jx add app` commands historically, so managed via GitOps etc.

## use helmfile?

One possible solution might be to add a single step the boot pipeline to invoke helmfile: https://github.com/roboll/helmfile#configuration

e.g. if there was an `addons/helmfile.yaml` file in the file system then run helmfile as a step on the file.

It looks like helmfile already supports tillerless + helm3 and lets us define namespaces and orders etc.

We may need to build our own step to wrap helmfile so that we can do some of the things we do with `jx step helm apply` right now

* should we support access to secrets in vault / local file system via URLs or Parameters injection?
* do we need to support exposing the properties from the `jx-requirements.yml` as values.yaml files that can be reused inside the helmfile templates?
* do we need to improve access to certs / domain after we've set those up in boot?
* allow the use of version streams to manage versions of things if no version is specified in the `helmfile.yaml` - like we do with helm charts requirements.yaml file right now (in `jx step helm apply`)

[...]

PR #6623, opened 2020-01-28
feat: adds a command to generate a helmfile.yaml from a j…

rawlingsj:
…x-apps.yml

this is an experimental feature which is isolated, if the idea matures then we may want to add the application types to a jx-requirements.yaml and remove the use of jx-apps.yml.  The reason that's not done here is so that we don't let the experiment affect existing core functionality.

I've also included a top level experiment command we can add these types of labs efforts to make it clear they are experimental and will likely change.

example usage:
```
jx step create helmfile --values ../secrets.yaml --values ../jx-requirements.values.yaml.gotmpl --outputDir ./apps
```

relates to the app extensions for boot [enhancement](https://github.com/jenkins-x/enhancements/pull/1) and issue #6442

Signed-off-by: James Rawlings <rawlingsj80@gmail.com>


jstrachan:
@rawlingsj I guess the primary user of this command would be the `jx boot` pipelines for helmfile; so I wonder if its better to move this command inside `jx step create helmfile` so its with all the other pipeline steps; rather than a top level `jx create` which end users might see via tab completion etc?

rawlingsj:
> I guess the primary user of this command would be the jx boot pipelines for helmfile; so I wonder if its better to move this command inside jx step create helmfile so its with all the other pipeline steps; rather than a top level jx create which end users might see via tab completion etc?

yeah good call, lemme do that now

Issue #6500, opened 2020-01-13
What is the purpose of specifying the boot version in the version stream?


hferentschik:
There is a circular dependency between _jenkins-x-boot-config_ and _jenkins-x-versions_. The former specifies a version stream version in its `jx-requirements.yml` and the latter has an entry for the version stream in `git/github.com/jenkins-x/jenkins-x-boot-config.yml`. What's the purpose of this? Something needs to be at the root of the boot process. 

I guess one use-case might be running `jx boot` from outside a _jenkins-x-boot-config_. In this scenario, boot can determine the boot version from the currently checked version stream and check that version of the boot config out.

In the case where you are running `jx boot` in an actual checkout of the boot config, it makes less sense. I can, for example, check out version 1.0 of the boot config and point it to a version stream which in turn references 0.9 of the boot config. The boot process will create the initial env repo based on the 0.9 tag (`step_verify_environments#createDevEnvironmentRepository`), but due to the cherry-pick logic in `helpers#ForkAndPullRepo`) we would still cherry-pick all the commits between 0.9 to 1.0 of the boot config and push it to the env repo (see #6498).

hferentschik:
Maybe we need to make a decision how `jx boot` should be used, aka from any directory checking out the boot config repo or from within an actual boot config repo.

In the former case there is an issue that one cannot modify the requirements file.

PR #6642, opened 2020-01-28
fix: allow `jx boot --no-update-git` to not mess with git

jstrachan:
when testing out a possible PR on the git boot repository referenced in the `jx-requirements.yml` even using `--no-update-git` I was finding that boot would:

* lookup my local fork of the git repository and since its not in the version stream assume 'master' was the branch I wanted (despite being in a fork/branch already
* re-apply all changes from master to my local checkout causing merge hell
* then `jx step verify env` tries to push to master the local changes merged with upstream causing merge conflicts etc.

This PR fixes this by disabling the bit of `jx boot` which switches from the current gitRef -> `master` if the `--no-update-git` is provided. That flag also passes in an env var so that `jx step verify env` doesn't try to push local changes to master of the dev repo (which is only really intended for real installs from the real upstream git repo - not from local forks trying changes out in boot config).

There's an additional minor change for helmfile to disable lazily creating the dev Environment if helmfile is used; which breaks helmfile/helm 3 which wants to create/update that resource via helm

### DIFF
--- a/pkg/boot/constants.go
+++ b/pkg/boot/constants.go
@@ -14,4 +14,8 @@ const (
 	VersionsRepoURLEnvVarName = "VERSIONS_REPO_URL"
 	// VersionsRepoBaseRefEnvVarName is the env var name used in the pipeline to reference the ref of versions repo
 	VersionsRepoBaseRefEnvVarName = "VERSIONS_BASE_REF"
+	// DisablePushUpdatesToDevEnvironment is the env var name used to disable pushing changes to the local git clone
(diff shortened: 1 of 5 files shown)


hferentschik:
Hmm, this really comes down to the circular dependency between Boot and the version stream (see #6500) and imo should be addressed by removing the dependency altogether as opposed to having more flags and conditional logic.

jstrachan:
@hferentschik we need a flag to effectively override the version stream when working on a fork & branch of a boot config git repo

PR #6643, opened 2020-01-28
fix: Mount Maven settings.xml secret into metapipeline pod

### LINKED ISSUE #6637: Maven builders in meta pipeline should mount settings.xml
### Summary 

When using the maven buildpack, the maven builders in the meta pipeline do not mount the `settings.xml` from the `jenkins-maven-settings` secret.

This causes builds to fail for projects that use a parent pom that must be downloaded from the local jx nexus rather than from central.

1. Start a pipeline to build a project using the maven buildpack

1. Observe that `settings.xml` is not being mounted to `/root/.m2/` in the `step-create-tekton-crds` container in the meta pipeline pod

### Expected behavior

The `step-create-tekton-crds` container should have the `jenkins-maven-settings` secret mounted to `/root/.m2/settings.xml` in the meta pipeline pod.

When `mvn versions:set -DnewVersion=\$(cat VERSION)` is run in the `step-create-tekton-crds` container, it should download dependencies from the jx nexus rather than from central.

### Actual behavior

Meta pipeline fails at `step-create-tekton-crds`:
```
running command: echo \$(jx-release-version) > VERSION

running command: mvn versions:set -DnewVersion=\$(cat VERSION)
[INFO Scanning for projects...
Downloading from central: https://repo.maven.apache.org/maven2/com/foo/parent-pom/1.0.0/parent-pom-1.0.0.pom
[ERROR [ERROR Some problems were encountered while processing the POMs:
[FATAL Non-resolvable parent POM for com.foo:test-app:1.0-SNAPSHOT: Could not find artifact com.foo:parent-pom:pom:1.0.0 in central (https://repo.maven.apache.org/maven2) and 'parent.relativePath' points at wrong local POM @ line 6, column 13
 @ 
[ERROR The build could not read 1 project -> [Help 1
[ERROR   
[ERROR   The project com.foo:test-app:1.0-SNAPSHOT (/workspace/source/pom.xml) has 1 error
[ERROR     Non-resolvable parent POM for com.foo:test-app:1.0-SNAPSHOT: Could not find artifact com.foo:parent-pom:pom:1.0.0 in central (https://repo.maven.apache.org/maven2) and 'parent.relativePath' points at wrong local POM @ line 6, column 13 -> [Help 2
[ERROR 
[ERROR To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR Re-run Maven using the -X switch to enable full debug logging.
[ERROR 
[ERROR For more information about the errors and possible solutions, please read the following articles:
[ERROR [Help 1 http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
[ERROR [Help 2 http://cwiki.apache.org/confluence/display/MAVEN/UnresolvableModelException
error: failed to set the version on release pipelines: failed to run '/bin/sh -c mvn versions:set -DnewVersion=$(cat VERSION)' command in directory '/workspace/source', output: ''
```

### Jx version

```
NAME               VERSION
jx                 2.0.1152
Kubernetes cluster v1.14.9
kubectl            v1.14.8
git                2.21.0 (Apple Git-122.2)
```

### Jenkins type

- [ ] Classic Jenkins

### Kubernetes cluster

On prem cluster created via Rancher.

abayer:
#### Submitter checklist

#### Description

We need to be mounting the `jenkins-maven-settings` secret into the metapipeline pod to deal with updating versions in POMs with parents only available from private repositories.

/assign @hferentschik 

fixes #6637

### DIFF
--- a/pkg/tekton/metapipeline/metapipeline.go
+++ b/pkg/tekton/metapipeline/metapipeline.go
@@ -35,6 +35,9 @@ const (
 	createTektonCRDsStepName = "create-tekton-crds"
 
 	tektonBaseDir = "/workspace"
+
+	mavenSettingsSecretName = "jenkins-maven-settings" // #nosec
+	mavenSettingsMount      = "/root/.m2/"
 )
 
 // CRDCreationParameters are the parameters needed to create the Tekton CRDs
@@ -133,6 +136,14 @@ func createPipeline(params CRDCreationParameters) (*syntax.ParsedPipeline, error
 		},
 		Options: &syntax.StageOptions{
 			RootOptions: &syntax.RootOptions{
+				Volumes: []*corev1.Volume{{
+					Name: mavenSettingsSecretName,
+					VolumeSource: corev1.VolumeSource{
+						Secret: &corev1.SecretVolumeSource{
+							SecretName: mavenSettingsSecretName,
+						},
+					},
+				}},
 				ContainerOptions: &corev1.Container{
 					Resources: corev1.ResourceRequirements{
 						Limits: corev1.ResourceList{
@@ -144,6 +155,10 @@ func createPipeline(params CRDCreationParameters) (*syntax.ParsedPipeline, error
 							"memory": resource.MustParse("256Mi"),
 						},
 					},
+					VolumeMounts: []corev1.VolumeMount{{
+						Name:      mavenSettingsSecretName,
+						MountPath: mavenSettingsMount,
+					}},
 				},
 			},
 		},


pkg/tekton/metapipeline/metapipeline.go
```diff
@@ -35,6 +35,9 @@ const (
 	createTektonCRDsStepName = "create-tekton-crds"
 
 	tektonBaseDir = "/workspace"
+
+	mavenSettingsSecretName = "jenkins-maven-settings" // #nosec
```
hferentschik: Is there really no other way? The meta pipeline should be agnostic of these things and it should be possible to configure this in a generic way. Having Maven/Nexus specific things in the meta-pipeline like this seems wrong.

abayer: We already effectively do - the default `setVersion` for the maven build pack calls `mvn`, which is why we need `builder-maven` for the step image. It's janky, but it is what it is.

jstrachan:
@abayer am cool with fixing this via the above PR - but I wonder if longer term we can define what volume mounts we wanna do in the underlying build packs? e.g. add the volume mount metadata into this file: https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes/blob/master/packs/maven/pipeline.yaml

then we could configure different secret mounts inside the build pack git repo for different kinds of builds without changing the meta pipeline?

abayer:
@jstrachan I think the metapipeline is always going to be a little weird in that regard, since it doesn't know anything about what build pack is being used etc until it's already running, which is too late to get the secret here.

pkg/tekton/metapipeline/metapipeline.go
abayer: This, fwiw, is why in the pre metapipeline days, we always had to have `jenkins-maven-settings` mounted in the `pipelinerunner`.

hferentschik: I see.

PR #6647, opened 2020-01-28
fix: Get structure for rerun PipelineRun properly

### LINKED ISSUE #6646: PipelineRuns rerun by Lighthouse due to Tekton race condition won't be handled by build controller
If Lighthouse has rerun a `PipelineRun` due to our buddy the Tekton race condition, the new `PipelineRun` has a different name than the original. This means that there's no `PipelineStructure` with the same name, so logging and the build controller don't know what to do with it.

Took me a while to determine that this was why I'd see some pipelines never report as finished in Lighthouse tests, despite the fact that they _did_ finish. Their activity had the metapipeline reported correctly, but didn't have the build pipeline included at all. I eventually managed to capture the build controller logs from a case where this happened and found lots of warnings due to not finding the `PipelineRun`'s `PipelineStructure`, realized this was only happening with `PipelineRun`s that had been rerun by Lighthouse (this doesn't happen with Prow, where the whole build from metapipeline onwards is rerun), and went ah-ha!

I'm verifying a fix now and should have a PR up in the next day.

abayer:
#### Submitter checklist

#### Description

If Lighthouse has rerun a `PipelineRun` due to our buddy the Tekton race condition, the new `PipelineRun` has a different name than the original. This means that there's no `PipelineStructure` with the same name, so logging and the build controller don't know what to do with it. So when we're looking up `PipelineStructure`s for a `PipelineRun`, first try to look by the `PipelineRun` name, and then by the `Pipeline` name associated with the `PipelineRun`. The `Pipeline` name will always be the same for the rerun `PipelineRun` as it was for the original, so this will work.

/assign @hferentschik 

fixes #6646

### DIFF
--- a/pkg/tekton/pipelines.go
+++ b/pkg/tekton/pipelines.go
@@ -8,8 +8,10 @@ import (
 	"time"
 
 	jenkinsio "github.com/jenkins-x/jx/pkg/apis/jenkins.io"
-	v1 "github.com/jenkins-x/jx/pkg/client/clientset/versioned/typed/jenkins.io/v1"
+	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	clientv1 "github.com/jenkins-x/jx/pkg/client/clientset/versioned/typed/jenkins.io/v1"
 	"github.com/jenkins-x/jx/pkg/kube/naming"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/rand"
 
@@ -120,7 +122,7 @@ func CreateOrUpdateTask(tektonClient tektonclient.Interface, ns string, created 
 	return answer, nil
 }
 
-func nextBuildNumberFromActivity(activityInterface v1.PipelineActivityInterface, gitInfo *gits.GitRepository, branch string) (string, error) {
+func nextBuildNumberFromActivity(activityInterface clientv1.PipelineActivityInterface, gitInfo *gits.GitRepository, branch string) (string, error) {
 	labelMap := labels.Set{
 		"owner":      gitInfo.Organisation,
 		"repository": gitInfo.Name,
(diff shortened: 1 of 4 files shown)


abayer:
/hold

Holding overnight while I have a set of manual tests running to verify this does what it's supposed to.

pkg/tekton/pipelines.go
```diff
@@ -493,6 +496,33 @@ func ApplyPipeline(jxClient versioned.Interface, tektonClient tektonclient.Inter
...
 }
 
+// StructureForPipelineRun finds the PipelineStructure for the given PipelineRun, trying its name first and then its
+// Pipeline name, returning an error if no PipelineStructure can be found.
+func StructureForPipelineRun(jxClient versioned.Interface, ns string, run *pipelineapi.PipelineRun) (*v1.PipelineStructure, error) {
+	structure, err := jxClient.JenkinsV1().PipelineStructures(ns).Get(run.Name, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// Try again but with the Pipeline name for this run.
+			pipelineName := run.Labels[pipeline.GroupName+pipeline.PipelineLabelKey]
```
mrageh: Why do you combine `pipeline.GroupName+pipeline.PipelineLabelKey` to search for the label?

abayer: To get the combination of `tekton.dev` and `/pipeline` for the full label. =)

abayer:
/hold cancel

Yay, my overnight tests did hit the scenario that was failing (a rerun `PipelineRun`) without the error.

pkg/tekton/pipelines.go
```diff
@@ -493,6 +496,33 @@ func ApplyPipeline(jxClient versioned.Interface, tektonClient tektonclient.Inter
...
+			// Fall back on the PipelineRef.Name if there isn't a label.
+			if pipelineName == "" {
+				pipelineName = run.Spec.PipelineRef.Name
+			}
+			// If we still have no name, error out.
+			if pipelineName == "" {
+				return nil, fmt.Errorf("couldn't find a Pipeline name for PipelineRun %s", run.Name)
+			}
+			structure, err = jxClient.JenkinsV1().PipelineStructures(ns).Get(pipelineName, metav1.GetOptions{})
```
hferentschik: If we know that lookup by pipeline run name can fail and the "better" option is by pipeline name, why don't we do the lookup always by the latter. Less API calls in average and less nested logic.

abayer: Done. =)

PR #6652, opened 2020-01-29
docs: update replicate step example to use the correct flag

rawlingsj:


hferentschik:
@rawlingsj, do you want to update the commit with sign-off or can you tell me how I skip/override the DCO check?

PR #6658, opened 2020-01-29
fix: Check if the requirements config was read from the existing file

### LINKED ISSUE #6653: Step helm apply does not load team settings
### Summary 
In my staging environment, some of the helm chart relay on vault to populate secrets. Unfortunately, `jx step helm apply` loads only default requirements. Even if the GitOps staging environment doesn't have `jx-requirements.yaml` file. It does not reach the [dev environment team settings](https://github.com/jenkins-x/jx/blob/6f606365f4ce912207b1b822f8426e7e037e9f6c/pkg/cmd/step/helm/step_helm_apply.go#L400). The reason why this piece of code is unreachable is that the [LoadRequirementsConfig ](https://github.com/jenkins-x/jx/blob/6f606365f4ce912207b1b822f8426e7e037e9f6c/pkg/config/install_requirements.go#L514) never return an error. At the end of the function, [the default config is returned](https://github.com/jenkins-x/jx/blob/6f606365f4ce912207b1b822f8426e7e037e9f6c/pkg/config/install_requirements.go#L546). I would like to [highlight the default behavior of LoadRequirementsConfigFile](https://github.com/jenkins-x/jx/blob/6f606365f4ce912207b1b822f8426e7e037e9f6c/pkg/config/install_requirements.go#L583-L586) will not return the error "file not exist", because it is [consumed by FileExists](https://github.com/jenkins-x/jx/blob/6f606365f4ce912207b1b822f8426e7e037e9f6c/pkg/util/files.go#L43-L45). The consumer of the LoadRequirementsConfig API will never know if the config comes from defaults or it has been read from file.

For example when the preview environment spawns the combination of `jx step helm build` and `jx preview --app $APP_NAME --dir ../..` correctly replace the vault placeholders. 

Run pipeline in a staging environment (GitOps version of JX).

### Expected behavior

The secrets in values.yaml are replaced with the value from Hashicorp vault.

### Actual behavior

All occurrence of `vault:my/secret/path` are replaced.

### Jx version

```
jx version
NAME               VERSION
jx                 2.0.1142
Kubernetes cluster v1.14.8
kubectl            v1.17.1
git                2.25.0
```

### Jenkins type

- [ ] Classic Jenkins

### Kubernetes cluster

My own cluster that runs on AWS.

MacBook and container gcr.io/jenkinsxio/builder-go

RafalKorepta:
#### Submitter checklist

#### Description

The [LoadRequirementsConfig ](https://github.com/jenkins-x/jx/blob/6f606365f4ce912207b1b822f8426e7e037e9f6c/pkg/config/install_requirements.go#L514) API does not return an error when the `jx-requirements.yaml` file does not exist. Instead, it returns a [default RequirementsConfig struct](https://github.com/jenkins-x/jx/blob/master/pkg/config/install_requirements.go#L503-L509) and not valid file path. This PR tries to fix the issue so that [getRequirements](https://github.com/jenkins-x/jx/blob/6f606365f4ce912207b1b822f8426e7e037e9f6c/pkg/cmd/step/helm/step_helm_apply.go#L388-L409) loads team dev environement requirements, when `jx-requirements.yaml` file does not exist.

fixes #6653

### DIFF
--- a/pkg/cmd/opts/common.go
+++ b/pkg/cmd/opts/common.go
@@ -8,43 +8,15 @@ import (
 	"strings"
 	"time"
 
-	"github.com/jenkins-x/jx/pkg/config"
-
-	"github.com/jenkins-x/jx/pkg/kube/cluster"
-
-	gojenkins "github.com/jenkins-x/golang-jenkins"
-	"github.com/jenkins-x/jx/pkg/cloud/gke"
-	"github.com/jenkins-x/jx/pkg/prow"
-	"github.com/jenkins-x/jx/pkg/versionstream"
-	"github.com/spf13/viper"
-
-	"github.com/jenkins-x/jx/pkg/secreturl"
-	"github.com/spf13/pflag"
-
-	"github.com/heptio/sonobuoy/pkg/client"
-	"github.com/jenkins-x/jx/pkg/cmd/clients"
-	"github.com/jenkins-x/jx/pkg/io/secrets"
-	"github.com/jenkins-x/jx/pkg/vault"
-
-	"github.com/jenkins-x/jx/pkg/kube/resources"
-	"github.com/jenkins-x/jx/pkg/kube/services"
-	"github.com/pkg/errors"
-
 	vaultoperatorclient "github.com/banzaicloud/bank-vaults/operator/pkg/client/clientset/versioned"
-	jenkinsv1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
-	"github.com/jenkins-x/jx/pkg/auth"
-	"github.com/jenkins-x/jx/pkg/client/clientset/versioned"
-	"github.com/jenkins-x/jx/pkg/gits"
-	"github.com/jenkins-x/jx/pkg/helm"
-	"github.com/jenkins-x/jx/pkg/kube"
-	"github.com/jenkins-x/jx/pkg/kustomize"
-	"github.com/jenkins-x/jx/pkg/log"
-	"github.com/jenkins-x/jx/pkg/table"
-	"github.com/jenkins-x/jx/pkg/util"
+	"github.com/heptio/sonobuoy/pkg/client"
 	certmngclient "github.com/jetstack/cert-manager/pkg/client/clientset/versioned"
 	istioclient "github.com/knative/pkg/client/clientset/versioned"
 	kserve "github.com/knative/serving/pkg/client/clientset/versioned"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
 	tektonclient "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
 	"gopkg.in/AlecAivazis/survey.v1"
 	"gopkg.in/AlecAivazis/survey.v1/terminal"
@@ -52,6 +24,29 @@ import (
 	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/client-go/kubernetes"
 	prowjobclient "k8s.io/test-infra/prow/client/clientset/versioned"
+
+	gojenkins "github.com/jenkins-x/golang-jenkins"
+	jenkinsv1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	"github.com/jenkins-x/jx/pkg/auth"
+	"github.com/jenkins-x/jx/pkg/client/clientset/versioned"
+	"github.com/jenkins-x/jx/pkg/cloud/gke"
+	"github.com/jenkins-x/jx/pkg/cmd/clients"
+	"github.com/jenkins-x/jx/pkg/config"
+	"github.com/jenkins-x/jx/pkg/gits"
+	"github.com/jenkins-x/jx/pkg/helm"
+	"github.com/jenkins-x/jx/pkg/io/secrets"
+	"github.com/jenkins-x/jx/pkg/kube"
+	"github.com/jenkins-x/jx/pkg/kube/cluster"
+	"github.com/jenkins-x/jx/pkg/kube/resources"
+	"github.com/jenkins-x/jx/pkg/kube/services"
+	"github.com/jenkins-x/jx/pkg/kustomize"
+	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/prow"
+	"github.com/jenkins-x/jx/pkg/secreturl"
+	"github.com/jenkins-x/jx/pkg/table"
+	"github.com/jenkins-x/jx/pkg/util"
+	"github.com/jenkins-x/jx/pkg/vault"
+	"github.com/jenkins-x/jx/pkg/versionstream"
 )
 
 // LogLevel represents the logging level when reporting feedback
@@ -554,7 +549,10 @@ func (o *CommonOptions) Helm() helm.Helmer {
 
 			// check helmfile featureflag but default to existing behaviour if there's any issues
 			var helmer helm.Helmer
-			r, _, _ := config.LoadRequirementsConfig("")
+			r, _, err := config.LoadRequirementsConfig("")
+			if err != nil {
+				r = config.NewRequirementsConfig()
+			}
 			if r.Helmfile {
 				helmer = o.NewHelm(false, "helm", true, false)
 			} else {
--- a/pkg/cmd/step/expose/step_expose.go
+++ b/pkg/cmd/step/expose/step_expose.go
@@ -9,6 +9,15 @@ import (
 	"text/template"
 
 	"github.com/ghodss/yaml"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/api/extensions/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/helm/pkg/chartutil"
+
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
 	"github.com/jenkins-x/jx/pkg/cmd/opts"
 	"github.com/jenkins-x/jx/pkg/cmd/templates"
@@ -18,14 +27,6 @@ import (
 	"github.com/jenkins-x/jx/pkg/kube/services"
 	"github.com/jenkins-x/jx/pkg/log"
 	"github.com/jenkins-x/jx/pkg/util"
-	"github.com/pkg/errors"
-	"github.com/spf13/cobra"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/api/extensions/v1beta1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/helm/pkg/chartutil"
 )
 
 var (
@@ -107,6 +108,10 @@ func (o *StepExposeOptions) Run() error {
 	if err != nil {
 		return errors.Wrapf(err, "failed to load requirements YAML")
 	}
+	if requirements == nil {
+		return errors.Errorf("unable to find requirements file in '%s'", o.Dir)
+	}
+
 	if requirements.Cluster.Namespace == "" {
 		requirements.Cluster.Namespace = ns
 	}
--- a/pkg/cmd/step/helm/step_helm_apply.go
+++ b/pkg/cmd/step/helm/step_helm_apply.go
@@ -7,12 +7,14 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/jenkins-x/jx/pkg/cmd/opts/step"
-	"github.com/jenkins-x/jx/pkg/platform"
-
 	"github.com/google/uuid"
+	"github.com/mholt/archiver"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
 	"github.com/jenkins-x/jx/pkg/cmd/opts"
+	"github.com/jenkins-x/jx/pkg/cmd/opts/step"
 	"github.com/jenkins-x/jx/pkg/cmd/templates"
 	"github.com/jenkins-x/jx/pkg/config"
 	"github.com/jenkins-x/jx/pkg/gits"
@@ -22,12 +24,10 @@ import (
 	"github.com/jenkins-x/jx/pkg/kube"
 	"github.com/jenkins-x/jx/pkg/kube/naming"
 	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/platform"
 	"github.com/jenkins-x/jx/pkg/secreturl/fakevault"
 	"github.com/jenkins-x/jx/pkg/util"
 	"github.com/jenkins-x/jx/pkg/vault"
-	"github.com/mholt/archiver"
-	"github.com/pkg/errors"
-	"github.com/spf13/cobra"
 )
 
 // StepHelmApplyOptions contains the command line flags
@@ -393,10 +393,6 @@ func (o *StepHelmApplyOptions) getRequirements() (*config.RequirementsConfig, st
 	// Try to load first the requirements from current directory
 	requirements, requirementsFileName, err := config.LoadRequirementsConfig(o.Dir)
 
-	_, err = os.Stat(requirementsFileName)
-	if os.IsNotExist(err) {
-		_, err = os.Stat(filepath.Join(o.Dir, requirementsFileName))
-	}
 	if err == nil {
 		return requirements, requirementsFileName, nil
 	}
--- a/pkg/cmd/step/helm/step_helm_apply.go
+++ b/pkg/cmd/step/helm/step_helm_apply.go
@@ -248,7 +248,7 @@ func (o *StepHelmApplyOptions) Run() error {
 	}
 
 	requirements, requirementsFileName, err := o.getRequirements()
-	if err != nil {
+	if err != nil || requirements == nil {
 		return errors.Wrap(err, "loading the requirements")
 	}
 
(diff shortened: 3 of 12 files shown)


pkg/cmd/step/helm/step_helm_apply.go
```diff
@@ -389,6 +389,11 @@ func (o *StepHelmApplyOptions) Run() error {
 func (o *StepHelmApplyOptions) getRequirements() (*config.RequirementsConfig, string, error) {
 	// Try to load first the requirements from current directory
 	requirements, requirementsFileName, err := config.LoadRequirementsConfig(o.Dir)
+
+	_, err = os.Stat(requirementsFileName)
```
ccojocar: I think that we should skip the `Stat` when the `err` returned by the `LoadRequirementsConfig` is not nil, and directly load the requirements from the team settings.

RafalKorepta: The thing is that LoadRequirementsConfig IMHO should return an error when `jx-requirments.xml` was not found. The current implementation will never return an error because it's swallowed by [FileExists function](https://github.com/jenkins-x/jx/blob/6f606365f4ce912207b1b822f8426e7e037e9f6c/pkg/util/files.go#L43-L45). Then the [LoadRequirementsConfigFile](https://github.com/jenkins-x/jx/blob/6f606365f4ce912207b1b822f8426e7e037e9f6c/pkg/config/install_requirements.go#L583-L586) will return an error which is nil. Could you check my explanation in the issue #6653? 

I'm not sure if changing files.FileExists to not swallow the error is the correct approach to solve this issue.

abayer:
Looks like this is causing legit failures in legacy static master tests.

RafalKorepta:
@abayer Where I can find source code for tests that failed?

abayer:
It's running the `create-spring` test from https://github.com/jenkins-x/bdd-jx - what seems to be happening is that the staging environment never gets updated. The test cluster, for legacy reasons, is created with `jx install` and uses static Jenkins masters (the `tekton` context uses `jx install` but with modern Tekton-driven pipelines). So there's definitely no `jx-requirements.yml`, and no `BootRequirements` field set in the team settings. It is still possible that this is just weird luck, so I am going to rerun again, but it's failed twice already.

RafalKorepta:
I have a question about bdd tests. Are they run in a cluster that is managed by GitOps repositories? Does this cluster have a Hashicorp vault? Can I modify those tests so that example spring application will use secrets from the vault in charts repo?

abayer:
It varies - `bdd` and `tekton` are legacy tests, running in clusters without vault, installed by `jx install`. `boot-vault`, on the other hand, runs similar tests but in a cluster that, surprisingly given the name, used `jx boot` and has vault. As I said, the actual code for those tests live in https://github.com/jenkins-x/bdd-jx - we don't currently have differing behavior in those tests for whether vault is present or not, but that could be done.

pkg/cmd/step/helm/step_helm_apply.go
ccojocar: I agree that it should return an error when the file doesn't exit. Not quite sure if there are any places in boot relying on this default requirements when file doesn't exit. 

It is worth trying to change the `LoadRequirementsConfigFile` to return an error when the file doesn't exit.

RafalKorepta: I agree. Today I will try to fix that.

hferentschik:
hi, just wondering what the state is on this PR. Looks like there are quite a few side effects by the changes introduced. For example, I see:

```
Creating cluster pr-6658-34-static-gkebdd
error: **-requirements.yml file not found
Pipeline failed on stage 'e2e-tests' : container 'step-e2e-tests'. The execution of the pipeline has stopped.
```

This seems to be related to what we are trying to change here. The change might not necessarily be wrong though, but maybe we don't see yet the full picture?

The failing integration tests might be the best place to start from now since they are easiest to reproduce. Here are the current test failures:
```
        	            	runtime.goexit
        	            	failed to load boot requirements
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify.(*StepVerifyPackagesOptions).Run
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_packages.go:91
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify.(*StepVerifyPreInstallOptions).Run
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_preinstall.go:177
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify_test.TestStepVerifyPreInstallNoTLS
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_preinstall_integration_test.go:136
        	            	testing.tRunner
        	            	runtime.goexit
        	Test:       	TestStepVerifyPreInstallNoTLS
=== FAIL: pkg/cmd/step/verify TestStepVerifyRequirements (0.67s)
=== PAUSE TestStepVerifyRequirements
=== CONT  TestStepVerifyRequirements
Using namespace '**' from context named '' on server ''.
    step_verify_requirements_integration_test.go:50: 
        	Error Trace:	step_verify_requirements_integration_test.go:50
        	            	**-requirements.yml file not found
        	            	github.com/jenkins-x/**/pkg/config.LoadRequirementsConfig
        	            		/workspace/source/pkg/config/install_requirements.go:532
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify.(*StepVerifyRequirementsOptions).Run
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_requirements.go:85
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify_test.TestStepVerifyRequirements
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_requirements_integration_test.go:49
        	            	testing.tRunner
        	            	runtime.goexit
        	            	failed to load boot requirements
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify.(*StepVerifyRequirementsOptions).Run
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_requirements.go:87
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify_test.TestStepVerifyRequirements
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_requirements_integration_test.go:49
        	            	testing.tRunner
        	            	runtime.goexit
        	Test:       	TestStepVerifyRequirements
        	Messages:   	Command failed: verify.StepVerifyRequirementsOptions{StepOptions:step.StepOptions{CommonOptions:(*opts.CommonOptions)(0xc000c76900), DisableImport:false, OutDir:""}, Dir:"/tmp/test-step-verify-requirements465172405"}
    step_verify_requirements_integration_test.go:62: found version 2.0.1604 for jenkins-x-platform requirement
DONE 59 tests, 4 skipped, 3 failures in 688.218s
```

hferentschik:
@RafalKorepta, are you able to run `make test-integration` locally?

RafalKorepta:
Hi @hferentschik, I had problems to run integration tests. Where I can find set-up documentation? 

ps. I created an issue to leverage kind in tests that need kubernetes cluster working #6681.

hferentschik:
>  I had problems to run integration tests. 

What did you try? All you need really is any Jenkins X install/cluster and a kube context connected to it. Then `make test-integration` should work.

Did you have any specific error?

> Where I can find set-up documentation?

There is https://jenkins-x.io/docs/contributing/code/#integration-tests, but that describes just how you mark a test as an integration test using the _integration_ build tag. It does not talk about any prerequisites. Maybe one should start by improving the docs? Interested ;-)

For what it's worth, I can reproduce the above errors locally by running `make integration` (and being connected to a Jenkins X cluster).

hferentschik:
> I created an issue to leverage kind in tests that need kubernetes cluster working #6681.

Nice idea, I have been toying with this idea myself. However, I am not sure how feasible it is. We have/had support for Minikube and Minishift which are about to deprecate/remove. 

I think Kind is more promising than Minishift/Minikube, but there are still several issues which would need to be addressed. ATM, the Jnekins X install relies on some cloud-specific things which you would not have in a Kind setup, but probably most problematic would be the lack of an ingress. This would make any kind of webhook setup with a git provider very hard.  

I would love to have something like Kind for running integration tests, but atm I think there are plenty of roadblocks ahead if one would try. That said, give it a go and see where it goes.

hferentschik:
> The LoadRequirementsConfig API does not return an error when the jx-requirements.yaml file does not exist.

Are you sure about this, or is the problem that the error is ignored where this function is called in _commons.go_? Maybe some tests would help?

RafalKorepta:
```
// LoadRequirementsConfigFile loads a specific project YAML configuration file
func LoadRequirementsConfigFile(fileName string) (*RequirementsConfig, error) {
        config := NewRequirementsConfig()
	exists, err := util.FileExists(fileName)
	if err != nil || !exists {
		return config, err
	}
```
@hferentschik In this function the if statement will return stub of requirements configuration struct and nill error when fileName does not exist. When the exists is false to error is nil, because of os.IsNotExist(err) part of FileExists helper function.
```
// FileExists checks if path exists and is a file
func FileExists(path string) (bool, error) {
	_, err := os.Stat(path)
	if err == nil {
		return true, nil
	}
	if os.IsNotExist(err) {
		return false, nil
	}
```

I see throughout the code almost every place using util.FileExists adjust for that behavior besides `step_help_apply.go`.  In the end, I can't use vault secrets in the helm chart by referencing them `vault:my/secret`.

hferentschik:
>  In this function the if statement will return stub of requirements configuration struct and nill error when fileName does not exist. When the exists is false to error is nil, because of os.IsNotExist(err) part of FileExists helper function.

good catch, thanks for the explanation.

hferentschik:
I wonder if the problem is not that this if statement handles two things and should be split apart

```
   config := NewRequirementsConfig()
	exists, err := util.FileExists(fileName)
	if err != nil || !exists {
		return config, err
	}
```

If there is an error the current return is ok, but if !exist we need to generate an error, so that callers higher in the call chain can act accordingly. Have you tried that?

RafalKorepta:
> If there is an error the current return is ok, but if !exist we need to generate an error, so that callers higher in the call chain can act accordingly. Have you tried that?

I didn't. I tried to minimalize the change. I go from the `step_helm_apply.go` perspective, but @ccojocar suggests that I should fix the `LoadRequirementsConfigFile` https://github.com/jenkins-x/jx/pull/6658#discussion_r372882671. Doing that I got stuck with unable to run tests locally.

pkg/config/install_requirements.go
```diff
@@ -516,90 +515,50 @@ func LoadRequirementsConfig(dir string) (*RequirementsConfig, string, error) {
...
 }
 
 // LoadRequirementsConfigFile loads a specific project YAML configuration file
 func LoadRequirementsConfigFile(fileName string) (*RequirementsConfig, error) {
-	config := NewRequirementsConfig()
-	exists, err := util.FileExists(fileName)
-	if err != nil || !exists {
-		return config, err
+	config := &RequirementsConfig{}
+	_, err := os.Stat(fileName)
```
hferentschik: woud one not need a _IsDir_ check here as well?

pkg/config/install_requirements.go
```diff
@@ -516,90 +515,50 @@ func LoadRequirementsConfig(dir string) (*RequirementsConfig, string, error) {
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "checking if file %s exists", fileName)
 	}
+
 	data, err := ioutil.ReadFile(fileName)
 	if err != nil {
-		return config, fmt.Errorf("Failed to load file %s due to %s", fileName, err)
+		return nil, fmt.Errorf("Failed to load file %s due to %s", fileName, err)
```
hferentschik: hmm,  I guess is the file is a directory you would get an error here

pkg/cmd/opts/common.go
```diff
@@ -8,50 +8,45 @@ import (
...
-	gojenkins "github.com/jenkins-x/golang-jenkins"
-	"github.com/jenkins-x/jx/pkg/cloud/gke"
-	"github.com/jenkins-x/jx/pkg/prow"
-	"github.com/jenkins-x/jx/pkg/versionstream"
-	"github.com/spf13/viper"
-
-	"github.com/jenkins-x/jx/pkg/secreturl"
-	"github.com/spf13/pflag"
-
+	vaultoperatorclient "github.com/banzaicloud/bank-vaults/operator/pkg/client/clientset/versioned"
```
hferentschik: this here is just a import ordering change, right? Do you have any custom setting for that? Just wondering?

pkg/cmd/step/expose/step_expose.go
```diff
@@ -107,7 +108,8 @@ func (o *StepExposeOptions) Run() error {
 	if err != nil {
 		return errors.Wrapf(err, "failed to load requirements YAML")
 	}
-	if requirements.Cluster.Namespace == "" {
+
+	if o.Namespace != "" {
```
hferentschik: Trying to understand how this is related to the problem of reading the requirements file? Is this trying to fix yet another issue?

pkg/cmd/opts/common.go
RafalKorepta: No. In my previous work, I used to sort them as:

- std lib
- line break
- external dependency
- line break
- internal dependency

After that, I run `go import` in the goland IDE, so that it sorts them in alphabetical order.

pkg/cmd/step/expose/step_expose.go
RafalKorepta: This is my mistake. I had a problem with a unit test, which was fixed by adding `jx-requirements.yml` in `test_data` folder. I will remove it in the next commit.

hferentschik:
@RafalKorepta which tests can you not run locally? unit tests (`make test`) or integration tests (`make test-integration`)? The former should have no external dependencies and should work all the time. The integration tests require a Jenkins X cluster. The current error there is:

```
=== Failed
=== FAIL: pkg/cmd/step/verify TestStepVerifyPreInstallNoKanikoLazyCreate (5.53s)
no need to reconnect to cluster
Verifying the kubernetes cluster before we try to boot Jenkins X in namespace: new-**-ns
Trying to lazily create any missing resources to get the current cluster ready to boot Jenkins X
Attempting to lazily create the deploy namespace new-**-ns
Namespace new-**-ns created 
Verifying Ingress...
WARNING: there is no context defined in your Kubernetes configuration - we may be inside a test case or pod?
No kube context - probably in a unit test or pod?
Using namespace '**' from context named '' on server ''.
    step_verify_preinstall_integration_test.go:112: 
        	Error Trace:	step_verify_preinstall_integration_test.go:112
        	            	**-requirements.yml file not found
        	            	github.com/jenkins-x/**/pkg/config.LoadRequirementsConfig
        	            		/workspace/source/pkg/config/install_requirements.go:532
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify.(*StepVerifyPackagesOptions).Run
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_packages.go:89
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify.(*StepVerifyPreInstallOptions).Run
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_preinstall.go:177
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify_test.TestStepVerifyPreInstallNoKanikoLazyCreate
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_preinstall_integration_test.go:111
        	            	testing.tRunner
        	            	runtime.goexit
        	            	failed to load boot requirements
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify.(*StepVerifyPackagesOptions).Run
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_packages.go:91
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify.(*StepVerifyPreInstallOptions).Run
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_preinstall.go:177
        	            	github.com/jenkins-x/**/pkg/cmd/step/verify_test.TestStepVerifyPreInstallNoKanikoLazyCreate
        	            		/workspace/source/pkg/cmd/step/verify/step_verify_preinstall_integration_test.go:111
        	            	testing.tRunner
        	            	runtime.goexit

```

So this seems to be a similar error to the one we see in the BDD tests. 

@RafalKorepta do you have a dev/test cluster available? I will try to investigate your pull request locally. Hard to tell what's going on just looking at the code. `LoadRequirementsConfigFile` is also called from boot, so I assume the changed behaviour from your PR has impacts there.

pkg/config/install_requirements.go
RafalKorepta: I think you are right. I will add a test and implementation for that in the next commit.

pkg/config/install_requirements.go
RafalKorepta: DITTO

RafalKorepta:
@hferentschik I have problems with integration tests (`make test-integration`). I was trying to run it against kind cluster without success. I will check the minikube.

hferentschik:
> I was trying to run it against kind cluster without success. I will check the minikube.

Not sure how far you get with these. Let me know.

That said, I find at least one reason for the test failures. You change unearthed another bug _step_verify_preinstall.go_. One needs to add something along these lines:

```
po.Dir = o.Dir
```

before po.Run is executed. Basically just above the if statement. 

```
	po := &StepVerifyPackagesOptions{}
	po.CommonOptions = o.CommonOptions
	po.Packages = []string{"kubectl", "git", "helm"}
	if !o.DisableVerifyPackages {
		err = po.Run()
		if err != nil {
			return err
		}
		log.Logger().Info("\n")
	}
```

If not `StepVerifyPackagesOptions` runs against a different directory than `StepVerifyPreInstallOptions` if a custom directory is specified. 

I am running tests locally atm to see if this will fix all integration test failures or not. At least it would be a step forward.

hferentschik:
@RafalKorepta, let's see where #6771 gets us. It makes two changes which make at least for me the integration test failures go away. Let's see what that does to the BDD tests.

hferentschik:
@RafalKorepta, getting closer #6771

pkg/cmd/opts/common.go
hferentschik: no worries

pkg/cmd/step/expose/step_expose.go
hferentschik: thanks, I was a bit baffled about this part.

hferentschik:
@RafalKorepta I think we sorted it out now. See #6771. I need to clean my changes up a bit early next week.

I guess we have two options. I clean up and create a pull request against your repo and you incorporate the changes into your PR or we close this PR and are going to merge my PR (will keep your commits of course).

Thanks for the hard work. The bug you fixed was basically shadowing some other issues in the code and by you fixing `LoadRequirementsConfig` we hit several other issues.

WDYT?

RafalKorepta:
I will close my issue and you can merge your changes ;) Thank you!

PR #6663, opened 2020-01-30
feat: allow the bdd suite to be configured

garethjevans:
#### Submitter checklist

fixes #


abayer:
/override boot-vault

It's gonna claim to have failed but it didn't actually fail for a legit reason. I think it barfed out due to the prod cluster issues, so overriding.

PR #6664, opened 2020-01-31
[wip] multi cluster support

jstrachan:
note this PR depends on the helmfile + apps support: https://github.com/jenkins-x/jx/pull/6660 - once that merges we can rebase this PR (which will then be quite small ;)

for documentation see https://github.com/jenkins-x/enhancements/tree/master/proposals/2/docs


polothy:
So, I was reading https://github.com/jenkins-x/enhancements/blob/master/proposals/2/docs/multi-cluster.md and I had a couple of questions:

1. What are in `Staging` and `Production` remote environments? Are those full `jx boot` repos?  Or closer to one of these standard environments, https://github.com/jenkins-x/default-environment-charts
2. Is it at all possible to have PRs built **in** the remote environments/clusters?  We currently use a pattern to deploy our infrastructure with our applications and building the PR in the remote cluster would allow us to preview infrastructure changes (highly requested from our developers).

Thanks so much for improving multi-cluster support.  The proposal looks like a very good improvement over the environment controller! 👍 :shipit:

jstrachan:
1. yeah, remote Staging & Production are git repos using boot & helmfile, which are cut down to just these apps 
https://github.com/jenkins-x/jenkins-x-boot-helmfile-config/blob/master/jx-apps.yml

2. Yeah we can do either really. I liked the idea of keeping apps that build images in the Dev cluster; but we can run pipelines in any cluster. It’s currently defined by the `Scheduler` referenced by the `SourceRepository` in each cluster

polothy:
Hey @jstrachan - sorry, I cannot figure out how to send feedback on other parts of your proposal, so I hope you don't mind I jot them down here.  Questions are as follows:

1. Do you plan to support `hemfile` for regular JX environments so we can get Helm 3 support everywhere?  EG: a new or updated one of these: https://github.com/jenkins-x/default-environment-charts
2. From what I understand, the `jx-apps.yaml` generates the two `helmfile` (one for apps and one for system)?  Not sure I understand the need for the abstraction and generation of the files.  EG: does this prevent us from using certain `helmfile` features?  What's gained by not having the two `helmfiles` and modifying those when you need new apps or system installs?

Thanks! This is all very exciting.  Cannot believe you are already taking this way beyond POC, you are a machine! 🚀 💯

polothy:
> Yeah we can do either really. I liked the idea of keeping apps that build images in the Dev cluster; but we can run pipelines in any cluster. It’s currently defined by the Scheduler referenced by the SourceRepository in each cluster

Right there with you on building images in dev only.  This is really only PRs for the environment, where you bump the version of AppFoo to 1.2.3.

jstrachan:
@polothy thanks! Any feedback anywhere is great ;)

1. absolutely - would love all environments (local or remote) to use helmfile & helm 3

2. good question. We could make the `jx-apps.yml` file be a real helmfile too; though am liking the concise yaml (if you compare `jx-apps.yml` to the generated `helmfile.yaml` files you’ll see what I mean ;).

It might be simpler to let the `jx-apps.yml` to allow an optional list of helmfiles to be applied after the system & apps?

Incidentally the system charts are for dealing with ingress/DNS/certs/TLS. Most apps (the `apps` folder) need to know what the domain is & the TLS certs etc; so the system charts sort that out - then `jx step verify ingress` runs to check that is all good, then the `apps` are installed

jstrachan:
@polothy btw I’d love this...
https://github.com/jenkins-x/jx/issues/6672

rawlingsj:
> It might be simpler to let the jx-apps.yml to allow an optional list of helmfiles to be applied after the system & apps?

+1 
There are lots of great helmfile features, which as folks get used to it I'm sure they'd want to use.  `jx-apps.yaml` is a really great way to add apps to boot but it is easy for folks just to add their own helmfiles if they wanted to.  Under the covers the `jx boot` pipeline just runs vanilla `helmfile sync`

jstrachan:
@rawlingsj maybe we could add a final pipeline step of `helmfile sync` in a `helmfiles/` folder with an empty `helmfile.yaml` file - then folks have a place to add any extra vanilla helmfile stuff without modifying the pipeline? 

Mind you it’s trivial to do that yourself via `jx create step` ;)

rawlingsj:
actually I think having a `helmfiles/ `  dir is a really nice idea, then folks can just drop their own helmfiles in

polothy:
We currently install _many_ cluster wide tools using a `helmfile`.  Just trying to figure out how these files all work and how we can add our own very custom charts and values.  I think some we would actually want to add to the `system` helmfile (or maybe even before `system` 🤔 ) because they are security related.

Totally right that we can just modify the pipeline and do a `helmfile sync` wherever we need.  I had not checked out the pipeline in the boot POC repo yet.  That's also perfectly fine for us.

Starting to get the impression that the `jx-apps.yaml` is probably more for distributing the JX apps/charts/install and maybe less about our random charts that we just want to add to the cluster.  That's totally fine and makes sense for distributing JX and all the things.


---Switching topics here---

I did notice [this helmfile](https://github.com/jenkins-x-labs/boot-helmfile-poc/blob/master/helmfile.yaml) and it seems like it isn't being used.  It looks like from the pipeline that the system and app helmfiles are generated, then the system helmfile is run, a JX command is run and then the apps helmfile is run.

I was wondering if you are interested in the [helmfile hooks](https://github.com/roboll/helmfile#hooks)?  The thought being, you could use the hook to run the JX command like so:

```yaml
- name: NAME
  namespace: NAMESPACE
  chart: foo/bar
  version: 0.1.0
  atomic: true
  wait: true
  labels:
    initPhase: "true"
  needs:
  - some_other/chart
  hooks:
  - events: ["postsync"]
    showlogs: true
    command: jx
    args:
    - step
    - create
    - install
    - values
    # And so on...
```

I'm not even close to knowing if that would work for you but, this would keep all the steps in the helmfile and maybe we could use that root helmfile to control the whole sync, like so:

```yaml
helmDefaults:
  # values, values, values...

helmfiles:
  - helmfiles/my_crazy_init.yaml
  - system/helmfile.yaml
  - apps/helmfile.yaml
  - helmfiles/cluster_tools.yaml
  - helmfiles/rule_the_world.yaml
```

That way JX can own/manage the lifecycle of the system and app helmfiles and we can shove in our stuff where we need to rather easily.  Pipeline is just a single `hemlfile sync`.

Just bouncing ideas around, like I said, this is hot stuff, I like what you have already and where you are taking it 🍻

polothy:
> @polothy btw I’d love this...
> #6672
> 
> actually to implement the diff, we’d want to run the PR pipeline in Staging / Production

oooOOooOOoo yesssss

Seems like it wouldn't be hard to implement - I just don't anything running _yet_.  Soon though, soon I get to play with the awesome toys.

jstrachan:
@polothy its totally fine for folks to do vanilla helmfile stuff; but hopefully we can convince folks to use Jenkins X Apps for lots of common things like nginx/gloo/istio/flagger/prometheus and the like. Mostly as lots of these tools need to be configured with secrets + domain/TLS/certs stuff; so we can figure out that binding between an off-the-shelf helm chart and the `jx-requirements.yml` which defines common ingress/storage/network stuff so that folks don't have to keep binding charts to their environments. configuring charts is undifferentiated heavy lifting requiring folks to grok all the internals of each charts `values.yaml` - we can certainly help include default bindings for common charts in the version stream so we can share that stuff. I'm hoping Jenkins X can keep improving the tooling around chart configuration too; so its easier to add popular OSS charts.

In terms of the [top level helmfile.yaml](https://github.com/jenkins-x/jenkins-x-boot-helmfile-config/blob/master/helmfile.yaml) thats mostly so you can easily uninstall everything in one simple command: https://github.com/jenkins-x/enhancements/blob/master/proposals/2/docs/getting-started.md#uninstall

incidentally this git repo is a cleaner example of what boot config looks like with helmfile: https://github.com/jenkins-x/jenkins-x-boot-helmfile-config (its the git repo used if you run `jx boot --helmfile`)

Thanks for the heads up on helm hooks, they look great. We could take advantage of those for sure; though I'm also liking the flexibility we get with pipelines. e.g. by default the pipeline steps are defined in the build pack: https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes/blob/master/packs/environment-helmfile/pipeline.yaml#L18 so that the default `jenkins-x.yml` file is [pretty much empty](https://github.com/jenkins-x/jenkins-x-boot-helmfile-config/blob/master/jenkins-x.yml) until you add a custom step via `jx create step` which greatly simplifies maintenance (if we need to modify one of the pipeline steps for everyone's boot install).  Also we're pretty used to having different pipeline steps for PRs versus release builds which the Jenkins X pipelines sort out nicely for us - along with having a nice composition, inheritence + override model so its easy for folks to disable steps or add custom pre/post steps or replace steps in a simple way.

polothy:
Hey @jstrachan - thanks for the continued feedback, this is extremely helpful!  After reviewing `jenkins-x-boot-helmfile-config` repo and that environment build pack, I'm starting to see the master plan here.  I really like it.  I especially like how the `jenkins-x-boot-helmfile-config` went on a diet and is much lighter weight than the original.  This is all great, full steam ahead!  We are starting to focus on boot real soon and this helmfile/multi-cluster/multi-team route looks well suited to our needs.  I'm really excited!

jstrachan:
@polothy many thanks!  I’m really excited too - it feels much simpler yet more flexible and powerful letting teams decide what apps they want to install & in which clusters / namespaces. 

I’m particularly happy that the boot config git repository now becomes a few simple YAML files that are specific to your dev cluster (or remote staging / production cluster). Hopefully soon all environments can use boot, helmfile & helm 3

vbehar:
hi, just to add a few more comments from our current usage of helmfile: we would need to be able to customize the `helmfile.yaml` more than what is currently allowed. to give you a few use-cases:
- we have releases with a bigger timeout, because we know deploying a new version might trigger the cluster autoscaler to create a new node, which will take time
- we have releases using the `force: true` flag, because they have a job resource, and in kube a job’s image is immutable...
- we use hooks to send some notifications. there might be different ways to implement that.

btw this is for deploying our own applications built and promoted by Jenkins X - not an extra 3rd party app.

vbehar:
for remote clusters, there is also the use-case of multiple prod clusters, with 1 per region for example. so we'll need to be able to add something to filter the apps to deploy on a specific cluster, and an input value to be able to generate per-cluster specific settings

polothy:
> for remote clusters, there is also the use-case of multiple prod clusters, with 1 per region for example.

We will also have multiple clusters, likely 1 per AWS region.

jstrachan:
I'm gonna close this PR for now - we're using this approach for folks to test out multi cluster: https://jenkins-x.io/docs/labs/boot/

PR #6678, opened 2020-02-01
fix: helmfile create namespace helm repo

rawlingsj:


rawlingsj:
overriding lint as there's an issue ATM but @abayer has a proposed fix which is being reviewed https://github.com/jenkins-x/jx/pull/6680

/override lint

PR #6550, opened 2020-01-17
fix: various problem due to non-conventional github repository name

aure-olli:
#### Submitter checklist

- [x] Change is code complete and matches issue description.
- [x] Change is covered by existing or new tests.

#### Description

Prior to this commit, the following problems where happening when a repository had a name not made of [a-z-] characters
- `jx step tag` would write the wrong docker image name (`strings.ToLower` was used (via `ToValidImageName`) instead of `naming.ToValidName`)
- `jx get build logs` would fail to list the running pipelines. This is a weird error, due to the fact that the labels of the `PipelineRun` and the `PipelineActivity` are not formatted the same way.
- `jx preview` and `jx promote` would fail to recognize the service (no formatting was performed)

#### Special notes for the reviewer(s)

Hopefully, non conventional repo names should work fine now.

PR #6693, opened 2020-02-05
fix: non-conventional github repository names don't appear in `jx get applications`

aure-olli:
#### Submitter checklist

#### Description

Complemetary to #6550
Fixes another problem for repository repositories with other characters than `[a-z-]`.
The application name was directly retrieved from the repository name and not formatted, thus it couldn't be displayed in `jx get applications`.


abayer:
/ok-to-test
/lgtm

As always, good catch. =)

PR #6699, opened 2020-02-06
feat: `helm lint` calls pass variables too

aure-olli:
#### Submitter checklist

#### Description

Complementary to #6427 and #6682 all calls to `helm lint` (including `jx step helm release`) also pass some variables
- `tags.jx-lint=true`
- `global.jxLint=true`
- `global.jxTypeEnv=lint`

This avoids having `helm lint` errors when the templates are looking for variables like `.Values.global.jxPreview` due to `global` being `nil`

### DIFF
--- a/pkg/cmd/opts/helm_test.go
+++ b/pkg/cmd/opts/helm_test.go
@@ -16,7 +16,7 @@ import (
 	"github.com/jenkins-x/jx/pkg/util"
 )
 
-// Test_HelmInitRecursiveDependencyBuild_extraction test that chart achives
+// Test_HelmInitRecursiveDependencyBuild_extraction tests that chart achives
 // are properly extracted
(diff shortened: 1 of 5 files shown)


aure-olli:
The best would to have the `helm lint` environment identical to the one used for `helm template`. Unfortunately it doesn't seem possible right now.

abayer:
/ok-to-test

This should be the last time you need someone to give you an `/ok-to-test` - I've sent you an invite to the `jenkins-x` org. =) Sadly, you still won't have access to the build logs (we still have work to do on scrubbing secrets), but your PRs will be built immediately. Thank you so much for your continuing contributions!

hferentschik:
What we don't have with this change or the changes for  #6427 and #6682 are reflections of this in the docs. Really it would be nice to have these variables documented and explained in the docs so that they actually become usable for everyone.

PR #6700, opened 2020-02-06
fix: REPLACE_ME build pack placeholders should work with modern syntax

### LINKED ISSUE #6698: REPLACE_ME placeholders in buildpacks using modern syntax aren't replaced
If your build pack is written in modern syntax, i.e., under `pipelines.release.pipeline`, not via lifecycles, `REPLACE_ME_*` placeholders are...not replaced. The logic that currently does that replacement is specific to/only run for transforming the lifecycle syntax into the modern model behind the scenes. This...was a mistake. We should be replacing the buildpack placeholders from _any_ syntax.

abayer:
#### Submitter checklist

#### Description

We were replacing the `REPLACE_ME_*` placeholders in directories in lifecycle-based `pipeline.yaml` syntax, but not in `pipeline`-based syntax in `pipeline.yaml`, so let's, y'know, do that.

/assign @hferentschik 

fixes #6698


pkg/tekton/syntax/pipeline.go
```diff
@@ -1091,6 +1097,126 @@ func (j *ParsedPipeline) GetPossibleAffinityPolicy(name string) *corev1.Affinity
...
+		for _, nested := range s.Loop.Steps {
+			nested.replacePlaceholdersInStep(args)
+			loopSteps = append(loopSteps, nested)
+		}
+		s.Loop.Steps = loopSteps
+	}
+}
+
+// modifyStep allows a container step to be modified to do something different
+func (s *Step) modifyStep(params StepPlaceholderReplacementArgs) {
```
hferentschik: Gee, I was not aware of how tightly the buildpack setup is tied in here. So changes to build packs might have very unexpected side effects. Either when we try to change them or when a user tries to customize. 

I realize though that we are just moving code here.

Would it potentially not better to move this somehow out and let the user make a choice, by for example choosing a specific build-pack w/ or w/o kaniko? I seem to recall @garethjevans saying that it would be nice to get this out, but that it would not be so easy?

Issue #6369, opened 2019-12-20
Lock the gcloud version 


ccojocar:
### Summary 

It seems that we don't lock the gcloud version. It always installs latest on MacOS: 

https://github.com/jenkins-x/jx/blob/8a4dac4e0bc5882e91d2d66d371a16c6f97485ee/pkg/cmd/opts/install.go#L778

See for instance eksctl is locked: 
https://github.com/jenkins-x/jenkins-x-versions/blob/bdab4d2a65b04658d30707877a9072f05738ba9d/packages/eksctl.yml#L1

### Jx version

The output of `jx version` is:

```
COPY OUTPUT HERE
```

### Jenkins type

- [x ] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

deanesmith:
@ccojocar, what version do you recommend we lock in at?

ccojocar:
Probably we can pick the latest from here https://cloud.google.com/sdk/docs/release-notes. I am not sure if there are any backward compatibility issues.

sladyn98:
I could take a shot at this. https://cloud.google.com/sdk/docs/release-notes The latest version from this sounds cool  ?

deanesmith:
@sladyn98 Contributions are strongly encouraged and this is quite valuable.

ccojocar:
@sladyn98 You can lock first the version in the `gcloud.yml` file similarly with the other CLIs https://github.com/jenkins-x/jenkins-x-versions/tree/master/packages, and then you can change the jx to download this specific version for gcloud in this method: https://github.com/jenkins-x/jx/blob/cdcf046be3172cc8b398c2a6f85f570226d3663f/pkg/cmd/opts/install.go#L779 You can use as an example the `InstallKustomize`: https://github.com/jenkins-x/jx/blob/cdcf046be3172cc8b398c2a6f85f570226d3663f/pkg/cmd/opts/install.go#L226

deanesmith:
Thank you @sladyn98  for your contribution.

hervelemeur:
@hferentschik any reason for reopening this issue while the PR made by @sladyn98 has been merged?

Issue #6661, opened 2020-01-30
Remove use of Brew to install binary dependencies on macOS


hferentschik:
in _opts/install.go_ we handle binary dependencies for macOS in the following form:

```
	if runtime.GOOS == "darwin" && !o.NoBrew {
		return o.RunCommand("brew", "install", "terraform")
	}
```

This has several issues:

* It is not integrated with the version stream nad just installs the latest version
* It creates a different installation path just for macOS (the other form of install is not tested)
* It is hard to control the version used (user might use Brew independently from jx) 

A better approach imo would be to use direct download of the binaries for all OS types and installing all binaries into _JX_HOME/bin_. _util.Command_ already prepends _JX_HOME/bin_ to command executions, so for the cases where we call these binaries from within _jx_ we are already covered.

PR #6705, opened 2020-02-06
[wip] feat: Install fixed gcloud version

sladyn98:
#### Submitter checklist

#### Description
Downloads and installs the latest version of gcloud 

Fixes : #6369


hferentschik:
@sladyn98, thanks for the PR. You will need to sign-off the commit in order to pass the DCO check. 

See:

hferentschik:
The semantic pull request status check fails because your commit is not following the conventional commits format. When you update the pull request could you please change your commit message? See also https://jenkins-x.io/docs/contributing/code/#the-commit-message

pkg/cmd/opts/install.go
```diff
@@ -780,7 +780,28 @@ func (o *CommonOptions) InstallGcloud() error {
 	if runtime.GOOS != "darwin" || o.NoBrew {
 		return errors.New("please install missing gcloud sdk - see https://cloud.google.com/sdk/downloads#interactive")
 	}
-	return o.RunCommand("brew", "cask", "install", "google-cloud-sdk")
+	latestVersion := "279.0.0"
```
hferentschik: We should not hard code version for external binaries here. That's what the version stream is for. See for example how it is done for the Helm install - https://github.com/jenkins-x/jx/blob/ea98d9eb2f0b2b0be06e0126e384e050f303f9cc/pkg/cmd/opts/install.go#L355

For this to work you will need to create a pull request in this repo first https://github.com/jenkins-x/jenkins-x-versions/tree/master/packages where you add a _ gcloud.yml_ specifying the version.

sladyn98: @hferentschik Thanks for the feedback.  I already have made a PR to that repository https://github.com/jenkins-x/jenkins-x-versions/pull/811

sladyn98:
@hferentschik I signed off the commit and fixed the semantics, should all the commits have the same semantics?

abayer:
FYI, the lint failure is legit:

```
Gofmt errors in files:
./pkg/cmd/opts/install.go
diff -u ./pkg/cmd/opts/install.go.orig ./pkg/cmd/opts/install.go
--- ./pkg/cmd/opts/install.go.orig	2020-02-10 05:49:14.534180481 +0000
+++ ./pkg/cmd/opts/install.go	2020-02-10 05:49:14.534180481 +0000
@@ -790,7 +790,7 @@
 	}
 	clientURL := fmt.Sprintf("https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-%s-%s-%s.tar.gz", stableVersion, runtime.GOOS, runtime.GOARCH)
 	binDir, err := util.JXBinLocation()
-	binary:="gcloud"
+	binary := "gcloud"
 	if err != nil {
 		return err
 	}
```

pkg/cmd/opts/install.go
```diff
@@ -780,7 +780,35 @@ func (o *CommonOptions) InstallGcloud() error {
 	if runtime.GOOS != "darwin" || o.NoBrew {
```
hferentschik: I think this if statement is not quite correct anymore. With the code below we can install `gcloud` for any OS into _JX_HOME/bin_, right?

hferentschik: Also in the current form the use of the _NoBrew_ flag does not make sense anymore, since we are not using brew in this function at all.

pkg/cmd/opts/install.go
```diff
@@ -780,7 +780,35 @@ func (o *CommonOptions) InstallGcloud() error {
...
 		return errors.New("please install missing gcloud sdk - see https://cloud.google.com/sdk/downloads#interactive")
 	}
-	return o.RunCommand("brew", "cask", "install", "google-cloud-sdk")
```
hferentschik: There is a conceptual difference here which is quite important. `brew cask install` will install `gcloud` in a way that it is automatically in the user's path. He is directly able to execute it. This is not the case for the current code installing into _JX_HOME/bin_. The user would have to add the jx bin directory to his path before being able to use it. 

Do we install _gcloud_ in order for the user to use it directly or do we need to install _gcloud_ because `jx` needs to call it. For the latter installing into _JX_HOME/bin_ is no issue, since we can control the PATH when we are calling it. If we want to help the user to install _gcloud_ for general use this might not be the best option. 

Eventually, the use of Homebrew needs to go (see #6661), but maybe a better solution, for now, is to keep using `brew` for macOS and in the other cases use the approach below (basically aligning the code with _InstallHelm_). @ccojocar WDYT?

Another thing I was hoping to solve as part of #6661 is how we version the binary. As a Jenkins X user you might be working with multiple Jenkins X installs in slightly different versions. As you switch between various clusters you would theoretically have to change the versions in _JX_HOME/bin_. For that to happen the binaries would need to be installed by namespace one way or the other. At least one would need to check/validate the binary version before using it.

The whole area is quite tricky and I am not quite sure how to move forward here. The generic install code below is valuable for sure, I am just not sure what the best way is to integrate it.

pkg/cmd/opts/install.go
```diff
@@ -780,7 +780,35 @@ func (o *CommonOptions) InstallGcloud() error {
...
+	versionResolver, err := o.GetVersionResolver()
```
hferentschik: What's about adding an integration test which shows that this code works. Something along these lines:

https://github.com/jenkins-x/jx/pull/6494/files#diff-a6e8c21bcfd8a25876b6a9d4ee96e191

sladyn98: Sounds cool :+1: I will give it a shot.

pkg/cmd/opts/install.go
sladyn98: Yeah with the current code the binary gets downloaded into the `JX_HOME/bin` . So IMO there is not point of the NoBrew flag anymore.

pkg/cmd/opts/install.go
sladyn98: > At least one would need to check/validate the binary version before using it.

I am quite new to the jenkins X codebase here so forgive me for the inexperience but validation of the binary before use for different versions sounds a viable solution since the user would end up changing the versions "theoretically".

hferentschik:
So I did have a quick look at the our codebase, and we do indeed call the `gcloud` binary from within `jx`, so this proposed change is addressing this use case. It would mean that for the cases where we call `jx` we will use the version we are going to install into _JX_HOME/bin_. See _helper.go_ in the _gke_ package. So it makes sense to get this merged.

It won't address installing `gcloud` for the user for general use.

pkg/cmd/opts/install.go
hferentschik: > I am quite new to the jenkins X codebase here so forgive me for the inexperience 

No problem.

> but validation of the binary before use for different versions sounds a viable solution

Potentially viable. I think, however, we can defer this problem to #6661. It is probably best to "fix" this for all binary installs together. It is really an orthogonal issue. Maybe once we have this PR merged, you are interested to take this further? 
For now let's install the binary into _JX_HOME/bin_ the same way the other functions in this file do it. I was just providing some additional context related to this change.

hferentschik:
@sladyn98, running `make fmt` should sort out your lint issue.

pkg/cmd/opts/install.go
sladyn98: @hferentschik I am definitely interested to take this further, if you have any related issues to take a look at like #6661 I would love to get started.

hferentschik:
@sladyn98, just touching base. Are you still working on this pull request? Are you trying to add an integration test?

sladyn98:
Yeah sorry I just had exams so will be pushing the test soon

hferentschik:
> Yeah sorry I just had exams so will be pushing the test soon

no problem at all :-) Just checking.

nehagup:
Hey @sladyn98 I've recently worked on similar issue for adding support to install kustomize. Let me know if you need any help writing test cases for gcloud.

pkg/cmd/opts/install.go
```diff
@@ -780,7 +780,35 @@ func (o *CommonOptions) InstallGcloud() error {
...
+	if err != nil {
+		return err
+	}
+	stableVersion, err := versionResolver.StableVersionNumber(versionstream.KindPackage, "gcloud")
+	if err != nil {
+		return err
+	}
+	clientURL := fmt.Sprintf("https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-%s-%s-%s.tar.gz", stableVersion, runtime.GOOS, runtime.GOARCH)
```
sladyn98: This url returns a 404 error because I guess the architecture for `amd64` does not have a download link.
We have the https://cloud.google.com/sdk/docs/quickstart-linux which provides us with the following links to donwload the sdk but when we leave it to a customizable link that link does not exist. https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-281.0.0-linux-x86_64.tar.gz

hferentschik:
thanks for the pull request update @@sladyn98. Your second commit is not fulfilling the conventional commit message format, so the semantic pull request check is failing. Also, you need to sign off all commits.

IMO, the best is if you rebase your pull request and squash both commits into a single commit. Test and production code change belong together anyway. Let me know if you need help with that.

Last but not least, there is a formatting issue, use `make fmt` to make sure the Go code is properly formatted.

pkg/cmd/opts/install.go
hferentschik: Thought we agreed on removing this?

hferentschik: Also, I am on macOS and for me the test is not even properly executing since it returns an error here. If anything it should be `!o.NoBrew` (I know this inverted logic is tricky)

pkg/cmd/opts/gocloud_integration_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	err = o.InstallGcloud()
+	if err != nil {
+		log.Logger().Warnf("Error in installing gcloud: %s", err)
+	}
+	
+	if !assert.FileExists(t, filepath.Join(testJXHome, "bin", "gcloud")) || err != nil {
+		t.Errorf("Gcloud was not installed in the temp dir of test TestCommonOptions_EnsureGCloudBinaryInstalled %s: %s", testJXHome, err,)
+	} 
+}
```
hferentschik: Missing new line

pkg/cmd/opts/install.go
hferentschik: We will need to change the URL accordingly. For example you can map _amd64_ to _x86_64_

pkg/cmd/opts/gocloud_integration_test.go
```diff
@@ -0,0 +1,38 @@
...
+			log.Logger().Warnf("Unable to remove temporary directory %s: %s", testJXHome, err)
+		}
+	}()
+
+	err = o.InstallGcloud()
+	if err != nil {
+		log.Logger().Warnf("Error in installing gcloud: %s", err)
+	}
+	
+	if !assert.FileExists(t, filepath.Join(testJXHome, "bin", "gcloud")) || err != nil {
```
hferentschik: the assert is the opposite to what it should be and `assert.FileExists(t, filepath.Join(testJXHome, "bin", "gcloud")` should be sufficient.

pkg/cmd/opts/gocloud_integration_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	defer func() {
+		err := testhelpers.CleanupTestJxHomeDir(origJXHome, testJXHome)
+		if err != nil {
+			log.Logger().Warnf("Unable to remove temporary directory %s: %s", testJXHome, err)
+		}
+	}()
+
+	err = o.InstallGcloud()
+	if err != nil {
```
hferentschik: In the final test you don't want to have this logging. The assertion should be good enough

pkg/cmd/opts/gocloud_integration_test.go
```diff
@@ -0,0 +1,38 @@
...
+func TestCommonOptions_EnsureGCloudBinaryInstalled(t *testing.T) {
+	o := opts.NewCommonOptionsWithFactory(clients.NewFactory())
+	o.NoBrew = true
+
+	origJXHome, testJXHome, err := testhelpers.CreateTestJxHomeDir()
+	if err != nil {
+		t.Errorf("Failed to create a test JX Home directory")
+	}
+
+	defer func() {
```
hferentschik: 👍

pkg/cmd/opts/install.go
sladyn98: Agreed, I might have missed it while pushing the change.

pkg/cmd/opts/gocloud_integration_test.go
sladyn98: Doesn't `assert.FileExists` return true, so in this case if it returns false we throw the error ?

sladyn98:
> Seems there are still some pieces missing. In particular, the actual download does not seem to work yet. @nehagup was also looking into this issue. Maybe she can help?

Yeah the actual download does not seem to happen, it just times out :(

sladyn98:
@hferentschik I guess my PR now matches the semantics ?

hferentschik:
>  I guess my PR now matches the semantics ?

Yes, that part is sorted now.

pkg/cmd/opts/gocloud_integration_test.go
```diff
@@ -0,0 +1,36 @@
+package opts_test
```
hferentschik: you will need to classify your test, either 'unit' or 'integration'. unit tests are standalone and don't needs any connection to the outside world and integration tests need a cluster connection or some other form of external communication, aka downloading things from the internet.

Without classifying the test it won't pass the linting check.

hferentschik: from the build log:

```
Showing logs for build jenkins-x-jx-pr-6705-lint-hd552-94 stage ci and container step-ensure-test-classification
The following 1 test files are not classified with a valid Go build tag [unit|integration]
./hack/../pkg/cmd/opts/gocloud_integration_test.go
Pipeline failed on stage 'ci' : container 'step-ensure-test-classification'. The execution of the pipeline has stopped.
```

pkg/cmd/opts/gocloud_integration_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	defer func() {
+		err := testhelpers.CleanupTestJxHomeDir(origJXHome, testJXHome)
+		if err != nil {
+			log.Logger().Warnf("Unable to remove temporary directory %s: %s", testJXHome, err)
+		}
+	}()
+
+	err = o.InstallGcloud()
+
```
hferentschik:
```
assert.NoError(t, err)
assert.FileExists(t, filepath.Join(testJXHome, "bin", "gcloud"))
```

pkg/cmd/opts/install.go
```diff
@@ -777,10 +777,39 @@ func (o *CommonOptions) InstallMinishift() error {
...
+	if err != nil {
+		return err
+	}
+	stableVersion, err := versionResolver.StableVersionNumber(versionstream.KindPackage, "gcloud")
+	if err != nil {
+		return err
 	}
-	return o.RunCommand("brew", "cask", "install", "google-cloud-sdk")
+	architecture := runtime.GOARCH
+	if runtime.GOARCH == "amd64" {
```
hferentschik: do any other GOARCH strings need mapping? I guess that's the one we really need.

sladyn98: According to what is available on the gcloud sdk website an amd32 might need a different mapping if that architecture needs downloading.
More specifically this link : `https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-282.0.0-linux-x86.tar.gz`

pkg/cmd/opts/install.go
```diff
@@ -777,10 +777,42 @@ func (o *CommonOptions) InstallMinishift() error {
...
+		return err
+	}
+	architecture := runtime.GOARCH
+	if runtime.GOARCH == "amd64" {
+		architecture = "x86_64"
+	}
+	if runtime.GOARCH == "amd32" {
+		architecture = "x86"
+	}
+	clientURL := fmt.Sprintf("https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-%s-%s-%s.tar.gz", stableVersion, runtime.GOOS, architecture)
```
nehagup: Hey @sladyn98 you can probably try using this url.
```
clientURL := fmt.Sprintf("https://storage.googleapis.com/cloud-sdk-release/google-cloud-sdk-%s-%s-x86.tar.gz", stableVersion, runtime.GOOS)
	if strings.Contains(runtime.GOARCH, "64") {
		clientURL = fmt.Sprintf("https://storage.googleapis.com/cloud-sdk-release/google-cloud-sdk-%s-%s-x86_64.tar.gz", stableVersion, runtime.GOOS)
	}
```

pkg/cmd/opts/install.go
nehagup: The idea after un-tzipping the downloaded file is to copy just the binary in the JX bin folder and delete the rest.

go.mod
```diff
@@ -171,3 +171,5 @@ replace github.com/Azure/azure-sdk-for-go => github.com/Azure/azure-sdk-for-go v
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v10.14.0+incompatible
 
 replace github.com/banzaicloud/bank-vaults => github.com/banzaicloud/bank-vaults v0.0.0-20190508130850-5673d28c46bd
+
+go 1.13
```
hferentschik: we have not upgraded to Go 1.13 yet. We want to do this in a dedicated move. This change would force everyone onto 1.13

pkg/cmd/opts/install.go
```diff
@@ -777,10 +777,38 @@ func (o *CommonOptions) InstallMinishift() error {
...
+	clientURL := fmt.Sprintf("https://storage.googleapis.com/cloud-sdk-release/google-cloud-sdk-%s-%s-x86.tar.gz", stableVersion, runtime.GOOS)
+	if strings.Contains(runtime.GOARCH, "64") {
+		clientURL = fmt.Sprintf("https://storage.googleapis.com/cloud-sdk-release/google-cloud-sdk-%s-%s-x86_64.tar.gz", stableVersion, runtime.GOOS)
+	}
+	binDir, err := util.JXBinLocation()
+	binary := "gcloud"
+	if err != nil {
+		return err
+	}
+	fileName, flag, err := packages.ShouldInstallBinary(binary)
```
hferentschik: This one is tricky. @nehagup was also running into it. We don't want to use _ShouldInstallBinary_  here. For some reason it checks generally in the global PATH whether a given binary is installed. This means if you have any the binary installed anywhere, it would already return _false_. Our aim it to install into _JX_HOME/bin_ if it does not exist there, independent of other versions of the binary on the PATH.

```go
	pgmPath, err := exec.LookPath(fileName)
	if err == nil {
		log.Logger().Debugf("%s is already available on your PATH at %s", util.ColorInfo(fileName), util.ColorInfo(pgmPath))
		return
	}
```

I think we could remove the above code from  `ShouldInstallBinary` to rectify this in general, but that might affect some other callers of this method. 

The alternative is what @nehagup has done for installing `kustomize` - https://github.com/jenkins-x/jx/blob/master/pkg/cmd/opts/install.go#L237
One just checks whether the binary exists at the expected path.

sladyn98: Yeah I guess checking if the file exists in the `JX_HOME/bin` makes sense, rather than checking the whole path. The function `ShouldInstallBinary` can be changed later if required.

pkg/cmd/opts/install.go
```diff
@@ -777,10 +777,38 @@ func (o *CommonOptions) InstallMinishift() error {
...
+	if err != nil || !flag {
+		return err
+	}
+	fullPath := filepath.Join(binDir, fileName)
+	tarFile := fullPath + ".tgz"
+	err = packages.DownloadFile(clientURL, tarFile)
```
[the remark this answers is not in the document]
sladyn98: Yeah the gcloud binary does live under the `google-cloud-sdk/bin/gcloud` , but UnTargz would decompress the entire ziped package. Instead could we copy the file from there onto the `JX_HOME/bin` directory, or is there a better way to do it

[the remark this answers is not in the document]
nehagup: Currently we're doing it by copying to the binary to `JX_HOME/bin`.

pkg/cmd/opts/gocloud_integration_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"github.com/jenkins-x/jx/pkg/cmd/clients"
+	"github.com/jenkins-x/jx/pkg/cmd/opts"
+	"github.com/jenkins-x/jx/pkg/log"
+)
+
+func TestCommonOptions_EnsureGCloudBinaryInstalled(t *testing.T) {
+	o := opts.NewCommonOptionsWithFactory(clients.NewFactory())
+	o.NoBrew = true
```
nehagup: Since we're not having brew checked on installation, I'm not sure why we need this.

pkg/cmd/opts/gocloud_integration_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	origJXHome, testJXHome, err := testhelpers.CreateTestJxHomeDir()
+	if err != nil {
+		t.Errorf("Failed to create a test JX Home directory")
```
nehagup: We can simply `assert.NoError(t, err, "Failed to create a test JX Home directory")`

pkg/cmd/opts/gocloud_integration_test.go
sladyn98: Yeah removing :+1:

pkg/cmd/opts/gocloud_integration_test.go
sladyn98: Thanks :)

pkg/cmd/opts/install.go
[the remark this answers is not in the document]
sladyn98: You mean is that already done or I need to do it?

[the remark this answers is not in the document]
nehagup: In every installation we need to copy the respective binary to the current JXBIN directory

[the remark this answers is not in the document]
sladyn98: Thanks I'll get it working and push the changes :)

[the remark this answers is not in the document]
sladyn98: @nehagup 
```
srcFolder := "copy/from/path"
destFolder := "copy/to/path"
cpCmd := exec.Command("cp", "-rf", srcFolder, destFolder)
err := cpCmd.Run()
```
Would we use this to copy files ?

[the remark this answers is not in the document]
hferentschik: No, this is Os dependent and won't for example work on Windows. We have already helpers for that. See files.go in the utils package. There is for example a _CopyDir_.

hferentschik:
@sladyn98, do you have an update on this PR? Still working on it?

sladyn98:
@hferentschik Yes I have committed the new change

go.mod
nehagup: @sladyn98 it would be great if we can simply remove this file changes during adding commit.

go.sum
```diff
@@ -852,6 +852,7 @@ github.com/shurcooL/go v0.0.0-20180423040247-9e1955d9fb6e h1:MZM7FHLqUHYI0Y/mQAt
 github.com/shurcooL/go v0.0.0-20180423040247-9e1955d9fb6e/go.mod h1:TDJrrUr11Vxrven61rcy3hJMUqaf/CLWYhHNPmT14Lk=
 github.com/shurcooL/graphql v0.0.0-20180924043259-e4a3a37e6d42 h1:YIoQLhvoRcfiL0pyxqkESFZXa7jQrcfLTUSSUeyYMO8=
 github.com/shurcooL/graphql v0.0.0-20180924043259-e4a3a37e6d42/go.mod h1:AuYgA5Kyo4c7HfUmvRGs/6rGlMMV/6B1bVnB9JxJEEg=
+github.com/sladyn98/Jenkins v0.0.0-20190604142804-fdcaef37b9f6 h1:QUtFPzC1VaYvHYE5N0EAPCo58zb9Tgi6bpL7zGVP6Nc=
```
nehagup: @sladyn98 it would be great if we can simply remove this file changes during adding commit.

pkg/cmd/opts/install.go
```diff
@@ -777,10 +777,51 @@ func (o *CommonOptions) InstallMinishift() error {
...
+	if err != nil {
+		return err
+	}
+	err = util.UnTargz(tarFile, binDir, []string{binary, fileName})
```
nehagup: Please also remove the un-used files and tar-ball after we've copied the binary.

pkg/cmd/opts/install.go
```diff
@@ -777,10 +777,55 @@ func (o *CommonOptions) InstallMinishift() error {
...
-	return o.RunCommand("brew", "cask", "install", "google-cloud-sdk")
+	err = util.UnTargz(tarFile, binDir, []string{binary, fileName})
+	if err != nil {
+		return err
+	}
+	err := util.CopyFile(fileName + "/bin/gcloud",binDir)
+	if err != nil {
+		return err
+	}
+	os.Remove(tarfile)
```
nehagup: We can put this in a defer func().

hferentschik:
What is the state on this PR? @sladyn98 or @nehagup any of you still working on this?

pkg/cmd/opts/install.go
sladyn98: Yeah +1

sladyn98:
@hferentschik Yes I pushed a change and got a review for it, does this need more tests ?

hferentschik:
> Yes I pushed a change and got a review for it, does this need more tests ?

I'll take a look

hferentschik:
@sladyn98, I think there is a syntax error:

```
pkg/cmd/opts/install.go:786:6: no new variables on left side of :=
pkg/cmd/opts/install.go:790:12: undefined: tarfile
```

hferentschik:
@sladyn98 what happened with the integration test? The latest version of this PR does not contain any test file anymore. 

What's about we close this PR and you open a new one? Things get a bit confusing on this pr/thread.

sladyn98:
I guess the test file might have gotten removed. I will try adding it again

sladyn98:
@hferentschik I have added the files and made some changes

sladyn98:
@hferentschik Do you want me to close this PR and open a new one :)

hferentschik:
@sladyn98, I think we have to abandon the attempts to install cloud this way. Even if we get the single gcloud file installed into $JX_HOME/bin it won't work, since gcloud is actually not a single binary, but a python application with many files. We would have to extract the whole bin directory (or maybe even more) somehow into $JX_HOME. Possible, but probably not a good idea for now.

Sorry that it took so long to realize this.

hferentschik:
Closing this PR since we won't be able to install gcloud the same way as other single binary dependencies like kubectl, kutomize, etc.

PR #6716, opened 2020-02-08
fix: ignore github app check if credential-helper is set

garethjevans:
#### Submitter checklist

fixes #

### DIFF
--- a/pkg/cmd/step/git/credentialhelper/step_git_credentialhelper.go
+++ b/pkg/cmd/step/git/credentialhelper/step_git_credentialhelper.go
@@ -1,0 +1,158 @@
+package credentialhelper
+
+import (
+	"os"
+
+	"github.com/jenkins-x/jx/pkg/auth"
+	"github.com/jenkins-x/jx/pkg/cmd/opts/step"
+	"github.com/jenkins-x/jx/pkg/gits/credentialhelper"
+	"github.com/pkg/errors"
+
+	"github.com/jenkins-x/jx/pkg/cmd/helper"
+
+	"github.com/jenkins-x/jx/pkg/cmd/opts"
+	"github.com/jenkins-x/jx/pkg/cmd/templates"
+	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/spf13/cobra"
+)
+
+const (
+	optionGitHubAppOwner = "github-app-owner"
+)
+
+// StepGitCredentialHelperOptions contains the command line flags
+type StepGitCredentialHelperOptions struct {
+	step.StepOptions
+
+	OutputFile        string
+	GitKind           string
+	GitHubAppOwner    string
+}
+
+var (
+	StepGitCredentialHelperLong = templates.LongDesc(`
+		This pipeline step generates a Git credentials file for the current Git provider secrets
+
+`)
+
+	StepGitCredentialHelperExample = templates.Examples(`
+		# respond to a git credentials request
+		jx step git credentials-helper
+`)
+)
+
+func NewCmdStepGitCredentialHelper(commonOpts *opts.CommonOptions) *cobra.Command {
+	options := StepGitCredentialHelperOptions{
+		StepOptions: step.StepOptions{
+			CommonOptions: commonOpts,
+		},
+	}
+	cmd := &cobra.Command{
+		Use:     "credential-helper",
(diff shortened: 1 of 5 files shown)


garethjevans:
I'm experimenting on this PR at the moment

pkg/cmd/step/git/credentialhelper/step_git_credentialhelper.go
```diff
@@ -0,0 +1,159 @@
...
+
+const (
+	optionGitHubAppOwner = "github-app-owner"
+)
+
+// StepGitCredentialHelperOptions contains the command line flags
+type StepGitCredentialHelperOptions struct {
+	step.StepOptions
+
+	GitKind        string
```
hferentschik: I think we would not need these flags in this case. They would not be passed anyways, right?

pkg/cmd/step/git/credentialhelper/step_git_credentialhelper.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+// NewCmdStepGitCredentialHelper creates a new command
+func NewCmdStepGitCredentialHelper(commonOpts *opts.CommonOptions) *cobra.Command {
+	options := StepGitCredentialHelperOptions{
+		StepOptions: step.StepOptions{
+			CommonOptions: commonOpts,
+		},
+	}
+	cmd := &cobra.Command{
```
hferentschik: I think a dedicated command might not be a bad idea. Should probably be hidden in this case and we probably need to move some things around in order to not to duplicate code.

pkg/cmd/step/git/credentialhelper/step_git_credentialhelper.go
```diff
@@ -0,0 +1,159 @@
...
+	var credentialList []credentialhelper.GitCredential
+
+	cfg := authConfigSvc.Config()
+	if cfg == nil {
+		return nil, errors.New("no git auth config found")
+	}
+
+	for _, server := range cfg.Servers {
+		var auths []*auth.UserAuth
+		if o.GitHubAppOwner != "" {
```
hferentschik: What's about we remove this and always use _server.User_. Then we build the GitCredentials and pass the UserAuth.GithubAppOwner and use it as path component. When the request comes in we match against the path.

pkg/cmd/step/git/credentialhelper/step_git_credentialhelper.go
garethjevans: Yeah, am hoping I can get rid of these with a bit of a refactor

hferentschik: I think I know how. Maybe we can have a chat once I am home. Using the path matching might work out quite nicely and we even git rid of these flags which would be another win really. 

What's about doing the fix in the old command and just focus on the required changes to fix the issue, opposed to also extract a new command?

hferentschik: we probably need an issue in this repo as well.

pkg/cmd/step/git/credentialhelper/step_git_credentialhelper.go
```diff
@@ -0,0 +1,159 @@
...
+const (
+	optionGitHubAppOwner = "github-app-owner"
+)
+
+// StepGitCredentialHelperOptions contains the command line flags
+type StepGitCredentialHelperOptions struct {
+	step.StepOptions
+
+	GitKind        string
+	GitHubAppOwner string
```
hferentschik: @garethjevans, actually it might be the path is not passed in a get request. But is _GitHubAppOwner_ not effectively the organisation/owner of the repo in question? If so, the probably absolute easiest would be to pass it along. We should have all the information we need when we set the git config for the repo.

garethjevans: Yeah, GitHubAppOwner is effectively the organisation/owner.  Just so I understand you correctly, you mean setting this value inside `configureJxAsGitCredentialHelper`

hferentschik:
I guess this pull request is superseded by #6720, right? Should we close it?

garethjevans:
Closing this in favour of  #6720

PR #6720, opened 2020-02-10
feat: always pass the repo owner to the credential helper

garethjevans:
#### Submitter checklist

fixes #

### DIFF
--- a/pkg/cmd/step/git/credentials/step_git_credentials.go
+++ b/pkg/cmd/step/git/credentials/step_git_credentials.go
@@ -25,16 +25,20 @@ import (
 )
 
 const (
-	optionOutputFile     = "output"
+	optionOutputFile = "output"
+	// Deprecated
 	optionGitHubAppOwner = "github-app-owner"
+	optionRepoOwner      = "repo-owner"
 )
 
 // StepGitCredentialsOptions contains the command line flags
 type StepGitCredentialsOptions struct {
 	step.StepOptions
 
-	OutputFile        string
+	OutputFile string
+	// Deprecated
 	GitHubAppOwner    string
+	RepoOwner         string
 	GitKind           string
 	CredentialsSecret string
 	CredentialHelper  bool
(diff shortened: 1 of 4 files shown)


abayer:
Mind creating/linking with `fixes #...` an issue? =)

garethjevans:
fixes #6722

pkg/cmd/step/git/credentials/step_git_credentials.go
```diff
@@ -25,16 +25,20 @@ import (
 )
 
 const (
-	optionOutputFile     = "output"
+	optionOutputFile = "output"
+	// Deprecated
```
hferentschik: What would be our plan to eventually remove this? Should we at least create an issue to keep of track and maybe add a 'deprecation-followup' label of similar? One could then on a regular basis review these issues and evaluate of a certain deprecated feature can be removed. 

I think it is pretty much only the SaaS which uses this, so provided you can update in one go, I'd even be tempted to remove directly. WDYT?

pkg/cmd/step/verify/step_verify_environments.go
```diff
@@ -510,9 +510,9 @@ func (o *StepVerifyEnvironmentsOptions) pushDevEnvironmentUpdates(environmentRep
 
 	err = gitter.Push(localRepoDir, remoteName, true, "master")
 	if err != nil {
-		return errors.Wrapf(err, "unable to push %s to %s", localRepoDir, environmentRepo.URL)
+		return errors.Wrapf(err, "unable to push %s to %s", localRepoDir, environmentRepo.CloneURL)
 	}
-	log.Logger().Infof("Pushed Git repository to %s\n\n", util.ColorInfo(environmentRepo.HTMLURL))
+	log.Logger().Infof("Pushed Git repository to %s", util.ColorInfo(environmentRepo.HTMLURL))
```
hferentschik: 👍

pkg/gits/helpers.go
```diff
@@ -713,13 +716,15 @@ func ForkAndPullRepo(gitURL string, dir string, baseRef string, branchName strin
...
 
-func configureJxAsGitCredentialHelper(dir string, gitter Gitter) error {
+func configureJxAsGitCredentialHelper(dir string, gitter Gitter, repoOwner string) error {
 	// configure jx as git credential helper for this repo
 	jxProcessBinary, err := os.Executable()
 	if err != nil {
 		return errors.Wrapf(err, "unable to determine jx binary location")
 	}
-	return gitter.Config(dir, "--local", "credential.helper", fmt.Sprintf("%s step git credentials --credential-helper", jxProcessBinary))
+	config := []string{"--local", "credential.helper", fmt.Sprintf("%s step git credentials --credential-helper --repo-owner %s", jxProcessBinary, repoOwner)}
```
hferentschik: +1, definitely feels like the smallest possible change to make this work for your use case which at the same time is not completely irrational for the other users.

pkg/cmd/step/git/credentials/step_git_credentials.go
garethjevans: This command (unfortunately) is embedded elsewhere in the SaaS pipelines so I’m not going to be able to remove it in one go, but should be able to in a few weeks once we’ve migrated over. I’ll create a task to do that

garethjevans: Taks to remove #6723

Issue #6726, opened 2020-02-11
Enable integrated testing of the jx stack for PRs spanning multiple repositories


abayer:
Currently, there is no way to effectively run BDD tests for jx for a combination of PRs from multiple repositories, short of manually spinning up a cluster and running the BDD tests by hand. Given the number of cases where a PR to jx has a corresponding PR to jenkins-x-boot-config (or other repos, but that's the combination that's most common), and we can't really test one without the other. The result is that one of the PRs can be made to pass the repo-specific BDD tests without the other, basically by not actually exercising the changed/new behavior, and we don't discover the problems with that PR until it's merged and made it into the version stream and the other PR can actually be tested. This is painfully inefficient, and leads to issues slipping through.

Therefore, we need a way to test PRs against multiple repositories together.

We're going to target a relatively small number of repositories in the initial scope, primarily because these are the repositories that most often require such cross-repo changes:
* https://github.com/jenkins-x/jx
* https://github.com/jenkins-x/jenkins-x-boot-config
* https://github.com/jenkins-x/jenkins-x-versions
* https://github.com/jenkins-x/jenkins-x-platform

Once those are in place, we'll see what other repositories (such as https://github.com/jenkins-x-charts/jxboot-resources and https://github.com/jenkins-x/jenkins-x-boot-helmfile-config) make sense to incorporate next.

jstrachan:
@abayer sounds good to me.

Incidentally whenever the helmfile stuff merges, to do the same kinda thing with helmfile/helm 3 we won't need to do custom builds of `jenkins-x-platform` (as its no longer used) and don't need a custom fork/PR of `jenkins-x-boot-config` as there's nothing version specific in there any more. So we could just use a custom version stream + custom builder images (which we already have) + the jx PR.

BTW I hit an issue recently with builder images not being used in lighthouse for metapipeline client; so we maybe wanna figure out a way to parameterise that image a little easier too. We shouldn't need to do a new lighthouse release or anything; just figure out a way to modify its builder image used inside the metapipeline client. Hopefully the `jenkins-x-versions` changes above fixes that too?

Whether using helm 2 / helm 3; it might be nice to verify all images used in all pods throughout the bdd tests too; just in case something goes a bit wonky in the helm charts / values files etc. Its so easy for you to think you've swapped an image out and a different one sneaks in ;)

PR #6754, opened 2020-02-17
test: Add image tag overrides for all controllers/cronjobs to boot test

### LINKED ISSUE #6753: Add all jenkins-x-platform controllers/cronjobs' images to boot values.yaml.template override in jx BDD test
Stemming from #6736, one quick win is to add the controllers (other than the build controller, which is already there) and cronjobs to `boot-vault.platform.yaml.template` with `$VERSION` for the tag. This will guarantee that all of those controllers and cronjobs will run using the images built by the PR test context, regardless of what else we do in re: cross-repo PR testing.

abayer:
#### Submitter checklist

#### Description

Part of #6726 - valuable enough on its own even if we weren't to do anything else for cross-repo PR testing.

fixes #6753

### DIFF
--- a/jenkins-x-bdd.yml
+++ b/jenkins-x-bdd.yml
@@ -62,19 +62,58 @@ pipelineConfig:
 
               - name: build-and-push-image
                 command: /kaniko/executor
-                args: ['--dockerfile=/workspace/source/Dockerfile','--destination=gcr.io/jenkinsxio/jx:${inputs.params.version}','--context=/workspace/source','--cache-repo=jenkins-x-docker-registry.jx.svc.cluster.local:5000/','--cache=true','--cache-dir=/workspace','--skip-tls-verify-registry=jenkins-x-docker-registry.jx.svc.cluster.local:5000']
+                args:
+                  - --dockerfile=/workspace/source/Dockerfile.builder-jx
+                  - --destination=gcr.io/jenkinsxio/builder-jx:${inputs.params.version}
+                  - --context=/workspace/source
+                  - --cache-repo=jenkins-x-docker-registry.jx.svc.cluster.local:5000/
+                  - --cache=true
+                  - --cache-dir=/workspace
+                  - --skip-tls-verify-registry=jenkins-x-docker-registry.jx.svc.cluster.local:5000
 
               - name: build-and-push-nodejs
                 command: /kaniko/executor
-                args: ['--dockerfile=/workspace/source/Dockerfile.builder-nodejs','--destination=gcr.io/jenkinsxio/builder-nodejs:${inputs.params.version}','--context=/workspace/source','--cache-repo=jenkins-x-docker-registry.jx.svc.cluster.local:5000/','--cache=true','--cache-dir=/workspace','--skip-tls-verify-registry=jenkins-x-docker-registry.jx.svc.cluster.local:5000']
+                args:
+                  - --dockerfile=/workspace/source/Dockerfile.builder-nodejs
+                  - --destination=gcr.io/jenkinsxio/builder-nodejs:${inputs.params.version}
+                  - --context=/workspace/source
+                  - --cache-repo=jenkins-x-docker-registry.jx.svc.cluster.local:5000/
+                  - --cache=true
+                  - --cache-dir=/workspace
+                  - --skip-tls-verify-registry=jenkins-x-docker-registry.jx.svc.cluster.local:5000
 
               - name: build-and-push-maven
                 command: /kaniko/executor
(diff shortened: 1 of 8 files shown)


boot-vault.platform.yaml.template
```diff
@@ -46,3 +46,38 @@ controllerbuild:
     repository: gcr.io/jenkinsxio/builder-go
     tag: $VERSION
 {{- end }}
+
```
hferentschik: just so that I understand better, where and how is this template file used? Any chance it could be moved to somewhere else (I realize that you are just adding things in this case, but I really would love to reduce the clutter in the root directory of the project)

hferentschik: What's about these `CODECOV_TOKEN: $CODECOV_TOKEN` lines? Could they be removed? I think they might be leftovers from the codecov experiment.

jenkins-x-boot-lh-ghe.yml
```diff
@@ -61,7 +61,7 @@ pipelineConfig:
 
               - name: build-and-push-image
                 command: /kaniko/executor
-                args: ['--dockerfile=/workspace/source/Dockerfile','--destination=gcr.io/jenkinsxio/jx:${inputs.params.version}','--context=/workspace/source','--cache-repo=jenkins-x-docker-registry.jx.svc.cluster.local:5000/','--cache=true','--cache-dir=/workspace','--skip-tls-verify-registry=jenkins-x-docker-registry.jx.svc.cluster.local:5000']
+                args: ['--dockerfile=/workspace/source/Dockerfile','--destination=gcr.io/jenkinsxio/builder-jx:${inputs.params.version}','--context=/workspace/source','--cache-repo=jenkins-x-docker-registry.jx.svc.cluster.local:5000/','--cache=true','--cache-dir=/workspace','--skip-tls-verify-registry=jenkins-x-docker-registry.jx.svc.cluster.local:5000']
```
hferentschik: what does this "builder-" prefix change for this context? wonder whether it would make sense to break the arguments into one argument per line. Might make it easier to spot differences when changes are made.

boot-vault.platform.yaml.template
abayer: Yeah, I can definitely move these and nuke the codecov stuff.

jenkins-x-boot-lh-ghe.yml
abayer: Because for some reason we weren't building the image as `builder-jx` but as `jx` and then never using it. I actually should probably change the `Dockerfile` to use `builder-jx:latest` as the base too. And yeah, good call.

jenkins-x-bdd.yml
```diff
@@ -62,19 +62,58 @@ pipelineConfig:
 
               - name: build-and-push-image
                 command: /kaniko/executor
-                args: ['--dockerfile=/workspace/source/Dockerfile','--destination=gcr.io/jenkinsxio/jx:${inputs.params.version}','--context=/workspace/source','--cache-repo=jenkins-x-docker-registry.jx.svc.cluster.local:5000/','--cache=true','--cache-dir=/workspace','--skip-tls-verify-registry=jenkins-x-docker-registry.jx.svc.cluster.local:5000']
+                args:
+                  - --dockerfile=/workspace/source/Dockerfile.builder-jx
```
hferentschik: 👍, nice, I think this makes it easier to take it further changes

jx/bdd/boot-lh-ghe/ci.sh
```diff
@@ -53,7 +53,7 @@ cp ../jx/bdd/boot-lh-ghe/jx-requirements.yml .
 cp ../jx/bdd/boot-lh-ghe/parameters.yaml env
 
 cp env/jenkins-x-platform/values.tmpl.yaml tmp.yaml
-cat tmp.yaml ../boot-vault.platform.yaml > env/jenkins-x-platform/values.tmpl.yaml
+cat tmp.yaml ../jx/bdd/boot-vault.platform.yaml > env/jenkins-x-platform/values.tmpl.yaml
```
hferentschik: 👍

jx/bdd/boot-vault.platform.yaml.template
```diff
@@ -46,3 +40,38 @@ controllerbuild:
     repository: gcr.io/jenkinsxio/builder-go
     tag: $VERSION
 {{- end }}
+
+gcpreviews:
```
hferentschik: thank you for moving this

hferentschik:
yeah :-)

PR #6757, opened 2020-02-18
fix: FindKustomize

nehagup:
finds out if there is any existing kustomize resource within the cwd and sub-directories

FindKustomize returns the kustomize resources path found in cwd and sub-directories

### DIFF
--- a/pkg/kustomize/interface.go
+++ b/pkg/kustomize/interface.go
@@ -3,4 +3,6 @@ package kustomize
 // Kustomizer defines common kustomize actions used within Jenkins X
 type Kustomizer interface {
 	Version(extraArgs ...string) (string, error)
+	ContainsKustomizeConfig(dir string) bool
+	FindKustomizationYamlPaths(dir string) (resource []string)
 }
--- a/pkg/kustomize/kustomize_cli.go
+++ b/pkg/kustomize/kustomize_cli.go
@@ -1,7 +1,12 @@
 package kustomize
 
 import (
+	"os"
+	"path/filepath"
 	"regexp"
+	"strings"
+
+	"github.com/jenkins-x/jx/pkg/log"
 
 	"github.com/jenkins-x/jx/pkg/util"
 	"github.com/pkg/errors"
@@ -48,4 +53,32 @@ func extractSemanticVersion(version string) (string, error) {
 		return "", errors.Wrapf(err, "not able to extract a semantic version of kustomize version output")
 	}
 	return regex.FindString(version), nil
+}
+
+// ContainsKustomizeConfig finds out if there is any kustomize resource in the cwd or subdirectories
+func (k *KustomizeCLI) ContainsKustomizeConfig(dir string) bool {
+	if len(k.FindKustomizationYamlPaths(dir)) != 0 {
+		return true
+	}
+
+	return false
+}
+
+// FindKustomizationYamlPaths looks for the kustomization.yaml i.e. kustomize resources in present and sub-directories
+func (k *KustomizeCLI) FindKustomizationYamlPaths(dir string) []string {
+	fp, err := filepath.Abs(dir)
+	var resources []string
+	err = filepath.Walk(fp, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if strings.Contains(path, "kustomization.yaml") {
+			resources = append(resources, path)
+		}
+		return nil
+	})
+	if err != nil {
+		log.Logger().Errorf("problem finding kustomize resources %s ", err)
+	}
+	return resources
 }
--- a/pkg/kustomize/kustomize_cli_test.go
+++ b/pkg/kustomize/kustomize_cli_test.go
@@ -1,0 +1,36 @@
+// +build unit
+
+package kustomize_test
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/jenkins-x/jx/pkg/kustomize"
+)
+
+func TestKustomizeCLI_FindKustomize(t *testing.T) {
+	testDir, err := filepath.Abs(filepath.Dir("test_data/kustomize_dummy/base"))
+	assert.NoError(t, err, "failed to find test data")
+	wantedOutput := []string{
+		filepath.Join(testDir, "base", "charts", "kustomization.yaml"),
+		filepath.Join(testDir, "base", "kustomization.yaml"),
+		filepath.Join(testDir, "staging", "kustomization.yaml"),
+	}
+
+	k := kustomize.NewKustomizeCLI()
+	output := k.FindKustomizationYamlPaths(testDir)
+
+	assert.ElementsMatch(t, wantedOutput, output, "not able to find all of the kustomize resource")
+}
+
+func TestKustomizeCLI_ContainsKustomizeConfig(t *testing.T) {
+	testDir, err := filepath.Abs(filepath.Dir("test_data/kustomize_dummy/base"))
+	assert.NoError(t, err, "failed to find test data")
+
+	k := kustomize.NewKustomizeCLI()
+	assert.True(t, k.ContainsKustomizeConfig(testDir))
+	assert.False(t, k.ContainsKustomizeConfig(filepath.Join(testDir, "foo")))
+}


pkg/cmd/clients/factory.go
```diff
@@ -781,5 +781,5 @@ func (f *factory) CreateCertManagerClient() (certmngclient.Interface, error) {
 
 // CreateKustomizer creates a Kustomize client
 func (f *factory) CreateKustomizer() kustomize.Kustomizer {
-	return kustomize.NewKustomizeCLI()
+	return kustomize.NewKustomizeCLI("")
```
hferentschik: Where is the empty string as parameter coming from?

pkg/kustomize/interface.go
```diff
@@ -3,4 +3,7 @@ package kustomize
 // Kustomizer defines common kustomize actions used within Jenkins X
 type Kustomizer interface {
 	Version(extraArgs ...string) (string, error)
+	HasKustomize() bool
+	FindKustomize() (resource []string)
+	SetCWD(dir string)
```
hferentschik: What is the aim of _SetCWD_? It seems odd to see this in the Kustomizer interface.

pkg/kustomize/interface.go
```diff
@@ -3,4 +3,7 @@ package kustomize
 // Kustomizer defines common kustomize actions used within Jenkins X
 type Kustomizer interface {
 	Version(extraArgs ...string) (string, error)
+	HasKustomize() bool
```
hferentschik: _HasKustomizeConfig_ or _UsesKustomize_ or _ContainsKustomizeConfig_ might be better.

pkg/kustomize/kustomize_cli.go
```diff
@@ -10,16 +15,18 @@ import (
 // KustomizeCLI implements common kustomize actions based on kustomize CLI
 type KustomizeCLI struct {
 	Runner util.Commander
+	CWD    string
 }
 
 // NewKustomizeCLI creates a new KustomizeCLI instance configured to use the provided kustomize CLI in
 // the given current working directory
-func NewKustomizeCLI() *KustomizeCLI {
+func NewKustomizeCLI(cwd string) *KustomizeCLI {
```
hferentschik: I don't think it is a good idea to set the working directory in the constructor. If you need to change the current working directory you would probably be better off to pass it as a parameter to the function which requires it. See _git_cli_ for example. One parameter to most of its functions is the directory of the repo.

pkg/kustomize/interface.go
hferentschik: Also, you might want to pass the directory from which to search the search as a parameter. This way you don't need to pass it to the constructor.

pkg/kustomize/kustomize_cli.go
```diff
@@ -10,16 +15,18 @@ import (
 // KustomizeCLI implements common kustomize actions based on kustomize CLI
 type KustomizeCLI struct {
 	Runner util.Commander
+	CWD    string
```
hferentschik: KustomizeCLI should probably track the working directory that way

pkg/kustomize/kustomize_cli.go
```diff
@@ -49,3 +56,36 @@ func extractSemanticVersion(version string) (string, error) {
...
+func (k *KustomizeCLI) HasKustomize() bool {
+	if len(k.FindKustomize()) != 0 {
+		return true
+	}
+
+	return false
+}
+
+// FindKustomize looks for the kustomization.yaml i.e. kustomize resources in present and sub-directories
+func (k *KustomizeCLI) FindKustomize() (resource []string) {
```
hferentschik: Do you really need to export this function? Is this not just a helper for HasKustomize?

nehagup: I planned this in order to fetch paths of all customisation files so that we know which all directories to apply kustomize to.

pkg/kustomize/kustomize_cli.go
```diff
@@ -49,3 +54,31 @@ func extractSemanticVersion(version string) (string, error) {
...
+func (k *KustomizeCLI) ContainsKustomizeConfig(dir string) bool {
+	if len(k.FindKustomize(dir)) != 0 {
+		return true
+	}
+
+	return false
+}
+
+// FindKustomize looks for the kustomization.yaml i.e. kustomize resources in present and sub-directories
+func (k *KustomizeCLI) FindKustomize(dir string) (resource []string) {
```
hferentschik: _FindKustomizeYaml_ or maybe _FindKustomizationYaml_ or even _FindKustomizationYamlPaths_? WDYT? Anything a bit more descriptive.

pkg/kustomize/kustomize_cli_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"testing"
+
+	"github.com/jenkins-x/jx/pkg/cmd/testhelpers"
+	"github.com/jenkins-x/jx/pkg/kustomize"
+	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/util"
+)
+
+func TestKustomizeCLI_FindKustomize_HasKustomize(t *testing.T) {
+	origJXHome, testJXHome, err := testhelpers.CreateTestJxHomeDir()
```
hferentschik: I don't think that you need to create a tmp JX_HOME directory here. Something like this should do:

```
	tmpDir, err := ioutil.TempDir("", "TestKustomizeCLI_FindKustomize")
	require.NoError(t, err, "failed to create temporary directory")
	defer func() {
		_ = os.RemoveAll(tmpDir)
	}()
```

Technically _ testJXHome_ is also "just" a temporary directory, but it is also "more", so a reader of this test might get confused on why a tmp _JX_HOME_ directory should be required.

pkg/kustomize/kustomize_cli_test.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	defer func() {
+		err := testhelpers.CleanupTestJxHomeDir(origJXHome, testJXHome)
+		if err != nil {
+			log.Logger().Warnf("unable to remove temporary directory %s: %s", testJXHome, err)
+		}
+	}()
+
+	err = os.MkdirAll(filepath.Join(testJXHome, "bin/base/charts"), os.ModePerm)
```
hferentschik: I'd not use _bin_ as part of the test directory structure. Use _myApp_ or something. Readers of the test might otherwise get confused.

pkg/kustomize/kustomize_cli_test.go
```diff
@@ -0,0 +1,63 @@
...
+	err = os.MkdirAll(filepath.Join(testJXHome, "bin/base/templates"), os.ModePerm)
+	err = os.MkdirAll(filepath.Join(testJXHome, "bin/staging"), os.ModePerm)
+	_, err = os.Create(filepath.Join(testJXHome, "bin/base/Charts.yaml"))
+	_, err = os.Create(filepath.Join(testJXHome, "bin/base/values.yaml"))
+	_, err = os.Create(filepath.Join(testJXHome, "bin/base/kustomization.yaml"))
+	_, err = os.Create(filepath.Join(testJXHome, "bin/staging/values.yaml"))
+	_, err = os.Create(filepath.Join(testJXHome, "bin/staging/kustomization.yaml"))
+	_, err = os.Create(filepath.Join(testJXHome, "bin/base/charts/kustomization.yaml"))
+	if err != nil {
```
hferentschik: hmm, technically you only check the last os.Create call here on error. You need to check all.
Another maybe easier approach is to create this directory structure in _testdata_ and just run the test against this dir. You can have multiple directory structures/setups under _testdata_ running several tests. Easier than creating the structure on the fly.

Look in the source tree for _testdata_ directories and see how it is done there. _testdata_ directories are treated special by Go and won't be compiled or otherwise added to the final binary.

pkg/kustomize/kustomize_cli_test.go
```diff
@@ -0,0 +1,63 @@
...
+	jxBin, err := util.JXBinLocation()
+	if err != nil {
+		t.Errorf("error finding JXBin in TestKustomizeCLI_FindKustomize %s", err)
+	}
+
+	k := kustomize.NewKustomizeCLI()
+	dir := filepath.Join(testJXHome, "bin")
+	wantedOutput := []string{filepath.Join(jxBin, "base/charts/kustomization.yaml"),
+		filepath.Join(jxBin, "base/kustomization.yaml"), filepath.Join(jxBin, "staging/kustomization.yaml")}
+	output := k.FindKustomize(dir)
```
hferentschik: Would be nice to have a dedicated test for `ContainsKustomizeConfig` as well.

pkg/kustomize/interface.go
```diff
@@ -3,4 +3,6 @@ package kustomize
 // Kustomizer defines common kustomize actions used within Jenkins X
 type Kustomizer interface {
 	Version(extraArgs ...string) (string, error)
+	ContainsKustomizeConfig(dir string) bool
```
hferentschik: 👍

pkg/kustomize/kustomize_cli_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"path/filepath"
+	"reflect"
+	"testing"
+
+	"github.com/jenkins-x/jx/pkg/kustomize"
+)
+
+func TestKustomizeCLI_FindKustomize(t *testing.T) {
+	cwd, _ := os.Getwd()
+	tmpDir := filepath.Dir(filepath.Join(cwd, "..", "test_data", "kustomize_dummy", "base"))
```
hferentschik: I think you put your test_data directory into the wrong package. It should be in the same package as the this test file, then you don't have to do the cwd stuff and going up one directory (..)

pkg/kustomize/kustomize_cli_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	k := kustomize.NewKustomizeCLI()
+	wantedOutput := []string{filepath.Join(tmpDir, "base/charts/kustomization.yaml"), filepath.Join(tmpDir, "base/kustomization.yaml"), filepath.Join(tmpDir, "staging/kustomization.yaml")}
+	output := k.FindKustomizationYamlPaths(tmpDir)
+
+	if !reflect.DeepEqual(wantedOutput, output) {
```
hferentschik: Deep equal is not quite right here. You basically have to arrays and want to assert that both arrays have the same elements (not necessarily in order). Basically a set equal. There is an `assert.ElementsMatch` for that.

hferentschik:
```
	testDir := filepath.Join("test_data", "kustomize_dummy")
	absPath, err := filepath.Abs(testDir)
	require.NoError(t, err)
	wantedOutput := []string{
		filepath.Join(absPath, "base", "charts", "kustomization.yaml"),
		filepath.Join(absPath, "base", "kustomization.yaml"),
		filepath.Join(absPath, "staging", "kustomization.yaml"),
	}


	k := kustomize.NewKustomizeCLI()
	actualOutput := k.FindKustomizationYamlPaths(testDir)

	assert.ElementsMatch(t, wantedOutput, actualOutput)
```

hferentschik: Provided you move the test data into the right directory.

hferentschik:
@nehagup, the linting is failing since you have not classified your test file:

```
The following 1 test files are not classified with a valid Go build tag [unit|integration]
./hack/../pkg/kustomize/kustomize_cli_test.go
```

pkg/kustomize/kustomize_cli_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	assert.ElementsMatch(t, wantedOutput, output, "not able to find all of the kustomize resource")
+}
+
+func TestKustomizeCLI_ContainsKustomizeConfig(t *testing.T) {
+	testDir, err := filepath.Abs(filepath.Dir("test_data/kustomize_dummy/base"))
+	assert.NoError(t, err, "failed to find test data")
+
+	k := kustomize.NewKustomizeCLI()
+	if !k.ContainsKustomizeConfig(testDir) {
```
hferentschik: I think the following is more idiomatic and easier to read:

```
	assert.True(t, k.ContainsKustomizeConfig(testDir))
	assert.False(t, k.ContainsKustomizeConfig(filepath.Join(testDir, "foo")))
```

pkg/kustomize/kustomize_cli.go
```diff
@@ -49,3 +54,31 @@ func extractSemanticVersion(version string) (string, error) {
...
+	err = filepath.Walk(fp, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if strings.Contains(path, "kustomization.yaml") {
+			resources = append(resources, path)
+		}
+		return nil
+	})
+	if err != nil {
```
hferentschik: Not sure whether logging is the best approach here. Usually, one returns the error, but let's see

pkg/kustomize/kustomize_cli.go
```diff
@@ -49,3 +54,31 @@ func extractSemanticVersion(version string) (string, error) {
...
+func (k *KustomizeCLI) ContainsKustomizeConfig(dir string) bool {
+	if len(k.FindKustomizationYamlPaths(dir)) != 0 {
+		return true
+	}
+
+	return false
+}
+
+// FindKustomizationYamlPaths looks for the kustomization.yaml i.e. kustomize resources in present and sub-directories
+func (k *KustomizeCLI) FindKustomizationYamlPaths(dir string) (resource []string) {
```
hferentschik: I don't think you need to name the return value here.

PR #6771, opened 2020-02-19
fix: Check if the requirements config was read from the existing file

hferentschik:
based on PR #6658

### DIFF
--- a/pkg/config/install_requirements_test.go
+++ b/pkg/config/install_requirements_test.go
@@ -120,15 +120,9 @@ func TestRequirementsConfigMarshalInEmptyDir(t *testing.T) {
 	dir, err := ioutil.TempDir("", "test-requirements-config-empty-")
 
 	requirements, fileName, err := config.LoadRequirementsConfig(dir)
-	assert.NoError(t, err, "failed to load requirements file in dir %s", dir)
-
-	exists, err := util.FileExists(fileName)
-	assert.NoError(t, err, "failed to check file exists %s", fileName)
-	assert.False(t, exists, "file should not exist %s", fileName)
-
-	assert.Equal(t, config.WebhookTypeProw, requirements.Webhook, "requirements.WebhookTypeProw")
-	assert.Equal(t, false, requirements.Kaniko, "requirements.Kaniko")
-	assert.Equal(t, config.SecretStorageTypeLocal, requirements.SecretStorage, "requirements.SecretStorage")
+	assert.Error(t, err)
+	assert.Empty(t, fileName)
+	assert.Nil(t, requirements)
 }
 
 func TestRequirementsConfigIngressAutoDNS(t *testing.T) {
--- a/pkg/config/install_requirements_test.go
+++ b/pkg/config/install_requirements_test.go
@@ -3,11 +3,14 @@
 package config_test
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 
 	"github.com/jenkins-x/jx/pkg/cloud/gke"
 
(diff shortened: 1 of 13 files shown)


hferentschik:
@RafalKorepta, down to one BDD test failure. How this change relates to the Spring demo app not coming up, I'm currently not sure yet.

@abayer, any idea what the problem could be. The failure is in this case not really so helpful.

abayer:
I’d guess that the staging env pipeline is failing, either in PR or release. Only way to figure that out is probably to jump on the failed cluster before it gets deleted and check the logs.

pkg/config/install_requirements_test.go
```diff
@@ -356,3 +357,53 @@ func TestHelmfileRequirementValues(t *testing.T) {
...
+		{"a/b/c", false, false},
+		{"e", true, true},
+		{"e/f", true, true},
+		{"e/f/g", true, false},
+	}
+
+	for _, testCase := range gitPublicTests {
+		t.Run(testCase.requirementsPath, func(t *testing.T) {
+			dir, err := ioutil.TempDir("", "jx-test-load-requirements-config")
+			assert.NoError(t, err, "failed to create tmp directory")
```
RafalKorepta: NIT: `require.NoError` can stop the unit test. In my experience set up should fail fast and assertion should be used after calling the function under test.

pkg/config/install_requirements_test.go
```diff
@@ -356,3 +357,53 @@ func TestHelmfileRequirementValues(t *testing.T) {
...
+			defer func() {
```
RafalKorepta: NIT: I sometimes use the following snippet:
```
defer assert.NoError(t, os.RemoveAll(dir))
```

pkg/config/install_requirements_test.go
```diff
@@ -356,3 +357,53 @@ func TestHelmfileRequirementValues(t *testing.T) {
...
+			testPath := filepath.Join(dir, testCase.requirementsPath)
+			err = os.MkdirAll(testPath, 0700)
+			assert.NoError(t, err, "unable to create test path %s", testPath)
+
+			var expectedRequirementsFile string
+			if testCase.createRequirements {
+				expectedRequirementsFile = filepath.Join(testPath, config.RequirementsConfigFileName)
+				dummyRequirementsData := []byte("webhook: prow\n")
+				err := ioutil.WriteFile(expectedRequirementsFile, dummyRequirementsData, 0644)
+				assert.NoError(t, err, "unable to write requirements file %s", expectedRequirementsFile)
```
RafalKorepta: NIT: `require.NoError`

pkg/config/install_requirements_test.go
```diff
@@ -356,3 +357,53 @@ func TestHelmfileRequirementValues(t *testing.T) {
...
+		t.Run(testCase.requirementsPath, func(t *testing.T) {
+			dir, err := ioutil.TempDir("", "jx-test-load-requirements-config")
+			assert.NoError(t, err, "failed to create tmp directory")
+			defer func() {
+				_ = os.RemoveAll(dir)
+			}()
+
+			testPath := filepath.Join(dir, testCase.requirementsPath)
+			err = os.MkdirAll(testPath, 0700)
+			assert.NoError(t, err, "unable to create test path %s", testPath)
```
RafalKorepta: NIT: `require.NoError`

pkg/config/install_requirements_test.go
```diff
@@ -356,3 +357,53 @@ func TestHelmfileRequirementValues(t *testing.T) {
...
 	assert.FileExists(t, valuesFile, "file %s should exist", valuesFile)
 }
+
+func Test_LoadRequirementsConfig(t *testing.T) {
+	t.Parallel()
+
+	var gitPublicTests = []struct {
+		requirementsPath   string
+		createRequirements bool
+		loadSuccessful     bool
```
RafalKorepta: This is an unused filed. Or am I missing something?

hferentschik: Nope, not at all. I had initially something else in mind, but in the end, never needed this flag. Going to remove it.

pkg/config/install_requirements_test.go
hferentschik: Interesting, what's the difference? Is there a semantic difference? I think _require_ crossed my path at some stage, but I usually use assert. Well, personally I prefer to write a Gingko test if start a new test ;-)

pkg/config/install_requirements_test.go
hferentschik: ahhh, sorry missed this. that's the explanation first time around

hferentschik: Good one. I agree with your reasoning. I prefer to fail fast as well. I'll remember this for the future.

hferentschik:
@RafalKorepta, thanks for the review. Much appreciated. I cleaned this up a bit. Let's hope it passes  all tests.

PR #6785, opened 2020-02-23
test: Add test to verify that pipelineConfig env secrets get used

abayer:
#### Submitter checklist

#### Description

I had a problem using an env var defined in the `pipelineConfig` using a secret, and wanted to verify that this actually does work as expected. It does, so I was doing something else wrong, but adding a test doesn't seem like a bad idea. =)

n/a


hferentschik:
lol 7 files for a single test ;-)

/lgtm

PR #6788, opened 2020-02-24
fix: FileExists returns now false for directories

### LINKED ISSUE #6758: Wrong implementation of util.FileExists
The current implementation of util.FileExists is not behaving as expected/documented. The current implementation looks like:

```
// FileExists checks if path exists and is a file
func FileExists(path string) (bool, error) {
	_, err := os.Stat(path)
	if err == nil {
		return true, nil
	}
	if os.IsNotExist(err) {
		return false, nil
	}
	return true, errors.Wrapf(err, "failed to check if file exists %s", path)
}
```

The check should verify whether the specified path is a file (`fileInfo.IsDir()`), but it does not. 
Also, the last return seems suspicious. I don't think that a return value of _true_ should be ever accompanied by an error.

Changing and testing the implementation should be easy, but given its heavy use we need to make sure that changing it won't introduce a regression.

hferentschik:
- updating ccode and tests to the new behavior

fixes #6758

### DIFF
--- a/pkg/cmd/boot/boot.go
+++ b/pkg/cmd/boot/boot.go
@@ -200,7 +200,7 @@ func (o *BootOptions) Run() error {
 			}
 		}
 
-		bootCloneExists, err := util.FileExists(cloneDir)
+		bootCloneExists, err := util.DirExists(cloneDir)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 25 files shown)


hferentschik:
Thoughts on this? There is a chance that I missed to convert a `FileExists` to `DirExists` somewhere, but at least the underlying implementation is correct now.

PR #6789, opened 2020-02-24
fix: Use kube crd name to look up the docker registry image

### LINKED ISSUE #5322: Docker registry repository name error
### Summary 
We've imported a new Github repo with name design.mycompany.com and a docker registry repo called 
xxxx.dkr.ecr.eu-west-1.amazonaws.com/mycompany/design-mycompany-com is created. We use ECR docker registry AWS service. 

The logs of the jx import command are:
```
...
Let's ensure that we have an ECR repository for the Docker image mycompany/design-mycompany-com
Created ECR repository: xxx.dkr.ecr.eu-west-1.amazonaws.com/mycompany/design-mycompany-com
...
```
But when jx is deploying the image in staging, the pod is failing to be created with the following error:
```
Events:
  Type     Reason     Age                From                                                 Message
  ----     ------     ----               ----                                                 -------
  Normal   Scheduled  61s                default-scheduler                                    Successfully assigned jx-staging/jenkins-x-design-mycompany-com-5d565b94c6-h42k9 to ip-10-101-45-65.eu-west-1.compute.internal
  Normal   Pulling    18s (x3 over 60s)  kubelet, ip-10-101-45-65.eu-west-1.compute.internal  pulling image "xxx.dkr.ecr.eu-west-1.amazonaws.com/mycompany/design.mycompany.com:0.0.4"
  Warning  Failed     18s (x3 over 60s)  kubelet, ip-10-101-45-65.eu-west-1.compute.internal  Failed to pull image "xxx.dkr.ecr.eu-west-1.amazonaws.com/mycompany/design.mycompany.com:0.0.4": rpc error: code = Unknown desc = Error response from daemon: repository xxx.dkr.ecr.eu-west-1.amazonaws.com/mycompany/design.mycompany.com not found
  Warning  Failed     18s (x3 over 60s)  kubelet, ip-10-101-45-65.eu-west-1.compute.internal  Error: ErrImagePull
  Normal   BackOff    6s (x4 over 59s)   kubelet, ip-10-101-45-65.eu-west-1.compute.internal  Back-off pulling image "xxx.dkr.ecr.eu-west-1.amazonaws.com/mycompany/design.mycompany.com:0.0.4"
  Warning  Failed     6s (x4 over 59s)   kubelet, ip-10-101-45-65.eu-west-1.compute.internal  Error: ImagePullBackOff

```
It tries to get the image from mycompany/design.mycompany.com (wrong) instead of mycompany/design-mycompany-com (correct) docker repo.

Import repo and wait until after the first build is deployed in staging env.

### Expected behavior
Pull an image with the correct name of the docker registry repo created by jx import command.

### Workaround
I did a workaround replacing dots by dashes in chart deployment yaml:
```
image: "{{ .Values.image.repository | replace "design.mycompany.com" "design-mycompany-com" }}:{{ .Values.image.tag }}"
```
This works, but IMHO this should be fixed.

### Jx version

```
NAME               VERSION
jx                 2.0.666
jenkins x platform 2.0.1053
Kubernetes cluster v1.12.10-eks-825e5d
kubectl            v1.14.2
git                2.17.1
Operating System   Ubuntu 18.04.3 LTS

```

### Jenkins type

- [x] Serverless Jenkins X Pipelines (Tekton + Prow) and multiple cluster installation
- [ ] Classic Jenkins

### Kubernetes cluster
EKS cluster created with Cloudformation

Thanks for your support!

mrageh:
#### Submitter checklist

#### Description
Prior to this commit, we'd use the kubernetes compliant crd name to create the docker registry destination. But then later on in our code, we'd use the Git repository name to look up the docker registry image and it would fail in some cases. 

For example if the Git repo name contained `.`, the kubernetes compliant crd would replace all the dots with hyphens. So if we a customer had a repo called `example.company.repo` would be converted to `example-company-repo`. but then we'd use the wrong name to do the look up.

fixes #5322

### DIFF
--- a/pkg/cmd/step/create/step_create_task_test.go
+++ b/pkg/cmd/step/create/step_create_task_test.go
@@ -86,7 +86,7 @@ func TestGenerateTektonCRDs(t *testing.T) {
 		{
 			name:         "js_build_pack",
 			language:     "javascript",
-			repoName:     "js-test-repo",
+			repoName:     "js.test.repo",
 			organization: "abayer",
 			branch:       "build-pack",
 			kind:         "release",
--- a/pkg/tekton/syntax/pipeline.go
+++ b/pkg/tekton/syntax/pipeline.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	v1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	"github.com/jenkins-x/jx/pkg/kube/naming"
 	"github.com/jenkins-x/jx/pkg/log"
 	"github.com/jenkins-x/jx/pkg/util"
 	"github.com/jenkins-x/jx/pkg/versionstream"
@@ -1211,7 +1212,7 @@ func (s *Step) modifyStep(params StepPlaceholderReplacementArgs) {
 			sourceDir := params.WorkspaceDir
 			dockerfile := filepath.Join(sourceDir, "Dockerfile")
 			localRepo := params.DockerRegistry
-			destination := params.DockerRegistry + "/" + params.DockerRegistryOrg + "/" + params.GitName
+			destination := params.DockerRegistry + "/" + params.DockerRegistryOrg + "/" + naming.ToValidName(params.GitName)
 
 			args := []string{"--cache=true", "--cache-dir=/workspace",
 				"--context=" + sourceDir,
--- a/pkg/cmd/step/create/test_data/step_create_task/js_build_pack/pipeline.yml
+++ b/pkg/cmd/step/create/test_data/step_create_task/js_build_pack/pipeline.yml
@@ -7,7 +7,7 @@ metadata:
     build: "1"
     jenkins.io/pipelineType: build
     owner: abayer
-    repository: js-test-repo
+    repository: js.test.repo
   name: abayer-js-test-repo-build-pack-9l9zj-1
   namespace: jx
 spec:
--- a/pkg/cmd/step/create/test_data/step_create_task/js_build_pack/pipelineresources.yml
+++ b/pkg/cmd/step/create/test_data/step_create_task/js_build_pack/pipelineresources.yml
@@ -9,7 +9,7 @@ items:
     - name: revision
       value: v0.0.1
     - name: url
-      value: https://github.com/abayer/js-test-repo
+      value: https://github.com/abayer/js.test.repo
     type: git
   status: {}
 metadata: {}
--- a/pkg/cmd/step/create/test_data/step_create_task/js_build_pack/pipelinerun.yml
+++ b/pkg/cmd/step/create/test_data/step_create_task/js_build_pack/pipelinerun.yml
@@ -7,7 +7,7 @@ metadata:
     build: "1"
     jenkins.io/pipelineType: build
     owner: abayer
-    repository: js-test-repo
+    repository: js.test.repo
   name: abayer-js-test-repo-build-pack-9l9zj-1
 spec:
   params:
--- a/pkg/cmd/step/create/test_data/step_create_task/js_build_pack/structure.yml
+++ b/pkg/cmd/step/create/test_data/step_create_task/js_build_pack/structure.yml
@@ -5,7 +5,7 @@ metadata:
     build: "1"
     jenkins.io/pipelineType: build
     owner: abayer
-    repository: js-test-repo
+    repository: js.test.repo
   name: abayer-js-test-repo-build-pack-9l9zj-1
 pipelineRef: null
 pipelineRunRef: null
--- a/pkg/cmd/step/create/test_data/step_create_task/js_build_pack/tasks.yml
+++ b/pkg/cmd/step/create/test_data/step_create_task/js_build_pack/tasks.yml
@@ -9,7 +9,7 @@ items:
       jenkins.io/pipelineType: build
       jenkins.io/task-stage-name: from-build-pack
       owner: abayer
-      repository: js-test-repo
+      repository: js.test.repo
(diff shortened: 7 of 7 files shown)


pkg/tekton/syntax/pipeline.go
```diff
@@ -1211,7 +1212,7 @@ func (s *Step) modifyStep(params StepPlaceholderReplacementArgs) {
 			sourceDir := params.WorkspaceDir
 			dockerfile := filepath.Join(sourceDir, "Dockerfile")
 			localRepo := params.DockerRegistry
-			destination := params.DockerRegistry + "/" + params.DockerRegistryOrg + "/" + params.GitName
+			destination := params.DockerRegistry + "/" + params.DockerRegistryOrg + "/" + naming.ToValidName(params.GitName)
```
hferentschik: Is there a way to add a test for this in some form? Either as unit test or even as part of the bdd tests for eks (no idea where this stands atm).

@abayer is the expert on testing this syntax stuff and @dgozalo might now how we stand on bdd tests.

hferentschik:
You want to make sure to run `make fmt` to make sure that the code is properly formatted. Seems there is a linting issue.

hferentschik:
/test bdd
/test boot-vault

not quite sure what to make of the bdd test failures atm. Let's kick them off once more and see what happens.

hferentschik:
@mrageh linting issue fixed :-) The change looks reasonable. Some sort of test would be nice, but not sure how feasible a useful test is.

@abayer wdyt? Merge or do you have an idea for a test? Update/add something to _pipeline_test.go_?

hferentschik:
> Prior to this commit, we'd use the kubernetes compliant crd name to create the docker registry destination

@mrageh, do you have a pointer to the code where we create the docker registry?

mrageh:
@hferentschik unfortunately I don't know where the creation is happens, but you can see from the issue linked that the image name for the docker registry was Kuberenetes crd compliant.

pkg/tekton/syntax/pipeline.go
abayer: Probably the best way to do this would be in https://github.com/jenkins-x/jx/blob/master/pkg/cmd/step/create/step_create_task_test.go - specifically to change https://github.com/jenkins-x/jx/blob/master/pkg/cmd/step/create/step_create_task_test.go#L86-L93 to have a `.` in its `repoName`, like `js.test-repo`. That'll fail initially due to the content in https://github.com/jenkins-x/jx/tree/master/pkg/cmd/step/create/test_data/step_create_task/js_build_pack not matching, but you can then update the various files in there to use `js.test-repo` rather than `js-test-repo` for the repo (just search for `js-test-repo` in that directory and change everything that isn't the `--destination=...` flag for a kaniko step in `tasks.yml`).

Yes, this ideally would be tested from `pipeline_test.go`, but this is messier than it should be.

abayer: And fwiw, I do think it's fine to just modify the existing test case - there are so many test cases that end up overlapping in the same areas that tweaking an existing test here isn't going to decrease coverage. The "no `.`s in the repo name, replacing skaffold in the build pack with kaniko in the Tekton CRDs" case is already covered in basically the exact same way in the `maven_build_pack` test case, for example.

pkg/cmd/step/create/step_create_task_test.go
```diff
@@ -86,7 +86,7 @@ func TestGenerateTektonCRDs(t *testing.T) {
 		{
 			name:         "js_build_pack",
 			language:     "javascript",
-			repoName:     "js-test-repo",
+			repoName:     "js.test.repo",
```
hferentschik: @mrageh so what are we really proving here? AFAICT all occurrences of '-' are replaced with '.' for this test case. However, I cannot see where ' js.test.repo' would become 'js-test-repo' which would be what we want to prove, right?

Might be that this particular test you are modifying is not using Kaniko? If we cannot test it, I'd rather not change anything in the test files. Better to know there is no testing opposed to looking at the commit in a few months time and thinking there was a test added and wondering how it works. 

@abayer - wdyt?

The other thing here is that at best a change to _step_create_task_test.go_ will test the mechanics of _modifyStep_ and that the replacement occurs. It will not assert that the actual reported issue is resolved.

abayer: @hferentschik So the thing that's relevant here is what _hasn't_ changed: https://github.com/jenkins-x/jx/pull/6789/files#diff-6f8d554e2e7907cb7ea6450cf8c34aa0R288 =)

mrageh: @hferentschik check this file out https://github.com/jenkins-x/jx/blob/456b4f42baa6dae0dc539d54a500a7878d54fd01/pkg/cmd/step/create/test_data/step_create_task/js_build_pack/tasks.yml#L288, as you can see the docker registry name was successfully converted to `js-test-repo`. Which is what this PR fixes, the test is very obscure I know. ¯\_(ツ)_/¯

hferentschik: Thanks, that's what I was looking for. Since it had not changed I did not see it in the diff. 

> Which is what this PR fixes, the test is very obscure I know

Got it now. Very subtle indeed.

mrageh:
@hferentschik in general I'm not happy with the way we handle making the git repo name kubernetes compliant, because this logic is scattered throughout JX and if we miss a place it can cause bugs such as this to occur. 

I did talk to @abayer about having one central structure that takes a git repo name and converts to a crd valid name. And then reusing that thing throughout the codebase. But that would involve too much work.

Issue #6793, opened 2020-02-24
Lighthouse status reporting should find JX UI URL automatically


abayer:
Relates to #6791, but this is for the "real" solution. We need to be able to automatically determine whether the UI is installed, and if so, what its ingress's URL is. In practice, we can probably do something like looking through ingresses for one with either `jxui` or `sso-jenkins-x-ui` as the underlying service, and get the host and whether it's TLS from there, but I'm not 100% sure if that's right.

I would prefer not to rely on the plank config in the prow config map, since that still requires manual configuration - this really needs to get picked up automatically, and is a blocker for Lighthouse to go GA.

cc @dgozalo based on our earlier discussion.

PR #6795, opened 2020-02-24
fix: LH reporting: fix target URL, specify UI URL, better running description

### LINKED ISSUE #6791: Lighthouse commit status reporting via build controller doesn't do anything
We need to be specifying `TargetURL` to the GitHub API, not `URL`. That's an easy fix, but we also don't have a way to automatically determine whether the UI is installed and what its URL is, which I'll tackle in a separate issue.

### LINKED ISSUE #6792: Lighthouse commit status reporting shouldn't require specifying a full URL template
We currently allow you to specify a target URL template (helm-style) either via the command line or an environment variable, but in 99% of cases, this is kinda superfluous - we know the template will basically be `{{ .BaseURL }}/teams/{{ .Namespace }}/projects/{{ .Owner }}/{{ .Repository }}/{{ .Branch }}/{{ .Build }}` - i.e., the JX UI URL. So let's add the ability to just specify the one bit of that we don't already have (the base URL of the UI including http or https) as an argument to the build controller. There's a wider thing about detecting that URL automatically that will need to be resolved before we can move everyone to Lighthouse, but this will do for the moment.

### LINKED ISSUE #6794: Lighthouse status reporting should have a better description
At the minimum, it should say what stage(s) are currently running rather than just "pending". Not sure yet if something more detailed makes sense for success and failure, but let's start here.

abayer:
#### Submitter checklist

#### Description

This isn't the end state - #6793 needs to be dealt with. But this will allow for making changes in `env/jenkins-x-platform/values.tmpl.yaml` to specify a `--job-url-base=https://whatever.whatever` and then that'll be used with a default template for creating the logs URL. Also, we'll actually use the logs URL now. And the "pending" message will now tell you what stage(s) are currently running because I wanted more information there. =)

fixes #6791
fixes #6792
fixes #6794

### DIFF
--- a/pkg/cmd/controller/controller_build.go
+++ b/pkg/cmd/controller/controller_build.go
@@ -42,6 +42,10 @@ import (
 	"github.com/jenkins-x/jx/pkg/kube"
 )
 
+const (
+	defaultTargetURLTemplate = "{{ .BaseURL }}/teams/{{ .Namespace }}/projects/{{ .Owner }}/{{ .Repository }}/{{ .Branch }}/{{ .Build }}"
+)
+
 // ControllerBuildOptions are the flags for the commands
 type ControllerBuildOptions struct {
 	ControllerOptions
@@ -51,6 +55,7 @@ type ControllerBuildOptions struct {
 	GitReporting        bool
 	TargetURLTemplate   string
 	FailIfNoGitProvider bool
+	JobURLBase          string
 
 	EnvironmentCache *kube.EnvironmentNamespaceCache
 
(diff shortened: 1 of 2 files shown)


pkg/cmd/controller/controller_build.go
```diff
@@ -128,8 +133,9 @@ func NewCmdControllerBuild(commonOpts *opts.CommonOptions) *cobra.Command {
 	cmd.Flags().BoolVarP(&options.FailIfNoGitProvider, "fail-on-git-provider-error", "", false, "If enable then lets terminate quickly if we cannot create a git provider")
 
 	// optional git reporting flags
-	cmd.Flags().StringVarP(&options.TargetURLTemplate, "target-url-template", "", "", "The Go template for generating the target URL of pipeline logs/views if git reporting is enabled")
+	cmd.Flags().StringVarP(&options.TargetURLTemplate, "target-url-template", "", "", "The Go template for generating the target URL of pipeline logs/views if git reporting is enabled. If unspecified, a default will be used based on `--job-url-base`.")
 	cmd.Flags().BoolVarP(&options.GitReporting, "git-reporting", "", false, "If enabled then lets report pipeline success/failures to the git provider. Note this is purely tactical until we can do this natively inside tekton")
+	cmd.Flags().StringVarP(&options.JobURLBase, "job-url-base", "", "", "The base URL, such as 'https://dashboard.jenkins-x.live', for generating the target URL for pipeline logs if git reporting is enabled.")
```
hferentschik: So is the plan to change how the build controller is configured as well? Will there a related pull request for the build controller Helm chart?

pkg/cmd/controller/controller_build.go
```diff
@@ -1268,18 +1277,34 @@ func (o *ControllerBuildOptions) reportStatus(kubeClient kubernetes.Interface, n
 	if pipelineContext == "" {
 		pipelineContext = "jenkins-x"
 	}
-	description := status
+
+	var runningStages []string
```
hferentschik: Personally, I'd put that into a function _createStatusDescription_, but up to you.
This change is really independent of Lighthouse, right? This new description would be available with Prow as well, right?

pkg/cmd/controller/controller_build.go
abayer: This part is a short term fix specifically for being able to run Lighthouse on our own prod cluster to find any further problems, so we’re just going to configure this in the prod cluster. #6793 will need to be resolved before we go GA with Lighthouse, so that one would only need to configure either of these two flags if they’re not using the JX UI for some reason.

pkg/cmd/controller/controller_build.go
abayer: It could be used with Prow, in theory, but it would just result in duplicate reporting, since a Prow already does its own reporting controlled through Prow. That said, I’m gonna do a follow up like you describe.

PR #6797, opened 2020-02-25
fix: jx get applications when helm release == app name

jstrachan:
lets patch the code so that we can still detect the actual app name if we use the app name as the helm release name (and don't use a namespace prefix on the release).

this helps the code work with helm 2 and helm 3


abayer:
The `tekton` failure seems to be legit - this breaks `jx upgrade ingress`.

abayer:
oops, I looked at the wrong logs. Ignore me!

PR #6801, opened 2020-02-25
docs: provide more precise link for feedback when issue state changes

MarckK:
Signed-off-by: Kara de la Marck <MarckK@users.noreply.github.com>

#### Submitter checklist

#### Description
Giving users a more direct link to provide feedback.

fixes #


abayer:
For what it’s worth, I don’t think this is the change you meant - the only change here is to unit test data. I’m legitimately not sure where the actual config for this is.

MarckK:
Thanks, @abayer :) There is [this](https://github.com/jenkins-x/prow-config-tekton/blob/411f79cdfa8138290744b740acf220bdb2a3c603/prow/config.yaml), but it is archived.

Issue #6812, opened 2020-02-25
non-boot clusters nil pointer on `jx step helm apply`


abayer:
Note that this isn't in the version stream - it's caused by #6771, and is in fact causing the version stream update to fail.

The staging (and production) env pipelines in a `jx install` cluster with Tekton (not with static masters - we caught that problem before #6771 was merged, but due to #6811 we didn't catch the Tekton case) will fail with an error like:

```
Showing logs for build jenkins-x-versions-bot-test-env-hj4rl-1 stage from-build-pack and container step-build-helm-apply
Modified file /workspace/source/env/Chart.yaml to set the chart to version 1
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x248 pc=0x26dafd3]

goroutine 1 [running]:
github.com/jenkins-x/**/pkg/cmd/step/helm.(*StepHelmApplyOptions).Run(0xc00083b290, 0x0, 0x0)
	/workspace/source/pkg/cmd/step/helm/step_helm_apply.go:255 +0xde3
github.com/jenkins-x/**/pkg/cmd/step/helm.NewCmdStepHelmApply.func1(0xc0009d4c80, 0xa733d88, 0x0, 0x0)
	/workspace/source/pkg/cmd/step/helm/step_helm_apply.go:85 +0x6f
github.com/spf13/cobra.(*Command).execute(0xc0009d4c80, 0xa733d88, 0x0, 0x0, 0xc0009d4c80, 0xa733d88)
	/workspace/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:830 +0x2ae
github.com/spf13/cobra.(*Command).ExecuteC(0xc000493680, 0xc0005bfc00, 0x5ace800, 0xc0000b8000)
	/workspace/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914 +0x2fc
github.com/spf13/cobra.(*Command).Execute(...)
	/workspace/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
github.com/jenkins-x/**/cmd/**/app.Run(0x0, 0x0, 0x0, 0xc0008a3f88, 0xc00009e058)
	/workspace/source/cmd/**/app/**.go:19 +0x185
main.main()
	/workspace/source/cmd/**/**.go:11 +0x32
```

[...]

Issue #6811, opened 2020-02-25
jx repo BDD tests don't use PR image in staging/production env pipelines


abayer:
The staging and production env pipelines use the `charts` build pack, which [use gcr.io for the image to run on](https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes/blob/11589715ac7002f8af9b60718456757f1b687e4f/packs/charts/pipeline.yaml#L3). This means that the magic for substituting the newly built images in the pod templates for `go`, `maven`, etc doesn't work, and the staging/prod env pipelines will still use the latest image from the version stream in the `tekton` and `boot-vault` BDD tests. This can result in a failure (i.e., in this case, #6771 failing in non-boot clusters in the staging/prod env pipelines) not getting caught until the the version stream PR.

I don't have an easy answer for this - well, ok, I have an easy answer (switch the `charts` build pack to use the `go` pod template name instead of `gcr.io/...`) but that's a hack.

abayer:
A dirty way to handle this is https://github.com/jenkins-x/jx/pull/6813/commits/b974175462d60f03d506603b746a28f6b6391620 - this works, resulting in the PR failing when I would have expected it to, but yeah, it's dirrrrrty and not what I'd want to do long term.

PR #6813, opened 2020-02-25
IGNORE: trying to get tekton context to fail

abayer:
Specifically to demonstrate #6812, which has been masked by
reproduce #6812 enough to fix it and then go deal with #6811 in the future.

### DIFF
--- a/pkg/cmd/step/helm/step_helm_apply.go
+++ b/pkg/cmd/step/helm/step_helm_apply.go
@@ -406,18 +406,17 @@ func (o *StepHelmApplyOptions) getRequirements() (*config.RequirementsConfig, st
 		return nil, "", errors.Wrap(err, "getting the team setting from the cluster")
 	}
 
-	// TODO To stay backwards compatible with code after fixing #6653, we need to create a dummy RequirementsConfig here
-	// TODO This is a special case for static master and needs to be removed (HF)
-	if !teamSettings.IsJenkinsXPipelines() {
+	requirements, err = config.GetRequirementsConfigFromTeamSettings(teamSettings)
+	if err != nil {
+		return nil, "", errors.Wrap(err, "getting the requirements from team settings")
+	}
+	// TODO: Workaround for non-boot clusters. Remove when we get rid of jx install. (APB)
+	if requirements == nil {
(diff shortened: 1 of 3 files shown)


pkg/cmd/step/helm/step_helm_apply.go
```diff
@@ -406,18 +406,17 @@ func (o *StepHelmApplyOptions) getRequirements() (*config.RequirementsConfig, st
...
 	}
 
-	// TODO To stay backwards compatible with code after fixing #6653, we need to create a dummy RequirementsConfig here
-	// TODO This is a special case for static master and needs to be removed (HF)
-	if !teamSettings.IsJenkinsXPipelines() {
+	requirements, err = config.GetRequirementsConfigFromTeamSettings(teamSettings)
+	if err != nil {
+		return nil, "", errors.Wrap(err, "getting the requirements from team settings")
+	}
+	// TODO: Workaround for non-boot clusters. Remove when we get rid of jx install. (APB)
```
hferentschik: +1, that's what I had in mind as well. Is this not working?

pkg/jenkinsfile/pipeline.go
```diff
@@ -933,6 +933,10 @@ func (c *PipelineConfig) CreatePipelineForBuildPack(args CreatePipelineArguments
 	// If agent.container is specified, use that for default container configuration for step images.
 	containerName := c.Agent.GetImage()
 	if containerName != "" {
+		gcrRemoved := strings.TrimPrefix(containerName, "gcr.io/jenkinsxio/builder-")
```
hferentschik: Is this related?

abayer: See #6811 - the first commit in this PR is me finding a hackish way to actually test the PR image in the `charts` build pack. Just a hack, not to be merged - we need a real solution for #6811 down the road.

pkg/cmd/step/helm/step_helm_apply.go
abayer: It is. =)

abayer: Rather, it _is_ working.

PR #6817, opened 2020-02-26
fix: Don't pass target URL to GH status if not an http/https URL

abayer:
#### Submitter checklist

#### Description

Otherwise you get an error because the target URL has to be an http/https URL. Sigh. So if the resulting target URL from the template doesn't start with `http://` or `https://`, don't pass a target URL at all.

And yes, this all really needs to be redone. Still in the land of tactical fixes here. I just wanna unblock the version stream right now. =)

n/a


abayer:
Note the `boot-lh-ghe` context here - it’s an optional one at the moment, but it did get past the problem this is fixing, so woooo.

PR #6823, opened 2020-02-26
fix: add a `jx alpha boot` command

jstrachan:
so folks can try out the new helm 3 / helmfile work from the `jx` command line via a separate binary

see the enhancement process: https://github.com/jenkins-x/enhancements/pull/15

also https://github.com/jenkins-x/enhancements/tree/master/proposals/2

### DIFF
--- a/pkg/cmd/alpha/alpha.go
+++ b/pkg/cmd/alpha/alpha.go
@@ -1,0 +1,141 @@
+package alpha
+
+import (
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	jenkinsv1 "github.com/jenkins-x/jx/pkg/apis/jenkins.io/v1"
+	"github.com/jenkins-x/jx/pkg/cmd/helper"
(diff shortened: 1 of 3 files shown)


jstrachan:
@hferentschik squashed

pkg/cmd/alpha/alpha.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/jenkins-x/jx/pkg/cmd/templates"
+	"github.com/jenkins-x/jx/pkg/extensions"
+	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/util"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+// Options are the options to execute the alpha command
+type Options struct {
```
hferentschik: Personally I think we should get away from this options pattern, especially here where there is no need for it at all. 
I'd rather have a `PluginManager` of sorts with a `GetPluginSpecs(url string)` and `RunPlugin`.

PR #6803, opened 2020-02-25
fix: do not run helm init when using helm version > v2

yelhouti:
#### Submitter checklist

- [X] Change is code complete and matches issue description.
- [X] Change is covered by existing or new tests.

#### Which issue this PR fixes

fixes #6818

PR #6826, opened 2020-02-26
fix: do not run helm init when using helm version > v2

### LINKED ISSUE #6818: jx boot fails when using helm3 due to jx trying to run helm init
### Summary 
jx boot fails when using helm3 due to jx trying to run helm init. To fix this we just need to check the helm version and if it's < v3 not run helm init.
this is the main reason helm3 is not working (there are other small quirks and than can be worked around without touching the code).

workaround: edit jenkins-x.yml and add `--disable-verify-helm` in step verify-preinstall
```
export JX_HELM3=true
jx boot
# fill/fix jx-requirements.yml the rerun boot
cd ./jenkins-x-boot-config/
jx boot
```

### Expected behavior
It should try to run helm init as per: https://helm.sh/blog/helm-v3-beta/#notable-changes-since-helm-v2 :
> The helm init command has been removed. It performed two primary functions. First, it installed Tiller. This is no longer needed. Second, it setup directories and repositories where Helm configuration lived. This is now automated. If the directory is not present it will be created.
### Actual behavior
```
error: initializing helm with config: {jx true false true false  }: failed to run 'helm init --client-only' command in directory '', output: 'Error: unknown flag: --client-only'
```
### Jx version

```
jx                 2.0.1202
Kubernetes cluster v1.15.2
kubectl            v1.17.0
git                2.21.1
Operating System   "Fedora release 30 (Thirty)"

```

### Jenkins type

- [ ] Classic Jenkins

hferentschik:
fixes #6818

following up on pull request #6803

### DIFF
--- a/pkg/cmd/step/verify/step_verify_preinstall.go
+++ b/pkg/cmd/step/verify/step_verify_preinstall.go
@@ -281,25 +281,28 @@ func (o *StepVerifyPreInstallOptions) verifyHelm(ns string) error {
 	log.Logger().Debug("Verifying Helm...")
 	// lets make sure we don't try use tiller
 	o.EnableRemoteKubeCluster()
-	_, err := o.Helm().Version(false)
+	v, err := o.Helm().Version(false)
 	if err != nil {
 		err = o.InstallHelm()
 		if err != nil {
 			return errors.Wrap(err, "failed to install Helm")
 		}
+		v, err = o.Helm().Version(false)
 	}
-	cfg := opts.InitHelmConfig{
-		Namespace:       ns,
-		OnlyHelmClient:  true,
-		Helm3:           false,
-		SkipTiller:      true,
-		GlobalTiller:    false,
-		TillerNamespace: "",
-		TillerRole:      "",
-	}
-	err = o.InitHelm(cfg)
-	if err != nil {
-		return errors.Wrapf(err, "initializing helm with config: %v", cfg)
+	if util.StartsWith(v, "v2") {
+		cfg := opts.InitHelmConfig{
+			Namespace:       ns,
+			OnlyHelmClient:  true,
+			Helm3:           false,
+			SkipTiller:      true,
+			GlobalTiller:    false,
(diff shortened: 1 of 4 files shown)


hferentschik:
@yelhouti, I think this code will work in as it will skip the call to init in the case of a helm version >3, but using this binary, I still get several other issues when trying to boot. 

Are you sure this will get you all the way? 

@jstrachan, do you see any issue here with merging this? I think the changes to the _Version()_ command are an improvement regardless. After talking to you and trying a bit myself I also think there is more missing to be able to use Helm 3.

yelhouti:
@hferentschik I am able to install using helm3 I just tryed again, all you need is export JX_HELM3=true, as documented somewhere

hferentschik:
> this other uirck is that after running jx-install step you have to quickly delete the the env dev

which `jx install` install step are we talking about here?

hferentschik:
@jstrachan, so what's the state of the code which gets enabled via _JX_HELM3=true_ enabled?

jstrachan:
this PR avoids the need for `JX_HELM3`: https://github.com/jenkins-x/jx/pull/6829

jstrachan:
do we really not need to run `helm init` on helm 3? not sure why we'd not want to. We may also want to install some plugins when using helm 3 so not really sure why we're even trying to disable helm init.

BTW with helm 3 we can't use `--client-only` which no longer exists https://github.com/jenkins-x/jx/pull/6826/files#diff-7c93774fbe3af48e31734ea58f57f559R681

jstrachan:
doh - yeah `helm init` has gone in helm 3 ;)

jstrachan:
ignore my previous comments ;) all looks good

hferentschik:
> this PR avoids the need for JX_HELM3: #6829

so with this PR you won't need `JX_HELM3=true` anymore and it will automatically do the right thing?

hferentschik:
> doh - yeah helm init has gone in helm 3 ;)

lol

PR #6839, opened 2020-02-28
fix: lets just log a warning if the alpha commands can't be found

jstrachan:
this can happen if github is having issues or there are firewall problems accessing github

this avoids failing on all jx commands which is pretty fatal

### DIFF
--- a/pkg/cmd/alpha/alpha.go
+++ b/pkg/cmd/alpha/alpha.go
@@ -37,7 +37,7 @@ var (
 )
 
 // NewCmdAlpha creates the "jx alpha" command
-func NewCmdAlpha(commonOpts *opts.CommonOptions) (*cobra.Command, error) {
+func NewCmdAlpha(commonOpts *opts.CommonOptions) *cobra.Command {
 	options := &Options{
 		CommonOptions: commonOpts,
 	}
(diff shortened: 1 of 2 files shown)


pkg/cmd/alpha/alpha.go
```diff
@@ -56,7 +56,7 @@ func NewCmdAlpha(commonOpts *opts.CommonOptions) (*cobra.Command, error) {
 
 	plugins, err := options.getPluginSpecs()
 	if err != nil {
-		return cmd, err
+		log.Logger().Warnf("failed to discover alpha commands from github: %s", err.Error())
```
garethjevans: should we return at this point? I've not looked at the code so am unsure at what `plugins` will be

garethjevans: it looks like it could be nil

jstrachan: its fine for it to be nil as iterating over a nil list is fine in go

e.g.
https://play.golang.org/p/igI2_8v9StC

if its nil it just doesn't add any sub commands

jstrachan: @garethjevans any chance I can convince you to /lgtm again? :)

PR #6840, opened 2020-02-28
chore: Use an annotation to see if running stages have already been reported

abayer:
#### Submitter checklist

#### Description

Should improve performance in build controller + Lighthouse a bit.

I'm running this on prod and it looks to make a real difference in performance thanks to getting rid of getting creds from vault and making a GitHub API call for every completed step.

n/a

### DIFF
--- a/pkg/cmd/controller/controller_build.go
+++ b/pkg/cmd/controller/controller_build.go
@@ -1223,7 +1223,7 @@ func (o *ControllerBuildOptions) reportStatus(kubeClient kubernetes.Interface, n
 	repo := activity.Spec.GitRepository
 	gitURL := activity.Spec.GitURL
 	activityStatus := activity.Spec.Status
-	status, description := toScmStatusAndDescription(activity)
+	statusInfo := toScmStatusDescriptionRunningStages(activity)
 
 	fields := map[string]interface{}{
 		"name":        activity.Name,
@@ -1233,7 +1233,7 @@ func (o *ControllerBuildOptions) reportStatus(kubeClient kubernetes.Interface, n
 		"gitSHA":      sha,
 		"gitURL":      gitURL,
 		"gitBranch":   activity.Spec.GitBranch,
-		"gitStatus":   status,
+		"gitStatus":   statusInfo.scmStatus,
 		"buildNumber": activity.Spec.Build,
 		"duration":    util.DurationString(activity.Spec.StartedTimestamp, activity.Spec.CompletedTimestamp),
 	}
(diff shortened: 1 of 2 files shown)


pkg/cmd/controller/controller_build.go
```diff
@@ -1361,10 +1362,12 @@ func toScmStatusAndDescription(activity *v1.PipelineActivity) (string, string) {
 	}
 	stagesByStatus := activity.StagesByStatus()
 
+	runningStages := ""
 	if len(stagesByStatus[v1.ActivityStatusTypeRunning]) > 0 {
+		runningStages = strings.Join(stagesByStatus[v1.ActivityStatusTypeRunning], ",")
 		description = fmt.Sprintf("Pipeline running stage(s): %s", strings.Join(stagesByStatus[v1.ActivityStatusTypeRunning], ", "))
 	}
-	return scmStatus, description
+	return scmStatus, description, runningStages
```
hferentschik: how long does this _runningStages_ get? Is there a risk that this exceed the 64 char limit?

hferentschik: I might the stages as array/slice!?

pkg/cmd/controller/controller_build.go
```diff
@@ -1342,7 +1343,7 @@ func CreateReportTargetURL(templateText string, params ReportParams) string {
 	return buf.String()
 }
 
-func toScmStatusAndDescription(activity *v1.PipelineActivity) (string, string) {
+func toScmStatusDescriptionRunningStages(activity *v1.PipelineActivity) (string, string, string) {
```
hferentschik: Getting close where I would create a struct and return it, three strings :-(

pkg/cmd/controller/controller_build.go
abayer: So k8s annotations can be absurdly long - 256kb, if I remember correctly - so this isn't a concern. =)

pkg/cmd/controller/controller_build.go
abayer: That is fair. Lemme do that.

pkg/cmd/controller/controller_build.go
```diff
@@ -1342,29 +1343,42 @@ func CreateReportTargetURL(templateText string, params ReportParams) string {
 	return buf.String()
 }
 
-func toScmStatusAndDescription(activity *v1.PipelineActivity) (string, string) {
-	scmStatus := ""
-	description := ""
+type reportStatusInfo struct {
```
hferentschik: 👍

pkg/cmd/controller/controller_build.go
```diff
@@ -1342,29 +1343,42 @@ func CreateReportTargetURL(templateText string, params ReportParams) string {
...
+		info.description = "Pipeline failed"
 	}
 	stagesByStatus := activity.StagesByStatus()
 
 	if len(stagesByStatus[v1.ActivityStatusTypeRunning]) > 0 {
-		description = fmt.Sprintf("Pipeline running stage(s): %s", strings.Join(stagesByStatus[v1.ActivityStatusTypeRunning], ", "))
+		info.runningStages = strings.Join(stagesByStatus[v1.ActivityStatusTypeRunning], ",")
+		info.description = fmt.Sprintf("Pipeline running stage(s): %s", strings.Join(stagesByStatus[v1.ActivityStatusTypeRunning], ", "))
+		if len(info.description) > 63 {
+			info.description = info.description[:59] + "..."
```
hferentschik: :-), I guess not needed after all

pkg/kube/constants.go
```diff
@@ -284,6 +284,8 @@ const (
 	AnnotationGitURLs = "jenkins.io/git-urls"
 	// AnnotationGitReportState used to annotate what state has been reported to git
 	AnnotationGitReportState = "jenkins.io/git-report-state"
+	// AnnotationGitReportRunningStages used to annotate what stages were last reported to git as running
+	AnnotationGitReportRunningStages = "jenkins.io/git-report-running-stages"
```
hferentschik: 👍

PR #6843, opened 2020-02-29
fix: Properly set lastCommitSha and batch PR SHAs in activity for batches

### LINKED ISSUE #6842: PipelineActivity has wrong/insufficient data for batch builds
For a batch build, the `PipelineActivity` has a few problems:
* `lastCommitSha` ends up pointing to the target branch SHA, not the "first" PR SHA as we'd expect.
* There is no SHA at all for the other PRs in the batch.

This is part of what causes https://github.com/jenkins-x/lighthouse/issues/564 (along with another problem in Lighthouse). We need to update our batch conciliation logic to have `lastCommitSha` be the head SHA for the PR we're tying the whole activity to (i.e., the one `gitBranch` corresponds to), and we should store the head SHAs for the other PRs in the batch as well.

abayer:
#### Submitter checklist

#### Description

The whole code around batch and activity needs to get rewritten, but this is a short-term fix needed for https://github.com/jenkins-x/lighthouse/issues/564.

fixes #6842


hferentschik:
Is this only a Lighthouse issue? Did this issue also exist with Prow?

abayer:
The fields were missing/wrong with Prow as well, but nothing actually cares about those fields with Prow, while LH uses the activity to play the same role as the ProwJob in Prow. That means we need to be more thorough and accurate in the activity with Lighthouse than we do with Prow.

docs/apidocs/openapi-spec/openapiv2.yaml
```diff
@@ -1828,11 +1828,13 @@ definitions:
         description: Restrictions limits who can merge
     type: object
   com.github.jenkins-x.jx.pkg.apis.jenkins_io.v1.PullRequestInfo:
-    description: PullRequestInfo contains information about a PR, like its PR and
-      Build numbers
+    description: PullRequestInfo contains information about a PR, like its PR, Build
```
hferentschik: so that's generated right?

pkg/apis/jenkins.io/v1/types_pipeline.go
```diff
@@ -71,10 +71,11 @@ type BatchPipelineActivity struct {
 	ComprisingPulLRequests []PullRequestInfo `json:"pullRequestInfo,omitempty" protobuf:"bytes,3,opt,name=pullRequestInfo"`
 }
 
-// PullRequestInfo contains information about a PR, like its PR and Build numbers
+// PullRequestInfo contains information about a PR, like its PR, Build numbers, and SHA
 type PullRequestInfo struct {
 	PullRequestNumber        string `json:"pullRequestNumber,omitempty" protobuf:"bytes,1,opt,name=pullRequestNumber"`
 	LastBuildNumberForCommit string `json:"lastBuildNumberForCommit,omitempty" protobuf:"bytes,2,opt,name=lastBuildNumberForCommit"`
+	LastBuildSHA             string `json:"lastBuildSHA,omitempty" protobuf:"bytes,3,opt,name=lastBuildSHA"`
```
hferentschik: What's about adding some GoDoc to _LastBuildSHA _ and _LastBuildNumberForCommit_? _PullRequestNumber_ seems obvious, the other two not so much. At least not to me atm.

pkg/kube/activity.go
```diff
@@ -343,7 +343,7 @@ func (k *PipelineActivityKey) addBatchBuildData(activitiesClient typev1.Pipeline
 		listOptions.LabelSelector = selector
 		list, err := activitiesClient.List(listOptions)
 		if err != nil {
-			return errors.Wrapf(err, "there was a problem listing all PipelineActivities for PR-%s with lastCommitSha %s", sha, prNumber)
+			return errors.Wrapf(err, "there was a problem listing all PipelineActivities for PR-%s with lastCommitSha %s", prNumber, sha)
```
hferentschik: lol. Good catch

pkg/tekton/pipeline_info.go
```diff
@@ -193,14 +195,13 @@ func CreatePipelineRunInfo(prName string, podList *corev1.PodList, ps *v1.Pipeli
...
 					if shaRegexp.MatchString(value) {
-						lastCommitSha = value
+						shaFromGitInit = value
 					} else {
 						branch = value
 					}
 				}
 			}
 		}
-		var pullPullSha string
```
hferentschik: So you are just moving this up to the rest of the var declaration, right

pkg/tekton/pipeline_info.go
```diff
@@ -193,14 +195,13 @@ func CreatePipelineRunInfo(prName string, podList *corev1.PodList, ps *v1.Pipeli
 					gitURL = value
 				case "-revision":
 					if shaRegexp.MatchString(value) {
-						lastCommitSha = value
+						shaFromGitInit = value
```
hferentschik:
```
if strings.HasPrefix(container.Name, "build-step-git-source") || strings.HasPrefix(container.Name, "step-git-source")
```

I find this a bit worrying, that one needs to match against container names and pipeline steps. That seems fragile, or am I missing something here?

hferentschik: So first we try to get git informtion from the actual checkout and then override with env variables if set, right?

pkg/tekton/pipeline_info.go
```diff
@@ -244,14 +245,39 @@ func CreatePipelineRunInfo(prName string, podList *corev1.PodList, ps *v1.Pipeli
...
 		}
-		if lastCommitSha == "" && pullPullSha != "" {
-			lastCommitSha = pullPullSha
+	}
+
+	if pullBaseSha != "" {
+		pri.BaseSHA = pullBaseSha
+	}
+
+	if pullPullSha != "" {
```
hferentschik: So I think I knew that once, but what is a pullPullSha again?

pkg/tekton/pipeline_info.go
```diff
@@ -263,24 +289,6 @@ func CreatePipelineRunInfo(prName string, podList *corev1.PodList, ps *v1.Pipeli
 		buildNumber = 1
 	}
 
-	pri.BaseSHA = pullBaseSha
-
-	if lastCommitSha == "" {
```
hferentschik: okk, that was nicely nested code ;-)

docs/apidocs/openapi-spec/openapiv2.yaml
abayer: Yup.

pkg/apis/jenkins.io/v1/types_pipeline.go
abayer: Will do!

hferentschik: thanks

pkg/tekton/pipeline_info.go
abayer: Yup. I couldn't see a reason to handle `pullPullSha` differently than `pullBaseSha` in this respect - they're both pulling from env vars set by Prow/Lighthouse that'll be on all the containers, so why are we putting one in a different scope than the other?

pkg/tekton/pipeline_info.go
abayer: I can't remember all of the history here, but I think this logic basically predated us realizing all the information we need regarding the SHA etc is actually in Prow/Lighthouse-provided env vars. So I'm opting not to mess with this janky logic, but I am making it the fallback, rather than the current behavior where we don't set a bunch of other things if we found a SHA here.

This change specifically fixes the problem of the `lastCommitSha` for batch builds being the master SHA, not the first-PR-in-the-batch HEAD SHA, because batch builds always have call the git source step for master (and we don't actually care what ref the git source step is checking out for any PR builds, in practice, because our own bespoke git merge step always checks out master and merges onto it regardless).

pkg/tekton/pipeline_info.go
abayer: The `PULL_PULL_SHA` env var is set for pull request builds (hence the first `PULL`) with the PR's `HEAD` SHA (hence the `PULL_SHA` part).

The other relevant env vars are the following - the first three are always set, even for non-PR builds, which is confusing but hey.
* `PULL_BASE_REF` - i.e., `master`
* `PULL_BASE_SHA` - the SHA for master's `HEAD`.
* `PULL_REFS` - basically a comma-separated list of `(branch):(sha)`, with `master` always first and then the PR(s) as `1:(sha),2:(other sha),...`

And then just on PRs:
* `PULL_NUMBER` - which kinda speaks for itself.

pkg/tekton/pipeline_info.go
```diff
@@ -182,6 +184,11 @@ func CreatePipelineRunInfo(prName string, podList *corev1.PodList, ps *v1.Pipeli
 	}
 	containers, _, isInit := kube.GetContainersWithStatusAndIsInit(pod)
 	for _, container := range containers {
+		// We historically used the git source step automatically injected by Tekton to get the git URL and the sha or
```
hferentschik: 👍 that helps a bit, thanks

pkg/tekton/pipeline_info.go
hferentschik: maybe one day I will fully understand what you just wrote ;-) It's the batching I don't fully grasp atm. I get it conceptually, but not sure how it actually works and how it keeps state for everything.

pkg/tekton/pipeline_info.go
```diff
@@ -193,65 +200,107 @@ func CreatePipelineRunInfo(prName string, podList *corev1.PodList, ps *v1.Pipeli
...
 					}
 				}
 			}
 		}
-		var pullPullSha string
 		for _, v := range container.Env {
 			if v.Value == "" {
 				continue
 			}
+			// PULL_PULL_SHA is set by Prow/Lighthouse with the HEAD SHA of the PR being built, or the HEAD SHA of the
```
hferentschik: 👍

pkg/tekton/pipeline_info.go
abayer: Yeah, I admit that I don't entirely understand the batch logic in `kube/activity.go` - something else that probably needs to be rewritten. I'm also still not yet sure if the SHA this PR adds will actually suffice for Lighthouse and batch. We'll see.

abayer:
So https://github.com/jenkins-x/jx/pull/6843/commits/22475c4dac0c3cf25d4d1a17543b29a8e2839c53 is to fix a test that's actually going to fail everywhere now due to the time passed since the date used for a build used in `get_activity_test.go` has just passed 10,000 hours, making the spacing in the output of `jx get activity` change...and the test expected the spacing to never change. I've reworked it to, well, not depend on spacing any more.

PR #6848, opened 2020-03-02
feat: allow extra labels to be added to generated PRs

macox:
#### Description

Allows additional labels to be added to PR's when using `jx step create pr`.
The purpose of this functionality is to avoid unrelated `bump` PRs to be combined in to one PR, which can potential cause a failure in one change blocking the progress of another unrelated change.

fixes #

### DIFF
--- a/pkg/gits/operations/pull_request_op_test.go
+++ b/pkg/gits/operations/pull_request_op_test.go
@@ -48,8 +48,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-func TestCreatePullRequest(t *testing.T) {
 
+func setupTestPullRequestOperation(t *testing.T) (operations.PullRequestOperation) {
 	_, _, _, commonOpts, _ := getFakeClientsAndNs(t)
 
 	testOrgName := "testowner"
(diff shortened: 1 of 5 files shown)


pkg/gits/operations/pull_request_op_test.go
```diff
@@ -103,6 +103,87 @@ func TestCreatePullRequest(t *testing.T) {
...
+
+	o.GitURLs = []string{"testowner/testrepo"}
+	o.SrcGitURL = "testowner/testrepo"
+	o.Version = "3.0.0"
+
+	var results *gits.PullRequestInfo
+
+	logOutput := log.CaptureOutput(func() {
+		results, err = o.CreatePullRequest("test", func(dir string, gitInfo *gits.GitRepository) (strings []string, e error) {
+			return []string{"1.0.0", "v1.0.1", "2.0.0"}, nil
```
hferentschik: So this functions returns versions/tags?

pkg/gits/operations/pull_request_op_test.go
```diff
@@ -103,6 +103,87 @@ func TestCreatePullRequest(t *testing.T) {
...
+
+	pegomock.When(gitter.HasChanges(pegomock.AnyString())).ThenReturn(true, nil)
+
+	o.GitURLs = []string{"testowner/testrepo"}
+	o.SrcGitURL = "testowner/testrepo"
+	o.Version = "3.0.0"
+
+	var results *gits.PullRequestInfo
+
+	logOutput := log.CaptureOutput(func() {
```
hferentschik: I don't think asserting against user output makes for good tests. That's fragile at best.
Would you not go via the _fakeGitProvider_, go get to the appropriate fake repository and assert that there was a (fake) PR added and that this PR has labels set? I think that is a much more direct test.

hferentschik:
@macox, would you mind opening an issue for this as well?

hferentschik:
personally, I'd also squash the commits into a single consistent commit for this feature. Historically there is no value in the fact that you had a typo in your code which you have to fix.

pkg/gits/operations/pull_request_op_test.go
```diff
@@ -103,6 +103,87 @@ func TestCreatePullRequest(t *testing.T) {
...
+		gitter,
+		fakeGitProvider,
+		nil,
+		resources_test.NewMockInstaller(),
+	)
+
+	err := testhelpers.CreateTestEnvironmentDir(o.CommonOptions)
+	assert.NoError(t, err)
+
+	pegomock.When(gitter.HasChanges(pegomock.AnyString())).ThenReturn(true, nil)
```
hferentschik: I'd probably extract all of the above into some _setUp_ function of sorts so that it is easier to  find and focus on the actual test here.

hferentschik:
On a different note, how is this going to be used? Are we adding new labels (apart of _updatebot_) to each PR? I guess what I am concerned about is to mass create labels. Labels are an important feature when working with issues/pull request on GitHub. I am not sure whether I would be thrilled if the tool I am using keeps adding labels for each build it runs.

macox:
> Are we adding new labels (apart of _updatebot_) to each PR?

No, extra labels will only be added when required. The use case I see is that two independent work streams are updating a single repo. The these streams create PRs at a similar time they are combined in a single PR. This now means the independent work streams are now dependant on each other, and any failures are difficult to reason about their origin.

hferentschik:
> No, extra labels will only be added when required. 

Which is when? When it happens how will the created labels look like? Just version strings like "v1.0.0"? Not sure people would like to see labels to appear which seem to have a tie to versions. I can see that this gets confusing. Would it not at least make sense to have a prefix or something of sort so that it becomes who or what is creating/managing these labels?

macox:
> > No, extra labels will only be added when required.
> 
> Which is when? When it happens how will the created labels look like? Just version strings like "v1.0.0"? Not sure people would like to see labels to appear which seem to have a tie to versions. I can see that this gets confusing. Would it not at least make sense to have a prefix or something of sort so that it becomes who or what is creating/managing these labels?

I've not suggested where labels should be added, or that they should be version strings.  They could look like whatever is relevant to the particular use case. This functionality is to add the ability to add labels IF required. In my current use case adding an extra label is a big advantage as previously described, and will reflect the workstream/repo that is creating the PR.

pkg/gits/operations/pull_request_op_test.go
```diff
@@ -79,30 +79,83 @@ func TestCreatePullRequest(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.NotNil(t, results, "we must have results coming out of the PR creation")
+
+	prNumber := 1
+	assert.Equal(t, prNumber, *results.PullRequest.Number,"This should be PR number %d", prNumber)
+
+	assert.Equal(t, fmt.Sprintf("chore(deps): bump %s from %s to %s", prOpts.SrcGitURL, fromVersion, prOpts.Version),
+		results.PullRequestArguments.Title, "The PR title should contain the old and new versions")
+
+	assert.Equal(t, 1, len(results.PullRequest.Labels),"One label expected")
```
hferentschik: Look at that, we already have access to the labels via the returned pull request details :-) 👍

hferentschik: What's about adding an assert for explicitly 'update-bot'. This kind of codifies the label we expect here

pkg/gits/operations/pull_request_op_test.go
```diff
@@ -48,8 +48,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-func TestCreatePullRequest(t *testing.T) {
 
+func setupTestPullRequestOperation(t *testing.T) (operations.PullRequestOperation) {
```
hferentschik: 👍

hferentschik:
Seems formatting is a bit off as well

hferentschik:
> I've not suggested which labels should be added,

Fair enough, your test seemed to imply that to a certain degree, hence I asked.

PR #6849, opened 2020-03-02
feat: add a pipeline which scans the Go dependencies with whitesource

ccojocar:
#### Submitter checklist

#### Description

This adds a new pipeline which will scan the Go dependencies using the whitesource agent. This will execute
when a pull request is merged to master and it will run in parallel with the default release pipeline.

The dependencies are updated to whitesource service which will check if there is any CVE open for them. 
Note that currently this will only report the dependencies, there isn't any policy in place to fail the build in 
case there is any critical/high CVE.

This requires https://github.com/jenkins-x/environment-tekton-weasel-dev/pull/405 in oder to run.

fixes #


whitesource.config
```diff
@@ -0,0 +1,256 @@
+####################################################################
```
hferentschik: what do you think about making this config a dot file or even better imo move the config into _hacking_. I really would like to reduce the clutter in the root directory.

ccojocar: This is the default config provided by whitesource which is loaded from the root folder by the container which runs in the pipeline.

hferentschik: and the location is not configurable?

hferentschik: Will there be a change to the Prow config as well? Do you want this as a required status check? Will this even run on our prod cluster?

ccojocar: The PR which changes the dev environment is linked in the description. 

https://github.com/jenkins-x/environment-tekton-weasel-dev/pull/405

ccojocar: I guess that I can make the config path configurable  and when is not set can default to `.whitesource.config`

hferentschik: > I guess that I can make the config path configurable and when is not set can default to .whitesource.config

that would be awesome

ccojocar: I  changed the whitesource image to make to be able to setup the configuration path.

ccojocar: Please could you also review the prow pull request? I think this can be merged first.

PR #6869, opened 2020-03-05
fix: setHelmBinary should use selected helm

yelhouti:


hferentschik:
Would you be able to create an issue or generally provide more context? What is the error and how does it manifest itself. I know it seems trivial, but it helps a lot if you are not currently working on this.

pkg/cmd/opts/helm.go
```diff
@@ -661,7 +661,7 @@ func (o *CommonOptions) HelmInitDependencyBuild(dir string, chartRepos []string,
 	// lets stick with helm2 for this step
 	//
 	helmBinary := o.Helm().HelmBinary()
-	o.Helm().SetHelmBinary("helm")
+	o.Helm().SetHelmBinary(helmBinary)
```
hferentschik: What's about the comment here? It seems it was an explicit choice to use "helm". Is this not an issue anymore?

yelhouti: I didn't see any problem, using helm3, in this step, maybe I'm missing something, but for me it works... I missed the comment, does removing it work for you?

hferentschik: I think we would need to remove more than just the comment. As the code is written now it is a noop and can effectively be removed, right:

```
helmBinary := o.Helm().HelmBinary()
o.Helm().SetHelmBinary(helmBinary)
```

This gets the helm binary from the `Helmer` and then set the very value it just retrieved. So really if we want to do anything here it would be removing the comment + the two lines above. The explicit setting of "helm" was a workaround when using helm3 which you are saying is not needed anymore.

yelhouti: I think you are right.
For all future changes we should adopt the following though:

The way I understand the code is that helmBinary is the name of the executable to be used when executing the command, this is needed because in our docker image we have helm and helm3 (the two binaries), normally users (outside the pipeline) have on "helm" and call it directly. helm2 and helm3 differ in few way:
- no helm init for helm3
- --name is remove in helm3 and is part of the params
...
In theory we should always call helm but in the pipeline the JX_HELM3 param should be used to tell use to set helmBinary to helm3.
Whenever we check which helm version we are using before running a command, instead of checking against helmBinary we should check against helm.version.

I will first fix this PR, but someone needs to be attributed this task to fix all I just explained...

hferentschik:
@yelhouti, can you refresh my memory where we were on this?

yelhouti:
mainly the work I'm trying to do here to make helm3 work is redundant with what is being done with jxl, maybe I should focus my efforts there if I find any problems. we should maybe have a roadmap/milestones and start merging what is already in jxl

hferentschik:
> mainly the work I'm trying to do here to make helm3 work is redundant with what is being done with jxl, maybe I should focus my efforts there if I find any problems. 

Probably.

> we should maybe have a roadmap/milestones and start merging what is already in jxl

There is work/discussion ongoing on how to proceed with `jx` vs `jxl`. It might not be as simple as merging.

hferentschik:
@yelhouti, I am closing this pull request until there is a more concrete plan on how to move forward.

PR #6870, opened 2020-03-05
fix: add helm3 suppot for helm_cli install and template

yelhouti:
#### Submitter checklist

fixes #

### DIFF
--- a/pkg/helm/helm_cli.go
+++ b/pkg/helm/helm_cli.go
@@ -309,22 +309,12 @@ func (h *HelmCLI) BuildDependency() error {
 func (h *HelmCLI) InstallChart(chart string, releaseName string, ns string, version string, timeout int,
 	values []string, valueStrings []string, valueFiles []string, repo string, username string, password string) error {
 	var err error
-	currentNamespace := ""
-	if h.Binary == "helm3" {
-		log.Logger().Warnf("Manually switching namespace to for helm3 alpha - %s, this code should be removed once --namespaces is implemented", ns)
-		currentNamespace, err = h.getCurrentNamespace()
-		if err != nil {
-			return err
-		}
-
-		err = h.setNamespace(ns)
-		if err != nil {
-			return err
-		}
-	}
-
 	args := []string{}
-	args = append(args, "install", "--wait", "--name", releaseName, "--namespace", ns, chart)
+	if h.Binary == "helm3" {
+		args = append(args, "install", "--wait", releaseName, "--namespace", ns, chart)
+	} else {
+		args = append(args, "install", "--wait", "--name", releaseName, "--namespace", ns, chart)
+	}
 	repo, err = addUsernamePasswordToURL(repo, username, password)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


pkg/helm/helm_cli.go
```diff
@@ -423,7 +406,12 @@ func (h *HelmCLI) FetchChart(chart string, version string, untar bool, untardir
 // Template generates the YAML from the chart template to the given directory
 func (h *HelmCLI) Template(chart string, releaseName string, ns string, outDir string, upgrade bool,
 	values []string, valueStrings []string, valueFiles []string) error {
-	args := []string{"template", "--name", releaseName, "--namespace", ns, chart, "--output-dir", outDir, "--debug"}
+	args := []string{}
+	if h.Binary == "helm3" {
```
hferentschik: Do you actually really have a _helm3_ binary? This keeps confusing me. AFAIU there is no 'helm3' binary? Should this all not be handled to query/check the version of the existing _helm_ binary? I am reluctant to go down this road tbh.

hferentschik:
hi @yelhouti, are you aware of the [Jenkins X office hours](https://jenkins-x.io/community/#office-hours)? Turns out that in the next office hours on the 5th of March, there will be some discussion/talk around Helm 3 and  Helmfile (and more). Maybe you could attend? This might help you understand where Helm 3 support is heading and you might be able to ask some questions. 

If you cannot make it, make sure to check out the recording afterwards.

pkg/helm/helm_cli.go
yelhouti: I was wondering the same thing, but to keep it consistent with the other checks I used helm3. I think this is mainly because of the jx builder image having two binaries, helm and helm3, but I also think that the only place where helm3 should be visible in the code when calling the binary, and all the checks should be done through the helm.binVersion...

hferentschik: the thing is that the helm3 binary in the builders is not Helm 3 proper. At least I don't think so. 
@jstrachan might know!? 

Really this code needs to change to deal with Helm 3 proper where there is just one single Helm binary in different versions. This needs to be reflected in `jx`, but also in the builder images. 

Given this and your other pull request, what is the net win? Do you really end up with a Jenkins X install which is usable?

yelhouti: Not really, I still have to add JX_HELM3 as env variable in my buildpack and some other stuff, I was advised by @jstrachan to use jxl but so far no luck, I will let you know when I finish how I think we should proceed...

hferentschik: Any updates on this? Looks like this might be obsolete given your latest pull requests, or not?

yelhouti: This stays on hold until we merge the jxl with helmfile if it's ok for you. if not I will close it and reopen if needed...

hferentschik: Let's close it and reopen. Nothing will be lost, but it is easier to keep track of changes. I also think it will take quite some time until jxl changes will make it in here.

yelhouti: sure ! closing it

PR #6872, opened 2020-03-05
fix: KubeAuthConfigHandler.LoadConfig should never return an empty config

### LINKED ISSUE #6871: KubeAuthConfigHandler.LoadConfig should never return an empty AuthConfig
Literally none of the other config handlers do this - they return a nil `*AuthConfig` and an error in any situation where there isn't content for the `*AuthConfig`. So let's be consistent here.

abayer:
#### Submitter checklist

#### Description

This is to be consistent with the other config handlers, so we don't need to do special casing anywhere for empty configs returned by `LoadConfig`.

fixes #6871

### DIFF
--- a/pkg/auth/kube_config_handler.go
+++ b/pkg/auth/kube_config_handler.go
@@ -2,6 +2,7 @@ package auth
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/jenkins-x/jx/pkg/util"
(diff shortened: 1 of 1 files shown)


hferentschik:
@abayer might not be so easy :-)

abayer:
@hferentschik Yup, you're right. I'm going to abandon this. =)

hferentschik:
>  Yup, you're right. I'm going to abandon this. 

the API still is "broken" though.

Issue #6885, opened 2020-03-09
Use Kaniko directly instead of the hack that overrides skaffold build inside jx


yelhouti:
### Summary 
there is a hack in the code that modifies the step from skaffold to use kaniko directly, until this is fixed this hack should removed, and pipelines should be updated
https://github.com/jenkins-x/jx/blob/d4dc45c69d6cd912cb4cb9fa4f909ceeb89bd76f/pkg/tekton/syntax/pipeline.go#L1206-L1237

### Jx version

The output of `jx version` is:

```
2.0.1234
```

### Jenkins type

- [X] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

abayer:
So this is actually a pain to deal with because build packs aren't versioned, and static master users pick up the same build packs as Tekton-driven pipeline users do. Skaffold is needed for static master cases, so until we address the lack of versioning of build packs, we're stuck with the silly replacement logic.

yelhouti:
Thanks for the explanation :)

PR #6204, opened 2019-11-26
fix: Use debug image and /busybox/sh for /kaniko/executor

abayer:
#### Submitter checklist

- [x] Change is code complete and matches issue description.
- [x] Change is covered by existing or new tests.

#### Description

/bin/sh doesn't exist on Kaniko's `executor` or `warmer` images, which is why we've been special-casing them all along. But that ends up creating a number of other problems. There is a `debug-...` image for every `executor` image, with `/busybox/sh`, so let's try using the `debug-...` image with `/busybox/sh -c` as the command when we see `/kaniko/executor` as the step command.

This should also allow environment variables to be used in kaniko step arguments.

#### Special notes for the reviewer(s)

/assign @garethjevans 

#### Which issue this PR fixes

fixes #6197
fixes #6058

PR #6886, opened 2020-03-09
fix: make kaniko use runtime env vars (instead of meta-pipeline/compile time) value

yelhouti:
…le time) value

#### Submitter checklist

workaround until #6885 is fixed and all this code is removed


abayer:
So it used to be that the overridden kaniko step didn't read environment variables correctly, but I _think_ I fixed this? Lemme dig around in history to see what I can find.

abayer:
yessss - #6204 fixed that. So I expect this PR to cause some test failures in `step_create_task_test.go` since the expected `Task`s won't match the generated ones, but I can clean that up once we're sure it's generating valid `Task`s in the first place.

abayer:
Ok, so this doesn't actually work: the BDD tests are failing to push with `error pushing image: failed to push to destination gcr.io//bdd-spring-1583758665:0.0.1` - I think because the `ORG` env var doesn't actually exist. You should change that to `REPO_OWNER`.

yelhouti:
>  I think because the `ORG` env var doesn't actually exist. You should change that to `REPO_OWNER`.
Actually the repo owner is not necessarily the same as the $ORG, where ever I found $ORG in the code `$DOCKER_REGISTRY/$ORG/$APP_NAME` it looked like this. 
Maybe the test is wrong (input that should be provided is not)

pkg/cmd/preview/preview.go
```diff
@@ -1001,18 +1001,12 @@ func (o *PreviewOptions) getContainerRegistry(projectConfig *config.ProjectConfi
...
 	containerRegistry, err := o.getContainerRegistry(projectConfig)
+	dockerRegistryOrg := o.GetDockerRegistryOrg(projectConfig, o.GitInfo)
+
 	if err != nil {
 		return "", err
 	}
 
-	organisation := os.Getenv(ORG)
-	if organisation == "" {
-		organisation = os.Getenv(REPO_OWNER)
```
abayer: Huh, I was nervous at first looking at this, but now I see that this was all just fallbacks for setting `dockerRegistryOrg`, so this was just silly code in the first place. This feels like a real improvement!

yelhouti: thanks :)

pkg/cmd/step/buildpack/test_data/step_buildpack_apply/go_inherit/Jenkinsfile
```diff
@@ -23,7 +23,7 @@ pipeline {
             checkout scm
             sh "make linux"
             sh "export VERSION=$PREVIEW_VERSION && skaffold build -f skaffold.yaml"
-            sh "jx step post build --image $DOCKER_REGISTRY/$ORG/$APP_NAME:$PREVIEW_VERSION"
+            sh "jx step post build --image $DOCKER_REGISTRY/$DOCKER_REGISTRY_ORG/$APP_NAME:$PREVIEW_VERSION"
```
abayer: We probably don't need to be changing the `cmd/step/buildpack` tests - these are really testing the generation of Jenkinsfiles from build packs for static master setups, and in static masters, `ORG` is set by Jenkins, and `DOCKER_REGISTRY_ORG` is not being set, and we're not going to waste the effort in trying to change that since we're removing support for static masters in the near future. That said, we're not going to be changing the build packs any time soon because build packs aren't versioned, meaning existing static master users are still going to be picking up `master` in the build packs repos, so..yeah. Messy.

Regardless - it doesn't _really_ matter here since these tests are using buildpacks specified here in the test, not the ones from the build packs repos, so it's ok to change these since you already did. =)

pkg/cmd/step/create/step_create_task.go
```diff
@@ -868,6 +868,12 @@ func (o *StepCreateTaskOptions) modifyEnvVars(container *corev1.Container, globa
 			Value: o.DockerRegistry,
 		})
 	}
+	if kube.GetSliceEnvVar(envVars, "DOCKER_REGISTRY_ORG") == nil {
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  "DOCKER_REGISTRY_ORG",
+			Value: o.DockerRegistryOrg,
```
abayer: I think what we need to do here is something like this: 
```
        dockerRegistryOrg := o.DockerRegistryOrg
        if dockerRegistryOrg == "" && gitInfo != nil {
            dockerRegistryOrg = gitInfo.Organisation
        }
	if kube.GetSliceEnvVar(envVars, "DOCKER_REGISTRY_ORG") == nil {
		envVars = append(envVars, corev1.EnvVar{
			Name:  "DOCKER_REGISTRY_ORG",
			Value: dockerRegistryOrg,
		})
	}
```

yelhouti:
This work is pending until we see hos env vars are used when jenkins-x.yml if present, if env vars keep taking precedence we should drop the PR...

hferentschik:
@yelhouti @abayer - what's the state on this PR? More work needed or maybe obsolete?

yelhouti:
Due to the Corona outbreak, I am stuck with some clients, I will revue this as soon as I can, sorry guys...

hferentschik:
Any update on this? Would it be ok to close this pull request for now? We can always reopen later.

yelhouti:
Sure, I found a workaround, hope it keeps working, closing this

PR #6888, opened 2020-03-09
chore: Add warning for upcoming removal of static masters

### LINKED ISSUE #6887: Add log warning for static master removal to all jx commands run with static master cluster
With static master removal coming soon, we should start warning users that it's coming.

abayer:
#### Submitter checklist

#### Description

This message should show up in the following situations:
* Running locally, connected to a cluster using static master.
* Running locally, calling `jx create cluster` without `--skip-installation` and without flags for installing with Tekton.
* Running locally, calling `jx install` without flags for installing with Tekton.

It should never show up in pipelines, in controllers, or in any case where you're connected to a Tekton cluster.

fixes #6887

### DIFF
--- a/pkg/cmd/cmd.go
+++ b/pkg/cmd/cmd.go
@@ -18,12 +18,21 @@ package cmd
 
 import (
 	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"runtime"
+	"strconv"
+	"strings"
+	"syscall"
 
 	"github.com/jenkins-x/jx/pkg/cmd/alpha"
 	"github.com/jenkins-x/jx/pkg/cmd/deprecation"
 	"github.com/jenkins-x/jx/pkg/cmd/experimental"
 	"github.com/jenkins-x/jx/pkg/cmd/profile"
 	"github.com/jenkins-x/jx/pkg/cmd/ui"
+	"github.com/jenkins-x/jx/pkg/kube"
+	"github.com/jenkins-x/jx/pkg/kube/cluster"
 	"github.com/spf13/viper"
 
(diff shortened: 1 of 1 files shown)


abayer:
This is aimed to merge on March 11, once there's a blog post on the removal of static masters for this to link to.

pkg/cmd/cmd.go
```diff
@@ -348,6 +349,62 @@ func setLoggingLevel(cmd *cobra.Command, args []string) {
...
+
+		teamSettings := &env.Spec.TeamSettings
+		shouldWarn = !teamSettings.IsJenkinsXPipelines()
+	}
+
+	if shouldWarn {
+		cmd.Println("WARNING")
+		cmd.Println("")
+		cmd.Println("Traditional Jenkins masters are being deprecated and removed from all new Jenkins X releases starting in April, 2020.")
+		cmd.Println("We strongly recommend not using this configuration and instead using the default Tekton-based configuration instead for all Jenkins X activity going forward.")
```
MarckK: Log message has 'instead' twice. Possible replacement: "We strongly recommend using the default Tekton-based configuration for all Jenkins X activity going forward. We do not recommend using Jenkins masters."

abayer: How about `We strongly recommend using the default Tekton-based configuration instead for all Jenkins X activity going forward. We do not recommend using the Jenkins X workflow based around Jenkins masters.`? "We do not recommend using Jenkins masters" is a little too potentially misleading. =)

ethangj: I'm good with `We strongly recommend using the default Tekton-based configuration instead for all Jenkins X activity going forward. We do not recommend using the Jenkins X workflow based around Jenkins masters.`

Issue #6799, opened 2020-02-25
The README of jx project is quite outdated for the current version focusing on boot


ccojocar:
### Summary 

The README of jx project is quite outdate referencing commands and instructions from the early days of the project when the main focus was on classic Jenkins. 

Probably, it is time to revamp it with more actual instructions. This is usually the place where most of the people land for the first time.

PR #6899, opened 2020-03-10
chore: remove deprecated minishift/minikube commands

### LINKED ISSUE #4088: Remove Minishift/Minikube cluster options
### Summary 

Running Jenkins-X on Minishift/Minikube does not seem to be terribly useful given that these single node clusters won't be reachable from the outside, meaning that webhooks et all won't work.

There are also resource constraints. A full Jenkins-X install with all its components just takes too many resources for such a single node cluster.

Last but not least, Minishift is now in maintenance mode. It won't work with OpenShift 4 anymore.

mrageh:
We've deprecated the minishift command a while back, this commit removes it from jx.

#### Submitter checklist

fixes #4088


mrageh:
/hold 

Let's not merge this until we figure out how we're going to version all the deprecated commands

README.md
```diff
@@ -252,92 +248,21 @@ For example to add the [Gitea Git server](https://gitea.io/en-US/) to your Jenki
...
 * add Gitea as a Git server (via the `jx create git server gitea` command).
 * create a new user in Gitea (via the `jx create git user -n gitea` command).
 * create a new Git API token in Gitea (via the `jx create git token -n gitea -p password username` command).
-     
+
 ## Troubleshooting
 
 We have tried to collate common issues here with work arounds. If your issue isn't listed here please [let us know](https://github.com/jenkins-x/jx/issues/new).
- 
-### Cannot create cluster Minikube
```
hferentschik: 👍

hferentschik: I think there is another issue just around this README update. https://github.com/jenkins-x/jx/issues/6799

Maybe while on it, we should add some other information into the README. @dennisnewel

pkg/cmd/create/create_cluster.go
```diff
@@ -116,7 +112,6 @@ func NewCmdCreateCluster(commonOpts *opts.CommonOptions) *cobra.Command {
 	cmd.AddCommand(NewCmdCreateClusterEKS(commonOpts))
 	cmd.AddCommand(NewCmdCreateClusterGKE(commonOpts))
 	cmd.AddCommand(NewCmdCreateClusterMinikube(commonOpts))
```
hferentschik: That probably needs removing as well

pkg/cmd/create/create_etc_hosts.go
```diff
@@ -75,14 +75,6 @@ func (o *CreateEtcHostsOptions) Run() error {
 	if name == "" {
 		return util.MissingOption(name)
 	}
-	if o.IP == "" {
```
hferentschik: I think we safely can remove the whole command

README.md
dennisnewel: are you thinking about adding a deprecation/removal section or something? e.g. "We removed Minicube support in version 2.0.xxx"?

hferentschik: @dennisnewel, I was more thinking of whether we should link to a more general FAQ. I would have thought we have a FAQ section in the online docs, but that does not seem to be the case. I don't think the README is a good location for this type of troubleshooting question. If there is no FAQ in the docs, we should create one there.

Regarding Minishift/Minikube, I don't think any deprecation section is needed. Definitely not in the README.

pkg/cmd/create/create_etc_hosts.go
hferentschik: @mrageh, I meant we can remove the whole create_etc_host command.

mrageh:
Closing in favour of #6947

PR #6910, opened 2020-03-12
feat: add option request memory to devpod

### LINKED ISSUE #6918: add request-memory option for devpod
### Summary 
The same way we have a request cpu option for devpod and request memory should be added

### Jenkins type

- [ ] Classic Jenkins

yelhouti:
#### Submitter checklist

fixes #6918

### DIFF
--- a/pkg/cmd/create/create_devpod.go
+++ b/pkg/cmd/create/create_devpod.go
@@ -47,6 +47,7 @@ import (
 
 const (
 	optionRequestCPU    = "request-cpu"
+	optionRequestMemory = "request-memory"
 	devPodGoPath        = "/workspace"
 	devPodContainerName = "devpod"
 )
@@ -84,6 +85,7 @@ type CreateDevPodOptions struct {
 	Suffix          string
 	WorkingDir      string
 	RequestCpu      string
+	RequestMemory   string
 	Dir             string
 	Reuse           bool
 	Sync            bool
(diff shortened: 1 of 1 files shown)


hferentschik:
@yelhouti, looking good. Would you mind opening an issue as well and referencing it in the pull request and/or commit message?

Wonder whether this could be somehow made testable?

yelhouti:
@hferentschik I can definitely try to add some tests if needed, but since this a dev feature, and there is already a request-cpu, I just copy pasted and edited the code, this is really not a big deal...
Also I only tend to create issues for bugs, when doing new features or code cleanups I feel the PR is self documenting (not really an issue)

hferentschik:
>  I can definitely try to add some tests if needed, but since this a dev feature, and there is already a request-cpu, I just copy pasted and edited the code, this is really not a big deal...

ok. I am fine with that. Test coverage is an issue in the codebase, so I like to use every opportunity to nudge people to write tests :-) I think at least planting the seed is a win.

> Also I only tend to create issues for bugs, when doing new features or code cleanups I feel the PR is self documenting (not really an issue)

We try to create issues for everything. Issues are not only reserved for bugs. For that reason, there are also labels like kind/bug and kind/feature.
Being consistent this way makes searching easier (even though GitHub treats PR and issues partly the same way, they are not and you usually search in one or the other) and it makes fore a more consistent changelog generation. 

Also creating an issue helps you to reflect and put into words what you are trying to solve and gives the reviewer the context he needs to see the change in the bigger picture. 

Hope that clarifies a bit where I am coming from.

yelhouti:
> Hope that clarifies a bit where I am coming from.
Totally get you, here you go an issue has been added :)

PR #6911, opened 2020-03-12
fix: enforce GCP service account minimum name length Closes #6334 fixes #6334

### LINKED ISSUE #6334: Jenkins fail to create a service account on google cloud if the cluster is two character name
When installing jenkins-x on a two character name cluster on gke, it fail to create the service account because the it name must have at least 6 characters.

so for 'ci', the name will be 'ci-ko'

I don't really have the time to reproduce the issue for the console output, but it's pretty straight forward.

joshuasimon-taulia:
#### Submitter checklist

#### Description
when a gke cluster name length is 3 characters or less, fixes failed service account creation during jx/jxl boot create
```
Validating Kaniko secret in namespace jx
attempting to lazily create the deploy namespace jx
Configuring Kaniko service account cd-ko for project devtools-2341
Unable to find service account cd-ko, checking if we have enough permission to create
Creating service account cd-ko
error: failed to verify requirements in dir /var/folders/bp/4mgw9f094872yc6v7gyzqqt1frkjzj/T/helmboot-736612290: failed to lazily create the kaniko secret in: jx: creating the service account: failed to run 'gcloud iam service-accounts create cd-ko --project devtools-1234 --display-name cd-ko' command in directory '', output: 'ERROR: (gcloud.iam.service-accounts.create) argument NAME: Bad value [cd-ko]: Service account name must be between 6 and 30 characters (inclusive), must begin with a lowercase letter, and consist of lowercase alphanumeric characters that can be separated by hyphens.
```

Closes #6334 fixes #6334

### DIFF
--- a/pkg/cmd/create/install.go
+++ b/pkg/cmd/create/install.go
@@ -2240,7 +2240,7 @@ func (options *InstallOptions) ConfigureKaniko() error {
 			}
 		}
 
-		serviceAccountName := naming.ToValidNameTruncated(fmt.Sprintf("%s-ko", clusterName), 30)
+		serviceAccountName := naming.ToValidGCPServiceAccount(fmt.Sprintf("%s-ko", clusterName))
 		log.Logger().Infof("Configuring Kaniko service account %s for project %s", util.ColorInfo(serviceAccountName), util.ColorInfo(projectID))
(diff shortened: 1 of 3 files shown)


abayer:
/ok-to-test
/lgtm

Thanks!

PR #6913, opened 2020-03-13
fix: remove namespace switch used for helm3 alpha

yelhouti:

#### Submitter checklist

fixes #

### DIFF
--- a/pkg/helm/helm_cli.go
+++ b/pkg/helm/helm_cli.go
@@ -309,19 +309,6 @@ func (h *HelmCLI) BuildDependency() error {
 func (h *HelmCLI) InstallChart(chart string, releaseName string, ns string, version string, timeout int,
 	values []string, valueStrings []string, valueFiles []string, repo string, username string, password string) error {
 	var err error
-	currentNamespace := ""
-	if h.Binary == "helm3" {
-		log.Logger().Warnf("Manually switching namespace to for helm3 alpha - %s, this code should be removed once --namespaces is implemented", ns)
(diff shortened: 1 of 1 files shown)


pkg/helm/helm_cli.go
```diff
@@ -450,19 +430,6 @@ func (h *HelmCLI) Template(chart string, releaseName string, ns string, outDir s
 // UpgradeChart upgrades a helm chart according with given helm flags
 func (h *HelmCLI) UpgradeChart(chart string, releaseName string, ns string, version string, install bool, timeout int, force bool, wait bool, values []string, valueStrings []string, valueFiles []string, repo string, username string, password string) error {
 	var err error
-	currentNamespace := ""
-	if h.Binary == "helm3" {
```
hferentschik: I like seeing the removal of these comparisons against "helm3". This looks promising. What issue/behaviour do you try to fix/implement with this change? Could you capture it in an issue?

yelhouti: in helm3 alpha we used to switch the namespace manually in the kubeconfig file as the flag "--namespace" wasn't working, now that we use a later version in our container, there is no more need to modify the config. this basically juste a cleanup... not a real issue

hferentschik:
@garethjevans, looking at the code, you seemed to have been the one putting these "helm3" checks in the first place. It seems odd to me that we still keep some in place and that we deal with a "helm3" binary in general. However, since it does not seem to break anything obvious I think this change should be fine. WDYT?

PR #6917, opened 2020-03-13
[wip] refactor: Should install binary checks at JX/bin location

### LINKED ISSUE #6912: refactor : ShouldInstallBinary function checking binary on $PATH, not behaving as expected.
### Summary 
ShouldInstallBinary function is currently checking presence of binary on the $PATH variable and returning false if not present there (while it's present on JX/bin). 

### Expected behavior

It should be checking at JX/bin for the binary, if not present should return false.

nehagup:
Signed-off-by: Neha Gupta 

fixes #6912

### DIFF
--- a/pkg/cloud/amazon/cli.go
+++ b/pkg/cloud/amazon/cli.go
@@ -52,7 +52,7 @@ func InstallKops() error {
 		return err
 	}
 	binary := "kops"
-	fileName, flag, err := packages.ShouldInstallBinary(binary)
+	flag, err := packages.ShouldInstallBinary(binary)
 	if err != nil || !flag {
 		return err
 	}
@@ -61,7 +61,7 @@ func InstallKops() error {
 		return err
 	}
 	clientURL := fmt.Sprintf("https://github.com/kubernetes/kops/releases/download/%s/kops-%s-%s", latestVersion, runtime.GOOS, runtime.GOARCH)
-	fullPath := filepath.Join(binDir, fileName)
+	fullPath := filepath.Join(binDir, binary)
 	tmpFile := fullPath + ".tmp"
 	err = packages.DownloadFile(clientURL, tmpFile)
 	if err != nil {
--- a/pkg/cloud/openshift/cli.go
+++ b/pkg/cloud/openshift/cli.go
@@ -22,7 +22,7 @@ func InstallOc() error {
 		return err
 	}
 	binary := "oc"
-	fileName, flag, err := packages.ShouldInstallBinary(binary)
+	flag, err := packages.ShouldInstallBinary(binary)
 	if err != nil || !flag {
 		return err
 	}
@@ -47,7 +47,7 @@ func InstallOc() error {
 		clientURL += fmt.Sprintf("-%s-%s.tar.gz", runtime.GOOS, arch)
 	}
 
-	fullPath := filepath.Join(binDir, fileName)
+	fullPath := filepath.Join(binDir, binary)
 	tarFile := filepath.Join(binDir, "oc.tgz")
 	if extension == ".zip" {
 		tarFile = filepath.Join(binDir, "oc.zip")
@@ -67,7 +67,7 @@ func InstallOc() error {
 		if err != nil {
 			return err
 		}
-		f := filepath.Join(zipDir, fileName)
+		f := filepath.Join(zipDir, binary)
 		exists, err := util.FileExists(f)
 		if err != nil {
 			return err
@@ -81,7 +81,7 @@ func InstallOc() error {
 		}
 		err = os.RemoveAll(zipDir)
 	} else {
-		err = util.UnTargz(tarFile, binDir, []string{binary, fileName})
+		err = util.UnTargz(tarFile, binDir, []string{binary, binary})
 	}
 	if err != nil {
 		return err
--- a/pkg/cmd/opts/install.go
+++ b/pkg/cmd/opts/install.go
@@ -178,7 +178,7 @@ func (o *CommonOptions) InstallBrewIfRequired() error {
 		return nil
 	}
 
-	_, flag, err := packages.ShouldInstallBinary("brew")
+	flag, err := packages.ShouldInstallBinary("brew")
 	if err != nil || !flag {
 		return err
 	}
@@ -194,7 +194,8 @@ func (o *CommonOptions) InstallGlooctl() error {
 	if err != nil {
 		return err
 	}
-	fileName, flag, err := packages.ShouldInstallBinary("glooctl")
+	fileName := "glooctl"
+	flag, err := packages.ShouldInstallBinary("glooctl")
 	if err != nil || !flag {
 		return err
 	}
@@ -381,7 +382,7 @@ func (o *CommonOptions) InstallHelm() error {
 		return err
 	}
 
-	fileName, flag, err := packages.ShouldInstallBinary(binary)
+	flag, err := packages.ShouldInstallBinary(binary)
 	if err != nil || !flag {
 		return err
 	}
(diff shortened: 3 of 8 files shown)


hferentschik:
@nehagup, BDD test setup fails with:

```
Determined that provider is gke from CreateCluster config
running command: ** create cluster gke --project-id=*************3 --no-tiller -m=n1-standard-2 --min-num-nodes=3 --max-num-nodes=5 -z=europe-west1-c --skip-login --cluster-name pr-6917-8-static-gkebdd --batch-mode --labels cluster=static-gkebdd,branch=pr-6917 --git-provider-url https://github.beescloud.com --versions-repo https://github.com/jenkins-x/jenkins-x-versions.git --versions-ref master --git-username **** --git-public=false --git-api-token **************¬ --default-*****-password **************¬
error: run without batch mode or manually install missing dependencies [gcloud kubectl helm]
error: exit status 1
```

Need to look into the code how this comes about.

nehagup:
> @nehagup, BDD test setup fails with:
> 
> ```
> Determined that provider is gke from CreateCluster config
> running command: ** create cluster gke --project-id=*************3 --no-tiller -m=n1-standard-2 --min-num-nodes=3 --max-num-nodes=5 -z=europe-west1-c --skip-login --cluster-name pr-6917-8-static-gkebdd --batch-mode --labels cluster=static-gkebdd,branch=pr-6917 --git-provider-url https://github.beescloud.com --versions-repo https://github.com/jenkins-x/jenkins-x-versions.git --versions-ref master --git-username **** --git-public=false --git-api-token **************¬ --default-*****-password **************¬
> error: run without batch mode or manually install missing dependencies [gcloud kubectl helm]
> error: exit status 1
> ```
> 
> Need to look into the code how this comes about.

It's probably because the installations are being done after checking the presence of respective binaries in the $PATH.

hferentschik:
It's here where it happens - https://github.com/jenkins-x/jx/blob/87a1c11f8b70590c74276f0d445684d75b66ae17/pkg/cmd/opts/install.go#L918

This is called from the various `jx create cluster` commands. Not sure atm what's going on.

nehagup:
> It's here where it happens -
> 
> https://github.com/jenkins-x/jx/blob/87a1c11f8b70590c74276f0d445684d75b66ae17/pkg/cmd/opts/install.go#L918
> 
> This is called from the various `jx create cluster` commands. Not sure atm what's going on.

Hey @hferentschik can anyone help me out debug why this is happening?

hferentschik:
> Hey @hferentschik can anyone help me out debug why this is happening?

I'll contact you next week and we can have a look together.

pkg/cloud/amazon/cli.go
```diff
@@ -61,7 +61,7 @@ func InstallKops() error {
 		return err
 	}
 	clientURL := fmt.Sprintf("https://github.com/kubernetes/kops/releases/download/%s/kops-%s-%s", latestVersion, runtime.GOOS, runtime.GOARCH)
-	fullPath := filepath.Join(binDir, fileName)
+	fullPath := filepath.Join(binDir, binary)
```
hferentschik: I think you first would need to call `BinaryWithExtension(binary)` here. Otherwise, there is a discrepancy between the file we are writing/creating here and the file whose existence we check for in `ShouldInstallBinary`.
This effects Windows. On macOS and Linux this would work.

pkg/cloud/openshift/cli.go
```diff
@@ -22,7 +22,7 @@ func InstallOc() error {
 		return err
 	}
 	binary := "oc"
-	fileName, flag, err := packages.ShouldInstallBinary(binary)
+	flag, err := packages.ShouldInstallBinary(binary)
```
hferentschik: Same as above. Will this work with Windows? The downloaded tarball, does it contain an `oc.exe` or just plain `oc`?

pkg/cmd/opts/install.go
```diff
@@ -178,7 +178,7 @@ func (o *CommonOptions) InstallBrewIfRequired() error {
 		return nil
 	}
 
-	_, flag, err := packages.ShouldInstallBinary("brew")
+	flag, err := packages.ShouldInstallBinary("brew")
```
hferentschik: We cannot do that anymore. Now that we removed the lookup in the global PATH in `ShouldInstallBinary`, we would effectively check whether `brew` is installed in _JX_HOME_ and if not (and it will never be) we would start the brew install which is going to install brew globally.

Really, we should not install _Homebrew_ to start with, but if we want to keep the existing functionality and going in increments, we would need a new helper function _IsInPath_ which checks whether a given binary is on the global PATH. For Homebrew we would have to call this function then opposed to `ShouldInstallBinary`. We need to differentiate between the usecase of checking whether we installed something into JX_HOME or globally. Does that make sense?

hferentschik:
Regarding the integration test failure:

```
Installing gcloud
    install_gitops_integration_test.go:143: 
        	Error Trace:	install_gitops_integration_test.go:143
        	            	error installing gcloud: please install missing gcloud sdk - see https://cloud.google.com/sdk/downloads#interactive
        	            	
        	            	installing cloud provider dependencies
        	            	github.com/jenkins-x/**/pkg/cmd/create.(*InstallOptions).installCloudProviderDependencies
        	            		/workspace/source/pkg/cmd/create/install.go:2126
        	            	github.com/jenkins-x/**/pkg/cmd/create.(*InstallOptions).Run
        	            		/workspace/source/pkg/cmd/create/install.go:586
        	            	github.com/jenkins-x/**/pkg/cmd/create_test.TestInstallGitOps
        	            		/workspace/source/pkg/cmd/create/install_gitops_integration_test.go:142
        	            	testing.tRunner
        	            	runtime.goexit
        	            	installing cloud provider dependencies
        	            	github.com/jenkins-x/**/pkg/cmd/create.(*InstallOptions).Run
        	            		/workspace/source/pkg/cmd/create/install.go:588
        	            	github.com/jenkins-x/**/pkg/cmd/create_test.TestInstallGitOps
        	            		/workspace/source/pkg/cmd/create/install_gitops_integration_test.go:142
        	            	testing.tRunner
        	            	runtime.goexit
        	Test:       	TestInstallGitOps
```

it fails to install gcloud. I guess this becomes dependent on getting #6705 finally in. Maybe we should start there and sort this PR out.

pkg/packages/install.go
```diff
@@ -173,36 +175,31 @@ type InstallOrUpdateBinaryOptions struct {
 }
 
 // ShouldInstallBinary checks if the given binary should be installed
-func ShouldInstallBinary(name string) (fileName string, download bool, err error) {
-	fileName = BinaryWithExtension(name)
-	download = false
-	pgmPath, err := exec.LookPath(fileName)
```
hferentschik: one thing we did not consider is when this runs in cluster. There we know that the binaries are in the right version and we probably want to avoid downloading the binaries each time a pipeline step needs to call an external binary. 

Ideally, these binaries would be installed into _JX_HOME_ in the images as well ,but that's not the case. I am wondering whether we don't have to keep the `exec.LookPath` call but wrap it in a `cluster.IsInCluster()` check. So only if we are in cluster we do a `exec.LookPath` and returns _false_ if the binary can be found on the global path.

When not in cluster, aka on the users machine, I still believe we want to make sure all binaries we need are installed in the version we require into _JX_HOME_.

nehagup: Hey @hferentschik thanks! I won't be able to complete this PR as off now. Can we please close it, I'll work on it (if open) after sometime? 
Apologies for the delay.

hferentschik:
@nehagup, heads up. I am going to close this pull request in favour of #7045  which is a rebased version of this. I believe @daveconde will be making the missing required changes to get this merged.

PR #6921, opened 2020-03-16
fix: lets not fail an update if a depedency asset cannot be found

### LINKED ISSUE #6923: Dependency updates fail if a release asset cannot be located
### Summary 

When performing `jx step create pr versions`, whilst iterating through each item within the versions stream it encounters a candidate where its release asset cannot be located it fails the update completely.

cagiti:
If for some reason a dependency asset is unavailable then the update fails completely, this change enables the update to continue and outputs a warning if a release asset is not available.

fixes #6923


abayer:
@cagiti Looks good, but mind opening an issue? =)

cagiti:
> @cagiti Looks good, but mind opening an issue? =)

sure thing 👍

cagiti:
fixes #6923

PR #6939, opened 2020-03-18
fix: trim suffix from semver-ish version text to avoid OS dependant errors

### LINKED ISSUE #6825: "Unable to get the Jenkins X Version" error when executing "jx diagnose" command on Windows
### Summary 
When executing `jx diagnose` on Windows, I've got the following error:
````
Running in namespace: jx
ERROR: Unable to get the Jenkins X Version
error: Command failed 'jx version --no-version-check': NAME               VERSION
jx                 2.0.1176
Kubernetes cluster v1.15.7
kubectl            v1.14.8
git                2.23.0.windows.1
Operating System   Windows 10 Enterprise 1709 build 16299


verifying packages
error: failed to parse semantic version for current version 2.23.0.windows.1 for package git: Invalid character(s) found in patch number "0.windows.1" exit status 1
````

### Jx version

```
NAME               VERSION
jx                 2.0.1176
Kubernetes cluster v1.15.7
kubectl            v1.14.8
git                2.23.0.windows.1
Operating System   Windows 10 Enterprise 1709 build 16299
```

### Jenkins type

- [ ] Classic Jenkins

### Kubernetes cluster

AKS, manually created

Windows 10 Enterprise 1709 build 16299

### PR DESCRIPTION
**Pull request by hervelemeur**
#### Submitter checklist

#### Description
As Git for Windows doesn't follow semver (as seen for example [here](https://github.com/git-for-windows/git/releases/tag/v2.25.1.windows.1)), the `jx diagnose` command fails, as issued [here](https://github.com/jenkins-x/jx/issues/6825)

This fix trim suffixes from semver-ish version text to avoid errors.

fixes #6825

### DIFF
--- a/pkg/versionstream/version_data.go
+++ b/pkg/versionstream/version_data.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 )
 
@@ -130,7 +131,7 @@ func verifyError(name string, err error) error {
 	return err
 }
 
-// removes any whitespace and `v` prefix from a version string
+// removes any whitespace, `v` prefix and other suffix from a version string
 func convertToVersion(text string) string {
 	answer := strings.TrimSpace(text)
 	answer = strings.TrimPrefix(answer, "v")
@@ -138,6 +139,11 @@ func convertToVersion(text string) string {
 	if len(words) > 1 {
 		answer = words[0]
 	}
+	// Some apps might not exactly follow semver, like for example Git for Windows: 2.23.0.windows.1
+	// we're trimming everything after a semver from the answer
+	// to avoid error described in issue #6825
+	r := regexp.MustCompile(`^([0-9]+\.[0-9]+\.[0-9]+)(.*)`)
+	answer = r.ReplaceAllString(answer, "$1")
 	return answer
 }
 
--- a/pkg/versionstream/version_data_test.go
+++ b/pkg/versionstream/version_data_test.go
@@ -80,6 +80,7 @@ func TestExactPackageVersionRange(t *testing.T) {
 
 	AssertPackageVersion(t, resolver, "git", "2.1.1 (Apple Git-117)", false)
 	AssertPackageVersion(t, resolver, "git", "2.20.1 (Apple Git-117)", true)
+	AssertPackageVersion(t, resolver, "git", "2.23.0.windows.1", true)
 }
 
 func AssertPackageVersion(t *testing.T, resolver *versionstream.VersionResolver, name string, version string, expectedValid bool) {

### CODE REVIEW

File: pkg/versionstream/version_data.go [hunk 1]
```diff
@@ -134,6 +134,12 @@ func verifyError(name string, err error) error {
 func convertToVersion(text string) string {
 	answer := strings.TrimSpace(text)
 	answer = strings.TrimPrefix(answer, "v")
+
+	// Git for Windows doesn't follow semver, example: 2.23.0.windows.1
+	// we're trimming here the ".windows.<number>" suffix from the currentVersion to avoid error described in issue #6825
+	r := regexp.MustCompile(`\.windows\.[0-9]+$`)
```
hferentschik: I find this a bit to specific. _convertToVersion_ is really a generic function to extract a semantic version from the specified text. Adding this special case for how the git version looks seems a bit fishy imo.

What's about we turn this around and look via regular expression at the string _text_ and returns the match if it exists. If we want to keep the current function signature we can return the empty string if there is no match. In this case it will fail later on when we pass the returned string to `semver.Make`. Alternatively, we can change the signature to return the `string, error`.

Something like this might do 
```
regexp.Compile(`v?([0-9]+\.[0-9]+\.[0-9]+)`)
```

hferentschik: I find this a bit to specific. _convertToVersion_ is really a generic function to extract a semantic version from the specified text. Adding this special case for how the git version looks seems a bit fishy imo.

What's about we turn this around and look via regular expression at the string _text_ and returns the match if it exists. If we want to keep the current function signature we can return the empty string if there is no match. In this case it will fail later on when we pass the returned string to `semver.Make`. Alternatively, we can change the signature to return the `string, error`.

Something like this might do 
```
regexp.Compile(`v?([0-9]+\.[0-9]+\.[0-9]+)`)
```

hferentschik: WDYT?

hferentschik: BTW, seems you forgot to import regexp.

hervelemeur: @hferentschik I've pushed a more generic version:
If there is a semver in `text`, then trim everything after it or keep the string untouched if no semver is detected, so it's not "git windows" specific.

Is it what you had in mind?

pkg/versionstream/version_data.go
```diff
@@ -130,14 +131,19 @@ func verifyError(name string, err error) error {
...
 	answer := strings.TrimSpace(text)
 	answer = strings.TrimPrefix(answer, "v")
 	words := strings.Fields(answer)
 	if len(words) > 1 {
 		answer = words[0]
 	}
+	// Some apps might not exactly follow semver, like for example Git for Windows: 2.23.0.windows.1
+	// we're trimming everything after a semver from the answer
+	// to avoid error described in issue #6825
+	r := regexp.MustCompile(`^([0-9]+\.[0-9]+\.[0-9]+)(.*)`)
```
hferentschik: 👍

hervelemeur:
> Looking good. There is still a linting issue:
Thanks, I did'nt found where it was, and even had it fixed but not commited.

> Any chance you can then also squash all commits into a single commit?

No problem, done and pushed!

hferentschik:
@hervelemeur, thanks for the pull request. Much appreciated.

PR #6943, opened 2020-03-18
fix: ignore load jx-requirements.yml validation errors

### LINKED ISSUE #6936: jx boot swallows jx-requirements.yml validation errors
### Summary 

If you have syntax errors in `jx-requirements.yml`, they get swallowed and a generic error gets returned instead. This error obscures what is going wrong.

Make a syntax error in the `jx-requirements.yml` that is still valid YAML.
For example, add an additional field in the `environment.x.ingress.` such as below:

```yaml
- ingress:
    domain: staging.apps.ocp.kearos.net
    externalDNS: false
    namespaceSubDomain: ""
    tls:
      email: ""
      enabled: false
      production: false
  key: staging
  namespace: staging
  repository: env-staging
```

### Expected behavior

`jx boot` returning with an error, but logging the validation mistakes.

For example:

```bash
environments.1: Additional property namespace is not allowed
environments.2: Additional property namespace is not allowed)
```

### Actual behavior

```bash
error: unable to load jx-requirements.yml: jx-requirements.yml file not found
```

The culprit is the function `func LoadRequirementsConfig(dir string) (*RequirementsConfig, string, error) {}` in `config/install_requirements.go`.

### Jx version

```
NAME               VERSION
jx                 2.0.1244
Kubernetes cluster v1.11.0+d4cacc0
kubectl            v1.16.7
git                2.21.0 (Apple Git-122.2)
Operating System   Mac OS X 10.15.3 build 19D76
```

### Jenkins type

- [ ] Classic Jenkins

### Kubernetes cluster

OpenShift 3.11, how it was created -> https://joostvdg.github.io/openshift/rhos311-gcp-minimal/

Mac OS X 10.15.3 build 19D76

nodece:

fixes #6936

### DIFF
--- a/pkg/config/install_requirements.go
+++ b/pkg/config/install_requirements.go
@@ -521,11 +521,19 @@ func LoadRequirementsConfig(dir string) (*RequirementsConfig, string, error) {
 	}
 	for absolute != "" && absolute != "." && absolute != "/" {
 		fileName := filepath.Join(absolute, RequirementsConfigFileName)
-		config, err := LoadRequirementsConfigFile(fileName)
-		if err == nil {
-			return config, fileName, nil
-		}
 		absolute = filepath.Dir(absolute)
+
+		exists, err := util.FileExists(fileName)
+		if err != nil {
+			return nil, "", err
+		}
+
+		if !exists {
+			continue
+		}
+
+		config, err := LoadRequirementsConfigFile(fileName)
+		return config, fileName, err
 	}
 	return nil, "", errors.New("jx-requirements.yml file not found")
 }
--- a/pkg/config/install_requirements_test.go
+++ b/pkg/config/install_requirements_test.go
@@ -408,3 +408,14 @@ func Test_LoadRequirementsConfig(t *testing.T) {
 		})
 	}
 }
+
+func TestLoadRequirementsConfig_load_invalid_yaml(t *testing.T) {
+	testDir := path.Join(testDataDir, "jx-requirements-syntax-error")
+
+	absolute, err := filepath.Abs(testDir)
+	assert.NoError(t, err, "could not find absolute path of dir %s", testDataDir)
+
+	_, _, err = config.LoadRequirementsConfig(testDir)
+	requirementsConfigPath := path.Join(absolute, config.RequirementsConfigFileName)
+	assert.EqualError(t, err, fmt.Sprintf("validation failures in YAML file %s:\nenvironments.0: Additional property namespace is not allowed", requirementsConfigPath))
+}
--- a/pkg/config/test_data/jx-requirements-syntax-error/jx-requirements.yml
+++ b/pkg/config/test_data/jx-requirements-syntax-error/jx-requirements.yml
@@ -1,0 +1,35 @@
+autoUpdate:
+  enabled: true
+  schedule: 0 0 * * *
+bootConfigURL: https://github.com/cloudbees/jenkins-x-boot-config.git
+cluster:
+  environmentGitPublic: true
+  gitKind: github
+  gitName: github
+  gitPublic: true
+  gitServer: https://github.com
+  gke:
+    projectNumber: "1055835833001"
+  namespace: jx
+  provider: gke
+  registry: gcr.io
+  zone: us-east1-c
+environments:
+  - ingress:
+      cloud_dns_secret_name: external-dns-gcp-sa
+      externalDNS: true
+      namespaceSubDomain: -jx.
+      tls:
+        enabled: true
+        production: false
+    key: dev
+    namespace: staging # property namespace is not allowed
+githubApp:
+  enabled: true
+gitops: true
+kaniko: true
+repository: nexus
+secretStorage: vault
+velero:
+  namespace: velero
+webhook: lighthouse


pkg/config/install_requirements.go
```diff
@@ -520,10 +520,10 @@ func LoadRequirementsConfig(dir string) (*RequirementsConfig, string, error) {
 	for absolute != "" && absolute != "." && absolute != "/" {
 		fileName := filepath.Join(absolute, RequirementsConfigFileName)
 		config, err := LoadRequirementsConfigFile(fileName)
-		if err == nil {
-			return config, fileName, nil
+		if err != nil {
+			return nil, "", err
```
hferentschik: I don't think this is quite correct. If you look at the docs for this function, then you see that the intend is to walk up the filesystem to find the requirements file there. I am not saying this is good behaviour, just how it is. The reason for this is to be able to execute individual jx boot step in sub-directories of the requirements/dev repository. 

I am pretty sure that this change will brake this behaviour. 

To do what you are after, I think we need to split or at least pull out the file existence check from `LoadRequirementsConfigFile`. You need to be able to distinguish between the file not existing at all and the file not loadable due to a syntax error.

I think it would be great to add some tests around this as well. Should be quite straight forward, since _install_requirements_test.go_ already exists. One can just add a couple of more test scenarios to the _test_data_ directory. 

WDYT?

nodece: Thank you for your review! I know how to fix it.

nodece: Resolved it.

hferentschik:
@nodece, fyi - the unit test failure:

```
--- FAIL: TestRequirementsConfigMarshalExistingFile (0.02s)
    install_requirements_test.go:62: 
        	Error Trace:	install_requirements_test.go:62
        	            	stat /tmp/test-requirements-config-142991987/subdir/**-requirements.yml: no such file or directory
        	            	checking if file /tmp/test-requirements-config-142991987/subdir/**-requirements.yml exists
        	            	github.com/jenkins-x/**/pkg/config.LoadRequirementsConfigFile
        	            		/workspace/source/pkg/config/install_requirements.go:542
        	            	github.com/jenkins-x/**/pkg/config.LoadRequirementsConfig
        	            		/workspace/source/pkg/config/install_requirements.go:524
        	            	github.com/jenkins-x/**/pkg/config_test.TestRequirementsConfigMarshalExistingFile
        	            		/workspace/source/pkg/config/install_requirements_test.go:61
        	            	testing.tRunner
        	            	runtime.goexit
        	Test:       	TestRequirementsConfigMarshalExistingFile
        	Messages:   	failed to load requirements file in subDir: /tmp/test-requirements-config-142991987/subdir
    install_requirements_test.go:63: 
        	Error Trace:	install_requirements_test.go:63
        	Error:      	unable to find file ""
        	Test:       	TestRequirementsConfigMarshalExistingFile
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x19a pc=0x10b0814]
goroutine 15 [running]:
testing.tRunner.func1(0xc000161300)
	/usr/local/go/src/testing/testing.go:830 +0x392
panic(0x11dcf60, 0x1f712b0)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/jenkins-x/**/pkg/config_test.TestRequirementsConfigMarshalExistingFile(0xc000161300)
	/workspace/source/pkg/config/install_requirements_test.go:65 +0x944
testing.tRunner(0xc000161300, 0x13c0f58)
	/usr/local/go/src/testing/testing.go:865 +0xc0
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:916 +0x35a
````

File: pkg/config/install_requirements.go [hunk 2]
```diff
@@ -525,6 +525,10 @@ func LoadRequirementsConfig(dir string) (*RequirementsConfig, string, error) {
 		if err == nil {
 			return config, fileName, nil
 		}
+		_, isPathErr := err.(*os.PathError)
```
hferentschik: Not sure whether an explicit `util.FileExists` would be better here. It might mean that we check for existence twice, but I don't think that matters. This is really now depending on an implementation detail and there is really no guarantee that `LoadRequirementsConfigFile` is going to pass up the `PathError`. It could for example wrap the error.

nodece: I think `errors.Wrap` is used. We need to `errors.Unwrap`.

nodece: Fixed

pkg/config/install_requirements.go
hferentschik: I still would feel more comfortable if we prior to `config, err := LoadRequirementsConfigFile(fileName)` do an explicit file existence check:

```go
        absolute = filepath.Dir(absolute)
	exists, err := util.FileExists(fileName)
	if err != nil {
		return nil, "", err
	}

        if !exist {
            continue
       }

       config, err := LoadRequirementsConfigFile(fileName)
	if err == nil {
		return nil, "", err
	}
```

Something along these lines. I think that is easier to understand and it does also not rely on an implementation detail of `LoadRequirementsConfigFile`, namely that it does return the PathError directly. It might as well wrap it.

pkg/config/install_requirements_test.go
```diff
@@ -407,4 +407,7 @@ func Test_LoadRequirementsConfig(t *testing.T) {
 			}
 		})
 	}
+
+	_, _, err := config.LoadRequirementsConfig(path.Join(testDataDir, "jx-requirements-syntax-error"))
```
hferentschik: Let's put this test into its own test function and give it a name which tells what we test, something like `Test_LoadRequirementsConfig_load_invalid_yaml`

I also think we should not only assert that an error is returned, but also that is the error we expect. For example by looking at the error message.

pkg/config/test_data/jx-requirements-syntax-error/jx-requirements.yml
```diff
@@ -0,0 +1,35 @@
...
+    namespace: staging # syntax-error: property namespace is not allowed
+githubApp:
+  enabled: true
+gitops: true
+kaniko: true
+repository: nexus
+secretStorage: vault
+velero:
+  namespace: velero
+webhook: lighthouse
```
hferentschik: newlines missing

pkg/config/install_requirements.go
nodece: I feel great!

pkg/config/test_data/jx-requirements-syntax-error/jx-requirements.yml
nodece: Github ignored line 36. actually, I've added newlines.

pkg/config/install_requirements_test.go
```diff
@@ -408,3 +408,14 @@ func Test_LoadRequirementsConfig(t *testing.T) {
...
 }
+
+func TestLoadRequirementsConfig_load_invalid_yaml(t *testing.T) {
+	testDir := path.Join(testDataDir, "jx-requirements-syntax-error")
+
+	absolute, err := filepath.Abs(testDir)
+	assert.NoError(t, err, "could not find absolute path of dir %s", testDataDir)
+
+	_, _, err = config.LoadRequirementsConfig(testDir)
+	requirementsConfigPath := path.Join(absolute, config.RequirementsConfigFileName)
```
hferentschik: 👍

pkg/config/install_requirements.go
```diff
@@ -521,11 +521,19 @@ func LoadRequirementsConfig(dir string) (*RequirementsConfig, string, error) {
...
+		exists, err := util.FileExists(fileName)
+		if err != nil {
+			return nil, "", err
+		}
+
+		if !exists {
+			continue
+		}
+
+		config, err := LoadRequirementsConfigFile(fileName)
```
hferentschik: :+1

hferentschik:
@nodece, thanks for your pull request and patience ;-) Much appreciated.

nodece:
@hferentschik Thanks for your guidance! I am willing to participate with jx in GSoC 2020.  I looking forward to make more contributions.

PR #6958, opened 2020-03-24
fix: check helm version instead of binary for timeout

yelhouti:

#### Submitter checklist

This problem just appeared today be cause of jx-labs updating helm binary (helm) to version 3, and we should check for version instead of binary name

fixes #

### DIFF
--- a/pkg/helm/helm_cli.go
+++ b/pkg/helm/helm_cli.go
@@ -448,7 +448,7 @@ func (h *HelmCLI) UpgradeChart(chart string, releaseName string, ns string, vers
 		args = append(args, "--force")
 	}
 	if timeout != -1 {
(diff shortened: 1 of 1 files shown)


pkg/helm/helm_cli.go
```diff
@@ -448,7 +448,7 @@ func (h *HelmCLI) UpgradeChart(chart string, releaseName string, ns string, vers
 		args = append(args, "--force")
 	}
 	if timeout != -1 {
-		if h.Binary == "helm3" {
+		if h.BinVersion == V3 {
```
hferentschik: Seems ok. Would it make sense to change the other places where we do the check against the binary name of 'helm3' at the same time?

PR #6964, opened 2020-03-24
[wip] fix: Add expiration time for token to UserAuth when using k8s secrets

### LINKED ISSUE #6963: UserAuth should record token expiration time when available if using Kubernetes secrets
Relevant particularly with GitHub Apps - we should be looking at the secret, when using `KubeConfigAuthHandler` at least, for a consistent label with an expiration time on it, and when we see one, storing it so that the `GitProvider` implementations can decide if they need to refresh their stored token.

We should be able to do something similar for `VaultAuthConfigHandler`,  and presumably `ConfigMapVaultConfigHandler` (that'll probably be more complex since it just does secret replacement within a config map generated by `jx boot`, while `VaultAuthConfigHandler` marshals/unmarshals a config map stored in Vault - I'm still a bit confused about all that), by adding an expiration time field to the secrets in Vault, but I'm going to leave that for a followup.

cc @garethjevans @macox @ccojocar

abayer:
#### Submitter checklist

#### Description

This will allow for `GitProvider` implementations to check if the token they're configured with has expired, and if so, reload their auth config. I'll have a follow-up supporting this for GitHub shortly. As mentioned on the issue, we need to eventually support defining expiration time with Vault-stored git auth secrets as well, but I wanted to start with the simplest proof of concept.

fixes #6963

### DIFF
--- a/pkg/auth/kube_config_handler.go
+++ b/pkg/auth/kube_config_handler.go
@@ -3,6 +3,7 @@ package auth
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/jenkins-x/jx/pkg/util"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


hferentschik:
Would love to discuss this a bit more first and get a better understanding of what we are trying to do here.

hferentschik:
@abayer, do you still need this? Can we close this pull request or should this be merged?

abayer:
@hferentschik Yeah, I'm gonna close this. At some point in the non-distant future, I'm going to need to rewrite a lot of the GitHub auth stuff to natively handle GitHub Apps, once we start moving towards having a manifest-based GitHub App for jx in general, but I think the approach used there is going to be different from this.

PR #6966, opened 2020-03-25
chore: remove deprecated get commands

### LINKED ISSUE #6965: Remove deprecated get commands
### Summary 

The CLI command deprecation list [here](https://jenkins-x.io/commands/deprecation/), lists the `get clusters`, `get aws` and `get eks` commands as deprecated. But we still support them within Jenkins X despite passing the deprecation date.

We should remove these deprecated commands.

mrageh:
#### Submitter checklist

#### Description

The CLI command deprecation list [here](https://jenkins-x.io/commands/deprecation/), lists a lot of `get` commands as deprecated. But we still support them within Jenkins X despite passing the deprecation date.

This pull request takes care of removing those deprecated commands from Jenkins X.

fixes #6965


pkg/cmd/get/get_workflow.go
```diff
@@ -1,136 +0,0 @@
...
-		if len(o.Args) > 0 {
-			name = o.Args[0]
-		}
-	}
-
-	if name != "" {
-		return o.getWorkflow(name, jxClient, ns)
-	}
-
-	workflows, err := jxClient.JenkinsV1().Workflows(ns).List(metav1.ListOptions{})
```
hferentschik: I am not sure about this _workflow_ thing. If we remove the get command don't we have to delete everything around it as well, aka the WorkFlow type. The docs says `WorkflowSpec is the specification of the pipeline activity`. 

I guess this ties also into the workflow controller. I am still not sure what that's meant to do and whether this is still used atm in any form.

@abayer ^^

abayer: I literally have no clue with the workflow controller is for - it's not used in Tekton-based installs, though, and I don't see it in BDD tests of static masters, so we probably can get rid of it? @jstrachan or @rawlingsj would know more about what it was intended to be doing than I do.

mrageh: No idea what this `workflow` thing is, I think @rawlingsj tried to remove it from the Jenkins X platform because he thought it's also unused.

mrageh: I removed it from this PR for now, we can come back and review this

mrageh: Spoke to @jstrachan, it's an experiment we introduced in the early days to avoid lots of huge jenkins build pods being around waiting for the promotion to happen. The relevant PR is https://github.com/jenkins-x/jx/pull/1476

PR #6968, opened 2020-03-25
fix: jx upgrade boot failure when duplicate commits

### LINKED ISSUE #6970: jx upgrade boot can fail when duplicate commits in the upgrade
### Summary 

`jx upgrade boot` can fail with the following error if there are duplicate commit's contained within the upgrade of the boot config

```
error: failed to update boot configuration: failed to cherry pick upgrade commits: cherry-picking <commit_id>: git output: The previous cherry-pick is now empty, possibly due to conflict resolution.
If you wish to commit it anyway, use:

    git commit --allow-empty

Otherwise, please use 'git reset'
On branch <branch_name>
You are currently cherry-picking commit <commit_id>.
```

macox:
#### Description

`jx upgrade boot` fails with the following error if there are duplicate commit's 

```
error: failed to update boot configuration: failed to cherry pick upgrade commits: cherry-picking <commit_id>: git output: The previous cherry-pick is now empty, possibly due to conflict resolution.
If you wish to commit it anyway, use:

    git commit --allow-empty

Otherwise, please use 'git reset'
On branch <branch_name>
You are currently cherry-picking commit <commit_id>.
```

fixes #6970

### DIFF
--- a/pkg/gits/git_cli.go
+++ b/pkg/gits/git_cli.go
@@ -1223,6 +1223,11 @@ func (g *GitCLI) CherryPickTheirs(dir string, commitish string) error {
 	return g.gitCmd(dir, "cherry-pick", commitish, "--strategy=recursive", "-X", "theirs")
 }
 
+// CherryPickTheirsKeepRedundantCommits does a git cherry-pick of commit
+func (g *GitCLI) CherryPickTheirsKeepRedundantCommits(dir string, commitish string) error {
+	return g.gitCmd(dir, "cherry-pick", commitish, "--strategy=recursive", "-X", "theirs", "--keep-redundant-commits")
+}
+
 // Describe does a git describe of commitish, optionally adding the abbrev arg if not empty, falling back to just the commit ref if it's untagged
 func (g *GitCLI) Describe(dir string, contains bool, commitish string, abbrev string, fallback bool) (string, string, error) {
 	args := []string{"describe", commitish}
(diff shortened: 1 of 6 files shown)


abayer:
@macox - could you create an issue and do the `fixes #...` dance so we get this in the release notes properly? Thanks!

pkg/gits/git_cli.go
```diff
@@ -1223,6 +1223,11 @@ func (g *GitCLI) CherryPickTheirs(dir string, commitish string) error {
 	return g.gitCmd(dir, "cherry-pick", commitish, "--strategy=recursive", "-X", "theirs")
 }
 
+// CherryPickTheirsKeepRedundantCommits does a git cherry-pick of commit
+func (g *GitCLI) CherryPickTheirsKeepRedundantCommits(dir string, commitish string) error {
```
hferentschik: Would it make sense to have a test case? It might also help to understand what we try to fix here.

hferentschik: or maybe it would be enough to understand:

> jx upgrade boot fails with the following error if there are duplicate commit's

how come there are duplicate commits? How does one get into this state?

hferentschik:
The MockGitter would need updating as well.

pkg/gits/git_cli.go
macox: That is a good question. From what I can see....
A PR is created containing a commit. Another PR is then created containing a commit with the same change as the first. Both PR's are merged.
When cherry-picking the commits the second one throws an error as in the description. (btw. adding --allow-empty to the cherry-pick command didn't work.)

hferentschik: Thanks for the clarification.

macox:
> As mentioned before, I don't think cherry-picking is the right thing to do here in general and I also think that the way we evolve the GitCli interface is not really helpful. 
> 
> However, this is the way things are atm and the change is not making things worth by any means.

Totally agree

PR #6999, opened 2020-04-02
fix: adding terraform context

### LINKED ISSUE #6859: Add build context for bootstrapping and testing via Terraform script
As part of implementing the cloud resource creation enhancement for `jx boot` (https://github.com/jenkins-x/enhancements/tree/master/proposals/3) we need to make sure that cluster creation, Jenkins X install and basic workflows are working with clusters created by the Terraform scripts.

We probably need a similar context to _boot_vault_ with the difference that we create the cluster via Terraform

hferentschik:
fixes #6859


hferentschik:
@daveconde, @abayer - yeah, first successful test run with Terraform, Boot & Vault :-)

PR #7021, opened 2020-04-06
fix: Remove static Jenkins masters

### LINKED ISSUE #7020: Remove Jenkins static masters for April 20, 2020
As discussed at https://jenkins-x.io/blog/2020/03/11/tekton/ etc.

I think the key is making `jx install` default to "next gen" - i.e., gitops/tekton/prow/etc. Then we need to remove Jenkins-specific commands and behavior, which is a messier proposition.

abayer:
#### Submitter checklist

#### Description

First step is making it impossible to install static masters via `jx install`.

fixes #7020


abayer:
The desire here is that the `bdd` context fails. =)

pkg/cmd/create/install.go
```diff
@@ -446,8 +440,14 @@ func (options *InstallOptions) CheckFlags() error {
 		}
 	}
 
+	if options.BatchMode && !flags.NextGeneration && !flags.Tekton && !flags.Prow {
+		flags.Tekton = true
+		flags.Prow = true
+		flags.Kaniko = true
+		options.InitOptions.Flags.NoTiller = true
+	}
```
mrageh, line 448: Is this relevant to removing static jenkins master from `jx`?

mrageh: By this I mean this whole `if` block

abayer: Yup.

pkg/cmd/create/install.go
```diff
@@ -426,16 +425,11 @@ func (options *InstallOptions) CheckFlags() error {
...
 	flags := &options.Flags
 
-	if flags.NextGeneration && flags.StaticJenkins {
-		return fmt.Errorf("incompatible options '--ng' and '--static-jenkins'. Please pick only one of them. We recommend --ng as --static-jenkins is deprecated")
-	}
-
-	if flags.Tekton && flags.StaticJenkins {
-		return fmt.Errorf("incompatible options '--tekton' and '--static-jenkins'. Please pick only one of them. We recommend --tekton as --static-jenkins is deprecated")
+	if flags.StaticJenkins {
+		return fmt.Errorf("option '--static-jenkins' has been removed")
```
hferentschik, line 429: would we still need this? Can we not remove all traces of static Jenkins?

hferentschik:
ooi, how do you hunt down the places which need updating/changing? What are your search criteria?

abayer:
/hold

Holding until Monday

abayer:
dangit, can't get the no-longer-relevant `bdd` context to vanish.

pkg/cmd/create/install.go
hferentschik: Should we not remove the registration of these flags from Cobra. I mean these lines:

```golang
	cmd.Flags().BoolVarP(&flags.Prow, prowFlagName, "", false, "Enable Prow to implement Serverless Jenkins and support ChatOps on Pull Requests")
	cmd.Flags().BoolVarP(&flags.Tekton, tektonFlagName, "", false, "Enables the Tekton pipeline engine (which used to be called knative build pipeline) along with Prow to provide Serverless Jenkins. Otherwise we default to use Knative Build if you enable Prow")
```

We could still have the flags per se if it is not so easy to get rid of them initially, but what is the point of having these CLI flags if in the end we are setting them to true anyways?

pkg/cmd/create/install.go
```diff
@@ -1169,28 +1165,9 @@ func (options *InstallOptions) configureHelmRepo() error {
...
-			//determine which install type is configured
-			jenkinsInstallOption := ServerlessJenkins
-			if options.Flags.StaticJenkins {
-				jenkinsInstallOption = StaticMasterJenkins
-			}
-			log.Logger().Infof(util.QuestionAnswer("Configured Jenkins installation type", jenkinsInstallOption))
-		}
+		//determine which install type is configured
+		jenkinsInstallOption := ServerlessJenkins
+		log.Logger().Infof(util.QuestionAnswer("Configured Jenkins installation type", jenkinsInstallOption))
```
hferentschik, line 1170: This could be a bit misleading as well, since it somehow implies that there are other installation types, but that is not the case, right?

abayer: Yeah, that could be better worded, but I think it's fine.

hferentschik:
Yeah 👍

PR #7026, opened 2020-04-07
chore: remove deprecated workflows

### LINKED ISSUE #6971: Remove deprecated workflows
### Summary 

We introduced workflows as an experiment in https://github.com/jenkins-x/jx/pull/1476  to help us avoid lots of huge jenkins build pods being around waiting for the promotion to happen. 

Since then we've deprecated the workflows pipeline, so we should remove it from JX. 

There's also this PR https://github.com/jenkins-x/jenkins-x-platform/pull/6140, that removes workflows from one of our downstream dependencies `jenkins-x-platform`. That PR needs to be merged when we have a PR that removes workflows from the main `jx` repo.

mrageh:
#### Submitter checklist

#### Description

Workflows is an experimental feature introduced into JX early in it's life. It's something that we need to remove from JX as it's no longer used.

fixes #6971


mrageh:
This PR deletes some legacy unit tests from [`promote_test.go`](https://github.com/jenkins-x/jx/pull/7026/files#diff-410a27e429dfdc26b335511cab42e4e9), the setup for those tests was very brittle and hard to refactor. When I removed the workflows controller from the app, the tests started to fail because they relied on `ControllerWorkflowOptions`.  

These tests tested the promotion logic, which is not really what a unit test is supposed to do. And those test cases are covered by our BDD tests. 

If anyone can think of a reason why we should not remove these tests please let me know

Issue #7027, opened 2020-04-07
No error handling when retrieving logs from Google long term Storage


romainverduci:
### Summary 
Retrieving logs in Google Storage does not catch errors, making it hard to debug when there is a permission access issue or any other kind of problem.

See https://github.com/jenkins-x/jx/blob/master/pkg/cloud/gke/gcloud.go#L558, we only watch stdOut and do not check for exit status code nor stdErr:

```golang
func StreamTransferFileFromBucket(fullBucketURL string) (*bufio.Scanner, error) {
	args := []string{"cp", fullBucketURL, "-"}
	cmd := exec.Command("gsutil", args...)
	stdout, err := cmd.StdoutPipe()
	if err != nil {
		return nil, err
	}
	err = cmd.Start()
	if err != nil {
		return nil, err
	}
	scanner := bufio.NewScanner(stdout)
	scanner.Split(bufio.ScanLines)
	return scanner, err
}
```

hferentschik:
This won't be so easy to fix I think w/o breaking the intended streaming behaviour.
Instead of trying to handle errors implicitly, what's about adding explicit code trying to access the bucket prior to copying it? We could execute a `gsutil -q stat` prior to the cp to check whether the bucket exist and if not return the error. If the bucket exists and is readable, it should be safe to use the current behaviour. You might still get generic IO error, but I don't think this matters for our case.

romainverduci:
@hferentschik that's actually a good idea, I will give it a try. 
Also I was wondering if we really need streaming on that as it's not to get live logs but stored logs. If we don't we can just use cmd.Run() and the error handling would be easier.

PR #7028, opened 2020-04-07
fix: catch errors when retrieving logs from gs bucket

romainverduci:
#### Description
Potential error like bucket not found or anything was not catched as we were only streaming command stdOut. It was hard to debug when something is going wrong.

As we don't really want to stream logs as they are stored in a bucket, I'm wondering if we could just use cmd.Run() or wait for the command to finish and just return the output.
I would make the error handling better.

fixes https://github.com/jenkins-x/jx/issues/7027

### DIFF
--- a/pkg/cloud/gke/gcloud.go
+++ b/pkg/cloud/gke/gcloud.go
@@ -556,19 +556,36 @@ func (g *GCloud) DeleteAllObjectsInBucket(bucketName string) error {
 // StreamTransferFileFromBucket will perform a stream transfer from the GCS bucket to stdout and return a scanner
 // with the piped result
 func StreamTransferFileFromBucket(fullBucketURL string) (*bufio.Scanner, error) {
+	bucketAccessible, err := isBucketAccessible(fullBucketURL)
+	if !bucketAccessible || err != nil {
+		return nil, errors.Wrap(err, "can't access bucket")
+	}
+
 	args := []string{"cp", fullBucketURL, "-"}
 	cmd := exec.Command("gsutil", args...)
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "can't get cmd stdout")
 	}
 	err = cmd.Start()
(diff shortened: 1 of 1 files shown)


pkg/cloud/gke/gcloud.go
```diff
@@ -562,10 +563,31 @@ func StreamTransferFileFromBucket(fullBucketURL string) (*bufio.Scanner, error)
...
+
+	if err = cmd.Start(); err != nil {
+		return nil, err
+	}
+
+	if stderr != nil {
+		scanner := bufio.NewScanner(stderr)
+		scanner.Split(bufio.ScanLines)
+		var errorStream bytes.Buffer
+		for scanner.Scan() {
```
hferentschik, line 580: I don't think this is correct. So without any further doing, I would expect that you call `cmd.Start` which will start the `gsutil` command without waiting. Originally the idea is to then directly return the Scanner and let the caller of the method deal with scanning/processig/streaming the stdout. 

With your change,  I would expect that after `Start` we are not effectively waiting here until EOF in the error stream at which point you break out of this loop, check the stderr and act accordingly. This means effectively that you are back to just synchronously executing `gsutil`. You can have this easier via `cmd.Run`. 

Have you tested this? 

I'd try to split this function into two, a `StreamTransferFileFromBucket` and a `streamCmd` (or something). The idea is to make the streaming (stdout/stderr) handling testable and independent from `gsutil`.

hferentschik: If you want to make this a synchronous function, I'd rename `StreamTransferFileFromBucket` to `TransferFileFromBucket` and return the full log as string or[]byte or whatever. Let the caller wrap a `Scanner` around it.

hferentschik: @dgozalo - you wrote a lot of this initial logging code. What's your take on this?

hferentschik: @romainverduci see also https://github.com/jenkins-x/jx/issues/7027#issuecomment-610522671

PR #7031, opened 2020-04-07
fix: go.mod replace for spf13/cobra

abayer:
#### Submitter checklist

#### Description

github.com/antham/chyle v1.4.0 has a lousy dependency on spf13/cobra that we need to replace.

n/a


go.mod
```diff
@@ -167,3 +167,5 @@ replace github.com/Azure/azure-sdk-for-go => github.com/Azure/azure-sdk-for-go v
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v10.14.0+incompatible
 
 replace github.com/banzaicloud/bank-vaults => github.com/banzaicloud/bank-vaults v0.0.0-20190508130850-5673d28c46bd
+
+replace github.com/spf13/cobra => github.com/spf13/cobra v0.0.5
```
hferentschik, line 171: we still are going to need this, right? That's for when we want to go Go 1.13?

PR #7036, opened 2020-04-10
[wip] feat: install helm2 as a binary plugin

daveconde:
Install helm2 as a binary plugin to avoid issues of incorrect versions on user's paths

### DIFF
--- a/pkg/cmd/clients/factory.go
+++ b/pkg/cmd/clients/factory.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/heptio/sonobuoy/pkg/client"
 	sonoboy_dynamic "github.com/heptio/sonobuoy/pkg/dynamic"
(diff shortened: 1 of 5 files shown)


hferentschik:
> Install helm2 as a binary plugin to avoid issues of incorrect versions on user's paths

Is this really such a good idea? Should we not rather try to fix how we install binaries into JX_HOME/bin? I have a hard time understanding the reasoning behind this.

We can avoid version issues, by not executing against binaries on the PATH, but under a fixed location, aka _JX_HOME/bin_. 

I am also concerned that this is yet introducing another way of doing things.

hferentschik:
@daveconde - have you looked at https://github.com/jenkins-x/jx/pull/6917? What's the plan with the other binaries we are installing?

hferentschik:
@daveconde, I am going to close this PR for now. AFAIU, we are going for the install into _JX_HOME/bin_.

Issue #6484, opened 2020-01-10
Update the Go version to 1.13.x (1.13.5) in the build and builders


ccojocar:
### Summary 
It seems that jx builds now with Go version `1.13.5`. We should update the Go version in the build and all builders.

```
$ake build
CGO_ENABLED=0 GO111MODULE=on go build -ldflags " -X github.com/jenkins-x/jx/pkg/version.Version=2.0.399-dev+d8d557a82 -X github.com/jenkins-x/jx/pkg/version.Revision='d8d557a82' -X github.com/jenkins-x/jx/pkg/version.Branch='master' -X github.com/jenkins-x/jx/pkg/version.BuildDate='20200110-10:48:49' -X github.com/jenkins-x/jx/pkg/version.GoVersion='1.13.5' """ -o build/jx cmd/jx/jx.go

$ build/jx version
NAME               VERSION
jx                 2.0.399-dev+d8d557a82
Kubernetes cluster v1.13.12-gke.8
kubectl            v1.16.3
helm client        Client: v2.16.1+gbbdfe5e
git                2.21.0 (Apple Git-122.2)
Operating System   Mac OS X 10.15.2 build 19C57


verifying packages

WARNING: jx version 2.0.721 is available in the version stream. We highly recommend you upgrade to it.

? Would you like to upgrade to the jx version? No

WARNING: could not find a stable package version for git from /Users/cosmin/.jx/jenkins-x-versions
For background see: https://jenkins-x.io/docs/concepts/version-stream/
Please lock this version down via the command: jx step create pr versions -k package -n git
```
### Jx version

The output of `jx version` is:

```
COPY OUTPUT HERE
```

### Jenkins type

- [x ] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

deanesmith:
Update builders, then update upstream dependencies

deanesmith:
Updated priority due to vulnerabilities.

PR #7038, opened 2020-04-13
chore: Bump antham/chyle to 1.6.0

abayer:
#### Submitter checklist

#### Description

This is part of #6484's prep work - antham/chyle 1.4.0 (and 1.5.0) depend on an arbitrary commit in spf13/cobra, and that ends up resulting in breaking dependency resolution in go 1.13. I initially worked around this by adding an explicit `require` for spf13/cobra 0.0.5, but antham/chyle 1.6.0 passes all our tests and depends on spf13/cobra 0.0.5 explicitly, so it's a better fix.

I also did a `go mod tidy` while I was here.

n/a


hferentschik:
Did you run a `make generate` after `go mod tidy`? Otherwise, you might get a surprise that _go.mod_ changes during the generate task (long story). There is a `make tidy-deps` which takes care of both.

abayer:
Oops, lemme do that.

abayer:
@hferentschik Ok, ran `make tidy-deps` - let's see how this does.

PR #7041, opened 2020-04-14
chore: remove deprecated `jx console` cmd

### LINKED ISSUE #6928: Remove deprecated console command
### Summary 

The CLI command deprecation list [here](https://jenkins-x.io/commands/deprecation/), lists the `jx console` command as deprecated. But we still support it within Jenkins X despite passing the deprecation date.

We should deprecate remove this command as it's been superseded by `jx ui`.

mrageh:
#### Submitter checklist

#### Description
Remove deprecated `jx console` command.

fixes #6928


hferentschik:
@mrageh - have you removed too much here:

```
pkg/cmd/open.go:72:25: o.ConsoleOptions.Open undefined (type ConsoleOptions has no field or method Open)
make: *** [Makefile:90: build] Error 2
```

mrageh:
Forgot to push up my rebase

abayer:
This should wait to be removed until we remove static masters.

hferentschik:
Since I just came across this. Here is something in jx-docs which would need updating as part of this - https://jenkins-x.io/docs/reference/components/docker-registry/. 

I am sure there is more, also related to the other commands we already removed. @mrageh, it might be worth going through the docs and searching for the commands which we previously removed. If any of the removed commands is used somewhere, we should update the docs.

@mrageh, probably also a good idea to update _deprecated.go_ and remove the deprecation warnings for the commands we already removed.

mrageh:
@hferentschik my plan is to do a full update of the docs and remove everything that's been deprecated.

PR #7042, opened 2020-04-14
chore: remove deprecated step commands

### LINKED ISSUE #7040: Remove deprecated step commands
### Summary 

The CLI command deprecation list [here](https://jenkins-x.io/commands/deprecation/), lists `step create Jenkins config`, `step pre build` and `step pre extend`.

We should remove these deprecated commands.

mrageh:
#### Submitter checklist

#### Description
Remove deprecated `jx step` commands, the full list of deprecated commands can be found [here](https://jenkins-x.io/commands/deprecation/).

fixes #7040


abayer:
`jx step post build` is [still used extensively](https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes/blob/master/packs/maven/pipeline.yaml#L15) in build packs, so it shouldn't be removed. I'm not sure how `jx step post install` is used, but I'm a little nervous about it.

garethjevans:
@mrageh AFAIK, the nexus commands are quite useful for anyone with a Java based workflow - we don't tend to use them as there is no Java in JX anymore but they might be useful for the community - @rawlingsj what are your thoughts on this?

hferentschik:
Sounds like we are getting a bit ahead of ourselves here then. Can we split this into a set of commands which are safe to delete and others we want to keep for a bit longer?

mrageh:
@garethjevans why did we add the `nexus` commands to the deprecation list, if we still use them?

rawlingsj:
> @mrageh AFAIK, the nexus commands are quite useful for anyone with a Java based workflow - we don't tend to use them as there is no Java in JX anymore but they might be useful for the community - @rawlingsj what are your thoughts on this?

Yeah I agree with @garethjevans these can still be useful for community folks, plus you never know we might need to release java libraries again in the future.  I'd keep these.

mrageh:
@rawlingsj why did we add the `nexus` command to the deprecation list, it's been there for months. So there's some sort of breakdown in communication.

mrageh:
Both `step nexus` and `step post build` commands have been restored

PR #7043, opened 2020-04-14
fix: Actually use a specific version of codegen dependencies

abayer:
#### Submitter checklist

#### Description

Because we weren't using `GO111MODULE=on`, we were actually running `go get -u -d k8s.io/kube-openapi/cmd/openapi-gen/...`, without the specific version (and the same for the `k8s.io/code-generator` ones). And due to https://github.com/kubernetes/gengo/pull/173 percolating through the
stack, the "latest" now won't build for us because it can't find `k8s.io/klog/v2`. I have no idea why that is, but yeah. So let's switch to always using `GO111MODULE=on` for codegen.

n/a


go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/antham/chyle v1.6.0
 	github.com/aws/aws-sdk-go v1.24.0
 	github.com/awslabs/goformation v0.0.0-20190320125420-ac0a17860cf1
-	github.com/banzaicloud/bank-vaults v0.0.0-20190508130850-5673d28c46bd
+	github.com/banzaicloud/bank-vaults v0.0.0-00010101000000-000000000000
```
hferentschik, line 26: I guess that's ok since we still have the explicit replace, right?

abayer: Yeah, I don't know why this decided to happen, but it did. The replace does seem to do the trick.

cmd/codegen/generator/open_api.go
```diff
@@ -166,7 +166,7 @@ type schemaWriterTemplateData struct {
 // InstallOpenApiGen installs the openapi-gen tool from the github.com/kubernetes/kube-openapi repository.
 func InstallOpenApiGen(version string, gopath string) error {
 	util.AppLogger().Infof("installing %s with version %s via 'go get' to %s", openApiGen, version, gopath)
-	err := util.GoGet(openApiGen, version, gopath, false, false, true)
+	err := util.GoGet(openApiGen, version, gopath, true, false, true)
```
hferentschik, line 169: @pmuir - you remember https://github.com/jenkins-x/jx/pull/4359/files? This change from @abayer is going to rewind this. Can you remember the details? I am trying to remember why we disabled modules again and what side effects we might get with this PR.

hferentschik:
@abayer that seems to work. I would be ok with it. WDYT?

abayer:
@hferentschik - fine by me.

PR #7045, opened 2020-04-14
[wip] refactor: Should install binary checks at JX/bin location

daveconde:
Rebased @nehagup changes on master


pkg/packages/install.go
```diff
@@ -173,36 +175,31 @@ type InstallOrUpdateBinaryOptions struct {
 }
 
 // ShouldInstallBinary checks if the given binary should be installed
-func ShouldInstallBinary(name string) (fileName string, download bool, err error) {
-	fileName = BinaryWithExtension(name)
-	download = false
-	pgmPath, err := exec.LookPath(fileName)
```
hferentschik, line 179: this needs to happen conditionally depending on whether this is running in cluster or not. Otherwise we get the build image downloading third-party dependencies. In a follow up step we can try to improve on this.

pkg/cmd/opts/install.go
```diff
@@ -170,7 +170,7 @@ func (o *CommonOptions) InstallBrewIfRequired() error {
 		return nil
 	}
 
-	_, flag, err := packages.ShouldInstallBinary("brew")
+	flag, err := packages.ShouldInstallBinary("brew")
```
hferentschik, line 173: We cannot do that anymore. Now that we removed the lookup in the global PATH in ShouldInstallBinary, we would effectively check whether brew is installed in JX_HOME and if not (and it will never be) we would start the brew install which is going to install brew globally.

Really, we should not install Homebrew to start with, but if we want to keep the existing functionality and going in increments, we would need a new helper function IsInPath which checks whether a given binary is on the global PATH. For Homebrew we would have to call this function then opposed to ShouldInstallBinary. We need to differentiate between the usecase of checking whether we installed something into JX_HOME or globally

PR #7062, opened 2020-04-16
fix: removing GitHub check verifyPrivateRepos during boot

### LINKED ISSUE #7059: Remove warning about need for paid subscriptions for GitHub orgs
When running `jx boot` you get asked:

```
If 'xyz' is an GitHub organisation it needs to have a paid subscription to create private repos. Do you wish to continue? Yes
```

This was to work around the fact that one was not able to create private repos for organisations on GitHub. This has changed and now even free organisations can have private repositories. Now we can remove this additional check.

hferentschik:
- Even free Github organisations can now create private repositories

fixes #7059


pkg/cmd/step/verify/step_verify_preinstall.go
```diff
@@ -760,25 +755,6 @@ func (o *StepVerifyPreInstallOptions) gatherGitRequirements(requirements *config
 	return nil
 }
 
-func (o *StepVerifyPreInstallOptions) verifyPrivateRepos(requirements *config.RequirementsConfig) error {
-	log.Logger().Infof("Environment repos will be %s, if you want to create %s environment repos, please set %s to %s in jx-requirements.yml", util.ColorInfo("private"), util.ColorInfo("public"), util.ColorInfo("environmentGitPublic"), util.ColorInfo("true"))
```
abayer, line 764: I feel like we should still be logging this in the else above.

hferentschik: ok, will do

PR #7065, opened 2020-04-16
feat: don't register the CRDs on controllers the installations take care of that already

### LINKED ISSUE #7058: Controllerbuild and Controllerrole register several CRDs when starting up
### Summary 
When we run both the `controllerrole` and `controllerbuild`, we are registering several CRDs.

This is also done previously when booting so it's not necessary to run it again. It also causes a problem when running a cluster with limited permissions.

Remove the cluster-wide permissions from the `controllerbuild` and `controllerrole` Service Accounts and run the controllers.

### Expected behavior
They should not attempt to register any CRD when running in Openshift as we are already doing that when installing Jenkins X.

### Actual behavior
It attempts to register the CRD.

### Jenkins type

- [ ] Classic Jenkins

dgozalo:
#### Submitter checklist

#### Description
It doesn't make sense to register CRDs again when we are running the controllers in a boot cluster as there's already a step that updates the CRDs in the Boot Pipeline.

This is making the controller fail when running in Openshift because the `ServiceAccounts` don't and shouldn't have any cluster wide permissions.

fixes #7058

### DIFF
--- a/pkg/cmd/controller/controller.go
+++ b/pkg/cmd/controller/controller.go
@@ -3,10 +3,9 @@ package controller
 import (
 	"github.com/jenkins-x/jx/pkg/cmd/controller/pipeline"
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
-	"github.com/spf13/cobra"
-
 	"github.com/jenkins-x/jx/pkg/cmd/opts"
 	"github.com/jenkins-x/jx/pkg/cmd/templates"
+	"github.com/spf13/cobra"
 )
 
 // ControllerOptions contains the CLI options
--- a/pkg/cmd/controller/controller_build.go
+++ b/pkg/cmd/controller/controller_build.go
@@ -147,15 +147,6 @@ func NewCmdControllerBuild(commonOpts *opts.CommonOptions) *cobra.Command {
 
 // Run implements this command
 func (o *ControllerBuildOptions) Run() error {
-	apisClient, err := o.ApiExtensionsClient()
-	if err != nil {
-		return err
-	}
-	err = kube.RegisterPipelineActivityCRD(apisClient)
-	if err != nil {
-		return err
-	}
-
 	jxClient, devNs, err := o.JXClientAndDevNamespace()
 	if err != nil {
 		return err
--- a/pkg/cmd/controller/controller_role.go
+++ b/pkg/cmd/controller/controller_role.go
@@ -89,19 +89,6 @@ func NewCmdControllerRole(commonOpts *opts.CommonOptions) *cobra.Command {
 }
 
 func (o *ControllerRoleOptions) Run() error {
-	apiClient, err := o.ApiExtensionsClient()
-	if err != nil {
-		return err
-	}
-	err = kube.RegisterEnvironmentCRD(apiClient)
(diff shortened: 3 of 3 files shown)


pkg/cmd/controller/controller.go
```diff
@@ -58,3 +63,17 @@ func NewCmdController(commonOpts *opts.CommonOptions) *cobra.Command {
...
 	return o.Cmd.Help()
 }
+
+func (o *ControllerOptions) shouldRegisterCRD(jxClient versioned.Interface, devNs string) (bool, error) {
+	devEnv, err := jxClient.JenkinsV1().Environments(devNs).Get("dev", metav1.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+
+	requirements, err := config2.GetRequirementsConfigFromTeamSettings(&devEnv.Spec.TeamSettings)
```
hferentschik, line 73: What's about we add an explicit `IsBootCluster` to `config2` or another appropriate place? It took me a while to figure out the intention here since there is the implicit assumption that if there is no requirements file in the team settings, we are dealing with a non-boot cluster.

pkg/cmd/controller/controller_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/jenkins-x/jx/pkg/client/clientset/versioned/fake"
+	"github.com/jenkins-x/jx/pkg/util"
+
+	"github.com/stretchr/testify/assert"
+
+	clients_test "github.com/jenkins-x/jx/pkg/cmd/clients/mocks"
+	"github.com/jenkins-x/jx/pkg/cmd/opts"
+)
+
+func TestShouldRegisterCRD_WithRequirementsFound(t *testing.T) {
```
hferentschik, line 23: 👍 nice test

pkg/cmd/controller/controller.go
dgozalo: The problem with that function is that it wouldn't be deterministic depending on the scope where it's called.

For example if it's called at the beginning of boot, it will return false, but in the scope of a controller, it will always be deterministic.

hferentschik: You could make it a function of this controller then? If it were not for the title of the pull request I would not know what this could is about. More so since as you just pointed out there are further constraints. 

Is this also something we need only for clusters created by `jx create cluster`? If so, is it worth adding a comment that this is maybe temporary or why we don't want to do this in a boot cluster?

dgozalo: Whether it was created with `jx create cluster` is irrelevant. When we boot a cluster, que add the requirements to the TeamSettings at some point in the process, this is what I'm looking for.

This while change comes because registering CRDs on a controller in a boot cluster doesn't make sense but I'm still not sure about a `jx install` cluster.

Maybe we could remove it altogether?

hferentschik: > Maybe we could remove it altogether?

That would be the preferred option, imo. Have you tried?

dgozalo:
> @dgozalo, do you mind adjusting the commit message? Now we don't register CRDs anymore and there is no check for boot.
> 
> Otherwise looking good.

Done.

PR #7068, opened 2020-04-17
feat: list pods instead of namespaces to avoid needing a cluster-wide permission

### LINKED ISSUE #7060: Modify the way the code checks that the user is connected to a cluster
### Summary 
Right now, the code attempts to list all the namespaces to check that the user is connected to a cluster.

This is not possible when you booting with limited permissions as this needs a cluster wide permission.

In Openshift, the user that will be booting could have very limited permissions and not be able to list namespaces other than the ones created by Jenkins X.

Create an Openshift cluster, create an user with very limited permissions (no cluster-wide permissions) and attempt to run `jx boot`.

### Expected behavior
It should correctly check that the user is connected to an Openshift cluster.

### Actual behavior
It fails because the user doesn't have any cluster-wide permissions can't and list namespaces at a cluster scope.

### Jenkins type

- [ ] Classic Jenkins

dgozalo:
#### Submitter checklist

#### Description
Listing namespaces requires a `clusterrole`. It will fail for clusters with tight permission restrictions, so we should list a resource that is local to a namespace.

fixes #7060

### DIFF
--- a/pkg/cmd/boot/boot.go
+++ b/pkg/cmd/boot/boot.go
@@ -567,11 +567,10 @@ func (o *BootOptions) verifyRequirements(requirements *config.RequirementsConfig
 }
 
 func (o *BootOptions) verifyClusterConnection() error {
-	client, err := o.KubeClient()
+	client, ns, err := o.KubeClientAndNamespace()
 	if err == nil {
-		_, err = client.CoreV1().Namespaces().List(metav1.ListOptions{})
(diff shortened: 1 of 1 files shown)


hferentschik:
@dgozalo - what's the state on this one? Is this still wip?

dgozalo:
> @dgozalo - what's the state on this one? Is this still wip?

It should be ready to go now.

hferentschik:
> It should be ready to go now.

@dgozalo do you want to take the 'wip' off?

PR #7073, opened 2020-04-17
feat: create a function to check the current permissions and split the deletion of old resources between cluster and non cluster resources

### LINKED ISSUE #7061: When installing a helm chart, Jenkins X attempts to delete stale cluster-wide resources even with limited permissions
### Summary 
When executing the code to install a Helm chart, Jenkins X attempts to delete any stale resources that have a new version.

This also applies to `clusterrole`and `clusterrolebinding` even if the booting user doesn't have any cluster-wide permissions and can't interact with these resources, making the booting process fail.

Attempt to install a Helm chart using `jx step helm apply` with an user that has no cluster-wide permissions.

### Expected behavior
It should avoid to interact with any cluster-wide resources or at least just warn that it couldn't delete them.

### Actual behavior
It fails because the user doesn't have enough permissions.

### Jenkins type

- [ ] Classic Jenkins

dgozalo:
#### Submitter checklist

#### Description
When adding a Helm chart, we have two "hooks" that will clean any old resources.

One of them attempt to clean cluster wide resources too, for which we may have no permissions to delete.

I'm modifying this to warn on failure instead of erring.

fixes #7061

### DIFF
--- a/pkg/helm/helm_template.go
+++ b/pkg/helm/helm_template.go
@@ -500,23 +500,33 @@ func (h *HelmTemplate) kubectlDeleteFile(ns string, file string) error {
 
 func (h *HelmTemplate) deleteOldResources(ns string, releaseName string, versionText string, wait bool) error {
 	selector := LabelReleaseName + "=" + releaseName + "," + LabelReleaseChartVersion + "!=" + versionText
-	return h.deleteResourcesAndClusterResourcesBySelector(ns, selector, wait, "older releases")
+	err := h.deleteNamespacedResourcesBySelector(ns, selector, wait, "older releases")
+	if err != nil {
+		return err
+	}
+	return h.deleteClusterResourcesBySelector(ns, selector, wait, "older releases")
 }
 
-func (h *HelmTemplate) deleteResourcesAndClusterResourcesBySelector(ns string, selector string, wait bool, message string) error {
-	kinds := []string{"all", "pvc", "configmap", "release", "sa", "role", "rolebinding", "secret"}
-	clusterKinds := []string{"clusterrole", "clusterrolebinding"}
-
+func (h *HelmTemplate) deleteNamespacedResourcesBySelector(ns string, selector string, wait bool, message string) error {
+	kinds := []string{"pvc", "configmap", "release", "sa", "role", "rolebinding", "secret"}
 	errList := []error{}
-
 	log.Logger().Debugf("Removing Kubernetes resources from %s using selector: %s from %s", message, util.ColorInfo(selector), strings.Join(kinds, " "))
 	errs := h.deleteResourcesBySelector(ns, kinds, selector, wait)
 	errList = append(errList, errs...)
+	return util.CombineErrors(errList...)
+}
 
-	selector += "," + LabelNamespace + "=" + ns
-	log.Logger().Debugf("Removing Kubernetes resources from %s using selector: %s from %s", message, util.ColorInfo(selector), strings.Join(clusterKinds, " "))
-	errs = h.deleteResourcesBySelector("", clusterKinds, selector, wait)
+func (h *HelmTemplate) deleteClusterResourcesBySelector(ns string, selector string, wait bool, message string) error {
+	clusterKinds := []string{"all", "clusterrole", "clusterrolebinding"}
+	errList := []error{}
+	hasPermissions, errs := kube.CanI(h.KubeClient, kube.Delete, kube.All)
 	errList = append(errList, errs...)
+	if hasPermissions {
+		selector += "," + LabelNamespace + "=" + ns
+		log.Logger().Debugf("Removing Kubernetes resources from %s using selector: %s from %s", message, util.ColorInfo(selector), strings.Join(clusterKinds, " "))
+		errs = h.deleteResourcesBySelector("", clusterKinds, selector, wait)
+		errList = append(errList, errs...)
+	}
 	return util.CombineErrors(errList...)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/helm/helm_template.go
```diff
@@ -532,7 +532,11 @@ func (h *HelmTemplate) deleteResourcesBySelector(ns string, kinds []string, sele
 		}
 		output, err := h.runKubectlWithOutput(args...)
 		if err != nil {
-			errList = append(errList, err)
+			if warn {
```
hferentschik, line 535: How dod you know that this error is caused by permission and not by any other error? Would you not have to inspect the error? 

I am not a big fan of this indirect coding. Would it not make more sense to have a function which checks whether you have cluster-wide permissions? I can see this being valuable for other use cases as well.

pkg/helm/helm_template.go
```diff
@@ -510,17 +510,17 @@ func (h *HelmTemplate) deleteResourcesAndClusterResourcesBySelector(ns string, s
...
 	selector += "," + LabelNamespace + "=" + ns
 	log.Logger().Debugf("Removing Kubernetes resources from %s using selector: %s from %s", message, util.ColorInfo(selector), strings.Join(clusterKinds, " "))
-	errs = h.deleteResourcesBySelector("", clusterKinds, selector, wait)
+	errs = h.deleteResourcesBySelector("", clusterKinds, selector, wait, true)
 	errList = append(errList, errs...)
 	return util.CombineErrors(errList...)
 }
 
-func (h *HelmTemplate) deleteResourcesBySelector(ns string, kinds []string, selector string, wait bool) []error {
+func (h *HelmTemplate) deleteResourcesBySelector(ns string, kinds []string, selector string, wait bool, warn bool) []error {
```
hferentschik, line 523: I think the real intention here is _ignoreError_. _warn_ is imo misleading.

pkg/helm/helm_template.go
```diff
@@ -510,17 +510,17 @@ func (h *HelmTemplate) deleteResourcesAndClusterResourcesBySelector(ns string, s
 	errList := []error{}
 
 	log.Logger().Debugf("Removing Kubernetes resources from %s using selector: %s from %s", message, util.ColorInfo(selector), strings.Join(kinds, " "))
```
hferentschik, line 512: I'd split _deleteResourcesAndClusterResourcesBySelector_ into two functions.

pkg/helm/helm_template.go
```diff
@@ -500,23 +500,35 @@ func (h *HelmTemplate) kubectlDeleteFile(ns string, file string) error {
 
 func (h *HelmTemplate) deleteOldResources(ns string, releaseName string, versionText string, wait bool) error {
 	selector := LabelReleaseName + "=" + releaseName + "," + LabelReleaseChartVersion + "!=" + versionText
-	return h.deleteResourcesAndClusterResourcesBySelector(ns, selector, wait, "older releases")
+	err := h.deleteNamespacedResourcesBySelector(ns, selector, wait, "older releases")
```
hferentschik, line 503: +1 , much nicer by splitting this function imo

pkg/helm/helm_template.go
```diff
@@ -500,23 +500,35 @@ func (h *HelmTemplate) kubectlDeleteFile(ns string, file string) error {
...
+	return util.CombineErrors(errList...)
+}
 
-	selector += "," + LabelNamespace + "=" + ns
-	log.Logger().Debugf("Removing Kubernetes resources from %s using selector: %s from %s", message, util.ColorInfo(selector), strings.Join(clusterKinds, " "))
-	errs = h.deleteResourcesBySelector("", clusterKinds, selector, wait)
+func (h *HelmTemplate) deleteClusterResourcesBySelector(ns string, selector string, wait bool, message string) error {
+	clusterKinds := []string{"all", "clusterrole", "clusterrolebinding"}
+	errList := []error{}
+	hasPermissions, errs := kube.CanI(kube.Delete, kube.All)
```
hferentschik, line 524: +1, nice one. This function will be useful in the future as well.

pkg/kube/permissions.go
```diff
@@ -0,0 +1,62 @@
...
+	// Watch represents the watch verb
+	Watch Verb = "watch"
+)
+
+// CanI will take a verb and a list of resources and it will check whether the current user / service account can
+// perform that verb against the resources in the Kubernetes cluster
+func CanI(verb Verb, resources ...Resource) (bool, []error) {
+	errList := []error{}
+	for _, resource := range resources {
+		c := exec.Command("kubectl", "auth", "can-i", "--all-namespaces", string(verb), string(resource)) // #nosec
```
hferentschik, line 44: There is no way to do this with the REST client? 
If we need to use _kubectl_, I think we should use the '-q' option and then just look at the exit status to determine whether an action is allowed. Checking against a string is fragile since there is no guarantee that is won't change and even more it won't work with an internationalized version of `kubectl` (if there is one).

hferentschik: Looking at https://kubernetes.io/docs/reference/access-authn-authz/authorization/, it seems that under the hood of `kubectl auth can-i` is the SelfSubjectAccessReview API. For this there are generated client-go functions I believe. One probably could do this via the REST API. WDYT?

pkg/kube/permissions.go
```diff
@@ -0,0 +1,65 @@
...
+	List Verb = "list"
+	// Update represents the update verb
+	Update Verb = "use"
+	// Watch represents the watch verb
+	Watch Verb = "watch"
+)
+
+// CanI will take a verb and a list of resources and it will check whether the current user / service account can
+// perform that verb against the resources in the Kubernetes cluster
+func CanI(kubeClient kubernetes.Interface, verb Verb, resources ...Resource) (bool, []error) {
```
hferentschik, line 40: 👍 that's awesome. So the result is in the returned object already. So sweet.

PR #6919, opened 2020-03-14
fix(sync): use commit sha instead of tag for ksync image

ankitm123:
Signed-off-by: ankitm123 <ankitmohapatra123@gmail.com>

#### Submitter checklist

- [x] Change is code complete and matches issue description.
- [x] Change is covered by existing or new tests.

#### Description
* use git commit as tags for ksync docker image
* refactor InstallSync to add tests
* improve error handling

#### Special notes for the reviewer(s)
/cc @hferentschik @abayer 
* If ksync is installed, jx will try to parse the git commit from the output of the command `ksync version`. 
* If ksync is not installed, then jx will get the commit sha for the latest tag/release for ksync.
It will use this commit sha (short sha of the first 7 characters) to pull the right ksync docker image 
#### Which issue this PR fixes
fixes #6262 and #6123

PR #7075, opened 2020-04-17
feat: split repo label into 4 labels (scheme, host, org and name parts)

jalonsoa:
If you have a large git host name and a large repo name then maybe you get error creating devpod because the label of the kubernetes resources is very large.

With this patch we try to reduce the label name & value getting only the first part of the git host name if the label name reach the 63 characters limit.

### DIFF
--- a/pkg/cmd/create/create_devpod.go
+++ b/pkg/cmd/create/create_devpod.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
@@ -105,6 +106,72 @@ type CreateDevPodOptions struct {
 	GitCredentials credentials.StepGitCredentialsOptions
 
 	Results CreateDevPodResults
+}
+
+// devPodLabels split gitInfo labels
+type devPodLabels struct {
+	gitSchemeLabelKey   string
+	gitSchemeLabelValue string
+	gitHostLabelKey     string
+	gitHostLabelValue   string
+	gitOrgLabelKey      string
+	gitOrgLabelValue    string
+	gitRepoLabelKey     string
+	gitRepoLabelValue   string
+
+	gitLabels map[string]string
+}
+
+// populateFromGitInfo populate this struct with the git repository elements (scheme, host, org and repo)
+func (d *devPodLabels) populateFromGitInfo(gitInfo *gits.GitRepository) error {
+	d.gitSchemeLabelKey = fmt.Sprintf("%s-scheme", kube.LabelDevPodGitPrefix)
+	d.gitSchemeLabelValue = naming.ToValidNameWithDotsTruncated(gitInfo.Scheme, 63)
+	d.gitHostLabelKey = fmt.Sprintf("%s-host", kube.LabelDevPodGitPrefix)
+	d.gitHostLabelValue = naming.ToValidNameWithDotsTruncated(gitInfo.Host, 63)
+
+	d.gitOrgLabelKey = fmt.Sprintf("%s-organisation", kube.LabelDevPodGitPrefix)
+	d.gitOrgLabelValue = naming.ToValidNameWithDots(gitInfo.Organisation)
+	d.gitRepoLabelKey = fmt.Sprintf("%s-name", kube.LabelDevPodGitPrefix)
+	d.gitRepoLabelValue = naming.ToValidNameWithDots(gitInfo.Name)
+
+	if len(d.gitOrgLabelValue) > 63 {
+		return errors.New("git organization label or value exceed length")
+	}
+
+	if len(d.gitRepoLabelValue) > 63 {
+		return errors.New("git repo label or value exceed length")
+	}
+
+	d.gitLabels = make(map[string]string)
+
+	if len(d.gitSchemeLabelValue) > 0 {
+		d.gitLabels[d.gitSchemeLabelKey] = d.gitSchemeLabelValue
+	}
+	if len(d.gitHostLabelValue) > 0 {
+		d.gitLabels[d.gitHostLabelKey] = d.gitHostLabelValue
+	}
+	if len(d.gitOrgLabelValue) > 0 {
+		d.gitLabels[d.gitOrgLabelKey] = d.gitOrgLabelValue
+	}
+	if len(d.gitRepoLabelValue) > 0 {
+		d.gitLabels[d.gitRepoLabelKey] = d.gitRepoLabelValue
+	}
+
+	return nil
+}
+
+// getLabels return a map with the labels
+func (d *devPodLabels) getLabels() map[string]string {
+	return d.gitLabels
+}
+
+// mergeLabels return a map with the labels
+func (d *devPodLabels) mergeLabels(labels map[string]string) map[string]string {
+	l := d.getLabels()
+	for k, v := range labels {
+		l[k] = v
+	}
+	return l
 }
 
 // NewCmdCreateDevPod creates a command object for the "create" command
@@ -237,21 +304,25 @@ func (o *CreateDevPodOptions) Run() error {
 	name := ""
 	podResources := client.CoreV1().Pods(ns)
 	var exposeServicePorts []int
+	var gitLabels devPodLabels
 
-	gitLabelKey := ""
-	gitLabelValue := ""
 	if importURL != "" && o.Reuse {
 		gitInfo, err := gits.ParseGitURL(importURL)
 		if err != nil {
 			log.Logger().Warnf("could not parse the git URL %s: %s", importURL, err.Error())
 		} else {
-			gitLabelKey = fmt.Sprintf("%s-%s-%s-%s", kube.LabelDevPodGitPrefix, gitInfo.Host, gitInfo.Organisation, gitInfo.Name)
-			gitLabelValue = naming.ToValidNameWithDots(importURL)
+			err = gitLabels.populateFromGitInfo(gitInfo)
+			if err != nil {
+				return errors.Wrapf(err, "populating labels from git info %s", err.Error())
+			}
+
 			// lets query to see if there is a DevPod already for this URL
 			matchLabels := map[string]string{
 				kube.LabelDevPodUsername: userName,
-				gitLabelKey:              gitLabelValue,
 			}
+
+			matchLabels = gitLabels.mergeLabels(matchLabels)
+
 			pod, err = o.findDevPodBySelector(podResources, matchLabels, dir)
 			if err != nil {
 				return errors.Wrapf(err, "finding DevPod by selector: %v", matchLabels)
@@ -342,9 +413,8 @@ func (o *CreateDevPodOptions) Run() error {
 		pod.Labels[kube.LabelPodTemplate] = label
 		pod.Labels[kube.LabelDevPodName] = name
 		pod.Labels[kube.LabelDevPodUsername] = userName
-		if gitLabelKey != "" && gitLabelValue != "" {
-			pod.Labels[gitLabelKey] = gitLabelValue
-		}
+
+		pod.Labels = gitLabels.mergeLabels(pod.Labels)
 
 		if len(pod.Spec.Containers) == 0 {
 			return fmt.Errorf("no containers specified for label %s with pod: %#v", label, pod)
@@ -639,12 +709,13 @@ func (o *CreateDevPodOptions) Run() error {
 				name = pod.Name
 				create = false
 
-				// lets add the label if its not already
-				if gitLabelKey != "" && gitLabelValue != "" && pod.Labels[gitLabelKey] != gitLabelValue {
-					pod.Labels[gitLabelKey] = gitLabelValue
+				newLabels := gitLabels.mergeLabels(pod.Labels)
+
+				if !reflect.DeepEqual(pod.Labels, newLabels) {
+					pod.Labels = newLabels
 					_, err = podResources.Update(pod)
 					if err != nil {
-						log.Logger().Warnf("Failed to update label of pod %s: %s", pod.Name, err.Error())
+						log.Logger().Warnf("Failed to update git labels of pod %s: %s", pod.Name, err.Error())
 					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


pkg/cmd/create/create_devpod.go
```diff
@@ -247,6 +247,11 @@ func (o *CreateDevPodOptions) Run() error {
 		} else {
 			gitLabelKey = fmt.Sprintf("%s-%s-%s-%s", kube.LabelDevPodGitPrefix, gitInfo.Host, gitInfo.Organisation, gitInfo.Name)
 			gitLabelValue = naming.ToValidNameWithDots(importURL)
+			// try to minimize label size reducing git host to first part
+			if len(gitLabelKey) > 63 || len(gitLabelValue) > 63 {
+				gitLabelKey   = fmt.Sprintf("%s-%s-%s-%s", kube.LabelDevPodGitPrefix, gitInfo.Host[0:strings.Index(gitInfo.Host, ".")], gitInfo.Organisation, gitInfo.Name)
+				gitLabelValue = naming.ToValidNameWithDots(strings.Replace(gitLabelValue,gitInfo.Host,gitInfo.Host[0:strings.Index(gitInfo.Host, ".")])
```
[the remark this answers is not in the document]
jalonsoa: Sorry, fixed !

ankitm123:
You should squash ur commits btw: https://jenkins-x.io/docs/contributing/code/#squash-and-rebase. Also, can you wrap this in a function, and include some unit tests. That would help to increase the test coverage.

pkg/cmd/create/create_devpod.go
```diff
@@ -158,6 +158,20 @@ func NewCmdCreateDevPod(commonOpts *opts.CommonOptions) *cobra.Command {
 	return cmd
 }
 
+// GetGitLabel try to reduce the label key and value size if they exceed 64 characters in length
+func GetGitLabel(gitInfo *gits.GitRepository, importURL string) (string, string) {
+	gitLabelKey := fmt.Sprintf("%s-%s-%s-%s", kube.LabelDevPodGitPrefix, gitInfo.Host, gitInfo.Organisation, gitInfo.Name)
+	gitLabelValue := naming.ToValidNameWithDots(importURL)
```
hferentschik, line 164: Would it not be easier to just call `naming.ToValidNameTruncated` regardless? Something like:

```go
gitLabelKey := naming.ToValidNameTruncated(fmt.Sprintf("%s-%s-%s-%s", kube.LabelDevPodGitPrefix, gitInfo.Host, gitInfo.Organisation, gitInfo.Name), 63)
gitLabelValue := naming.ToValidNameTruncated(naming.ToValidNameWithDots(importURL), 63)
```

pkg/cmd/create/create_devpod.go
```diff
@@ -158,6 +158,20 @@ func NewCmdCreateDevPod(commonOpts *opts.CommonOptions) *cobra.Command {
...
+
+	// try to minimize label size reducing git host to first part
+	if len(gitLabelKey) > 63 || len(gitLabelValue) > 63 {
```
hferentschik, line 167: See above, if you use `ToValidNameTruncated` you get the name truncated when needed and the full name when the length is above 63 characters. In this case you don't have to do another check.

pkg/cmd/create/create_devpod.go
```diff
@@ -158,6 +158,20 @@ func NewCmdCreateDevPod(commonOpts *opts.CommonOptions) *cobra.Command {
...
+		gitLabelKey = fmt.Sprintf("%s-%s-%s-%s", kube.LabelDevPodGitPrefix, gitInfo.Host[0:strings.Index(gitInfo.Host, ".")], gitInfo.Organisation, gitInfo.Name)
```
hferentschik, line 168: Does it matter how the label key is truncated? What if even with the host part the length is exceeded? I guess by just truncating you might get potential clashes in label selection, right?

One wonders why this all needs to go into a single label? Would it not be better to have three separate labels, host, organisation and name? Each of which should be wrapped around a `ToValidNameTruncated`. WDYT?

pkg/cmd/create/create_devpod.go
jalonsoa: If you only truncate the length of the label name or label value, you can match the resources of another proyect. For exampl, if you have 2 proyects:

git.very-large-name-of-my-company.com/large-name-group/project-1
git.very-large-name-of-my-company.com/large-name-group/project-2

And you cut by length ... two labels are the same, but the resources inside kubernetes must be different.

If you cut by git hostname ... the projkect remain differents.

pkg/cmd/create/create_devpod.go
jalonsoa: You are right, this patch don't remove the problem of length, it only tries to reduce its possible occurrence.

And you are right again .. I agree with you, we must separate the host, the group and the project name in different labels, but not only at this point, in all the places where they are used to create labels.

This little patch is a quick win to reduce the ocurrence, but don't fix the problem

pkg/cmd/create/create_devpod.go
hferentschik: got you.

pkg/cmd/create/create_devpod.go
hferentschik: > You are right, this patch don't remove the problem of length, it only tries to reduce its possible occurrence.

Ok

> And you are right again .. I agree with you, we must separate the host, the group and the project name in different labels, but not only at this point, in all the places where they are used to create labels.

These labels are just used within the devpods, right? So we could just cleanup the devpod code to use individual labels, right? Would that not be the proper fix?

jalonsoa: Ok for me. I'm workking on it !

jalonsoa: Hi,
  four labels for scheme, host, org and repo is ok for you ? For example

jenkins.io/repo/scheme: https
jenkins.io/repo/host: git.very-large-name-of-my-company.com
jenkins.io/repo/organisation: large-name-group
jenkins.io/repo/name: project-1

j

pkg/cmd/create/create_devpod.go
```diff
@@ -158,6 +158,20 @@ func NewCmdCreateDevPod(commonOpts *opts.CommonOptions) *cobra.Command {
 	return cmd
 }
 
+// GetGitLabels generate label keys and values for git repository elements (scheme, host, org and repo)
+func GetGitLabels(gitInfo *gits.GitRepository) (string, string, string, string, string, string, string, string) {
```
hferentschik, line 162: Why would you export this function? It should be at the very least private. With all these return parameters I find this function very hard to use. What is really the intention here? It is really easy to mess up the assignment in this way.

pkg/cmd/create/create_devpod.go
```diff
@@ -238,19 +252,27 @@ func (o *CreateDevPodOptions) Run() error {
 	podResources := client.CoreV1().Pods(ns)
 	var exposeServicePorts []int
 
-	gitLabelKey := ""
-	gitLabelValue := ""
+	gitSchemeLabelKey := ""
```
hferentschik, line 255: That's a lot of variables. What do you think about introducing a struct which contains all the relevant labels? Something like _devPodLabels_. Then we can encapsulate all required variables in there and we can have something like `func (d *_devPodLabels_) populateFromGitInfo( *gits.GitRepository) error` to set the fields. 
We could also have `func (d *_devPodLabels_) matchLabels() map[string]string`. And maybe even `func (d *_devPodLabels_) addLabels(pod *corev1.Pod) `. WDYT?

pkg/cmd/create/create_devpod.go
```diff
@@ -158,6 +158,20 @@ func NewCmdCreateDevPod(commonOpts *opts.CommonOptions) *cobra.Command {
 	return cmd
 }
 
+// GetGitLabels generate label keys and values for git repository elements (scheme, host, org and repo)
+func GetGitLabels(gitInfo *gits.GitRepository) (string, string, string, string, string, string, string, string) {
+	gitSchemeLabelKey := fmt.Sprintf("%s/scheme", kube.LabelDevPodGitPrefix)
+	gitSchemeLabelValue := gitInfo.Scheme
+	gitHostLabelKey := fmt.Sprintf("%s/host", kube.LabelDevPodGitPrefix)
+	gitHostLabelValue := gitInfo.Host
```
hferentschik, line 166: for good measure we might want to wrap each of the values into a `ToValidNameTruncated`

pkg/cmd/create/create_devpod.go
jalonsoa: It's ok for me .. working on it !

pkg/cmd/create/create_devpod.go
jalonsoa: ummm .. We don't have problem with scheme or host, but org or repo .. we can clash with another repo if we cut by length on this variable.

Maybe, cut scheme and host, but return error if the length of org & repo variables exceed the length ?

pkg/cmd/create/create_devpod.go
jalonsoa: First step: populateFromGitInfo and getLabels it's ok ... but addLabels I'm not sure .. You must check if you modify the pod labels before update, sou you need to compare the pod labels with the merged labels (pod lobales + git labels) .. maybe `func (d *_devPodLabels_) mergeLabels( labels map[string]string) map[string]string` is interesting, but not addLabels(pod)

pkg/cmd/create/create_devpod.go
hferentschik: > but org or repo .. we can clash with another repo if we cut by length on this variable.

wonder what the concrete consequence, in this case, would be. Any idea?

>  Maybe, cut scheme and host, but return error if the length of org & repo variables exceed the length ?

Works for me. I guess now that the labels are split the scenario becomes quite unlikely really.

pkg/cmd/create/create_devpod.go
hferentschik: Up to you, these were just suggestions initially. Hard to tell the details w/o actually cutting the code.

hferentschik:
@jalonsoa - ooi, is your pull request using pull request #6919 using as well or how do you make sure that the right ksync image is used?

pkg/cmd/create/create_devpod.go
```diff
@@ -107,6 +108,72 @@ type CreateDevPodOptions struct {
...
+}
+
+// populateFromGitInfo populate this struct with the git repository elements (scheme, host, org and repo)
+func (d *devPodLabels) populateFromGitInfo(gitInfo *gits.GitRepository) error {
+	d.gitSchemeLabelKey = naming.ToValidNameTruncated(fmt.Sprintf("%s/scheme", kube.LabelDevPodGitPrefix), 63)
+	d.gitSchemeLabelValue = naming.ToValidNameTruncated(gitInfo.Scheme, 63)
+	d.gitHostLabelKey = naming.ToValidNameTruncated(fmt.Sprintf("%s/host", kube.LabelDevPodGitPrefix), 63)
+	d.gitHostLabelValue = naming.ToValidNameTruncated(gitInfo.Host, 63)
+
+	d.gitOrgLabelKey = fmt.Sprintf("%s/organisation", kube.LabelDevPodGitPrefix)
```
hferentschik, line 132: I think this will create an invalid label key with to '/':

```
error: finding DevPod by selector: map[jenkins-io-repo-host:github-com jenkins-io-repo-scheme:https jenkins.io/devpod_user:hardy jenkins.io/repo/name:go-eks jenkins.io/repo/organisation:hf-bee]: converting label selector: map[jenkins-io-repo-host:github-com jenkins-io-repo-scheme:https jenkins.io/devpod_user:hardy jenkins.io/repo/name:go-eks jenkins.io/repo/organisation:hf-bee]: invalid label key "jenkins.io/repo/organisation": a qualified name must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]') with an optional DNS subdomain prefix and '/' (e.g. 'example.com/MyName')
```

I think we need to create new labels.

jalonsoa: You are right. We must sanitize the label keys (I use now the ToValidNameTruncated function to all labels keys now)

jalonsoa:
@hferentschik about ksync .. the pullrequest that you mention is about select the correct ksync image / binary to download when you don't have the ksync installed, isn't ? This pull request  don't have problems with this.

I review the code of jx sync command (and the pull request) and this command use only the LabelDevPodUsername ("jenkins.io/devpod_user") label for select all devpod pods from this developer and use this information to remove all old ksync of this developer/template. Then run the ksync command with this parameters:

o.RunCommand("ksync", "create", "--name", name, "-l", "jenkins.io/devpod="+name, reload, "-n", ns, dir, remoteDir)

So .. I think that the othe pull request don't affect our change

hferentschik:
> about ksync .. the pullrequest that you mention is about select the correct ksync image / binary to download when you don't have the ksync installed, isn't ? This pull request don't have problems with this.

Right, that was my problem. I had at that stage no _ksync_ installed and was basically first running into the issues fixed by #6919. So technically they might not depend on each other, but they are related. Either way, #6919 is about to merge. I will give pull request another spin once it is in.
Looking at it all seems fine, so hopefully, it should be in shortly.

pkg/cmd/create/create_devpod.go
```diff
@@ -107,6 +108,72 @@ type CreateDevPodOptions struct {
...
+	d.gitHostLabelKey = fmt.Sprintf("%s-host", kube.LabelDevPodGitPrefix)
+	d.gitHostLabelValue = naming.ToValidNameWithDotsTruncated(gitInfo.Host, 63)
+
+	d.gitOrgLabelKey = fmt.Sprintf("%s-organisation", kube.LabelDevPodGitPrefix)
+	d.gitOrgLabelValue = naming.ToValidNameWithDots(gitInfo.Organisation)
+	d.gitRepoLabelKey = fmt.Sprintf("%s-name", kube.LabelDevPodGitPrefix)
+	d.gitRepoLabelValue = naming.ToValidNameWithDots(gitInfo.Name)
+
+	if len(d.gitOrgLabelValue) > 63 {
+		return errors.New("git organization label or value exceed length")
```
hferentschik, line 138: 👍

PR #7079, opened 2020-04-19
refactor: add missing error checks

### LINKED ISSUE #6851: Run errcheck as part of the checks when merging a PR
### Summary 
There are some places in the code base which lack error handling. For instance: https://github.com/jenkins-x/jx/blob/master/pkg/helm/tiller.go#L51

It would be a good idea to run errcheck (https://github.com/kisielk/errcheck) as part of the checks when merging a PR, this will make the code base more consistent and improve the code quality.

Sample when running 
`CGO_ENABLED=0 GO111MODULE=on errcheck ./...`
```
....
pkg/table/table.go:53:15:	fmt.Fprint(out, t.Separator)
pkg/table/table.go:58:15:	fmt.Fprint(out, col)
pkg/table/table.go:60:15:	fmt.Fprint(out, util.Pad(col, " ", l, align))
pkg/table/table.go:63:13:	fmt.Fprint(out, "\n")
pkg/terraform/terraform.go:32:11:	os.Setenv("GOOGLE_CREDENTIALS", serviceAccountPath)
pkg/terraform/terraform.go:113:18:	defer file.Close()
pkg/util/cache.go:42:18:	writeTimeToFile(timecheckFileName, time.Now())
pkg/util/commands.go:141:11:	os.Setenv("PATH", PathWithBinary(c.Dir))
pkg/util/commands.go:170:11:	os.Setenv("PATH", PathWithBinary(c.Dir))
pkg/util/downloads.go:33:17:	defer out.Close()
pkg/util/downloads.go:40:23:	defer resp.Body.Close()
pkg/util/downloads.go:54:10:	os.Chmod(filepath, 0755)
pkg/util/downloads.go:138:23:	defer resp.Body.Close()
pkg/util/downloads.go:172:27:	defer response.Body.Close()
pkg/util/downloads.go:221:23:	defer resp.Body.Close()
pkg/util/downloads.go:257:21:	defer zreader.Close()
pkg/util/downloads.go:260:20:	defer reader.Close()
....
```
### Jx version
2.0.1225

### Jenkins type

- [ ] Classic Jenkins

ankitm123:

#### Submitter checklist

#### Which issue this PR fixex

fixes #6851


pkg/cmd/clients/factory.go
```diff
@@ -741,7 +741,7 @@ func (f *factory) CreateVaultOperatorClient() (vaultoperatorclient.Interface, er
 func (f *factory) CreateHelm(verbose bool,
 	helmBinary string,
 	noTiller bool,
-	helmTemplate bool) helm.Helmer {
+	helmTemplate bool) (helm.Helmer, error) {
```
hferentschik, line 744: Would it be worth to split this change up and address first everything apart of this change to `CreateHelm`? We could explicitly ignore the error. This would reduce the changeset quite a bit.

One reason for that is that I believe that we really never (or at least rarely) jump into this branch which creates the error which we then propagate. Also imagining a future where we move towards Helm3 and all Tiller related code is obsolete, this particular piece of code might even go away. WDYT?

ankitm123: That makes sense. You are right that once helm3 comes, this code would be removed. Fun fact: there are still 200+ places in the code base (excluding defer) where errors are not being checked, so I am going to focus on those now :)

hferentschik: > Fun fact: there are still 200+ places in the code base (excluding defer) where errors are not being checked, so I am going to focus on those now 

Gee

ankitm123:
@hferentschik any idea why the tests fail?

ankitm123:
Seems like unit and integration tests pass now, but lint still fails. Locally when I run make lint, it fails for `make generate-docs`, I see this weird thing in git diff (after running `make generate-docs`):
```
-       openapi "github.com/jenkins-x/jx/pkg/client/openapi/all"
+       openapi "github.com/ankitm123/jx/pkg/client/openapi/all"
```
But I think this is just a local build issue, but I am not sure ...

abayer:
The lint failure seems to have been the container crashing - I just relaunched it.

abayer:
Ok, that's weird - the lint job crashed in the same way again:

```
Showing logs for build jenkins-x-jx-pr-7079-lint-5kmtw-62 stage ci and container step-lint
Installing GolangCI-Lint
golangci/golangci-lint info checking GitHub for tag 'v1.20.0'
golangci/golangci-lint info found version: 1.20.0 for v1.20.0/linux/amd64
golangci/golangci-lint info installed /workspace/go/bin/golangci-lint
level=info msg="[lintersdb] Active 11 linters: [deadcode errcheck gofmt goimports gosec interfacer misspell structcheck typecheck unconvert unused]"
level=info msg="[loader] Go packages loading at mode 575 (deps|files|imports|compiled_files|exports_file|name|types_sizes) took 3m46.395093777s"
level=info msg="[runner/filename_unadjuster] Pre-built 0 adjustments in 315.226665ms"
./hack/linter.sh: line 36:    71 Killed                  golangci-lint run --no-config --disable-all -E misspell -E unconvert -E deadcode -E unconvert -E gosec -E gofmt -E goimports -E structcheck -E interfacer -E typecheck -E errcheck -E unused --skip-dirs vendor --timeout 15m --verbose --build-tags build
```

ankitm123:
I added two additional linters - errcheck and unused, do you think that is causing the errors, may be those test containers need more resources, or may be we need to increase the timeout period?

ankitm123:
ok, the containers are getting killed because of the high memory usage of golangci lint, I will make some changes to the settings. Related issue: https://github.com/golangci/golangci-lint/issues/337

ankitm123:
Seems like lint is randomly failing. Enabling more linters will make it fail more. I think there are some options we can try:
- Request more memory. golangcl-lint needs lot of memory (I posted the memory issue in this PR). The jobs request only 2G memory, and the limit is at 3GB, probably double them?
- Split it into more jobs. We have just one lint job, may be break that into multiple jobs.
wdyt? @abayer @hferentschik

jenkins-x-lint.yml
```diff
@@ -20,11 +20,11 @@ pipelineConfig:
               containerOptions:
                 resources:
                   limits:
+                    cpu: 4
+                    memory: 8Gi
```
hferentschik, line 24: @abayer do you have any issue with increasing the limit here?

pkg/cmd/gc/gc_gke.go
```diff
@@ -155,9 +157,16 @@ func (o *GCGKEOptions) Run() error {
...
 
 	path := util.UrlJoin(dir, "gc_gke.sh")
-	util.FileExists(path)
+	_, err = util.FileExists(path)
+	if err != nil {
+		return errors.Wrapf(err, "checking if file %s exists", path)
+	}
 
-	os.Remove(path)
+	//Todo: May be delete this file only if it exists?
```
hferentschik, line 165: Good catch. I think this is the original intention. I don't ting the existing code

```go
util.FileExists(path)
os.Remove(path)
```

makes much sense. My guess would be that the delete was meant to go into an if block under the existence check.

pkg/cmd/get/get_pipeline.go
```diff
@@ -152,6 +155,7 @@ func (o *GetPipelineOptions) Run() error {
 	table := createTable(o)
 
 	for _, j := range names {
+		//Todo: Which error are we trying to catch here?
```
hferentschik, line 158: none for what I can tell

pkg/cmd/get/get_pipeline.go
```diff
@@ -176,7 +180,8 @@ func (o *GetPipelineOptions) dump(jenkinsClient gojenkins.JenkinsClient, name st
 
 	if job.Jobs != nil {
 		for _, child := range job.Jobs {
-			o.dump(jenkinsClient, job.FullName+"/"+child.Name, table)
+			//Todo: Recursive call
+			o.dump(jenkinsClient, job.FullName+"/"+child.Name, table) //nolint:errcheck
```
hferentschik, line 184: I guess the idea is to find child jobs !?

I am wondering though whether `get pipelines` is not Static Master specific anyways and hence obsolete. My guess is we should delete the whole command.

@abayer - wdyt?

pkg/cmd/step/restore/step_restore_from_backup.go
```diff
@@ -93,7 +93,7 @@ func (o *FromBackupOptions) Run() error {
 	// check if a velero schedule exists
 	scheduleExists, err := velero.DoesVeleroBackupScheduleExist(apiClient, o.Namespace)
 	if err != nil {
-		errors.Wrap(err, "when trying to check for velero schedules")
+		return errors.Wrap(err, "when trying to check for velero schedules")
```
hferentschik, line 96: lol, apparently that happened quite often in the codebase. Wrapping the error, but not returning it. Definitely shaking out some bugs here 👍

pkg/cmd/step/step_split_monorepo.go
```diff
@@ -206,9 +206,10 @@ func (o *StepSplitMonorepoOptions) Run() error {
 							return err
 						}
 					}
+					//ToDo: Why are we ignoring errors?
```
hferentschik, line 209: No idea. My best guess would be that the comment implies that for example _.gitignore_ might already be added to version control and when calling `git add` on it you would get an error. However, this is not the case. You can call `git add` as often as you want. So if this call would return an error it would be an error in the whole logic/execution of executing the actual `git add`. So imo I don't think the error should be ignored here. 

Also, `split monorepo` is another one of these commands which I believe have no use. In fact, it is in the list of deprecated commands (_deprecated.go_). For all, I care we can remove it. Not sure whether we should do it as part of this pull request though. @abayer thoughts? 
If we keep it, I would handle the error and remove the comment.

pkg/pipeline_events/elasticsearch.go
```diff
@@ -90,7 +90,10 @@ func (e ElasticsearchProvider) SendRelease(r *v1.Release) error {
 				r.Namespace: r.CreationTimestamp.String(),
 			},
 		}
-		e.SendIssue(&esissue)
+		err = e.SendIssue(&esissue)
```
hferentschik, line 93: hmmm, did not even know this existed. The file/functionality itself. I am almost certain it can safely be deleted. @abayer? 

Just wrapping the error works for me as well.

ankitm123:
I did see a few jenkins static related functionalities, but was not sure, if I could remove them. Either way, they go beyond the scope of this PR, so I just used //nolint:errcheck for those.

pkg/cmd/get/get_pipeline.go
ankitm123: Removed

pkg/cmd/gc/gc_gke.go
ankitm123: Fixed 👍

ankitm123:
So after rebasing, the lint failed again. So, I decided to split the linting into two files - linter.sh and linter-extra.sh, keeping the increased limits for both of those linting jobs.

hack/linter-extra.sh
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+if ! [ -x "$(command -v golangci-lint)" ]; then
+	echo "Installing GolangCI-Lint"
+	${DIR}/install_golint.sh -b $GOPATH/bin v1.20.0
+fi
+
+export GO111MODULE=on
```
hferentschik, line 17: Personally, I find it hard to see what the difference between lint and lint-extra is. I fear this might confuse others as well. What do we gain? 

AFAICT, the bulk of the scripts are identical. So if we really want to split into multiple lint steps in `Makefile` and `jenkins-x-lint.yaml`, then I think it should be a single linter script and you pass in by argument which checks to run. 

If this is just about splitting into more tasks and contexts, one thing I want to do for a long time it to break out a `codegen` context (`jenkins-x-codegen.yaml) and so the code generation checks separately. That's technically really not linting and it has confused people before. I'd rather do that than splitting lint and lint-extra (unless there is something I am missing here)

pkg/cmd/gc/gc_gke.go
```diff
@@ -155,9 +157,16 @@ func (o *GCGKEOptions) Run() error {
...
 
 	path := util.UrlJoin(dir, "gc_gke.sh")
-	util.FileExists(path)
-
-	os.Remove(path)
+	exists, err := util.FileExists(path)
+	if err != nil {
+		return errors.Wrapf(err, "checking if file %s exists", path)
+	}
+	if exists {
```
hferentschik, line 164: 👍  sweet

pkg/cmd/get/get_pipeline.go
```diff
@@ -123,7 +123,10 @@ func (o *GetPipelineOptions) Run() error {
 			if err != nil {
 				return err
 			}
-			o.dump(jenkins, job.Name, &table)
+			err = o.dump(jenkins, job.Name, &table)
```
hferentschik, line 126: 👍

pkg/pipeline_events/elasticsearch.go
```diff
@@ -90,7 +92,10 @@ func (e ElasticsearchProvider) SendRelease(r *v1.Release) error {
...
+		if err != nil {
```
hferentschik, line 96: 👍

hack/linter-extra.sh
ankitm123: lint -> all the old linters, lint-extra -> new linters (ununsed and errcheck). It's a bit difficult for me to know where the script really fails. I tried various combinations, and this is the one which succeeded. I think it might make sense to coordinate over slack, if that's an option. I can try making the changes, and if you can get me the logs, that would help us debug it 🥇 wdyt? Let me try to separate the code gen from linting, and see what happens.

ankitm123:
@hferentschik combining all the linters into one lint file, made it fail again. Can we increase the memory further. There is also the option to use a lower value of GOGC, this will reduce the memory usage, but at the expense of making the tests run longer. https://github.com/golangci/golangci-lint#memory-usage-of-golangci-lint
The jx code base is much bigger than the golangci-lint repo, so it makes sense why it breaks. Also in the future, we should add more linters (staticcheck, go cyclo), which will make it impossible to lint the code base imo. Wdyt?

ankitm123:
So the lint passed after I set GOGC=10 💯

jenkins-x-lint.yml
abayer: I do not, no.

jenkins-x-codegen.yml
```diff
@@ -0,0 +1,32 @@
+buildPack: none
```
hferentschik, line 1: 👍

hferentschik: we will need to enable the new context in the configuration for the repo.

Makefile
```diff
@@ -213,6 +213,9 @@ importfmt: get-fmt-deps
 lint: ## Lint the code
 	./hack/gofmt.sh
 	./hack/linter.sh
+
+.PHONY: code-generate
```
hferentschik, line 217: 👍

jenkins-x-codegen.yml
```diff
@@ -0,0 +1,32 @@
...
+                    cpu: 4
+                    memory: 8Gi
+                  requests:
+                    cpu: 3
+                    memory: 6Gi
+            steps:
+              - name: generate
+                image: golang:1.12.10
+                command: ./hack/generate.sh
+                dir: /workspace/go/src/github.com/jenkins-x/jx
```
hferentschik, line 32: nitpick, missing newline

hferentschik:
@ankitm123 - fyi https://github.com/jenkins-x/environment-tekton-weasel-dev/pull/581. Once that's merging the new context should appear and we can rebuild the whole PR. If all looks good we can merge and then we can make it a required context.

hferentschik:
@ankitm123 :

```
Showing logs for build jenkins-x-jx-pr-7079-codegen-gvrfd-189 stage ci and container step-generate
/bin/sh: 1: cd: can't cd to /workspace/go/src/github.com/jenkins-x/**
Pipeline failed on stage 'ci' : container 'step-generate'. The execution of the pipeline has stopped.
```

I think we need the copy parts from the lint status check as well to move the sources to the "right" location. At least while we are still on Go 1.12.

ankitm123:
@hferentschik great, the codegen test works now, gonna remove that copy part from the lint (it's not needed any more imo).

jenkins-x-lint.yml
hferentschik: :-)

jenkins-x-codegen.yml
```diff
@@ -0,0 +1,46 @@
...
+              containerOptions:
+                resources:
+                  limits:
+                    cpu: 4
+                    memory: 8Gi
+                  requests:
+                    cpu: 3
+                    memory: 6Gi
+            steps:
+              - name: mk-jx-project-dir
```
hferentschik, line 29: :+1

cagiti:
This is a great change, but I think it would be useful to log an error for missing error checks in one PR before returning the error, because it alters behaviour of the codebase (albeit incorrect_). Then a subsequent PR which returns the error would be required (_when the impact is understood_), and arguably could be a case for a minor version increment.

ankitm123:
> This is a great change, but I think it would be useful to log an error for missing error checks in one PR before returning the error, because it alters behaviour of the codebase (albeit incorrect_). Then a subsequent PR which returns the error would be required (_when the impact is understood_), and arguably could be a case for a minor version increment.

You are right. Honestly, I thought that this PR might lead to some cases, where the downstream code relies on _no errors_ being returned. There are about 90 places in the codebase where I have skipped errchecks (using nolint). Not catching the error can lead to unintended results though, for instance we still dont handle errors for any of the defer functions and I am not sure what kind of bugs they have introduced (for instance we dont know if those files are being closed properly). Are you seeing any places where handling the errors have led to some issues? Would be nice to re-evaluate the logic there.

PR #7091, opened 2020-04-21
fix: Handle Terraform SA display name format in gc_gke

abayer:
#### Submitter checklist

#### Description

The `displayName` in our boot-created SAs is like `pr-206-5-boot-gke-vt`, but the ones created via Terraform end up as `Velero service account for cluster pr-7065-46-tf-boot`. Which is fine! Except that our GC keys on `displayName` having a known suffix, and these...don't. So let's fall back on parsing the name from the email when there are spaces in the `displayName`.

n/a

### DIFF
--- a/pkg/cmd/gc/gc_gke.go
+++ b/pkg/cmd/gc/gc_gke.go
@@ -49,7 +49,7 @@ var (
 		jx gc gke
 `)
 
-	ServiceAccountSuffixes = []string{"-vt", "-ko", "-tf", "-dn", "-ex", "-jb", "-st", "-tk", "-vo", "-bc"}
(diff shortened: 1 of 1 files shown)


pkg/cmd/gc/gc_gke.go
```diff
@@ -507,12 +507,16 @@ func (o *GCGKEOptions) getServiceAccounts() ([]serviceAccount, error) {
 		return nil, err
 	}
 
+	var modifiedSAs []serviceAccount
 	for _, sa := range serviceAccounts {
-		if sa.DisplayName == "" {
+		// If there isn't a display name or the display name contains spaces, which is the case for Terraform-created
```
hferentschik, line 512: I think the better way to do this is anyways to use the email. That is something in a fixed known format. The display name is probably one of the worst things to use in this case.

hferentschik:
@abayer, thanks for fixing this. I was close on your tail on this one. 

/lgtm

PR #7092, opened 2020-04-22
fix: simplify jx version

### LINKED ISSUE #7081: jx version does not print version
### Summary 
Running `jx version` with a non-existent namespace, returns this error:
```
error: namespaces "<namespace-name>" not found
```
Running `kubectl version --short` gives the right output:
```
Client Version: v1.16.2
Server Version: v1.15.6
```
I am not sure why namespace information is required for a version command.

Change namespace to a non-existing namespace
`kubectl config set-context --current --namespace=non-existent-ns`
and then run
`jx version`

### Expected behavior
jx version prints out that the name space does not exist, but prints the rest of the information (jx binary version, kubernetes server version, kubectl version, git version and os info).
### Actual behavior
jx version prints out that the namespace does not exist, nothing else.
### Jx version
2.0.1286

```
NAME               VERSION
jx                 2.0.1286
Kubernetes cluster v1.15.6
kubectl            v1.16.2
Operating System   Mac OS X 10.15.4 build 19E287
```

### Jenkins type

- [ ] Classic Jenkins

### Kubernetes cluster
LKE (Linode Kubernetes Engine)

Mac OS X 10.15.4 build 19E287

ankitm123:

#### Submitter checklist

#### Description
This PR:
- Refactors the version command, making it feel more like traditional K8s commands like helm and kubectl.
- Uses the information passed during the building of the binary (build date, branch, goversion etc...) and displays them 
- Adds a new --short flag to only display the binary version
- Moves version to it's own directory/package
- Deprecates the --version flag (I am not sure what date to put, so I went with July 1, 2020), what is the deprecation policy (what timeline are we talking about here)? 
- Adds some basic unit tests to check that `jx version` works properly

The output from `jx version` looks like this now:
```
Version        2.0.1065-dev+3064c5b7c
Commit         3064c5b7c
Branch         jx-7081
Build date     20200527-07:47:07
Go version     1.13.8
Git tree state clean
```
Output of `jx version --short` looks like this:
```
Version 2.0.1065-dev+3064c5b7c
```

Running jx --version, returns this:
```
2.0.1065-dev+3064c5b7c
 Deprecated will be removed on July 1, 2020. Please use version instead
```

fixes #7081


pkg/cmd/version/version.go
```diff
@@ -30,39 +30,40 @@ type VersionOptions struct {
...
+	cmd.Flags().BoolVarP(&vOpts.HelmTLS, "helm-tls", "", false, "Whether to use TLS with helm")
+	cmd.Flags().BoolVarP(&vOpts.NoVersionCheck, "no-version-check", "n", false, "Disable checking of version upgrade checks")
+	cmd.Flags().BoolVarP(&vOpts.NoVerify, "no-verify", "", false, "Disable verification of package versions")
+	cmd.Flags().StringVarP(&vOpts.Namespace, "namespace", "", "", "The namespace to use to look for currently installed platform version")
 	return cmd
 }
 
 func (o *VersionOptions) Run() error {
 	ns := o.Namespace
+	fmt.Println("Namespace: ", ns)
```
hferentschik, line 61: I guess that's a leftover from debugging?

pkg/cmd/version/version.go
```diff
@@ -30,39 +30,40 @@ type VersionOptions struct {
...
 	if ns == "" {
 		var err error
 		_, ns, err = o.JXClientAndDevNamespace()
 		if err != nil {
-			return err
+			log.Logger().Warnf("Error with jx client: %v", err)
```
hferentschik, line 66: I am not sure whether this is the right thing here to do since the namespace is used in `GetPackageVersions`. The latter really should be called against the dev namespace of a Jenkins X install. With your change we pass the empty namespace. This is at the very least changing intent. 

For all I can tell, it also does not matter if I call `jx version` from a non jx namespace, as long as there is a Jenkins X system namespace at all. Problematic is to be connected to an invalid context or a cluster w/o Jenkins X installation.

There is an ambiguity here which I would love to resolve, but I think it would need more doing and a discussion across the team on what to do. There are two commands `jx --version` and `jx version`. The former literally just returns the version of the `jx` binary. No connection to the cluster. Simple. `jx version` on the other hand does more. It does report the version of the binary, but then it also checks for the cluster, git version, etc. A lot of users get caught by that.

I think to really "fix" that we need to change the behaviour of `jx version` to be the same as `jx --version`. We can then either deprecate/remove `jx --version` in the long run or just keep both. I think in the space we are with `jx` (helm, kustomize, cloud, ...) most people expect the version command to just return the version of the binary. 

The current functionality of `jx version` is basically a duplicate or at least very similar to `jx diagnose` (which might need some updating). That's what I would refer people to use instead of the current `jx version`.

@jenkins-x/core thoughts?

ankitm123: I think it should mimic what `kubectl version` and `helm version` do. They just print the client and server versions and exit. Version should not do more than that imo. The issue I had was it failed to print anything when there was an invalid namespace, which is a bit odd, as `kubectl version` does not throw error for that use case. 
I agree that version does a lot more than what it should do, just this much information should get printed imo.
```
NAME               VERSION
jx                 2.0.1065-dev+52231531f
Kubernetes cluster v1.15.6
kubectl            v1.16.2
```

pkg/cmd/version/version.go
ankitm123: Yes, classic rookie mistake, will remove.

pkg/cmd/version/version.go
hferentschik: I am not so sure about `Kubernetes cluster` and `kubectel`. With the former, you won't know whether the cluster which it reports is the actual Jenkins X cluster. Your current context might be set to something else. Also, it does imo not provide much useful context for `jx`. Same for `kubectl`, not sure why I need the version for that. But trying to do less is already a step into the right direction.

ankitm123: Right, I agree. We should move:
```
Kubernetes cluster v1.15.6
kubectl            v1.16.2
```
to `jx diagnose`. And make it like `helm version` (v2). Print the binary information, and the server side information. That would be ideal imo

hferentschik:
@ankitm123, sorry for all the state changes on this pull request. I accidentally approved the wrong pull request. I guess we are still discussing the best way forward on this.

pkg/cmd/version/version.go
hferentschik: > Right, I agree. We should move:

> Kubernetes cluster v1.15.6
> kubectl            v1.16.2
> to jx diagnose

+1, that makes more sense imo. Diagnose is where you try to collect as much useful information as possible.

> Print the binary information, and the server-side information. That would be ideal imo.

Right, but I am not sure what we have atm in terms of determining the server-side version of Jenkins X. And of course this it would require to find the Jenkins X namespace. I guess you would print the client-side version and then try to get some information about the server-side and if for some reasons, not possible print a warning of sorts. Similar to what Helm 2 does with Tiller. Persoanlly, I could live with just the client-side version for now.

> As an example, say the current context is set incorrectly, then the behavior should be:

> client: <jx-binary-version>
> server: <err message/warning message>

Exactly.

ankitm123: I am also thinking that client side version should be sufficient, there is no associated version for jx namespace anyways. We should remove the `--version` option to make it consistent with the other kubernetes cli tools imo. wdyt?

hferentschik:
Ok, now we can get top back one I guess ;-) To start with it would need a rebase.

pkg/cmd/version/version.go
hferentschik: > We should remove the --version option to make it consistent with the other kubernetes cli tools imo. wdyt

I guess we would deprecate it first, opposed to removing it directly.

hferentschik: So a plan of action could be:

1. Fix `jx diagnose` and ensure it contains all output we want. I would think it might make sense to have something like `jx diagnose --long` or similar which when enabled runs the additional `kubectl` commands which are currently in `jx diagnose`. Also, looking at the current implementation of `jx diagnose`, it uses os.exec to call `jx` commands itself. That seems wrong and depending on your PATH can have unwanted results. Might be worth do tackle `jx diagnose` in a dedicated issue which outlines the planned changes.
1. Once `jx diagnose` behaves as described above, we can change `jx version` to just return the version of the CLI. This is inline with other tools in the K8S space
1. Deprecate `jx --version` and eventually remove it

ankitm123: Makes sense to deprecate it first, and then remove it 👍

ankitm123: I will open an issue for jx diagnose, and a PR for that. For the version command, this is what I am thinking:
- `jx version --short`:
```
version: v1.16.2
``` 
- `jx version`:
```
version: X.Y.Z
commit: commitid
git tree state: always clean!
```
Pretty much what the other K8s tools do :), wdyt?

hferentschik: I am most interested in the short version, but nothing if it is not too complicated to make your suggestion work I am fine with it as well. For development versions, we already add the commit sha to the version - https://github.com/jenkins-x/jx/blob/22ce2138110e61c4497f02efc8205dd2fc8de718/Makefile#L33. Looks like this `2.1.33-dev+343f641ad`. To do what you are suggesting we need to change this a bit and always (even for release builds where there is an explicit VERSION set) compile in the build SHA.  

Regarding `git tree state`, I am not even really sure what that means. I guess that's whether you compile the binary with outstanding changes in the working tree. Not sure how valuable that is in the end. 

Important for me is that the version command is not doing network calls and is independent of your current Kube context. All information should be really be compiled into the binary.

hferentschik: @ankitm123 given our latest plans what shall we do with this pull request? Keep it open or close for now?

ankitm123: I think we should keep it open, I started refactoring `jx diagnose` in PR #7178. Once that is completed and merged, I will complete this. I did remove the code which was doing all the extra stuff in this PR (communicating to the kubernetes cluster, printing kubectl versions etc), but as we decided, we refactor jx diagnose, and then once that PR is merged, we merge this PR, wdyt?

hferentschik: Might be worth playing a bit with the `diagnose` options. Not sure whether we always want to print all the pvcs, pods, etc.

ankitm123: Thought abt the same thing, there is no way to only print version or configmaps or status using `jx diagnose`, making those `o.showOption("pods")` redundant.  I was thinking something along the lines of `jx diagnose <resource>`. I will be updating #7192 this weekend.

ankitm123: @hferentschik the only thing left in this PR is to figure out how to mark `--version` as deprecated. I found the place where we set it in the codebase (cmd.go), 
```
rootCommand.Version = version.GetVersion()
rootCommand.SetVersionTemplate("{{printf .Version}}\n")
```
but seems like 
```
cmd.Flags().MarkDeprecated("version", "")
```
does not work for this flag. I am trying to figure that out. As a hack, I can always add that deprecation information to the version template, but I would not prefer not to (it's ugly).

.github/issue_template.md
```diff
@@ -14,6 +14,14 @@ The output of `jx version` is:
 COPY OUTPUT HERE
 ```
 
+### Diagnostic information
+
+The output of `jx diagnose --show=version` is:
```
hferentschik, line 19: does this require #7192 to be ready and merged yet? 
Right now I get:

```
jx diagnose --show=version
Running in namespace: jx
ERROR: Unable to get the Jenkins X Version
error: Command failed 'jx version --no-version-check': Error: unknown flag: --no-version-check


Options:
      --short=false: Print version
Usage:
  jx version [flags] [options]
Use "jx options" for a list of global command-line options (applies to all commands). exit status 1
```

Makefile
```diff
@@ -15,8 +15,10 @@ ROOT_PACKAGE := github.com/$(ORG_REPO)/v2
 GO_VERSION := $(shell $(GO) version | sed -e 's/^[^0-9.]*\([0-9.]*\).*/\1/')
 GO_DEPENDENCIES := $(call rwildcard,pkg/,*.go) $(call rwildcard,cmd/jx/,*.go)
 
-BRANCH     := $(shell git rev-parse --abbrev-ref HEAD 2> /dev/null  || echo 'unknown')
-BUILD_DATE := $(shell date +%Y%m%d-%H:%M:%S)
+BRANCH         := $(shell git rev-parse --abbrev-ref HEAD 2> /dev/null  || echo 'unknown')
+BUILD_DATE     := $(shell date +%Y%m%d-%H:%M:%S)
+GIT_TREE_STATE := $(shell test -z "`git status --porcelain`" && echo "clean" || echo "dirty")
```
hferentschik, line 20: nice :-)

pkg/cmd/version/version_test.go
```diff
@@ -0,0 +1,59 @@
...
+var versiontests = []struct {
+	short bool
+	desc  string
+}{
+	{true, "Version with short flag"},
+	{false, "Normal version"},
+}
+
+// Basic unit tests to check that all the different flags properly for jx version sub command
+func Test_ExecuteVersion(t *testing.T) {
```
hferentschik, line 25: 👍

pkg/cmd/version/version.go
```diff
@@ -0,0 +1,63 @@
...
+		Short: "Print the version information",
+		Run: func(cmd *cobra.Command, args []string) {
+			options.Cmd = cmd
+			options.Args = args
+			err := options.Run()
+			helper.CheckErr(err)
+		},
+	}
+
+	cmd.Flags().BoolVar(&options.short, "short", false, "Print version")
```
hferentschik, line 36: wonder whether it would make sense to have a `--machine-readable` flag or maybe more a `-o` flag which would be more used in other commands already. Right now the output of the version commands is:

```
$ jx version
Version        2.1.48-dev+3064c5b7c
Commit         3064c5b7c
Branch         jx-7081
Build date     20200527-20:10:18
Go version     1.13.11
Git tree state clean
```

resp

```
jx version --short
Version 2.1.48-dev+3064c5b7c
```

Sometimes for scripting, it is nice to have an easy way to just get to a specific piece of information without having to use _sed_, _awk_ or similar. So having the option to write the result as for example json could be handy. Alternatively, `-short` could just print the version without anything else. This way at least the version can be directly retrieved in a script. WDYT?

BTW, I like this change a lot. Well done!

ankitm123: I was thinking along the same lines, but we should think about a list of commands which should have the -o flag (or may be make it global), and what format we plan to support (json, yaml, toml etc..). It should be a separate ticket imo (There will be lot of changes), thoughts?

.github/issue_template.md
ankitm123: Ah, I see the error, I removed the `--no-version-check`, and it complains about that. I am going to add those flags I removed, back to the version command, and deprecate them. Also is there another command which does a version check? If not, I need to write one as part of this PR. I dont think `jx version` should really check if there are newer versions of jx, wdyt? Actually, should we even do a version check?

pkg/cmd/version/version.go
hferentschik: Another thing which comes to mind, I am wondering how useful the 'Branch' value is. While working on the codebase you would almost always be on a branch, in the pipeline, however, we are on a detached HEAD state. So I guess the branch name would become 'unkown'. This would, in particular, be the case for release builds which is probably not a good idea.

hferentschik: One more - maybe we can format the build date a bit nicer - '20200527-20:10:18', something along the lines of `2020-04-06T16:33:17Z`. Some separators in the date at least.

.github/issue_template.md
hferentschik: > Actually, should we even do a version check?

No, I don't think this is needed.

hferentschik: Should the command in the template just be `jx diagnose`? For `jx diagnose --show=version` I get:

```
jx diagnose --show=version
Flag --show has been deprecated, use jx diagnose <object> instead. This will be removed on July 1, 2020
Running in namespace: jx
NAME                          VERSION
jx                            2.1.60-dev+83850f2b5
Kubernetes cluster            v1.14.10-gke.36
kubectl (installed in JX_BIN) v1.13.2
helm client                   2.12.2
git                           2.20.1
Operating System              Mac OS X 10.14.6 build 18G4032
Error: unknown flag: --show
Usage:
  version [flags]

Flags:
  -h, --help    help for version
      --short   Print version

error: unknown flag: --show
```

Makefile
```diff
@@ -38,10 +39,10 @@ BUILD_TIME_CONFIG_FLAGS ?= ""
 # Full build flags used when building binaries. Not used for test compilation/execution.
 BUILDFLAGS :=  -ldflags \
   " -X $(ROOT_PACKAGE)/pkg/version.Version=$(VERSION)\
-		-X $(ROOT_PACKAGE)/pkg/version.Revision='$(REV)'\
-		-X $(ROOT_PACKAGE)/pkg/version.Branch='$(BRANCH)'\
-		-X $(ROOT_PACKAGE)/pkg/version.BuildDate='$(BUILD_DATE)'\
-		-X $(ROOT_PACKAGE)/pkg/version.GoVersion='$(GO_VERSION)'\
+		-X $(ROOT_PACKAGE)/pkg/version.Revision=$(REV)\
+		-X $(ROOT_PACKAGE)/pkg/version.BuildDate=$(BUILD_DATE)\
```
hferentschik, line 43: :+1

pkg/cmd/diagnose.go
```diff
@@ -44,60 +87,135 @@ func (o *DiagnoseOptions) Run() error {
...
+				return err
+			}
 		}
-	}
 
-	if o.showOption("pvc") {
-		err := printStatus(o, "Kubernetes PVCs", "kubectl", "get", "pvc", "--namespace", ns)
-		if err != nil {
-			return err
+		if o.showOption("status") {
```
hferentschik, line 112:
```jx diagnose --show=status
Flag --show has been deprecated, use jx diagnose <object> instead. This will be removed on July 1, 2020
Running in namespace: jx
NAME                          VERSION
jx                            2.1.60-dev+83850f2b5
Kubernetes cluster            v1.14.10-gke.36
kubectl (installed in JX_BIN) v1.13.2
helm client                   2.12.2
git                           2.20.1
Operating System              Mac OS X 10.14.6 build 18G4032
Error: unknown flag: --show
Usage:
  status [node] [flags]

Aliases:
  status, status

Examples:
  # displays the current status of the Kubernetes cluster
  jx status

Flags:
  -h, --help          help for status
  -n, --node string   the named node to get

error: unknown flag: --show
```

hferentschik: For pods and ingresses it works!?

pkg/cmd/diagnose.go
```diff
@@ -44,60 +87,135 @@ func (o *DiagnoseOptions) Run() error {
...
 	}
 
-	if o.showOption("status") {
-		err := printStatus(o, "Jenkins X Status", "jx", "status")
-		if err != nil {
-			return err
+	table.Render()
+
+	switch len(o.Args) {
+	//ToDo: This wont be required after the deprecated --show flag is removed
```
hferentschik, line 103: Maybe moving all the code which is temporary or can be later removed into a dedicated function? WDYT?

pkg/cmd/opts/version.go
```diff
@@ -84,7 +84,7 @@ func (o *CommonOptions) GetPackageVersions(ns string, helmTLS bool) (map[string]
 				if v != "" {
 					switch i {
 					case 0:
-						table.AddRow("kubectl", info(v))
+						table.AddRow("kubectl (installed in JX_BIN)", info(v))
```
hferentschik, line 87: +1, an important distinction :-)

pkg/cmd/version/version.go
```diff
@@ -0,0 +1,61 @@
+package version
```
hferentschik, line 1: +1 for moving into a dedicated package. It should be an ongoing task to whenever we touch commands to further decompose this massive _cmd_ package.

ankitm123: Agreed :100:

pkg/cmd/opts/version.go
ankitm123: I saw some confusion from new users in the community slack channel, so decided to make it clearer :)

hferentschik:
@ankitm123, just touching base on this pull request again. If they are open so long it is easy to loose track.

I'd like to merge this asap. I think the only comment which is currently unresolved is that the _version_ gets printed twice in `jx diagnose`. Have you seen https://github.com/jenkins-x/jx/pull/7092#pullrequestreview-426089248? 
If you want to address this at a later point that's fine with me as well. I think there is already so much improvement in this pull request that I could live with some "duplicated" version output ;-)

ankitm123:
@hferentschik I will get it done by EOD today.  I am looking at it now.

hferentschik:
> I will get it done by EOD today. I am looking at it now.

sweet

pkg/cmd/diagnose.go
ankitm123: makes sense, will do :+1:

ankitm123:
@hferentschik moved the deprecated show related code to a separate function, and fixed the issue with `unknown flag --show`. This was occurring for version and status options.

The reason we were getting that error was because --show was being passed to `jx version`. Previously, we used to execute `jx version --no-version-check` from the diagnose command, but instead a cleaner way to achieve this is to just use NewCmdVersion(o *options).Execute(). The only downside is that it passes the flags passed to the calling function. This was fixed by clearing the flags before executing version/status command.

Issue #7035, opened 2020-04-09
jx boot fails on Windows


ACB1337:
### Summary
When running jx boot on Windows it tries to execute /bin/sh in step validate-git. This obviously fails.

### Steps to reproduce the behavior
execute jx boot on Windows in powershell

### Expected behavior
set up jenkins x

### Actual behavior
```
Booting Jenkins X
WARNING: failed to load ConfigMap jenkins-x-docker-registry in namespace jx: failed to get configmap jenkins-x-docker-registry in namespace jx, configmaps "jenkins-x-docker-registry" not found
WARNING: failed to load ConfigMap jenkins-x-docker-registry in namespace jx: failed to get configmap jenkins-x-docker-registry in namespace jx, configmaps "jenkins-x-docker-registry" not found

STEP: validate-git command: /bin/sh -c jx step git validate in dir: E:\dev\environment-test-dev\workspace\source\workspace\source\env

error: failed to interpret pipeline file jenkins-x.yml: failed to run '/bin/sh -c jx step git validate' command in directory 'workspace\source\workspace\source\env', output: ''
```
### Jx version

The output of `jx version` is:

```
WARNING: Failed to retrieve team settings: failed to setup the dev environment for namespace 'default': the server could not find the requested resource (post environments.jenkins.io) - falling back to default settings...
NAME               VERSION
jx                 2.0.1274
Kubernetes cluster v1.14.10-gke.27
kubectl            v1.15.5
git                2.19.0.windows.1
Operating System   Windows 10 Pro 1809 build 17763
```

### Jenkins type

- [x] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

### Kubernetes cluster
GKE cluster tried with jx create gke --skip-install and manual creation

### Operating system / Environment
Windows 10/Powershell

nodece:
I need install `sh` into `Windows`.

nodece:
You need install `sh` into `Windows`.

nodece:
Assume installed `bash` or `sh` into `Windows`.

The following is working in `Windows 10/Powershell`:  

```
bash
jx create gke --skip-install 
```

ACB1337:
if you execute bash in Powershell you are starting WSL. So no it is not working on Windows it is working on Linux.

nodece:
Sorry @ACB1337 , I provide a wrong message.

We cannot run `/bin/sh` in `Windows`. Need to consider the `Windows` platform.

[...]

PR #7096, opened 2020-04-22
fix: use sh instead of /bin/sh in Windows

nodece:

#### Description
Due to windows file struct, We cannot found `/bin/sh` in Windows. 

fixes https://github.com/jenkins-x/jx/issues/7035

### DIFF
--- a/pkg/cmd/step/create/step_create_task.go
+++ b/pkg/cmd/step/create/step_create_task.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/step/create/step_create_task.go
```diff
@@ -1401,8 +1402,13 @@ func (o *StepCreateTaskOptions) runStepCommand(step *syntax.Step) error {
 
 	commandText := strings.Replace(step.GetFullCommand(), "\\$", "$", -1)
 
+	name := "/bin/sh"
+	if runtime.GOOS == "windows" {
+		name = "sh"
```
hferentschik, line 1407: I wonder whether it would be safe to always use "sh". This really should work on any OS. @abayer wdyt?

abayer: Seems possible? I’m a little wary, but I may just be excessively conservative.

nodece: Maybe we can use `cmd `.  WDYT?

hferentschik: not sure about `cmd`. How would that work? I was just trying to about the special case for Windows. A simple 'sh' should work on macOS and Linux as well. But maybe @abayer is right and we should be a bit more conservative here.

hferentschik: I don't have a way atm to verify that this actually works on Windows, so I guess we just see what happens.

PR #7104, opened 2020-04-23
feat: Should install binary checks at JX/bin location

### LINKED ISSUE #6912: refactor : ShouldInstallBinary function checking binary on $PATH, not behaving as expected.

daveconde:
- fixes #6912
- Adds --all option to jx install dependencies
- Removes unused binaries
- Drops brew usage where possible


Dockerfile.builder-go
```diff
@@ -2,3 +2,5 @@
 FROM gcr.io/jenkinsxio/builder-go:latest
 
 COPY build/linux/jx /usr/bin/jx
+
+RUN jx install dependencies --all
```
hferentschik, line 6: That I like :-)

pkg/cloud/amazon/cli.go
```diff
@@ -44,32 +38,3 @@ func InstallEksCtlWithVersion(version string, skipPathScan bool) error {
 func InstallEksCtl(skipPathScan bool) error {
 	return InstallEksCtlWithVersion("", skipPathScan)
 }
-
-// InstallKops installs kops
-func InstallKops() error {
```
hferentschik, line 49: +1 for removing

pkg/cmd/create/install_dependencies.go
```diff
@@ -37,19 +38,11 @@ var (
...
 		"helm3",
-		"kops",
-		"kvm",
-		"kvm2",
 		"ksync",
 		"oc",
-		"virtualbox",
-		"xhyve",
-		"hyperv",
 		"terraform",
```
hferentschik, line 45: I don't think we need `terraform` either. There is a `terraform.go`, but none of the functions is used, so that could be deleted at the same time. `jx` should not call `terraform`.

pkg/cmd/create/install_dependencies.go
```diff
@@ -37,19 +38,11 @@ var (
...
 		"gcloud",
 		"helm",
-		"ibmcloud",
 		"tiller",
 		"helm3",
-		"kops",
-		"kvm",
-		"kvm2",
 		"ksync",
 		"oc",
```
hferentschik, line 44: Ideally, we would not install this either imo. I think right now it is only used for some code for `jx install openshift`, but I doubt that is working. To keep things simple we might just keep it for now.

pkg/packages/packages.go
```diff
@@ -25,6 +25,21 @@ const IamAuthenticatorAwsVersion = "1.12.7"
...
+const Helm3Version = "3.2.0"
+
+// KustomizeVersion binary version to use
+const KustomizeVersion = "3.5.1"
+
+// GlooVersion binary version to use
+const GlooVersion = "1.3.23"
+
+// TerraformVersion binary version to use
+const TerraformVersion = "0.12.24"
```
hferentschik, line 41: I don't think we need that.

pkg/cmd/create/install_dependencies.go
```diff
@@ -37,19 +38,10 @@ var (
...
-		"virtualbox",
-		"xhyve",
-		"hyperv",
-		"terraform",
```
hferentschik, line 51: actually, I have second thoughts around this now. Right now it is unused, but there is a chance we utilize for the distribution. We might need to discuss first how we want to go about the distro.

hferentschik:
this pull request needs a rebase now

pkg/cmd/create/install_dependencies.go
daveconde: Lets drop it for now

PR #7108, opened 2020-04-24
feat: add authorAvatarUrl and authorURL to PA

romainverduci:


hferentschik:
@romainverduci, I think a `make fmt` would help in your case ;-)

pkg/cmd/controller/controller_build.go
```diff
@@ -435,6 +437,8 @@ func (o *ControllerBuildOptions) completeBuildSourceInfo(activity *v1.PipelineAc
 		if len(gitCommits) > 0 {
 			if gitCommits[0] != nil && gitCommits[0].Author != nil {
 				activity.Spec.Author = gitCommits[0].Author.Login
+				activity.Spec.AuthorAvatarURL = gitCommits[0].Author.AvatarURL
```
hferentschik, line 440: hmmm, I would be a bit more defensive on all this. Also with the line above. If you look at `GitCommit` is is a pointer and the `Author` is a pointer to a `GitUser`. This could can panic if `author` is `nil`. I am pretty sure that can happen and depends for example if the email you use for your commits it public on GitHub. 
We had issues like this before. I suggest to take height for potential `nil`s.

romainverduci: Thanks for the proactive review, I'll take that into account !
I'm first trying to make it work to fix my issues then I will make it good :)

hferentschik: I did not see that there is already a `nil` check here `if gitCommits[0] != nil && gitCommits[0].Author != nil` so things should be good.

hferentschik:
/lgtm

(would be nice with an issue as well :-))

PR #7114, opened 2020-04-27
fix: Add context to GitLab and BitBucket commit status functions

### LINKED ISSUE #7113: Include context in GitLab and BitBucket commit status function return values
The function that creates a "standard" commit status record from a GitLab one in jx doesn't include the "context" (in the case of GitLab and the two BitBuckets, the field is `name`, but it's the same as `context` in GitHub, etc). We should include that.

abayer:
#### Submitter checklist

#### Description

The only direct reason for this is the Lighthouse BDD tests, but it's still something we should have in there anyway.

fixes #7113

### DIFF
--- a/pkg/gits/gitlab.go
+++ b/pkg/gits/gitlab.go
@@ -515,6 +515,7 @@ func fromCommitStatus(status *gitlab.CommitStatus) *GitRepoStatus {
 	return &GitRepoStatus{
 		ID:          string(status.ID),
 		URL:         status.TargetURL,
+		TargetURL:   status.TargetURL,
 		State:       jxState,
 		Description: status.Description,
 		Context:     status.Name,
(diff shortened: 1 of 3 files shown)


abayer:
/hold

Holding until I'm sure I actually have all the GitLab provider changes needed for BDD.

pkg/gits/gitlab.go
```diff
@@ -292,9 +292,12 @@ func (g *GitlabProvider) UpdatePullRequest(data *GitPullRequestArguments, number
 	base := data.Base
 
 	o := &gitlab.UpdateMergeRequestOptions{
-		Title:        &title,
-		Description:  &body,
-		TargetBranch: &base,
```
mrageh, line 297: Why was the `TargetBranch` removed?

PR #7130, opened 2020-04-30
fix: Sleep five seconds after GitHub repo creation

### LINKED ISSUE #7129: Sporadic race condition on new GitHub repo creation and initial push to master
In our BDD tests, we've long seen errors pushing to a newly created GitHub repository very rarely, but in the last week, it's become much more frequent. The repo is created properly, but seems to not be ready to be pushed to quite as fast as we try to push to it.

abayer:
#### Submitter checklist

#### Description

This should prevent errors caused by GitHub not yet actually having the git repo proper ready and exposed.

fixes #7129

### DIFF
--- a/pkg/gits/github.go
+++ b/pkg/gits/github.go
@@ -343,6 +343,8 @@ func (p *GitHubProvider) CreateRepository(owner string, name string, private boo
 		}
 		return nil, errors.New(msg)
 	}
+	// Sleep 5 seconds to ensure repository exists enough to be pushed to.
+	time.Sleep(5 * time.Second)
 	return toGitHubRepo(name, owner, repo), nil
 }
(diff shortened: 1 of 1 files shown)


pkg/gits/github.go
```diff
@@ -343,6 +343,8 @@ func (p *GitHubProvider) CreateRepository(owner string, name string, private boo
 		}
 		return nil, errors.New(msg)
 	}
+	// Sleep 5 seconds to ensure repository exists enough to be pushed to.
+	time.Sleep(5 * time.Second)
```
hferentschik, line 347: hmmm, for what it's worth, imo sleeping is just a crutch. I'd try to do an active poll loop in some form.

Issue #7112, opened 2020-04-27
Checking if preview available for non HTTP applications


aure-olli:
### Summary 

We have gRPC applications, and we always have troubles with Jenkins-X trying the ingress in `jx preview`, using HTTP requests. Unfortunately, there is currently no way to disable this, except giving another name to the service.

I propose:
- An option to disable checking
- An option to set the protocol (HTTP, HTTPS, GRPC, GRPC-usecured, TCP, ...)
- An option to manually set the service to try
- An option to manually set the ingress to try
- Options to manually set the host, port and path to try
- Reading the `nginx.ingress.kubernetes.io/backend-protocol` annotation of the ingress to choose the protocol ([doc](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#backend-protocol))
- Trying to guess the protocol with the service port (443 -> HTTPS, 50051 -> GRPC, rest is HTTP)

### Steps to reproduce the behavior

Create a gRPC application, with a gRPC ingress (`nginx.ingress.kubernetes.io/backend-protocol: GRPC`) and make a preview.

### Expected behavior

`jx preview` should acknowledge that the gRPC server is running.

### Actual behavior

`jx preview` fails because it receives "502 Bad Gateway" from the ingress when trying HTTPS requests.

### Jx version

The output of `jx version` is:

```
NAME               VERSION
jx                 2.0.1243
Kubernetes cluster v1.14.10-gke.27
kubectl            v1.17.0
helm client        2.14.3
git                2.17.1
Operating System   Ubuntu 18.04.4 LTS
```

### Discussion

Does my solution seem reasonable ? I propose the options:
- `--skip-availability-check`
- `--availability-check-protocol` with values `HTTP`, `HTTPS`, `GRPC`, `GRPC-unsecured` and `TCP`
- `--availability-check-service`
- `--availability-check-ingress`
- `--availability-check-host`
- `--availability-check-port`
- `--availability-check-path`

aure-olli:
@aure-olli,  nice one. Makes sense. In the current situation, a `skip-availability-check` makes sense, since it is not a given that the "exposed service" is HTTP based. 

Would you mind creating an issue for just this PR and the `skip-availability-check` flag? The new issue could also reference #7112 for the bigger picture. 

[...]

PR #7143, opened 2020-05-05
feat: Allow to disable availability check in preview

aure-olli:
Priori to this commit, `jx preview` would systematically check for availability of the service, even if the latter wasn't designed to reply to HTTP requests, and consequently would fail.

This commits adds a `--skip-availability-check` flag to prevent this behavior.

This is a first step toward #7112

### DIFF
--- a/pkg/cmd/preview/preview.go
+++ b/pkg/cmd/preview/preview.go
@@ -105,6 +105,8 @@ type PreviewOptions struct {
 	PreviewHealthTimeoutDuration  time.Duration
 
 	HelmValuesConfig config.HelmValuesConfig
+
+	SkipAvailabilityCheck bool
 }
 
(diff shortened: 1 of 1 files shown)


hferentschik:
@aure-olli,  nice one. Makes sense. In the current situation, a `skip-availability-check` makes sense, since it is not a given that the "exposed service" is HTTP based. 

Would you mind creating an issue for just this PR and the `skip-availability-check` flag? The new issue could also reference #7112 for the bigger picture. 

Regarding #7112, I am wondering whether we need to discuss first a bit more about the expectations of a "preview". ATM, it is pretty much an exposed HTTP(S) service. I am wondering whether the assumption is not a bit limiting. WDYT?

aure-olli:
fix #7155

OK I created a related issue.

I move the discussion back to #7112

PR #7168, opened 2020-05-12
fix: Verify pre install fails with incorrect kubectl or helm version on the path

### LINKED ISSUE #7169: jx step verify preinstall fails due to helm version
### Summary 

Running `jx step verify preinstall` fails with `FATAL: Your current helm version v3 is not supported. Please downgrade to helm v2.` when helm 3 is on the path. JX now uses helm from ~/.jx/bin and should not fail in this case.

daveconde:
fixes #7169 

This PR installs helm and kubectl in to ~/.jx/bin as part of `jx step verify preinstall` it also forces ~/.jx/bin to the start of the path.

Signed-off-by: David Conde

### DIFF
--- a/pkg/cmd/step/verify/step_verify_preinstall.go
+++ b/pkg/cmd/step/verify/step_verify_preinstall.go
@@ -32,6 +32,7 @@ import (
 	"github.com/jenkins-x/jx/pkg/kube/cluster"
 	"github.com/jenkins-x/jx/pkg/kube/naming"
 	"github.com/jenkins-x/jx/pkg/log"
+	"github.com/jenkins-x/jx/pkg/packages"
 	"github.com/jenkins-x/jx/pkg/util"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -172,6 +173,11 @@ func (o *StepVerifyPreInstallOptions) Run() error {
 	}
 	log.Logger().Info("\n")
 
+	err = o.installMissingDependencies()
+	if err != nil {
+		return err
+	}
+
 	po := &StepVerifyPackagesOptions{}
 	po.CommonOptions = o.CommonOptions
 	po.Packages = []string{"kubectl", "git", "helm"}
@@ -276,6 +282,14 @@ func (o *StepVerifyPreInstallOptions) Run() error {
 	log.Logger().Infof("Cluster looks good, you are ready to '%s' now!", info("jx boot"))
 	fmt.Println()
 	return nil
+}
+
+// installMissingDependencies installs missing deps like helm and kubectl
+func (o *StepVerifyPreInstallOptions) installMissingDependencies() error {
+	var deps []string
+	deps = packages.AddRequiredBinary("kubectl", deps)
+	deps = packages.AddRequiredBinary("helm", deps)
+	return o.DoInstallMissingDependencies(deps)
 }
 
 // EnsureHelm ensures helm is installed
--- a/pkg/util/commands.go
+++ b/pkg/util/commands.go
@@ -266,19 +266,22 @@ func (c *Command) run() (string, error) {
 	return text, err
 }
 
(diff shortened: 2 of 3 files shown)


pkg/cmd/step/verify/step_verify_preinstall.go
```diff
@@ -172,6 +173,11 @@ func (o *StepVerifyPreInstallOptions) Run() error {
 	}
 	log.Logger().Info("\n")
 
+	err = o.installMissingDependencies()
```
hferentschik, line 176: do we really want to install or just say they are not installed? I guess what bugs me is that this is supposed to be a _verify_, not an install step.

pkg/util/commands.go
```diff
@@ -269,14 +269,17 @@ func (c *Command) run() (string, error) {
...
 	if mvnBinDir != "" {
 		answer += string(os.PathListSeparator) + mvnBinDir
 	}
 	for _, p := range paths {
-		answer += string(os.PathListSeparator) + p
+		if p != "" {
+			answer += string(os.PathListSeparator) + p
+		}
 	}
+	binDir, _ := JXBinLocation()
```
hferentschik, line 282: maybe _jxBinDir_ to make it a bit cleaner in the final line where you put things together

hferentschik: A test for this might not be bad either?

pkg/util/commands.go
```diff
@@ -269,14 +269,17 @@ func (c *Command) run() (string, error) {
...
-	answer := path + string(os.PathListSeparator) + binDir
 	mvnBinDir, _ := MavenBinaryLocation()
+	answer := ""
 	if mvnBinDir != "" {
 		answer += string(os.PathListSeparator) + mvnBinDir
 	}
 	for _, p := range paths {
-		answer += string(os.PathListSeparator) + p
+		if p != "" {
+			answer += string(os.PathListSeparator) + p
```
hferentschik, line 279: maybe assign this to something new like "customPaths" or "providedPath"? The last line of this function is quite confusing imo. It takes a while to understand what and how you put things together. WDYT?

pkg/util/commands.go
```diff
@@ -269,14 +269,17 @@ func (c *Command) run() (string, error) {
...
+		}
 	}
+	binDir, _ := JXBinLocation()
+	answer = binDir + string(os.PathListSeparator) + path + answer
```
hferentschik, line 283: I see, so the JX_HOME/bin was not added first, but after the main PATH. That is of course wrong.

pkg/util/commands.go
```diff
@@ -269,14 +269,17 @@ func (c *Command) run() (string, error) {
 // PathWithBinary Sets the $PATH variable. Accepts an optional slice of strings containing paths to add to $PATH
```
hferentschik, line 269: The godoc seems wrong. This function is not setting a PATH variable here.

pkg/cmd/step/verify/step_verify_preinstall.go
daveconde: I think the fact that they are an internal implementation detail for JX now that they are in ~/.jx/bin would make it ok to auto install. Is there a better location to do this you would suggest that would remove it out of the verify? Ideally longer terms we'd remove the need to call out to external binaries.

pkg/util/commands.go
daveconde: I'll make that a bit clearer to read and update the godoc

pkg/util/commands.go
daveconde: test is incoming, I've made the variables in here clearer

pkg/cmd/step/verify/step_verify_preinstall.go
hferentschik: The alternative is to fail with an error and let the user fix it. Given that we want to install into JX_HOME/bin it would not make much sense to just display `jx install dependencies`, so I guess it is the best we can do for now.

Issue #6374, opened 2019-12-21
jx edit buildpack does not update the dev repo


vfarcic:
### Summary 

After executing `jx edit buildpack` to use a forked repo, the changes are not reflected in the `dev` repo created with `jx boot`.

### Steps to reproduce the behavior

* Enter into the `dev` repo

```bash
jx edit buildpack \
    --url https://github.com/[REPLACE_WITH_YOUR_GITHUB_USER]/jenkins-x-kubernetes \
    --ref master \
    --batch-mode
```

### Expected behavior

The change of the buildpack address is stored in the `dev` repo.

### Actual behavior

The change of the buildpack address is NOT stored in the `dev` repo.

### Jx version

The output of `jx version` is:

```
NAME               VERSION
jx                 2.0.1093
Kubernetes cluster v1.15.4-gke.22
kubectl            v1.17.0
helm client        Client: v2.14.3+g0e7f3b6
git                2.20.1 (Apple Git-117)
Operating System   Mac OS X 10.14.6 build 18G2022
```

### Jenkins type

- [X] Serverless Jenkins X Pipelines (Tekton + Prow)
- [ ] Classic Jenkins

### Kubernetes cluster

Any

### Operating system / Environment

Any

hervelemeur:
Following [our Slack discussion](https://kubernetes.slack.com/archives/C9MBGQJRH/p1588181950110200?thread_ts=1588089186.068200&cid=C9MBGQJRH), I've also tried to configure jenkins-x with custom buildpacks, but it didn't work, no change at all.

```
$ jx edit buildpack -u https://github.com/my-org/jenkins-x-kubernetes
? Defaulting workload build pack: Kubernetes Workloads: Automated CI+CD with GitOps Promotion
Setting the team build pack to kubernetes-workloads repo: https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes.git ref: master
```

JX version:
```
NAME               VERSION
jx                 2.0.1284
Kubernetes cluster v1.15.7
kubectl            v1.17.3
helm client        2.16.3
git                2.21.1 (Apple Git-122.3)
Operating System   Mac OS X 10.15.3 build 19D76
```
On AKS cluster, manually created.

@vfarcic FYI

/remove-lifecycle rotten

vfarcic:
[...]

Issue #6303, opened 2019-12-09
Setting custom buildpacks for team using jx boot


babadofar:
### Summary 
jx boot does not have a way to set a custom buildpack .

 The settings in https://github.com/jenkins-x-charts/jxboot-resources/blob/master/jxboot-resources/templates/dev-env.yaml#L23-L24 needs to be parametrized. 

### Steps to reproduce the behavior
In `jx-requirements.yml`  set the url to the buildpack you need 

    teamSettings:
      buildPackRef: master
      buildPackUrl: https://myurl.com/jenkins-x-weirdness.git`

### Expected behavior
Jx will use the buildpacks from your rep
### Actual behavior
No change
### Jx version

The output of `jx version` is:

```
NAME               VERSION
jx                 2.0.1045
Kubernetes cluster v1.14.8-eks-b7174d
kubectl            v1.16.2
helm client        Client: v2.14.3+g0e7f3b6
git                2.18.0
Operating System   Mac OS X 10.14.6 build 18G1012

```

### Jenkins type

- [ X] Serverless Jenkins X Pipelines (Tekton + Prow)

### Kubernetes cluster
EKS, created using `jx create cluster `

sstubbs:
This would be useful as currently I'm setting the buildpack via `jx edit buildpack` which sometimes resets to default when nodes are replaced.

a-y-khan:
My organization would find this useful too

babadofar:
Looks like this has been fixed.
By this commit
https://github.com/jenkins-x-charts/jxboot-resources/commit/612eaa6b0a45148dbda8a2304b5cba2536b81c1b

You can set custom build pack location in jxboot-resources/values.yaml, as shown here https://github.com/jenkins-x-charts/jxboot-resources/blob/612eaa6b0a45148dbda8a2304b5cba2536b81c1b/jxboot-resources/values.yaml#L31

```  
gitops:
  dev:
...
    buildPackName: kubernetes-workloads
    buildPackRef: master
    buildPackUrl: https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes.git  
```


The documentation https://jenkins-x.io/docs/reference/components/build-packs/#creating-new-build-packs is misleading, however.

PR #7172, opened 2020-05-12
feat: Add buildpack config to requirements file

### LINKED ISSUE #7126: The command "jx edit buildpack" does nothing
### Summary 
Following [our Slack discussion](https://kubernetes.slack.com/archives/C9MBGQJRH/p1588181950110200?thread_ts=1588089186.068200&cid=C9MBGQJRH), I've tried to configure jenkins-x with custom buildpacks, but it didn't work, no change at all.

```
$ jx edit buildpack -u https://github.com/my-org/jenkins-x-kubernetes
? Defaulting workload build pack: Kubernetes Workloads: Automated CI+CD with GitOps Promotion
Setting the team build pack to kubernetes-workloads repo: https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes.git ref: master
```

- fork the jenkins-x-kubernetes buildpack repository
- execute this command:
```
$ jx edit buildpack -u https://github.com/my-org/jenkins-x-kubernetes
```

### Expected behavior
JX is configured to use the custom buildpacks, and this configuration is saved in the jx boot dev repository.
### Actual behavior
Nothing happens.
### Jx version

```
NAME               VERSION
jx                 2.0.1284
Kubernetes cluster v1.15.7
kubectl            v1.17.3
helm client        2.16.3
Operating System   Mac OS X 10.15.3 build 19D76
```
### Kubernetes cluster

AKS cluster, manually created.

@vfarcic FYI
Relate to https://github.com/jenkins-x/jx/issues/6374

mrageh:
#### Submitter checklist

#### Description
When we moved JX over to boot and Helm 2, a regression was introduced where `jx edit buildpack` stopped working. Because JX team settings were no longer used to populate the team settings for a CRD. So the default buildpack info from the values.yaml file were used.

This commit introduces the build pack configuration to the requirements file, which will allow our users to specify custom build packs. We will deprecate the `jx edit buildpack` command in a future commit.

fixes #7126 and #6374


hferentschik:
@mrageh - I guess this is still wip atm, right? Adding fields to the requirements alone will not solve any issue, right?

pkg/config/install_requirements.go
```diff
@@ -243,6 +243,16 @@ type IngressConfig struct {
 	DomainIssuerURL string `json:"domainIssuerURL,omitempty"`
 }
 
+// BuildPackConfig contains build pack info
+type BuildPackConfig struct {
+	// BuildPackName
+	BuildPackName string `json:"buildPackName,omitempty"`
```
hferentschik, line 249: I am not a fan of `BuildPackName`. I think this is ambiguous and misleading (`jx edit buildpack` as well for that matter). A buildpack is the configuration used for building a given type of project. For example, when importing a Go project, you end up with a `jenkins-x.yaml` which looks like that:

```yaml
buildPack: go
```

See also https://jenkins-x.io/docs/reference/components/build-packs/.

When it comes to this `BuildPackConfig` we afaiu are not talking about configuring a specific buildpack, but rather change the lookup location for buildpacks in general, right? So we are more talking about a `BuildPackGroup` I would think. Or maybe `BuildPackLocation[Name|URL|Ref]`

I understand this is taking the `jx edit buildpack` concept 1-to-1 into requirements, but I don't think this is a good idea.

Maybe worth having a wider discussion around this.

dgozalo: These keys are already being used to configure the dev environment in https://github.com/jenkins-x-charts/jxboot-resources/blob/master/jxboot-resources/values.yaml#L43-L45.

This is just making it configurable via `jx-requirements.yml`.

There's some disparity, yes. The `buildPackName` in `jxboot-resources` refers to the name of the **group** where the **buildPack** property in a given `jenkins-x.yml` file refers to a given pack within the group.

If we change this to `buildPackGroup`, we'd need to change the property in `jxboot-resources` too.

mrageh: @hferentschik I think what you suggested is something we should do in a follow up PR. But not this one as it's moving existing functionality to the `jx-requirements.yml` file.

mrageh:
@hferentschik we can merge this PR as is, it will resolve our issue once I make changes the `jx-bootresources` and `jx-boot-config` repos

pkg/config/install_requirements.go
hferentschik: >  I think what you suggested is something we should do in a follow-up PR. But not this one as it's moving existing functionality to the jx-requirements.yml file.

but here is exactly the rub, we cannot just change the requirements at will. This is the interface/API to our users. We have already seen how hard it is to change things. We have deprecated fields, badly grouped fields, custom serialization to name just a few things.

hferentschik: >  I think what you suggested is something we should do in a follow up PR. But not this one as it's moving existing functionality to the jx-requirements.yml file.

So just by adding these fields here, it is already configurable? How so? Are the values of the requirements file interpolated into the charts via the Helm templating?

hferentschik: @mrageh at least we should have a wider discussion around this. What would it take to rename in the Helm chart? Can this be done without breaking things?

@dgozalo 
> If we change this to buildPackGroup, we'd need to change the property in jxboot-resources too.
Can this be done?

hferentschik: There is imo another discussion we should have. I initially thought that if we add this information into `jx-requirements.yaml` we would the consume it somehow directly from there. What this seems to do is to somehow render this into the TeamSettings. We also add the requirements itself to the TeamSettings and then we look it up for some things and extract values from the stored requirements. At the same time, we also update the TeamSettings via interpolation into the jxboot-resources Chart. That is imo confusing, to say the least.

hferentschik:
> we can merge this PR as is, it will resolve our issue once I make changes the jx-bootresources and jx-boot-config repos

If this were an internal implementation detail sure, but that's not the case here.

pkg/config/install_requirements.go
mrageh: @hferentschik let's hold of on this PR for now as you are right, the `jx-requirements.yml` is our public api and we should hold off on merging this.

hferentschik:
I tried to summarize my thoughts around this issue here - https://github.com/jenkins-x/jx/issues/7126#issuecomment-628537697

mrageh:
@hferentschik Updated the PR based on your comment here - https://github.com/jenkins-x/jx/issues/7126#issuecomment-628537697

pkg/config/install_requirements.go
```diff
@@ -243,6 +243,22 @@ type IngressConfig struct {
 	DomainIssuerURL string `json:"domainIssuerURL,omitempty"`
 }
 
+// BuildPackConfig contains build pack info
+type BuildPackConfig struct {
```
hferentschik, line 247: +1

pkg/config/install_requirements.go
```diff
@@ -243,6 +243,22 @@ type IngressConfig struct {
...
+	// Location contains location config
+	Location LocationConfig `json:"location,omitempty"`
+}
+
+// LocationConfig contains buildpack location
+type LocationConfig struct {
+	// Name
+	Name string `json:"name,omitempty"`
+	// GitURL
```
hferentschik, line 256: +1

hferentschik:
@mrageh - can you run `make generate` as well?

pkg/config/install_requirements.go
```diff
@@ -243,6 +243,22 @@ type IngressConfig struct {
...
 }
 
+// BuildPackConfig contains build pack info
+type BuildPackConfig struct {
+	// Location contains location config
+	Location LocationConfig `json:"location,omitempty"`
+}
+
+// LocationConfig contains buildpack location
+type LocationConfig struct {
```
hferentschik, line 253: After the discussion with @jstrachan yesterday, I think that BuildPackLibrary is the best term for what we are trying to model here. 

It would btw, be nice to add a test to _install_requirements_test.go which shows how these structs serialize. The naming of these structs is one thing, but how this actually is written into jx-requirements.yaml is driven by the annotations.

hferentschik: Did we not say we are going to use the term library?

mrageh: yup PR has been updated, just pushed passing test up

pkg/buildnum/.tmp_gomock_reflect_199170717/prog.go
```diff
@@ -0,0 +1,46 @@
+
+package main
```
hferentschik, line 2: I think you added accidentally a temporary file from code generation

hferentschik: Actually a whole bunch of them

pkg/config/install_requirements.go
```diff
@@ -243,6 +243,22 @@ type IngressConfig struct {
...
 }
 
+// BuildPackConfig contains build pack info
+type BuildPackConfig struct {
+	// Location contains location config
+	BuildPackLibrary BuildPackLibrary `json:"buildPackLibrary,omitempty"`
+}
+
+// BuildPackLibrary contains buildpack location
+type BuildPackLibrary struct {
```
hferentschik, line 253: :-) 👍

pkg/config/test_data/build_pack_library.yaml
```diff
@@ -0,0 +1,5 @@
+buildPack:
+  buildPackLibrary:
```
hferentschik, line 2: Sweet, thanks for the test

pkg/config/install_requirements.go
```diff
@@ -473,6 +489,8 @@ type RequirementsConfig struct {
 	AutoUpdate AutoUpdateConfig `json:"autoUpdate,omitempty"`
 	// BootConfigURL contains the url to which the dev environment is associated with
 	BootConfigURL string `json:"bootConfigURL,omitempty"`
+	// BuildPackConfig contains custom build pack settings
+	BuildPack BuildPackConfig `json:"buildPack,omitempty"`
```
hferentschik, line 493: what's about making this 'buildPacks' opposed to 'buildPack' - making it clearer that this is the configuration for build packs in general and not for a given one.

hferentschik:
@mrageh - linting fails due to the checked in top files from code generation.

pkg/buildnum/.tmp_gomock_reflect_199170717/prog.go
mrageh: whoops

mrageh: removed them

pkg/config/install_requirements.go
mrageh: I spoke to @dgozalo about this and we both thought `buildPack` made more sense as it can only hold a single thing in this current time period.

mrageh: So the plural name can be misleading, but in the future it could make more sense if we start supporting multiple build packs in the future.

hervelemeur:
@hferentschik see this comment: https://github.com/jenkins-x/jx/issues/6303#issuecomment-630213155

There is apparently already a way to specify it in the dev repository.

pkg/config/install_requirements.go
hferentschik: The way I saw it was that we have configuration for build packs in general under this node. ATM, that would only be the build pack library location. So for me, "build pack" is in fact a bit misleading, since this is not a single build pack thing, but an overarching setting for build packs. But let's go for it if you think so.

mrageh: Agreed and changed

hferentschik:
might be worse to do a manual rebase as well to make sure things are all good with the go 1.13 upgrade

hferentschik:
@abayer, @mrageh - a bit unclear to me atm what exactly goes wrong with these two failing contexts. Any ideas?

mrageh:
None at all, here's the error message 

```
WARNING: Failed to perform tests on cluster pr-7172-95-boot-vault: failed to run 'make test-quickstart-golang-http test-app-lifecycle' command in directory '/tmp/jenkins-x/bdd-**', output: ''
error: failed to run 'make test-quickstart-golang-http test-app-lifecycle' command in directory '/tmp/jenkins-x/bdd-**', output: ''
```

```
error: failed to run 'make test-quickstart-golang-http' command in directory '/tmp/jenkins-x/bdd-**', output: ''
Pipeline failed on stage 'e2e-tests' : container 'step-e2e-tests'. The execution of the pipeline has stopped.
```

abayer:
@hferentschik fyi, logs weren't being stored properly - I fixed that with https://github.com/jenkins-x/environment-tekton-weasel-dev/pull/598 and, in case anyone else ever matches the weird bespoke setup we had on weasel, https://github.com/jenkins-x/jx/pull/7214.

abayer:
So the problem seems to be that the staging pipeline is failing on the `setup-jx-git-credentials` step because it's using the version stream to get the image version (via https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes/blob/master/packs/charts/pipeline.yaml#L3) and that's not overridden by the magic for tweaking things in `jenkins-x-platform` to use the PR build's image. We've got a bit of a chicken and the egg problem here.

pkg/config/install_requirements.go
```diff
@@ -243,6 +243,22 @@ type IngressConfig struct {
 	DomainIssuerURL string `json:"domainIssuerURL,omitempty"`
 }
 
+// BuildPackConfig contains build pack info
+type BuildPackConfig struct {
+	// Location contains location config
+	BuildPackLibrary BuildPackLibrary `json:"buildPackLibrary,omitempty"`
```
abayer, line 249: Aaaaah - re my comment about the staging pipeline using the version stream image and failing with 
```
Showing logs for build cb-kubecd/environment-pr-7172-113-boot-vault-staging/master #1 promotion stage from-build-pack and container step-setup-jx-git-credentials
error: error getting Requirements from TeamSettings to determine if in GitHub app mode: validation failures in requirements from team settings:
(root): Additional property buildPacks is not allowed
```

The requirements has
```
      buildPacks:
        buildPackLibrary: {}
```
in it, because `BuildPackLibrary` is never "empty" per se, since it's not a pointer. So I think if this is changed to a pointer, it won't be put in the serialized requirements file if there's nothing in it.

abayer: ^^^ @hferentschik - does that sound right?

pkg/config/install_requirements.go
```diff
@@ -473,6 +489,8 @@ type RequirementsConfig struct {
 	AutoUpdate AutoUpdateConfig `json:"autoUpdate,omitempty"`
 	// BootConfigURL contains the url to which the dev environment is associated with
 	BootConfigURL string `json:"bootConfigURL,omitempty"`
+	// BuildPackConfig contains custom build pack settings
+	BuildPacks BuildPackConfig `json:"buildPacks,omitempty"`
```
abayer, line 493: Actually, this should be a pointer too.

hferentschik:
> So the problem seems to be that the staging pipeline is failing on the setup-jx-git-credentials step because it's using the version stream to get the image version (via https://github.com/jenkins-x-buildpacks/jenkins-x-kubernetes/blob/master/packs/charts/pipeline.yaml#L3) and that's not overridden by the magic for tweaking things in jenkins-x-platform to use the PR build's image. We've got a bit of a chicken and the egg problem here.

So why did it work for #7211

pkg/config/install_requirements.go
hferentschik: jupp, good catch.

pkg/config/install_requirements.go
hferentschik: I think this being the pointer is key

abayer: Yeah, but it's best practices to make any `omitempty` field that's not a string/bool/int/etc be a pointer anyway. =)

pkg/config/install_requirements.go
hferentschik: > . So I think if this is changed to a pointer, it won't be put in the serialized requirements file if there's nothing in it.

I think it is actually `BuildPackConfig` which needs to be a pointer. See also `AzureConfig` and `GKEConfig`. Easy to test. Create an empty `RequirementsConfig`, serialize it, deserialize it and ensure that the ``BuildPackConfig` is nil. And for good measure one can assert against the generated file as well.

hferentschik: @mrageh - does that make sense? Please add the test as well, to make sure we don't end up with another surprise.

pkg/config/install_requirements.go
hferentschik: > Yeah, but it's best practices to make any omitempty field that's not a string/bool/int/etc be a pointer anyway.

agreed. All should be pointers really. The GitHub lib does a better job there.

pkg/config/install_requirements.go
hferentschik: @mrageh - seems tests are passing now. Are you still working on this additional test or do you want to merge w/o?

PR #7173, opened 2020-05-12
fix: Various BitBucket Server fixes

### LINKED ISSUE #6828: Error when parsing Git URL on Bitbucket Server
### Summary 

Some commands like `jx import` or `jx create devpod` need to clone a Git repo in order to work.  
If you use Bitbucket Server, this repo URL isn't parsed correctly in this step.

In the Bitbucket web interface, the clone URL is something like `https://git.***.fr/scm/<PROJECT>/<REPO>.git`.  
These commands are failing because they don't have the right clone URL (it is missing the `/scm` part) : `fatal: repository 'https://git.***.fr/<PROJECT>/<REPO>.git/' not found`.

I think it is related to this line in the code : https://github.com/jenkins-x/jx/blob/1976d34ff6ac7df0c8874c9f99c32586e6f3b524/pkg/gits/git_url.go#L203-L204

I tested these cases : install Jenkins X with Bitbucket Server and either : 
- run `jx create devpod` 
- run `jx import` with `webhook: lighthouse`

### Expected behavior

The repository is correctly cloned.

### Actual behavior

It fails with an error like 

```
Cloning into '<REPO>'...
fatal: repository 'https://git.***.fr/<PROJECT>/<REPO>.git/' not found
command terminated with exit code 128
```

### Jx version

```
NAME               VERSION
jx                 2.0.1202
Kubernetes cluster v1.17.2
kubectl            v1.16.0
helm client        Client: v2.16.3+g1ee0254
git                2.17.1
Operating System   Ubuntu 18.04.4 LTS
```

### Jenkins type

- [ ] Classic Jenkins

### Kubernetes cluster

On premise

```
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.4 LTS
Release:	18.04
Codename:	bionic
```

### LINKED ISSUE #7184: GitProvider BitBucket Server doesn't include pull title in pull request
We just don't set the title. Easy fix.

### LINKED ISSUE #7187: BitBucket Server owner/organisation/project usage is inconsistent
Particularly in relation to other providers (at least GitHub and GitLab), where "owner" is taken to mean "organization", and `Project` is generally not used. So let's clean that up.

### LINKED ISSUE #7188: BitBucket Server GetPullRequest doesn't set a bunch of fields
`UpdatePullRequestStatus` does, but not `GetPullRequest`.

abayer:
#### Submitter checklist

#### Description

* Makes sure that whenever we call `*GitRepository.HttpCloneURL`, we pass the git kind to it so that we can determine if the clone URL needs `/scm` added to it for BitBucket Server.
* Include the PR title in the returned `GitPullRequest` from various BitBucket Server functions.
* Make use of owner/org/project in BitBucket Server provider consistent with GitHub and GitLab providers.
* Populate PR data properly.

fixes #6828 
fixes #7184 
fixes #7187 
fixes #7188

### DIFF
--- a/pkg/cmd/create/create_devpod.go
+++ b/pkg/cmd/create/create_devpod.go
@@ -278,7 +278,7 @@ func (o *CreateDevPodOptions) Run() error {
 			log.Logger().Warnf("could not find git URL in dir %s due to: %s", dir, err.Error())
 		} else {
 			gitKind, _ := o.GitServerKind(gitInfo)
-			importURL = gitInfo.HttpCloneURL(gitKind)
+			importURL = gits.HttpCloneURL(gitInfo, gitKind)
 		}
 	}
 	label := o.Label
--- a/pkg/cmd/create/create_devpod.go
+++ b/pkg/cmd/create/create_devpod.go
@@ -277,7 +277,8 @@ func (o *CreateDevPodOptions) Run() error {
 		if err != nil {
 			log.Logger().Warnf("could not find git URL in dir %s due to: %s", dir, err.Error())
 		} else {
-			importURL = gitInfo.HttpCloneURL()
+			gitKind, _ := o.GitServerKind(gitInfo)
+			importURL = gitInfo.HttpCloneURL(gitKind)
 		}
 	}
 	label := o.Label
--- a/pkg/cmd/step/step_custom_pipeline.go
+++ b/pkg/cmd/step/step_custom_pipeline.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/jenkins-x/jx/pkg/cmd/opts/step"
+	"github.com/jenkins-x/jx/pkg/gits"
 
 	"github.com/jenkins-x/jx/pkg/cmd/helper"
 
@@ -160,7 +161,7 @@ func (o *StepCustomPipelineOptions) Run() error {
 				return err
 			}
 		} else {
-			gitURL := gitInfo.HttpCloneURL(gitKind)
+			gitURL := gits.HttpCloneURL(gitInfo, gitKind)
 			log.Logger().Infof("Using git URL %s and branch %s", util.ColorInfo(gitURL), util.ColorInfo(branch))
 
 			err = o.Retry(3, time.Second*10, func() error {
--- a/pkg/cmd/step/step_custom_pipeline.go
+++ b/pkg/cmd/step/step_custom_pipeline.go
@@ -106,6 +106,11 @@ func (o *StepCustomPipelineOptions) Run() error {
 		return err
 	}
 
+	gitKind, err := o.GitServerKind(gitInfo)
+	if err != nil {
+		return err
+	}
+
 	branch, err := o.Git().Branch(o.Dir)
 	if err != nil {
 		return err
@@ -155,7 +160,7 @@ func (o *StepCustomPipelineOptions) Run() error {
 				return err
 			}
 		} else {
-			gitURL := gitInfo.HttpCloneURL()
+			gitURL := gitInfo.HttpCloneURL(gitKind)
 			log.Logger().Infof("Using git URL %s and branch %s", util.ColorInfo(gitURL), util.ColorInfo(branch))
 
 			err = o.Retry(3, time.Second*10, func() error {
--- a/pkg/gits/bitbucket_server.go
+++ b/pkg/gits/bitbucket_server.go
@@ -1143,8 +1143,11 @@ func (b *BitbucketServerProvider) GetRelease(org string, name string, tag string
 }
 
 func (b *BitbucketServerProvider) AddCollaborator(user string, organisation string, repo string) error {
-	log.Logger().Infof("Automatically adding the pipeline user as a collaborator is currently not implemented for bitbucket. Please add user: %v as a collaborator to this project.", user)
-	return nil
+	options := make(map[string]interface{})
+	options["name"] = user
+	options["permission"] = "REPO_WRITE"
+	_, err := b.Client.DefaultApi.SetPermissionForUser(organisation, repo, options)
+	return err
 }
 
 func (b *BitbucketServerProvider) ListInvitations() ([]*github.RepositoryInvitation, *github.Response, error) {
--- a/pkg/gits/bitbucket_server.go
+++ b/pkg/gits/bitbucket_server.go
@@ -473,10 +473,17 @@ func getMergeCommitSHAFromPRActivity(prActivity map[string]interface{}) *string 
 	var activity []map[string]interface{}
 	var mergeCommit map[string]interface{}
 
-	mapstructure.Decode(prActivity["values"], &activity)     //nolint:errcheck
-	mapstructure.Decode(activity[0]["commit"], &mergeCommit) //nolint:errcheck
-	commitSHA := mergeCommit["id"].(string)
-
+	mapstructure.Decode(prActivity["values"], &activity) //nolint:errcheck
+	for _, act := range activity {
+		if act["action"] == "MERGED" {
+			mapstructure.Decode(act["commit"], &mergeCommit) //nolint:errcheck
+			break
+		}
+	}
+	commitSHA := ""
+	if _, ok := mergeCommit["id"]; ok {
+		commitSHA = mergeCommit["id"].(string)
+	}
 	return &commitSHA
 }
 
--- a/pkg/gits/bitbucket_server.go
+++ b/pkg/gits/bitbucket_server.go
@@ -492,7 +492,6 @@ func getLastCommitFromPRCommits(prCommits map[string]interface{}) *bitbucket.Com
 
 func (b *BitbucketServerProvider) UpdatePullRequestStatus(pr *GitPullRequest) error {
 	var bitbucketPR bitbucket.PullRequest
-	var prCommits, prActivity map[string]interface{}
 
 	prID := *pr.Number
 	projectKey, repo := parseBitBucketServerURL(pr.URL)
@@ -506,45 +505,15 @@ func (b *BitbucketServerProvider) UpdatePullRequestStatus(pr *GitPullRequest) er
 		return err
 	}
 
-	pr.State = &bitbucketPR.State
-	pr.Title = bitbucketPR.Title
-	pr.Body = bitbucketPR.Description
-	pr.Author = &GitUser{
-		Login: bitbucketPR.Author.User.Name,
-	}
-
-	if bitbucketPR.State == "MERGED" {
-		merged := true
-		pr.Merged = &merged
-		apiResponse, err := b.Client.DefaultApi.GetPullRequestActivity(projectKey, repo, prID)
-		if err != nil {
-			return err
-		}
-
-		err = mapstructure.Decode(apiResponse.Values, &prActivity)
-		if err != nil {
-			return err
-		}
-		pr.MergeCommitSHA = getMergeCommitSHAFromPRActivity(prActivity)
-	}
-	diffURL := bitbucketPR.Links.Self[0].Href + "/diff"
-	pr.DiffURL = &diffURL
-
-	apiResponse, err = b.Client.DefaultApi.GetPullRequestCommits(projectKey, repo, prID)
+	err = b.populatePullRequest(pr, &bitbucketPR)
 	if err != nil {
 		return err
 	}
-	err = mapstructure.Decode(apiResponse.Values, &prCommits)
-	if err != nil {
-		return err
-	}
-	pr.LastCommitSha = getLastCommitSHAFromPRCommits(prCommits)
-
 	return nil
 }
 
 func (b *BitbucketServerProvider) GetPullRequest(owner string, repo *GitRepository, number int) (*GitPullRequest, error) {
-	var bPR bitbucket.PullRequest
+	var bPR *bitbucket.PullRequest
 
 	apiResponse, err := b.Client.DefaultApi.GetPullRequest(strings.ToUpper(owner), repo.Name, number)
 	if err != nil {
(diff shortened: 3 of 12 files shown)


abayer:
Woo, verified that this fixes `jx create quickstart`. Should be ready to merge.

pkg/cmd/create/create_devpod.go
```diff
@@ -277,7 +277,8 @@ func (o *CreateDevPodOptions) Run() error {
 		if err != nil {
 			log.Logger().Warnf("could not find git URL in dir %s due to: %s", dir, err.Error())
 		} else {
-			importURL = gitInfo.HttpCloneURL()
+			gitKind, _ := o.GitServerKind(gitInfo)
```
hferentschik, line 280: That really looks quite silly. So you pass _gitInfo_ into something to get the _gitKind_ which you then pass into _gitInfo_ to get the clone URL. To me, that shows that something is not quite right here. 

I feel `gitInfo.HttpCloneURL(gitKind)` is really a bad idea. I would expect `GitRepository` to be exactly the thing which hides/abstracts this from me. 

So what's about for example if we back the kind into the (various) "constructor" functions of `GitRepository`, in this case, `o.FindGitInfo(dir)`. It could be `o.GitInfoDir(dir, kind)`. 

BTW, there is nothing in the git config dir which would indicate/help us to determine whether we are dealing with BitBucket, is it? Also, if the directory got cloned via the HTTP URL to start with, we are able to even already handle this when we parse the clone URL from the git config, right? The issue would be if the clone URL in the git config is ssh based and we want to convert it to a `HttpCloneURL`. Is that correct?

BTW, is this 'scm' path segment BitBucket version dependent? Assuming BitBucket lies behind this URL - https://bitbucket.org/m2m/cumulocity-examples/src/develop/ - I am wondering why there is no 'scm' there? I am also wondering whether one can generally say "if bitbucket, then add 'scm'"?

pkg/cmd/step/step_custom_pipeline.go
```diff
@@ -106,6 +106,11 @@ func (o *StepCustomPipelineOptions) Run() error {
 		return err
 	}
 
+	gitKind, err := o.GitServerKind(gitInfo)
```
hferentschik, line 109: OOI, how does this work under the hood. Stepping into the code it first seems to look at the URL host to see whether it is one of the known git hosting sites, but then it continues to do `GetServiceKindFromSecrets` where I start wondering.

abayer: The important bit at this point is https://github.com/jenkins-x/jx/blob/master/pkg/kube/git_services.go#L99 and below - checking `jx-auth-config` to see if we’ve got creds for the particular host and if so, what provider kind it is. It does first check if it’s a known hosting provider, and after the auth config check falls back further to the secrets (no longer relevant in modern installs thanks to the auth config) and finally a super old legacy check for a CRD.

pkg/cmd/create/create_devpod.go
abayer: Yeah, there’s nothing in the git config. It’s fairly infuriating. BitBucket Cloud (I.e., bitbucket.org) doesn’t have the `scm` bit, but BitBucket Server always does.

The reason for the silliness of passing `gitInfo` to get the kind and then passing the kind to `gitInfo` is me trying to do minimal changes and leveraging the existing https://github.com/jenkins-x/jx/blob/master/pkg/cmd/opts/git.go#L174 in command options. That said, I still had to hack around that in one case, so lemme rethink this a bit.

abayer: Blech, I can't come up with a consistent way to pull this off other than this. In _most_ cases, we're going to be getting the right `https` clone URL from the git config, but not always - someone could clone with `git@` etc, at which point we're hosed if we don't do this. We'd still need to generate an http clone URL in those cases without any way of knowing that it needs the `scm`. We also don't want to be checking against the config map/secrets/etc every time we parse a git URL, so I don't think we can have `ParseGitURL` set the kind.

Sooooo I think this is what we're stuck with.

hferentschik: > BitBucket Cloud (I.e., bitbucket.org) doesn’t have the scm bit, but BitBucket Server always does.

And we can differentiate this? `KindBitBucketServer` != BitBucket Cloud?

abayer: Yup, `KindBitBucketServer` and `KindBitBucketCloud` are different things.

hferentschik: > We also don't want to be checking against the config map/secrets/etc every time we parse a git URL, so I don't think we can have ParseGitURL set the kind.

But in most cases I see, the parking of the URL is just a few lines above the line you use to determine the kind, right?
We could have multiple ParseURL methods w/ amd w/o the kind. Internally we store the kind and handle it if it is set otherways we use some sort of default. 
Personally, I have a bad feeling about this. I would poke at this a bit more, but if you think it should be merged, let's do it.

pkg/cmd/step/step_custom_pipeline.go
hferentschik: > to see if we’ve got creds for the particular host and if so, what provider kind it is.

and where does this kind come from? It is not determined in some way, it must be provided, right?

abayer: In the end, it comes from when git creds are defined and configuration in `jx-requirements.yml` driving that.

hferentschik: > In the end, it comes from when git creds are defined and configuration in jx-requirements.yml driving that.

would it not more direct to go via the requirements then?  That's a different discussion really, just wondering. It seems a bit odd to me to go all the way to the secrets and whatnot, but still ending up with some indirect information.

pkg/cmd/create/create_devpod.go
hferentschik: > I'm definitely open to alternatives, I just haven't been able to find something that's consistent and correct.

Here is another idea. See whether you like this. The rub is really imo that the separation of concerns is really bad here. What if one were to remove `HttpCloneURL` from `GitRepository` (only called 5 times really as we also see here in this PR). Then we make GitKind a proper type (which is in itself probably a good thing). It still could be a simple string, but making it a type we can add methods on it. We can then move `HttpCloneURL` to `GitKind` passing the GitInfo itself or the parts which are needed to construct the URL. That might be a better flow. WDYT?

abayer: Hmmm...I'm not sure we really want to switch git kind to a real type - seems like that may be more complexity than is worth adding for this. But moving `HttpCloneURL` to a standalone function taking a `GitRepository` and a git kind as arguments does seem more than reasonable.

hferentschik: > But moving HttpCloneURL to a standalone function taking a GitRepository and a git kind as arguments does seem more than reasonable.

Sounds like a possibility as well

hferentschik:
@abayer what's the state on this now?

abayer:
@hferentschik Should be ready to merge.

pkg/gits/git_url.go
```diff
@@ -26,7 +26,15 @@ func (i *GitRepository) PullRequestURL(prName string) string {
 }
 
 // HttpCloneURL returns the HTTPS git URL this repository
-func (i *GitRepository) HttpCloneURL() string {
+func (i *GitRepository) HttpCloneURL(kind string) string {
```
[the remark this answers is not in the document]
hferentschik: Thought you wanted to extract this out?

pkg/gits/bitbucket_server.go
```diff
@@ -370,7 +373,7 @@ func (b *BitbucketServerProvider) CreatePullRequest(data *GitPullRequestArgument
 			"repository": map[string]interface{}{
 				"slug": data.GitRepository.Name,
 				"project": map[string]interface{}{
-					"key": data.GitRepository.Project,
+					"key": strings.ToUpper(data.GitRepository.Organisation),
```
hferentschik, line 376: Why the need for capitalization? Maybe a comment would be nice somewhere why this is needed.
That said, I find this interesting. Here seems to be a provider (kind) specific behaviour again. If the org in case of BitBucket always needs to be upper-cased, it seems to be quite fragile having to remember to upper case everywhere where `data.GitRepository.Organisation` is used. Really `GitRepository` should either be aware of, or as with the HTTPCloneURL the `GitKind` should know how to handle this.

pkg/gits/bitbucket_server.go
```diff
@@ -658,7 +680,7 @@ func (b *BitbucketServerProvider) GetPullRequestCommits(owner string, repository
 	paginationOptions["start"] = 0
 	paginationOptions["limit"] = pageLimit
 	for {
-		apiResponse, err := b.Client.DefaultApi.GetPullRequestCommitsWithOptions(repository.Project, repository.Name, number, paginationOptions)
+		apiResponse, err := b.Client.DefaultApi.GetPullRequestCommitsWithOptions(strings.ToUpper(owner), repository.Name, number, paginationOptions)
```
hferentschik, line 683: I am a bit confused by switching from 'Project' to 'owner'. Or in another case to 'organisation'. Has there been a change in the API or was the existing code never working?

pkg/gits/bitbucket_server_test.go
```diff
@@ -50,9 +50,21 @@ var bitbucketServerRouter = util.Router{
 		"GET": "pr.json",
 		"PUT": "pr.json",
 	},
+	"/rest/api/1.0/projects/TEST-ORG/repos/test-repo/pull-requests/7": util.MethodMap{
```
hferentschik, line 53: Orgs are always capitalized in BitBucket? Maybe we can add a link to the API docs or something. What happens if the org is not capitalized?

pkg/gits/bitbucket_server.go
abayer: The original code didn’t work consistently, yeah.

pkg/cmd/step/step_custom_pipeline.go
abayer: I’ve debated going that route, but I think it is possible to add more secrets for other providers outside of requirements, which is...a thing. Next time I dive through this whole space for anything more than the most trivial of fixes to the “supported” providers (GH, GL, BBS), I.e., when I get around to adding BitBucket Cloud support, I’m going to first dedicate a few days to mapping out all the permutations of this junk to handle it properly.

pkg/gits/bitbucket_server.go
abayer: So I’m not 100% sure the API calls will fail when using lowercase for owner, but I am 100% sure that the BitBucket Server “project” will always be uppercase, just like JIRA projects or Confluence spaces. 

My changes regarding case in this PR are generally defensive (I know upper case will always work, so let’s do that) and in this case, `Project` is only relevant for BBS, so it definitely should be uppercase. I opted not to uppercase the values for the `Organisation` field because that one’s used across all providers...though it was made more confusing in that the BBS provider sometimes used `Project` where every other provider would use `Organisation`. Fun times.

pkg/gits/bitbucket_server.go
abayer: See my comment above for a bit more detail. Using `Project` in these API calls would only work consistently if the `repository` had been generated from the BBS provider directly, since otherwise `Project` isn’t set with the general purpose git url parsing stuff.

pkg/gits/bitbucket_server_test.go
abayer: Like I said above, the org (well, project in BBS terminology) is always capitalized in BBS. I had a BDD failure trying to call with `jxbdd` rather than `JXBDD`, e.g.

pkg/gits/bitbucket_server.go
hferentschik: > See my comment above for a bit more detail. Using Project in these API calls would only work consistently if the repository had been generated from the BBS provider directly, since otherwise Project isn’t set with the general purpose git url parsing stuff.

got you

pkg/cmd/step/step_custom_pipeline.go
hferentschik: > when I get around to adding BitBucket Cloud support, I’m going to first dedicate a few days to mapping out all the permutations of this junk to handle it properly.

Maybe at least create a place holder issue!?

pkg/gits/bitbucket_server.go
hferentschik: > So I’m not 100% sure the API calls will fail when using lowercase for owner, but I am 100% sure that the BitBucket Server “project” will always be uppercase, just like JIRA projects or Confluence spaces.

So, do we have a (BDD) testbed for this? Are you just developing against a temporary installation you have somewhere? Could we get BitBucket covered via BDD? That would by far be the best.

abayer: Yup, https://github.com/jenkins-x/lighthouse/pull/737 has me exercising the Lighthouse suite against BBS.

pkg/cmd/step/step_custom_pipeline.go
abayer: https://github.com/jenkins-x/jx/issues/7201

pkg/cmd/create/create_devpod.go
abayer: And done.

pkg/cmd/create/create_devpod.go
```diff
@@ -277,7 +277,8 @@ func (o *CreateDevPodOptions) Run() error {
...
+			importURL = gits.HttpCloneURL(gitInfo, gitKind)
```
hferentschik, line 281: 👍

pkg/gits/git_url_test.go
```diff
@@ -175,3 +175,70 @@ func TestGitInfoProviderURL(t *testing.T) {
 		assert.Equal(t, "https://github.com", info.ProviderURL(), "ProviderURL() for %s", u)
 	}
 }
+
+func TestGitInfoHttpCloneURL(t *testing.T) {
```
hferentschik, line 179: 👍

abayer:
/override tf-boot

I'm impatient waiting for it to finish destroying. =)

Issue #7176, opened 2020-05-13
binary plugins can only be used in k8s clusters+namespaces that have a Plugin


jstrachan:
so if you install a binary plugin you cannot use it in any other cluster/namespace.

this means we can't use binary plugins as a way to add plugins for spinning up/configuring new clusters

PR #7177, opened 2020-05-13
fix: support finding a plugin in the plugins dir

jstrachan:

binary plugins can't be found if you are connected to a kubernetes cluster which has no `Plugin` CRD installed.

This PR makes binary plugins installed into the `~/.jx/plugins/bin` dir equivalent to adding `jx-foo` to your `$PATH`; it lets folks install binary plugins without having to add them to your `$PATH`

see #7176

### DIFF
--- a/pkg/cmd/cmd.go
+++ b/pkg/cmd/cmd.go
@@ -19,8 +19,10 @@ package cmd
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


daveconde:
@jstrachan Would you have time to please pop in a test for this and split the logic out in to a few functions so we can get this merged?

Regarding preloading plugins via something like `jx install dependencies` did you want to do that in a follow up PR?

hferentschik:
What's the state on this pull request? Is that still needed?

ekeren:
@jstrachan, what is the status here?

jstrachan:
snowed on other stuff; it's still needed. 

I'll try find time to refactor it

jstrachan:
@daveconde I figure we'd try get this PR in first before adding another PR for another command to eagerly download Plugins to the local file system etc

ekeren:
@jstrachan - will you be also adding a test here?  (I see @daveconde asked for one)

jstrachan:
yeah

abayer:
@jstrachan codegen started failing for me on another PR out of the blue too. I couldn't reproduce locally.

abayer:
@jstrachan Specifically looks like it needs https://github.com/golang/tools/pull/239 due to a change in the tool.

abayer:
Think it's been fixed - retesting.

PR #7211, opened 2020-05-20
feat: remove the CRD creation from the changelog step and allow the creation of the controller's role and rolebindings on other namespaces

### LINKED ISSUE #7209: The changelog step attempts to register CRDs when it shouldn't
### Summary 
The `jx step changelog` command will attempt to register a few CRDs.

This is already covered in the `jx boot` pipeline and it should never be necessary to do so again, because it conflicts with clusters where cluster-wide permissions are not allowed.

Create a cluster without cluster-wide permissions, boot and watch a pipeline fail at this step.

### Expected behavior
It shouldn't register any CRDs.

### LINKED ISSUE #7210: The jx Helm chart for controllers should allow the creation of roles and rolebindings in additional namespaces
### Summary 
In order to make Jenkins X work on clusters without cluster-wide permissions, a certain set of objects must be created in the `staging` and `production` namespaces.

These resources are roles and rolebindings that will be needed because of the lack of cluster specific ones.

Install Jenkins X. For a given controller, the `role` and `rolebinding` objects will only be created in the `jx` namespace.

### Expected behavior
If configured, they should be created in other namespaces to account for the lack of cluster permissions.

dgozalo:
#### Submitter checklist

#### Description
We should not register any CRDs on any controller as they are being registered and updated by Boot. This makes the steps / controllers to fail on a cluster without cluster-wide permissions.

Also, allow the creation of `roles` and `rolebindings` for the `jx` Helm chart on different namespaces when configured.

fixes #7209 
fixes #7210

### DIFF
--- a/pkg/cmd/step/step_changelog.go
+++ b/pkg/cmd/step/step_changelog.go
@@ -205,28 +205,8 @@ func (o *StepChangelogOptions) Run() error {
 		o.BatchMode = true
 	}
 
-	apisClient, err := o.ApiExtensionsClient()
-	if err != nil {
-		return err
-	}
-	err = kube.RegisterPipelineActivityCRD(apisClient)
-	if err != nil {
-		return err
-	}
-	err = kube.RegisterGitServiceCRD(apisClient)
-	if err != nil {
-		return err
-	}
-	err = kube.RegisterReleaseCRD(apisClient)
-	if err != nil {
-		return err
-	}
-	err = o.RegisterUserCRD()
-	if err != nil {
-		return err
-	}
-
 	dir := o.Dir
+	var err error
 	if dir == "" {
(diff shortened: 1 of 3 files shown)


pkg/cmd/step/step_changelog.go
```diff
@@ -205,28 +205,8 @@ func (o *StepChangelogOptions) Run() error {
 		o.BatchMode = true
 	}
 
-	apisClient, err := o.ApiExtensionsClient()
```
hferentschik, line 208: how odd. Why was this ever here?

dgozalo: It's not the first place where I've seen it. It's legacy code added before we started registering CRDs on install / boot.

PR #7218, opened 2020-05-20
fix: Fix build log retrieval for BitBucket Server and GitLab

### LINKED ISSUE #7216: Build log retrieval on GitLab using gh-pages fails
You'll get a 503 with the resulting URL, because of https://gitlab.com/gitlab-org/gitlab/-/issues/36191. We need to change the log retrieval approach for GitLab (and unstash in general, actually) to go through the API.

### LINKED ISSUE #7217: Build log retrieval on BitBucket Server needs to have username in URL
GitHub raw content fetching works with just the token added to the URL, but BitBucket Server requires the username and token.

### LINKED ISSUE #7219: Build log retrieval on GitHub Enterprise still thinks it should use github.com url
The logic that determines the build log URL still thinks anything of kind github should be using `raw.githubusercontent.com`, but obviously it shouldn't be that for GitHub Enterprise.

abayer:
#### Submitter checklist

#### Description

GitHub allows fetching raw file content with a url like `https://TOKEN@...`, but BitBucket Server requires `https://USER:TOKEN@...`. We need to fetch GitLab raw files via the REST API, regrettably. See https://gitlab.com/gitlab-org/gitlab/-/issues/36191 for more details. So let's change the generated logs URL to the REST API url, and change the unstashing behavior to support the idea of adding a header, not just manipulating the URL.

And finally, the logic for determining whether the provider is github.com or GitHub Enterprise here wasn't working right, so let's tweak that.

fixes #7216 
fixes #7217 
fixes #7219

### DIFF
--- a/pkg/cloud/buckets/buckets.go
+++ b/pkg/cloud/buckets/buckets.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"net/url"
 	"strings"
 	"time"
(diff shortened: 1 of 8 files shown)


abayer:
Putting this on WIP until I've gotten the relevant BDD test working.

pkg/cloud/buckets/buckets.go
```diff
@@ -46,29 +47,36 @@ func KubeProviderToBucketScheme(provider string) string {
 
 // ReadURL reads the given URL from either a http/https endpoint or a bucket URL path.
 // if specified the httpFn is a function which can append the user/password or token if using a git provider
```
hferentschik, line 49: Might be worth updating the godocs here.

abayer: Done.

PR #7230, opened 2020-05-25
feat: Make `jx ui` accept 401 JXUI response.

### LINKED ISSUE #7229: jx ui command should accept 401 response from JXUI
### Summary 
Currently, in order to wait for the port-forwarded JXUI to become ready, the `jx ui` command checks for `/` to respond with 200.
In case the JXUI is basic auth protected (401 response) `jx ui` will always fail and stop forwarding the port after a timeout.

Although my basic auth PR also adds an Ingress JXUI cannot be accessed securely as long as TLS/DNS is not configured properly. Therefore the `jx ui` command is still needed but needs to be adjusted to accept a 401 JXUI response as well.

Install [JXUI from basic auth PR](https://github.com/cloudbees/jxui-frontend/pull/1491)

### Expected behavior
`jx ui` should port forward the UI locally - also when it responds with 401 (basic auth enabled).

### Actual behavior
`jx ui` port forwards properly first but then times out waiting for the UI to respond with HTTP status code 200.

### Jx version
2.0.1245

```
jx                 2.0.1245
Kubernetes cluster v1.14.10-gke.36
kubectl            v1.16.8
helm client        2.16.5
git                2.17.1
Operating System   Ubuntu 18.04.4 LTS
```

### Kubernetes cluster
* Own CJXD cluster
* bdd-jx test cluster

Ubuntu 18.04

mgoltzsche:
#### Submitter checklist

fixes #7229

### DIFF
--- a/pkg/cmd/ui/ui.go
+++ b/pkg/cmd/ui/ui.go
@@ -184,12 +184,13 @@ func (o UIOptions) waitForForwarding(localURL string) error {
 	return o.RetryUntilTrueOrTimeout(time.Minute, time.Second*3, func() (b bool, e error) {
 		log.Logger().Debugf("Checking the status of %s", localURL)
 		resp, err := http.Get(localURL) // #nosec
-		if err != nil || (resp != nil && resp.StatusCode != 200) {
+		respSuccess := resp != nil && (resp.StatusCode == 200 || resp.StatusCode == 401)
+		if err != nil || !respSuccess {
 			log.Logger().Debugf("Returned err: %+v", err)
 			log.Logger().Info(".")
 			return false, nil
 		}
-		return resp.StatusCode == 200, nil
+		return respSuccess, nil
 	})
 }
 


pkg/cmd/ui/ui.go
```diff
@@ -184,12 +184,13 @@ func (o UIOptions) waitForForwarding(localURL string) error {
 	return o.RetryUntilTrueOrTimeout(time.Minute, time.Second*3, func() (b bool, e error) {
 		log.Logger().Debugf("Checking the status of %s", localURL)
 		resp, err := http.Get(localURL) // #nosec
-		if err != nil || (resp != nil && resp.StatusCode != 200) {
+		respSuccess := resp != nil && (resp.StatusCode == 200 || resp.StatusCode == 401)
+		if err != nil || !respSuccess {
 			log.Logger().Debugf("Returned err: %+v", err)
```
ankitm123, line 189: If err is nil and response success is false, then the logger will not give any results which can be used for debugging, may be include the value of response success ... or just break it into two separate checks, one for error and one for respSuccess? thoughts?

mgoltzsche: true in this case it doesn't tell the user much. However it is already merged ;). On the other hand none of the errors would clarify to the user what the problem is other than not reaching JXUI which is obvious.

PR #7234, opened 2020-05-26
fix: allow to use external Vault instance for secrets

### LINKED ISSUE #7090: Allow for using existing HashiCorp Vault with Jenkins X
### Summary 
If I have created my own new Vault instance or have a pre-existing Vault instance, I want to be able to leverage it for managing Jenkins X secrets.  

Notes:
- Specify Vault instance in the `jx-requirements.yml` file 
- Provide meaningful feedback during pre-check if there is an issue with the specified Vault instance and halt the boot process
- Would need to add/enable Kubernetes Authenticator for Vault
- Provide rich documentation to describe how to do this

hferentschik:
- enhance client and vault factories to receive enough configuration to connect to external Vault instance
- using jx-install-config ConfigMap for storing Vault relevant configuration information
- updating InstallRequirements with additionally required fields
- making sure 'step boot vault' distinguishes between Jenkins X managed and external Vault instance
- syncing vault configuration from requirements file into jx-install-config ConfigMap during `jx step boot vault`

fixes #7090


pkg/cmd/deprecation/deprecated.go
```diff
@@ -171,45 +54,15 @@ var deprecatedCommands map[string]deprecationInfo = map[string]deprecationInfo{
...
 	},
-	"console": {
-		replacement: "jx ui",
-		date:        "Jun 1 2020",
-		info:        "Classic Jenkins console will be replaced by Jenkins X UI app",
-	},
-	"controller workflow": {
-		date: "Feb 1 2020",
-	},
 }
```
mrageh, line 74: I don't think the changes in this commit are relevant to #7090, would it be better to move it into a separate PR?

pkg/cmd/clients/fake/fake_factory.go
```diff
@@ -244,9 +244,9 @@ func (f *FakeFactory) CreateSystemVaultClient(namespace string) (vault.Client, e
 	return fake_vault.NewFakeVaultClient(), nil
 }
 
-// CreateVaultClient returns the given vault client for managing secrets
+// CreateInternalVaultClient returns the given vault client for managing secrets
 // Will use default values for name and namespace if nil values are applied
-func (f *FakeFactory) CreateVaultClient(name string, namespace string) (vault.Client, error) {
+func (f *FakeFactory) CreateInternalVaultClient(name string, namespace string) (vault.Client, error) {
```
mrageh, line 249: 👍

pkg/cmd/deprecation/deprecated.go
hferentschik: Can do. It was just a cleanup/chore prior to me adding the deprecation of the `jx [create|delete] vault`. I kept it separate since as you say it could potentially be a separate PR. Same for the first commit, which is also more of a cleanup.

mrageh: I think we should move both this commit and that one to separate PR's, it's more work and frustrating. But it will set a good example for others on how to create PR's that are focussed on solving one problem really well.

mrageh: I feel like I'm nit picking your work and it could certainly be argued that the value provided is not great. But it will set a great precedence.

abayer:
Sorry for using this PR as a testbed for https://github.com/jenkins-x/lighthouse/pull/772, but hey, I fixed the bug in that PR, so hey. =)

PR #7236, opened 2020-05-26
fix: Lock eksctl version to 0.11.1

### LINKED ISSUE #7235: Eksctl version is incorrect
### Summary 

Today I noticed that in some circumstances the eksctl version was using the latest version available on github. With the recent defaulting to k8s 1.16 in the 0.20 release of eksctl `jx boot` will fail.

Other binary versions are now fixed in `packages.go`. We should do the same for eksctl so we are sure the version we use is compatible with Jenkins X.

daveconde:
This PR fixes the eksctl version to 0.11.1 to keep it inline with other binaries that we install. The fix prevents `jx create cluster` from creating a 1.16 k8s cluster if the incorrect binary version is used.

fixes #7235

Also fixes a warning during ksync version checking.

Signed-off-by: David Conde

### DIFF
--- a/pkg/ksync/cli.go
+++ b/pkg/ksync/cli.go
@@ -79,10 +79,7 @@ func install(latestVersion, binDir, fileName string) error {
 func (kcli *CLI) Version() (string, error) {
 	args := []string{"version"}
 	kcli.Runner.SetArgs(args)
-	out, err := kcli.Runner.RunWithoutRetry()
-	if err != nil {
-		return "", err
-	}
+	out, _ := kcli.Runner.RunWithoutRetry()
 	return out, nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/ksync/cli.go
```diff
@@ -79,10 +79,7 @@ func install(latestVersion, binDir, fileName string) error {
 func (kcli *CLI) Version() (string, error) {
 	args := []string{"version"}
 	kcli.Runner.SetArgs(args)
-	out, err := kcli.Runner.RunWithoutRetry()
```
dgozalo, line 82: Why are we ignoring the error here?

daveconde: `ksync version` does both a client and server version check. In a case when no server is present a warning is returned. The version parsing is still possible with the warning included in the output. Unfortunately there is no extra flag to do a client only version check.

PR #7243, opened 2020-05-27
chore: removing commands not longer in codebase from deprecation list

### LINKED ISSUE #7242: Cleanup of deprecated.go
`deprecated.go` lists several commands which are already deprecated and even removed from the codebase. These commands should be removed.

hferentschik:
fixes #7242


hferentschik:
@daveconde @abayer - any idea why this would not merge or is it just slow?

hferentschik:
> any idea why this would not merge or is it just slow?

seem to be in merge pool now

PR #7245, opened 2020-05-27
refactor: moving pkg/util/error.go to pkg/util/errors/errors.go

### LINKED ISSUE #7244: CombineErrors in error.go in util package has a dependency to apimachinery
`CombineErrors` uses `k8s.io/apimachinery/pkg/util/errors` as a dependency creates an unwanted dependency. If one, for example, wants to use the functionality of `CombineErrors` in a non-Kube related code area and one wants to cast to the `Aggregate` to get to the individual errors, one needs to use Kubernetes related functions in a non-Kube related area.

The easiest path forward, in this case, is imo to cut&paste the `Aggregate` functionality from apimachinery into our codebase (including the tests). The functionality is stable, tested and small enough to not be a burden.

hferentschik:
- also making the package independent from Kube's apimachinery by copying the relevant code into jx

fixes #7244


pkg/cloud/factory/factory.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/jenkins-x/jx/v2/pkg/config"
 	"github.com/jenkins-x/jx/v2/pkg/kube"
 	"github.com/jenkins-x/jx/v2/pkg/log"
-	"github.com/jenkins-x/jx/v2/pkg/util"
+	jxerrors "github.com/jenkins-x/jx/v2/pkg/util/errors"
```
daveconde, line 13: The moving of the package and then having to alias the import feels a bit strange, I wonder if we could come up with a name that would avoid the need to alias?

pkg/versionstream/resolver.go
```diff
@@ -3,6 +3,8 @@ package versionstream
 import (
 	"fmt"
 
+	"github.com/jenkins-x/jx/v2/pkg/util/errors"
```
daveconde, line 6: it would be nice to keep this one consistent with the imports in the other files, pending the outcome of the previous comment :)

pkg/cloud/factory/factory.go
hferentschik: A better suggestion for the package name? Or do you suggest to not move at all? Admittedly this is not required,  but I still think we need to keep breaking up these large packages. The large packages have been identified before as an issue and I keep breaking them up as much as possible. See also https://github.com/jenkins-x/jx/pull/7234.

'error' does not work, so the most natural choice would be errors which of course might in some cases clash with `github.com/pkg/errors`. Personally, I don't feel the alias strange and we use `CombineErrors` quite sparingly so far. 

If we want to do something change something, I vote for another package name over keeping everything in one big package. That said, I am not quite sure what this other name should be.

daveconde: Maybe something like github.com/jx/v2/pkg/errorutil might be a good way to go from reading this https://rakyll.org/style-packages/

hferentschik: IMO this makes the package tree very flat since, after this rule, we should also have _stringutil_, _fileutil_, etc. I think most projects have infact the _errors_ packages and hence the need for aliases at times. 

That said, if you feel strongly about _errorutil_ it might be better than keeping all in the _util_ package. @abayer, @dgozalo, @mrageh, @garethjevans - thoughts?

abayer: I think this is a reasonable exception to make, since it’s such a common package name.

hferentschik: > I think this is a reasonable exception to make, since it’s such a common package name.

meaning you prefer _errorutil_ as well?

hferentschik:
Superseded by #7261

PR #7224, opened 2020-05-22
chore: migrate code.gitea.io/sdk/gitea to v0.12.0

6543:
migrate to new version & use new features

https://gitea.com/gitea/go-sdk/src/branch/master/docs/migrate-v0.11-to-v0.12.md

PR #7246, opened 2020-05-27
chore: migrate code.gitea.io/sdk/gitea to v0.12.0

hferentschik:
fixes: #7247

supersedes #7224


6543:
thanks for picking it up :)

hferentschik:
> thanks for picking it up :)

no problem. Hope it works now. This whole business with go modules can be quite tricky.

hferentschik:
@abayer - wdyt?

PR #7249, opened 2020-05-27
chore: allow LoadRequirements to be a little less restrictive loading.

garethjevans:
This is useful downstream where we use this function to load requirements
from the file system and we're not guarrenteed to have exactly the same
version of JX.

#### Submitter checklist

fixes #


garethjevans:
helps fix cloudbees/arcalos#1135

hferentschik:
Could we create an issue as well and maybe squash the commits?

PR #7267, opened 2020-06-01
fix: allow to use external Vault instance for secrets

### LINKED ISSUE #7090: Allow for using existing HashiCorp Vault with Jenkins X

hferentschik:
- enhance client and vault factories to receive enough configuration to connect to external Vault instance
- using jx-install-config ConfigMap for storing Vault relevant configuration information
- updating InstallRequirements with additionally required fields
- making sure 'step boot vault' distinguishes between Jenkins X managed and external Vault instance
- syncing vault configuration from requirements file into jx-install-config ConfigMap during `jx step boot vault`
- removing `jx get secret`
- deprecating `jx [create|delete] vault' 

fixes #7090

One interesting bit is that the ConfigMap ` jx-install-config` to keep state of the current Vault configuration. `jx boot` needs to ensure the right values are written to this map and commands needing access to Vault will first need to read this map in order to build the Vault client. 
An alternative approach would be to use the `jx-requirements.yaml` stored in the `TeamSettings` of the `Environment` CRD. Personally I found the use of  ` jx-install-config`


hferentschik:
:-)

samedguener:
@hferentschik Thanks for the contribution! 

One quick question: Is there any documentation on how to use this feature :) ?

PR #7279, opened 2020-06-04
chore: Switch to lighthouse-config library

### LINKED ISSUE #7280: Switch jx to use lighthouse-config library for "Prow" config
We'll be using that library in Lighthouse as well, so this lets us make changes to the config for jobs without needing to do circular dependency hell.

abayer:
#### Submitter checklist

#### Description

Using https://github.com/jenkins-x/lighthouse-config. The `k8s.io/test-infra/prow/plugins` config is hairier - I need to tease that apart before switching to it, but we should be fine just switching the `k8s.io/test-infra/prow/config` imports for now. This will let us make changes to the config, specifically things around presubmit and postsubmit definitions.

fixes #7280


hferentschik:
If you open an issue we are all good :-)

abayer:
@hferentschik done. =)

pkg/cmd/step/scheduler/test_data/step_scheduler_config_apply/gitops_all/config.yaml
```diff
@@ -23,6 +23,8 @@ branch-protection:
 deck:
   spyglass: {}
 gerrit: {}
+github:
+  LinkURL: null
```
hferentschik, line 27: why is this a capital 'L' in 'LinkURL', that sticks out a fair bit given that our usual naming

hferentschik: Is this something new we introduced somewhere?

pkg/pipelinescheduler/config_builder.go
```diff
@@ -6,10 +6,9 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	jenkinsv1 "github.com/jenkins-x/jx/v2/pkg/apis/jenkins.io/v1"
+	"github.com/jenkins-x/lighthouse-config/pkg/config"
```
hferentschik, line 9: what is the advantage of extracting the lighthouse config into its own repo opposed to having a dedicated _api_ package in lighthouse itself which we could depend on?
I guess I am concerned about the amount of repository we are having to maintain. Do you think _lighthouse-config_ will ever have a different lifecycle than lighthouse itself?

abayer: Versioning and avoiding circular dependencies. Since Lighthouse depends on jx for the metapipeline client, and extracting the metapipeline client from jx would be significantly tougher than extracting the config code from Lighthouse, this was the best way I could come up with to get jx to use Lighthouse’s config in the scheduler logic. The config will also release significantly less often than Lighthouse itself - as in I expect to make some changes in the next few weeks to add easier configurability of conditional execution, and then probably won’t be doing any releases in the foreseeable future, but if jx depends on Lighthouse, we’d need to either update jx’s dependency on Lighthouse every time Lighthouse releases (which in practice wouldn’t be an option since Lighthouse releases for each jx release!) or update the dependency manually, which would open up the possibility of forgetting to update jx when a relevant Lighthouse change was released.

I also just have an instinctive aversion to circular dependencies. =)

pkg/cmd/step/scheduler/test_data/step_scheduler_config_apply/gitops_all/config.yaml
abayer: I think it's a serialization change between the Prow version we've been depending on and the Prow version Lighthouse derives from.

abayer: Regardless, it's from the Prow code we derived from, not something we added for Lighthouse.

hferentschik: ok. Really sticks out :-(

pkg/pipelinescheduler/config_builder.go
hferentschik: Thanks for the explanation :-) I was not thinking about the circular dependency .

PR #7285, opened 2020-06-05
[wip] chore: allow GetRequirementsConfigFromTeamSettings to be less restrictive validating settings

romainverduci:
#### Submitter checklist

- [*] Change is code complete and matches issue description.
- [*] Change is covered by existing or new tests.

#### Description
This is useful downstream where we use this function to load requirements
from the file system and we're not guarrenteed to have exactly the same
version of JX.

fixes https://github.com/jenkins-x/jx/issues/7288

### DIFF
--- a/pkg/cloud/factory/factory.go
+++ b/pkg/cloud/factory/factory.go
@@ -40,7 +40,7 @@ func NewBucketProviderFromTeamSettingsConfiguration(factory clients.Factory) (bu
 	if err != nil {
 		return nil, errors.Wrap(err, "error obtaining the dev environment teamSettings to select the correct bucket provider")
 	}
-	requirements, err := config.GetRequirementsConfigFromTeamSettings(teamSettings)
+	requirements, err := config.GetRequirementsConfigFromTeamSettings(teamSettings, config.DefaultFailOnValidationError)
 	if err != nil || requirements == nil {
 		return nil, errorutil.CombineErrors(err, errors.New("error obtaining the requirements file to decide bucket provider"))
 	}
--- a/pkg/cmd/clients/factory.go
+++ b/pkg/cmd/clients/factory.go
@@ -628,7 +628,7 @@ func (f *factory) useVaultIngress(devNamespace string) (bool, error) {
 	if err != nil {
 		return false, errors.Wrapf(err, "getting team settings from namespace %s", devNamespace)
 	}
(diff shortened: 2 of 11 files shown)


abayer:
Mind opening an issue? Also, codegen is failing because jx now requires go 1.13, and you’re still using go 1.12, which puts an indirect dependency in go.mod that go 1.13 removes. =)

hferentschik:
I am not sure about this? It feels wrong. I have a feeling we are missing something here. Maybe it is a problem of the validation function itself. @romainverduci, would you mind describing in the issue how this issue occurs and what error it results in.

hferentschik:
@abayer, I think you wrote this validation logic. Do the latest changes around this look right to you?

hferentschik:
A couple of questions:

1. What is the aim of this validation? Just to verify that the YAML is valid or that it can be processed by the `jx` binary executing this function? 
1. Assuming you have an older version of `jx` and a newer version has a changed requirements file (new fields for example), what happens if I get a requirements file which
   1. does not contain any of the new elements? It should pass, right?
   1. does contain fields which are unknown to this version of `jx`? Should it not fail and is this what we are trying to prevent here?

romainverduci:
> I am not sure about this? It feels wrong. I have a feeling we are missing something here. Maybe it is a problem of the validation function itself. @romainverduci, would you mind describing in the issue how this issue occurs and what error it results in.

I updated the issue with some more details, I paste it here as well:

Concrete issue happening:
- There was a field added in the team settings in this PR: https://github.com/jenkins-x/jx/pull/7172
- It made import / quickstart failing because of:
```
error getting Requirements from TeamSettings to determine if in GitHub app mode: validation failuresi in requirements from team settings: (root): additional property buildPacks is not allowed
```
It looks like the validation failure comes from this method: https://github.com/jenkins-x/jx/blob/7ae7aac6e2af35b30cbe02193871f61d8169f835/pkg/config/install_requirements.go#L629

Adding a field in team settings should not fail this validation.

romainverduci:
@hferentschik I'm closing this PR as I agree this is a bad workaround and the quick fix for me is to just upgrade jx on both the cluster and client side. That being said there was already a PR with this approach here: https://github.com/jenkins-x/jx/pull/7249 that you may want to revert.
Let's follow up on the issue to have a proper long term fix.

PR #7306, opened 2020-06-09
[wip] poc: add deprecated information to generated docs

garethjevans:
#### Submitter checklist

fixes #

### DIFF
--- a/pkg/cmd/deprecation/deprecated.go
+++ b/pkg/cmd/deprecation/deprecated.go
@@ -8,84 +8,85 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// deprecatedCommands list of deprecated commands along with some more deprecation details
-var deprecatedCommands map[string]deprecationInfo = map[string]deprecationInfo{
+// DeprecatedCommands list of deprecated commands along with some more deprecation details
+var DeprecatedCommands = map[string]DeprecationInfo{
 	"install": {
-		replacement: "jx boot",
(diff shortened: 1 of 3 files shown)


hferentschik:
@garethjevans, seems the formatting is off. `make fmt` might help.

Also, can we open an issue describing what this is about?
Is it about making sure that commands which are marked as deprecated in `deprecated.go` are also marked in the reference docs  - https://jenkins-x.io/docs/reference/commands/?

garethjevans:
The markdown doc generator, does not generate docs for deprecated (or hidden) commands.  It would make sense that whilst the command is deprecated but still in use `jx install` is a good example, that we generate docs for them, but clearly mark them as deprecated.

pkg/cmd/deprecation/deprecated.go
```diff
@@ -8,84 +8,85 @@ import (
...
+	Date        string
+	Info        string
 }
 
 // DeprecateCommands runs recursively over all commands and set the deprecation message
 // on every command defined the deprecated commands map.
 func DeprecateCommands(cmd *cobra.Command) {
 	path := commandPath(cmd)
-	if deprecation, ok := deprecatedCommands[path]; ok {
-		cmd.Deprecated = deprecationMessage(deprecation)
```
garethjevans, line 88: As you can see, we are no longer marking the cmd as "Deprecated", but updating the LongExample.  This is not ideal as it won't display when the command is run, just when help is called

Issue #6957, opened 2020-03-24
Discovery - Kubernetes 1.16 support


deanesmith:
### Summary 
As a result of major cloud provider's bumping default versions of Kubernetes to 1.16 (and beyond), it would be prudent to assess how much effort is required to formally support Jenkins-X on Kubernetes 1.16.  

A reasonable inventory of related open issues can be found here -
https://github.com/jenkins-x/jx/issues?q=is%3Aopen+is%3Aissue+label%3Aarea%2Fkubernetes

The output of this task would be to establish how much effort is required for the additional support and to assess the risk and/or any upgrade nuances.

ankitm123:
We could use [pluto](https://github.com/FairwindsOps/pluto) to find deprecated Kubernetes apiVersions in the codebase and helm releases (helm2 and helm3).

ankitm123:
These are the versions of the charts that need to be updated in jenkins-x-platform to support kubernetes 1.16+ support (based on the error message I got after doing a `jx boot` on kubernetes 1.17.X cluster). 

- chartmuseum: 1.0.2 (https://github.com/helm/charts/pull/18123/files)
- docker-registry: 1.8.3 (https://github.com/helm/charts/pull/17454/files)
- heapster (this is deprecated, and should be replaced with metric-server): 2.4.0 (https://github.com/helm/charts/pull/17334/files)
- jenkins (Probably should be removed): - 

I will open a PR against the jenkins-x-platform. I will check if the other helm charts need to be updated as well.

garethjevans:
Both helm 2 & 3 support a pretty decent set of built-in objects. e.g. https://v2.helm.sh/docs/chart_template_guide/#built-in-objects.  We should be able to use the `Capabilities` to determine if the api is present. e.g. `Capabilities.APIVersions.Has $version`.

Need to careful to do it for versions, rather than resources as it looks like that was only added in 2.15

garethjevans:
@deanesmith do we have a view on what kubernetes versions we are attempting to support? v1.13 -v1.18?

deanesmith:
@garethjevans any improvement on the current support level but v1.13-v1.16 is miminal I believe.  What do you think?

ankitm123:
Should it not mirror what the cloud providers support? For example eks: https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#version-deprecation (Probably similar for gke and aks)
[...]

PR #7313, opened 2020-06-10
chore: Bump Kubernetes (and associated) dependencies

abayer:
#### Submitter checklist

#### Description

Perhaps the final part of #6957, and the first part of #7330. Bumps our k8s dependencies to 1.16.5 (more or less), bumps various other dependencies that also depend on `k8s.io/client-go` accordingly, and finally, bumps our Tekton go.mod dependency to 0.11.3, with various changes to how we construct pipelines accordingly.

This last part is, shockingly to me, backwards compatible: it works with our Tekton 0.8.0 version running in the Tekton controller, as well as working with Tekton 0.11.3 once we upgrade the chart. One small hack was needed for that last part - `/builder/home` has been changed to `/tekton/home` in 0.11.x, so I added a step to each task that symlinks `/tekton/home` to `/builder/home` if `/builder/home` doesn't exist, just to be sure existing pipelines won't fail.

Otherwise, this is shockingly clean. We're still on 0.9.1 of cert-manager here because that's what's needed for the legacy `jx upgrade ingress` stuff, but otherwise I went with as new a dependency version as I could, and things...passed.

Mostly #6957 and #7330, but the former probably won't really be done until we have a test matrix running against multiple k8s versions, and the latter won't be done until this has merged and we can upgrade the Tekton chart to 0.11.3.


cmd/codegen/app/generate_openapi.go
```diff
@@ -152,13 +152,20 @@ func (o *CreateClientOpenAPIOptions) Run() error {
...
-	defer cleanupFunc()
-
 	err = generator.InstallOpenApiGen(o.GeneratorVersion, gopath)
+	// Reset the go.mod and go.sum after installing openapi-gen
+	cleanupFunc()
 	if err != nil {
 		return errors.Wrapf(err, "error installing kubernetes openapi tools")
 	}
 
+	cleanupFunc, err = util.BackupGoModAndGoSum()
```
hferentschik, line 162: > This ends up requiring some trickery with `go.mod` - installing
`openapi-gen` ends up removing `k8s.io/api` and `k8s.io/apimachinery`
from `go.mod`, resulting in the openapi being generated against master
of those packages rather than the version we're actually depending on,
so we need to copy/restore `go.mod` and `go.sum` twice - before/after
installing `openapi-gen`, and before/after actually generating the openapi.

oh joy

go.mod
```diff
@@ -119,44 +112,54 @@ require (
...
+replace github.com/Azure/azure-sdk-for-go => github.com/Azure/azure-sdk-for-go v23.2.0+incompatible
 
-replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v10.14.0+incompatible
+replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.1+incompatible
+
+replace github.com/banzaicloud/bank-vaults => github.com/banzaicloud/bank-vaults v0.0.0-20191212164220-b327d7f2b681
+
+replace github.com/banzaicloud/bank-vaults/pkg/sdk => github.com/banzaicloud/bank-vaults/pkg/sdk v0.0.0-20191212164220-b327d7f2b681
+
+replace k8s.io/test-infra => github.com/abayer/test-infra v0.0.0-20200611142252-211a92405c22
```
hferentschik, line 161: Should this fork of test-infra not live under the jenkins-x org? I don't think anything like this should be in personal repos. WDYT?

abayer: Yeah, I’ll push that branch to the jenkins-x fork.

go.mod
```diff
@@ -119,44 +112,54 @@ require (
...
 	k8s.io/test-infra v0.0.0-20190131093439-a22cef183a8f
-	knative.dev/pkg v0.0.0-20191217184203-cf220a867b3d
+	knative.dev/pkg v0.0.0-20200207181514-32ea84581573
+	knative.dev/serving v0.12.1-0.20200210194206-365600fcbe27
 	sigs.k8s.io/yaml v1.1.0
 
 )
 
-replace k8s.io/api => k8s.io/api v0.0.0-20190528110122-9ad12a4af326
+replace k8s.io/api => k8s.io/api v0.16.5
```
hferentschik, line 139: Have you tried moving some of these replace statements back into the dependency list? It would be nice to get rid of them, but I guess we can experiment with this once the bulk of the move is in.

hferentschik:
Funny _import_test.go_ seems to be failing for me on this branch, but passes on master!? Have not looked into it. Might be unrelated since I have seen several unit tests acting up recently when run locally. _jsonschema_test.go_ is, for example, a test which fails for me randomly on master. I will see whether I can look into this this week. I just reacted to it since switching the branch came up with this new failure I have not seen yet.

go.mod
abayer: So something somewhere in the dependency tree keeps insisting on replacing the normal dependency list version with 1.17.2, so gotta keep the replace for now, sadly.

abayer:
@hferentschik Yeah, I'm seeing the same thing and have no clue why.

go.mod
abayer: Done - https://github.com/jenkins-x/test-infra/tree/for-k8s-1.16-deps

abayer:
@hferentschik To make it even weirder, `go test -v --tags=unit ./pkg/cmd/importcmd/...` works fine. So the only way I've seen it fail yet is via `make test`. Something tells me it's a parallelism thing.

abayer:
@hferentschik Aaaaah, no, it's not parallelism, it's that `TestSetPreviewNamespace` fails if you don't have a kubeconfig because it's not using a fake factory! Fixing in a new commit here.

pkg/cmd/importcmd/import_test.go
```diff
@@ -218,9 +217,10 @@ func TestImportOptions_GetOrganisation(t *testing.T) {
 
 func TestSetPreviewNamespace(t *testing.T) {
 	t.Parallel()
-	factory := clients.NewFactory()
+	factory := fake.NewFakeFactory()
```
abayer, line 220: @hferentschik Tada - the other changes below are because I just copied the relevant section from `TestSetPreviewNamespaceWithPermanentNs` since that test definitely worked, so better safe than sorry.

hferentschik: good stuff

go.mod
hferentschik: > So something somewhere in the dependency tree keeps insisting on replacing the normal dependency list version with 1.17.2, so gotta keep the replace for now, sadly.

Fair enough. Still a huge step forward. I think one of the reasons I failed with my attempts was that I at the same time tried to get rid of the replace statements and it always kept failing one way or another.

go.mod
```diff
@@ -119,44 +112,54 @@ require (
...
+replace github.com/Azure/azure-sdk-for-go => github.com/Azure/azure-sdk-for-go v23.2.0+incompatible
 
-replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v10.14.0+incompatible
+replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.1+incompatible
+
+replace github.com/banzaicloud/bank-vaults => github.com/banzaicloud/bank-vaults v0.0.0-20191212164220-b327d7f2b681
+
+replace github.com/banzaicloud/bank-vaults/pkg/sdk => github.com/banzaicloud/bank-vaults/pkg/sdk v0.0.0-20191212164220-b327d7f2b681
+
+replace k8s.io/test-infra => github.com/jenkins-x/test-infra v0.0.0-20200611142252-211a92405c22
```
hferentschik, line 161: +1

hferentschik:
@abayer, I am fine with this change. Just wondering whether there will be any downstream implications and whether we need to sync the merge od this in some form? 

@garethjevans, @cagiti - any potential impact on your side?

garethjevans:
@hferentschik yeah, there will be impact on the SaaS but we've got a few open PRs ready for this

abayer:
`boot-vault` failure was because we hit address limit on the GKE project. I've nuked some old runs that hadn't yet been gc'd to speed things up.

PR #7317, opened 2020-06-11
feat: let users define a preview namespace when importing / creating quickstarts

### LINKED ISSUE #7302: Let import application / create quickstart define how the preview namespace will be called
### Summary 
Right now, when you create a pull request for an application, it uses a certain pattern to create a new Namespace for preview environments.

We need to let applications decide which namespace to use so they can define a pre created namespace for all applications.

This way users would have a lot more control on how to name their environments and will let them reuse namespaces in case there's a problem with creating new namespaces on the fly.

dgozalo:
#### Submitter checklist

#### Description
This will allow users define which namespace to use for previews while checking that they can't set a permanent environment's ns as a preview namespace.

This will then modify the `values.yaml` file within the `charts/preview/` folder so it can be read and used by the `jx preview`command within a builder.

fixes #7302

### DIFF
--- a/pkg/cmd/importcmd/test_data/application_with_preview/charts/preview/values.yaml
+++ b/pkg/cmd/importcmd/test_data/application_with_preview/charts/preview/values.yaml
@@ -1,0 +1,19 @@
+cleanup:
+  Annotations:
+    helm.sh/hook: pre-delete
+    helm.sh/hook-delete-policy: hook-succeeded
+  Args:
+    - --cleanup
+expose:
+  Annotations:
+    helm.sh/hook: post-install,post-upgrade
+    helm.sh/hook-delete-policy: hook-succeeded
+  config:
+    exposer: Ingress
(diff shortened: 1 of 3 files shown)


pkg/cmd/importcmd/test_data/application_with_preview/charts/preview/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+    helm.sh/hook-delete-policy: hook-succeeded
+  config:
+    exposer: Ingress
+    http: true
+    tlsacme: false
+preview:
+  image:
+    pullPolicy: IfNotPresent
+    repository: null
+    tag: null
```
hferentschik, line 19: newline missing :-)

dgozalo: Fixed.

PR #7320, opened 2020-06-11
feat: add bot as repo collaborator during import

mgoltzsche:
as the title says

### DIFF
--- a/pkg/cmd/importcmd/import.go
+++ b/pkg/cmd/importcmd/import.go
@@ -422,6 +422,11 @@ func (options *ImportOptions) Run() error {
 			}
 			options.GetReporter().PushedGitRepository(options.RepoURL)
 		}
+
+		err = options.AddBotAsCollaborator()
+		if err != nil {
+			return err
+		}
 	}
 
 	if options.DryRun {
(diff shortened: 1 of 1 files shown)


mgoltzsche:
unfortunately this PR is still not ready: it made the bot accept the invitation at some point. However after some cleaning up now it does not anymore because apparently it is listing invitations for my personal user instead of the bot somehow.

pkg/cmd/importcmd/import.go
```diff
@@ -707,68 +712,66 @@ func (options *ImportOptions) CreateNewRemoteRepository() error {
...
 		// If the user creating the repo is not the pipeline user, add the pipeline user as a contributor to the repo
 		if options.PipelineUserName != options.GitUserAuth.Username && options.GitServer != nil && options.GitServer.URL == options.PipelineServer {
 			// Make the invitation
-			err := options.GitProvider.AddCollaborator(options.PipelineUserName, details.Organisation, details.RepoName)
+			err := options.GitProvider.AddCollaborator(options.PipelineUserName, options.Organisation, options.AppName)
+			if err != nil {
+				return err
+			}
+
+			authConfigSvc, err := options.GitAuthConfigService()
```
mgoltzsche, line 734: What tricked me here: Previously `options.GitLocalAuthConfigService()` was used (for quickstart) which used the local git credentials. In order to retrieve the pipeline bot's credentials from vault `options.GitAuthConfigService()` must be used and the credentials obtained using `authConfigSvc.Config().GetPipelineAuth()`.

romainverduci:
This looks good to me and is actually working when I'm testing it locally with both the UI and the CLI.
Would be nice to have a review from someone with more knowledge in this area.
@abayer @hferentschik maybe ?

PR #7348, opened 2020-06-16
fix: avoid nil pointer exception when creating sr

### LINKED ISSUE #7349: Nil pointer exception when creating sr
### Summary 

There is an uncontrolled nil pointer exception that occurs if the creation of a SourceRepository fails.

Try to create a jx environments with a very long name.

### Expected behavior

Either the environment is created or an error is reported.

### Actual behavior

Nil pointer exception:

```
Creating 01--st-romen-in-romen-rome-termediate-159172-721182 Environment in namespace **
Created environment 01--st-romen-in-romen-rome-termediate-159172-721182
Namespace 01--st-romen-in-romen-rome-termediate-159172-721182 created
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x2ebf59b]

goroutine 1 [running]:
github.com/jenkins-x/**/pkg/cmd/create.(*CreateEnvOptions).RegisterEnvironment(0xc000341800, 0xc000829400, 0x58dd700, 0xc0000341c0, 0x5857e20, 0xc000d751e0, 0x2, 0x0)
	*****************/pkg/cmd/create/create_env.go:323 +0x41b
github.com/jenkins-x/**/pkg/cmd/create.(*CreateEnvOptions).Run(0xc000341800, 0x0, 0x0)
	*****************/pkg/cmd/create/create_env.go:254 +0xbfa
github.com/jenkins-x/**/pkg/cmd/create.NewCmdCreateEnv.func1(0xc00078a780, 0xc00074e900, 0x0, 0xc)
	*****************/pkg/cmd/create/create_env.go:96 +0x6f
github.com/spf13/cobra.(*Command).execute(0xc00078a780, 0xc00074e840, 0xc, 0xc, 0xc00078a780, 0xc00074e840)
	/workspace/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:830 +0x2ae
github.com/spf13/cobra.(*Command).ExecuteC(0xc0003c2280, 0xc000703200, 0x57fa200, 0xc0000bc000)
	/workspace/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914 +0x2fc
github.com/spf13/cobra.(*Command).Execute(...)
	/workspace/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
github.com/jenkins-x/**/cmd/**/app.Run(0x0, 0x0, 0x0, 0xc000723f88, 0xc0000a0058)
	*****************/cmd/**/app/**.go:19 +0x185
main.main()
	*****************/cmd/**/**.go:11 +0x32
```

### Jx version

```
NAME               VERSION
jx                 2.0.1192+cjxd.7
Kubernetes cluster v1.14.10-gke.36
kubectl            v1.14.7
Operating System   Mac OS X 10.15.5 build 19F101

```

romenrg:
#### Submitter checklist

#### Description

When registering an environment (in `RegisterEnvironment` function in `create_env.go`), eventually a call to `kube.GetOrCreateSourceRepository` is made, which is expected to return the `sr` (but it might eventually fail). Immediately after this call, there was a line that tried to log the SourceRepository's name. However, if an error was produced and/or the `sr` value was nil, attempting to access fields was causing a nil pointer exception, as in the example included in the reported issue.

With this change, some new validations are introduced to make sure that the nil pointer exception doesn't occur and that an error or a log entry is produced instead.

fixes #7349

### DIFF
--- a/pkg/cmd/create/create_env.go
+++ b/pkg/cmd/create/create_env.go
@@ -320,7 +320,13 @@ func (o *CreateEnvOptions) RegisterEnvironment(env *v1.Environment, gitProvider 
 				return err
 			}
 			sr, err := kube.GetOrCreateSourceRepository(jxClient, devNs, gitInfo.Name, gitInfo.Organisation, gitInfo.HostURLWithoutUser())
-			log.Logger().Debugf("have SourceRepository: %s\n", sr.Name)
+			if err != nil {
+				return errors.Wrap(err, "getting or creating source repository")
+			} else if sr == nil {
+				log.Logger().Warnf("unexpected returned nil value for sr in GetOrCreateSourceRepository")
+			} else {
+				log.Logger().Debugf("have SourceRepository: %s\n", sr.Name)
+			}
 
 			err = o.GenerateProwConfig(devNs, devEnv)
 			if err != nil {


ankitm123:
can you create an issue for this PR @romenrg?

pkg/cmd/create/create_env.go
```diff
@@ -320,7 +320,13 @@ func (o *CreateEnvOptions) RegisterEnvironment(env *v1.Environment, gitProvider
 				return err
 			}
 			sr, err := kube.GetOrCreateSourceRepository(jxClient, devNs, gitInfo.Name, gitInfo.Organisation, gitInfo.HostURLWithoutUser())
-			log.Logger().Debugf("have SourceRepository: %s\n", sr.Name)
+			if err != nil {
+				return errors.Wrap(err, "getting or creating source repository")
+			} else if sr == nil {
```
ankitm123, line 325: What do you think about using `!=` instead of `==` for all the checks? Keeps the code base consistent. Thoughts? May be also eliminate the need for an `else if` statement?

pkg/cmd/create/create_env.go
```diff
@@ -320,7 +320,13 @@ func (o *CreateEnvOptions) RegisterEnvironment(env *v1.Environment, gitProvider
 				return err
 			}
 			sr, err := kube.GetOrCreateSourceRepository(jxClient, devNs, gitInfo.Name, gitInfo.Organisation, gitInfo.HostURLWithoutUser())
-			log.Logger().Debugf("have SourceRepository: %s\n", sr.Name)
+			if err != nil {
```
ankitm123, line 323: :100:

pkg/cmd/create/create_env.go
romenrg: I might be missing something, but I wanted to check 2 things: 1) that there are errors (err != nil) and, if so return the error with a message and 2) if there are no errors but for some reason sr is still nil, log that as a warning, since it seems like an unexpected behavior. Then, the "success" case, if there are no errors and sr has a value, print the source repository name safely.

romenrg:
Addressing @ ankitm123 requests:
- Created an issue and linked it in the description of this PR
- Squashed the 2 commits

pkg/cmd/create/create_env.go
romenrg: Oh, I think I now understood what you meant. Swap the content in my second and third conditions and replace the else if to be "!="? To me the current approach seems more natural to understand the condition, but I guess it's a matter of preference. My personal vote is to leave it as is in the PR, but if there are strong opinions in favor of your suggestions, I can make the change.

ankitm123: 1. :+1: 
2. Is there a case where this can be nil (I glanced at the function, dont see where it returns nil ...)? Also, if this is nil, should we not return an error? @abayer @romenrg

abayer: In this particular case, I don't think we could ever hit a nil without an error, but I'd keep this in to be safe.

ankitm123: should we return an error, or just log sr being nil (if it is harmless) - new to this part of the codebase ... ?

romenrg: IMHO the current approach is the safest bet and an improvement over what was already there. In the unlikely event that there is no error but sr is nil, logging the case will allow whoever encounters it to decide if a further improvement must be done to this piece of the codebase or not (depending on if it causes issues later in the code); but by having this log at least he/she will see a message explaining what happened, instead of a nil pointer exception.

ankitm123: I am ok with it :+1:

PR #7352, opened 2020-06-17
fix: uninstall jenkins-x

### LINKED ISSUE #7165: Unable to `jx uninstall`
`jx uninstall` hangs when running against EKS cluster. The problems seem to be that the deletion of the _PersistentVolumeClaims_ gets stuck in Terminating state.

```
? This action will permanently delete Jenkins X from the Kubernetes context foo. Please type in the name of the context to confirm: foo
Removing installation of Jenkins X in team namespace jx
WARNING: Not deleting jx-prow because the release is not installed

error: [failed to uninstall the jenkins-x helm chart in namespace jx: [failed to run 'kubectl delete pvc --ignore-not-found -l jenkins.io/chart-release=jenkins-x --namespace jx --wait' command in directory '', output: 'persistentvolumeclaim "jenkins" deleted
persistentvolumeclaim "jenkins-x-chartmuseum" deleted
persistentvolumeclaim "jenkins-x-docker-registry" deleted
persistentvolumeclaim "jenkins-x-nexus" deleted
```

Manually running `kubectl delete pvc --ignore-not-found -l jenkins.io/chart-release=jenkins-x --namespace jx` also hangs.

Could affect GKE as well. Only verified against EKS so far.

mrageh:
#### Submitter checklist

#### Description
When uninstalling `jx`, the PVC would hang in a terminating state. When we set the PVC, finalizer volume to `null`, the uninstall process completes as normal.

There is an open issue in [Kubernetes](https://github.com/kubernetes/kubernetes/issues/77258) that has not yet been fixed.

This PR sets the PVC finalizer to `null`, which fixes our issue and allows users to uninstall `jx`.

fixes #7165

### DIFF
--- a/pkg/cmd/uninstall/uninstall.go
+++ b/pkg/cmd/uninstall/uninstall.go
@@ -65,7 +65,6 @@ func NewCmdUninstall(commonOpts *opts.CommonOptions) *cobra.Command {
 }
 
 func (o *UninstallOptions) Run() error {
-
 	config, _, err := o.Kube().LoadConfig()
 	if err != nil {
 		return err
@@ -111,6 +110,11 @@ func (o *UninstallOptions) Run() error {
 	}
 
 	log.Logger().Infof("Removing installation of Jenkins X in team namespace %s", util.ColorInfo(namespace))
+
+	err = o.setPVCFinalizerToNull(namespace)
+	if err != nil {
+		return errors.Wrapf(err, "unable to nil PVC finalizers")
+	}
 
 	envMap, envNames, err := kube.GetEnvironments(jxClient, namespace)
 	if err != nil {
@@ -235,6 +239,7 @@ func (o *UninstallOptions) deleteNamespace(namespace string) error {
 // as it checks if the release is present before deleting, it can be forced to delete in case it doesn't find it because
 // of an unrelated error
 func (o *UninstallOptions) DeleteReleaseIfPresent(ns string, releaseName string, errors []error, force bool) []error {
+	log.Logger().Debugf("Deleting release '%s'", releaseName)
 	if err := o.Helm().StatusRelease(ns, releaseName); err == nil || force {
 		err := o.Helm().DeleteRelease(ns, releaseName, true)
 		if err != nil {
@@ -244,4 +249,29 @@ func (o *UninstallOptions) DeleteReleaseIfPresent(ns string, releaseName string,
 		log.Logger().Warnf("Not deleting %s because the release is not installed", releaseName)
 	}
 	return errors
+}
+
+func (o *UninstallOptions) setPVCFinalizerToNull(ns string) error {
+	client, err := o.KubeClient()
+	if err != nil {
+		return errors.Wrap(err, "getting the kube client")
+	}
+
+	list, err := client.CoreV1().PersistentVolumeClaims(ns).List(meta_v1.ListOptions{})
+
+	if err != nil {
+		return errors.Wrap(err, "getting PVC list")
+	}
+
+	for _, pvc := range list.Items {
+		currentPVC := pvc
+		currentPVC.Finalizers = nil
+		log.Logger().Debugf("Setting finalizer for PVC '%s' to nil", currentPVC.Name)
+		_, err = client.CoreV1().PersistentVolumeClaims(ns).Update(&currentPVC)
+		if err != nil {
+			return errors.Wrapf(err, "failed to update pvc '%s'", pvc.GetName())
+		}
+	}
+
+	return nil
 }


pkg/cmd/uninstall/uninstall.go
```diff
@@ -245,3 +252,19 @@ func (o *UninstallOptions) DeleteReleaseIfPresent(ns string, releaseName string,
...
 	return errors
 }
+
+func (o *UninstallOptions) setPVCFinalizerToNull(ns string, releaseName string) error {
+	client, err := o.KubeClient()
+	if err != nil {
+		return errors.Wrap(err, "getting the kube client")
+	}
+
+	list, _ := client.CoreV1().PersistentVolumeClaims(ns).List(meta_v1.ListOptions{})
```
dgozalo, line 262: I think we should not ignore this error, there may be a problem with the connection there. The list will be empty instead of nil so there won't be a panic, but I think we should handle all errors.

pkg/cmd/uninstall/uninstall.go
```diff
@@ -245,3 +252,19 @@ func (o *UninstallOptions) DeleteReleaseIfPresent(ns string, releaseName string,
...
+
+	for _, pvc := range list.Items {
+		pvc.Finalizers = []string{}
+		client.CoreV1().PersistentVolumeClaims(ns).Update(&pvc)
```
dgozalo, line 266: Same here, we should be handling the error on update.

pkg/cmd/uninstall/uninstall.go
```diff
@@ -236,6 +236,13 @@ func (o *UninstallOptions) deleteNamespace(namespace string) error {
 // of an unrelated error
 func (o *UninstallOptions) DeleteReleaseIfPresent(ns string, releaseName string, errors []error, force bool) []error {
 	if err := o.Helm().StatusRelease(ns, releaseName); err == nil || force {
+
+		err = o.setPVCFinalizerToNull(ns, releaseName)
+
+		if err != nil {
+			errors = append(errors, fmt.Errorf("failed to uninstall the %s helm chart in namespace %s: %s", releaseName, ns, err))
```
dgozalo, line 243: This error actually comes from the `setPVCFinalizerToNull`, the error message should be related.

pkg/cmd/uninstall/uninstall.go
```diff
@@ -236,6 +236,13 @@ func (o *UninstallOptions) deleteNamespace(namespace string) error {
 // of an unrelated error
 func (o *UninstallOptions) DeleteReleaseIfPresent(ns string, releaseName string, errors []error, force bool) []error {
 	if err := o.Helm().StatusRelease(ns, releaseName); err == nil || force {
+
+		err = o.setPVCFinalizerToNull(ns, releaseName)
+
+		if err != nil {
+			errors = append(errors, fmt.Errorf("Failed to set PVC finalizer to null in %s helm chart in namespace %s: %s", releaseName, ns, err))
```
ankitm123, line 243: Any reason we don't return when we see an error? curious ...
More specifically, should we even go to the next step (in this case DeleteRelease) if an error is encountered?

mrageh: Good point, let me fix that and return immediately!

hferentschik:
from the unit test logs:
```
# github.com/jenkins-x/**/v2/pkg/cmd/uninstall
pkg/cmd/uninstall/uninstall.go:274:11: Wrapf format %s has arg pvc of wrong type k8s.io/api/core/v1.PersistentVolumeClaim
```

pkg/cmd/uninstall/uninstall.go
mrageh: Fixed!

mrageh:
Fixed, this is ready to merge

dgozalo:
I think this could use a unit test. Just need to create a few PVCs using the mock and test that your code actually sets the Finalizers to nil.

pkg/cmd/uninstall/uninstall.go
```diff
@@ -236,6 +236,14 @@ func (o *UninstallOptions) deleteNamespace(namespace string) error {
 // of an unrelated error
 func (o *UninstallOptions) DeleteReleaseIfPresent(ns string, releaseName string, errors []error, force bool) []error {
 	if err := o.Helm().StatusRelease(ns, releaseName); err == nil || force {
+
+		err = o.setPVCFinalizerToNull(ns, releaseName)
```
hferentschik, line 240: Actually, out of interest, why do you call `setPVCFinalizerToNull` from here? The way `setPVCFinalizerToNull` is implemented, it is a global thing w/o any dependence to the release name.

pkg/cmd/uninstall/uninstall.go
```diff
@@ -245,3 +253,27 @@ func (o *UninstallOptions) DeleteReleaseIfPresent(ns string, releaseName string,
 	}
 	return errors
 }
+
+func (o *UninstallOptions) setPVCFinalizerToNull(ns string, releaseName string) error {
```
hferentschik, line 257: _releaseName_ is not really used. I think `setPVCFinalizerToNull` is really a "global" function to be executed somewhere from within `Run`.

hferentschik:
@mrageh, @dgozalo - fyi, I pushed another commit to this pull request. It removes the unused parameter from `setPVCFinalizerToNull` and moves the function call itself. 

Also adding `jx uninstall` as a task to the _tf-boot_ context cleanup. This way we get a test of it as well.

PR #7353, opened 2020-06-17
feat: attempt to prevent using provided requirements for a provisione…

cagiti:
…d jenkins-x cluster

The overall goal here is to attempt to prevent the use of the `--requirements` option of `jx boot` on a cluster which is being managed by the environment dev repository for that cluster. It is assumed that the `--requirements` option of the `jx boot` command should only be used for the initial boot.

See https://github.com/jenkins-x/jx/issues/7405 for more details

Updated message on help text for `--requirements` option
```sh
→ ./jx boot --help
Boots up Jenkins X in a Kubernetes cluster using GitOps and a Jenkins X Pipeline

For more documentation see: https://jenkins-x.io/docs/getting-started/setup/boot/

Aliases:
boot, bootstrap
Examples:
  # create a kubernetes cluster via Terraform or via jx
  jx create cluster gke --skip-installation

  # now lets boot up Jenkins X installing/upgrading whatever is needed
  jx boot

  # if we have already booted and just want to apply some environment changes without
  # re-applying ingress and so forth we can start at the environment step:
  jx boot --start-step install-env
Options:
  -a, --attempt-restore=false: attempt to boot from an existing dev environment repository
  -d, --dir='.': the directory to look for the Jenkins X Pipeline, requirements and charts
  -e, --end-step='': the step in the pipeline to end at
      --git-ref='': override the Git ref for the JX Boot source to start from, ignoring the versions stream. Normally specified with git-url as well
  -u, --git-url='': override the Git clone URL for the JX Boot source to start from, ignoring the versions stream. Normally specified with git-ref as well
  -v, --helm-log='': sets the helm logging level from 0 to 9. Passed into the helm CLI via the '-v' argument. Useful to diagnose helm related issues
      --no-update-git=false: disables any attempt to update the local git clone if its old
  -r, --requirements='': WARNING: this should only be used for the initial boot of a cluster: requirements file which will overwrite the default requirements file
  -s, --start-step='': the step in the pipeline to start from
      --versions-ref='master': the bootstrap ref for the versions repo. Once the boot config is cloned, the repo will be then read from the jx-requirements.yml
      --versions-repo='https://github.com/jenkins-x/jenkins-x-versions.git': the bootstrap URL for the versions repo. Once the boot config is cloned, the repo will be then read from the jx-requirements.yml
Usage:
  jx boot [flags] [options]
Use "jx options" for a list of global command-line options (applies to all commands).
```

### DIFF
--- a/pkg/cmd/boot/boot.go
+++ b/pkg/cmd/boot/boot.go
@@ -184,6 +184,13 @@ func (o *BootOptions) Run() error {
 		return err
 	}
 
+	// check if the user has passed in a requirements file to a cluster which has already been provisioned.
+	if o.RequirementsFile != "" {
+		if err := o.checkIfProvidedRequirementsArePossiblyStale(); err != nil {
+			return errors.Wrapf(err, "loading from provided requirements file")
+		}
+	}
+
 	if err := o.overrideRequirements(gitURL); err != nil {
 		return errors.Wrap(err, "overwriting the default requirements")
 	}
@@ -494,6 +501,17 @@ func existingBootClone(dir string) (bool, error) {
 		return false, err
 	}
 	return requirementsExist && pipelineExists, nil
+}
+
+func (o *BootOptions) checkIfProvidedRequirementsArePossiblyStale() error {
+	_, devEnv := o.GetDevEnv()
+	if devEnv != nil {
+		log.Logger().Warnf("It seems you're passing a requirements file to cluster which has already been provisioned.")
+		log.Logger().Warnf("We recommend you update the %s file at %s, using the updates provided within your local %s file.",
+			config.RequirementsConfigFileName, devEnv.Spec.Source, o.RequirementsFile)
+		return errors.New("attempting to boot cluster using a requirements file which is possibly stale")
+	}
+	return nil
 }
 
 func (o *BootOptions) overrideRequirements(defaultBootConfigURL string) error {
--- a/pkg/cmd/boot/boot.go
+++ b/pkg/cmd/boot/boot.go
@@ -102,7 +102,7 @@ func NewCmdBoot(commonOpts *opts.CommonOptions) *cobra.Command {
 	cmd.Flags().StringVarP(&options.StartStep, "start-step", "s", "", "the step in the pipeline to start from")
 	cmd.Flags().StringVarP(&options.EndStep, "end-step", "e", "", "the step in the pipeline to end at")
 	cmd.Flags().StringVarP(&options.HelmLogLevel, "helm-log", "v", "", "sets the helm logging level from 0 to 9. Passed into the helm CLI via the '-v' argument. Useful to diagnose helm related issues")
-	cmd.Flags().StringVarP(&options.RequirementsFile, "requirements", "r", "", "requirements file which will overwrite the default requirements file")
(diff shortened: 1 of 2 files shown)


pkg/cmd/boot/boot.go
```diff
@@ -184,6 +184,13 @@ func (o *BootOptions) Run() error {
 		return err
 	}
 
+	// check if the user has passed in a requirements file to a cluster which has already been provisioned.
+	if o.RequirementsFile != "" {
```
hferentschik, line 188: why not integrate this into the existing `overrideRequirements` function? That would be the natural place to do it, right?

I am also wondering whether the criteria of an empty `o.RequirementsFile` does not fall a bit short? 

```go
	requirements, requirementsFile, err := config.LoadRequirementsConfig(o.Dir, config.DefaultFailOnValidationError)
	if err != nil {
		return errors.Wrapf(err, "loading requirements from dir %q", o.Dir)
	}

	// overwrite the default requirements with provided requirements
	if o.RequirementsFile != "" {
		providedRequirements, err := config.LoadRequirementsConfigFile(o.RequirementsFile, config.DefaultFailOnValidationError)
		if err != nil {
			return errors.Wrapf(err, "loading requirements from file %q", o.RequirementsFile)
		}
		*requirements = *providedRequirements
	}
```

This is the code from `overrideRequirements`. We already try to find the requirements file in the current directory. I would do an `filepath.ABS` on both path (provided both are set) and if they are not identical generate the warning/error.  Probably needs to be paired with the `o.GetDevEnv()` check you added. But it would allow for the cases where the user does something like `jx boot -r ./jx-requirements.yml` or from a sub-directory `jx boot -r ../jx-requirements.yml`. WDYT?

hferentschik: Maybe we can write a test for this as well.

cagiti: I guess it comes down to the intention of the `--requirements` option, I've always been of the opinion that it should only be used for the initial boot sequence of a fresh cluster.

cagiti: > This is the code from `overrideRequirements`. We already try to find the requirements file in the current directory. I would do an `filepath.ABS` on both path (provided both are set) and if they are not identical generate the warning/error. Probably needs to be paired with the `o.GetDevEnv()` check you added. But it would allow for the cases where the user does something like `jx boot -r ./jx-requirements.yml` or from a sub-directory `jx boot -r ../jx-requirements.yml`. WDYT?

I think it's dangerous in general to want to specify a requirements file via the `--requirements` option when `boot`ing an already provisioned cluster as those changes aren't guaranteed to make their way into the environment dev repository unless I've misunderstood something?

deanesmith:
Would it be possible to get an issue for this to help with release notes?

pkg/cmd/boot/boot.go
hferentschik: Ok, so I am coming around here. Once there is an existing Jenkins X install, you should not be able to use the `-r` option.

cagiti:
- [ ] Add issue which describes the intent or this PR
- [ ] provide more feedback around the use of `--requirements` option

pkg/cmd/boot/boot.go
hferentschik: We probably should also add something to the help message:

`cmd.Flags().StringVarP(&options.RequirementsFile, "requirements", "r", "", "requirements file which will overwrite the default requirements file")`

Issue #7315, opened 2020-06-11
jx get vault-config fails


joostvdg:
### Summary 

Running `jx get vault-config` fails.

### Steps to reproduce the behavior

Have an existing cluster with Vault installed via `jx boot`, and a recent `jx` CLI `2.1.x`+.

### Expected behavior

```shell
Downloading https://releases.hashicorp.com/vault/1.4.2/vault_1.4.2_darwin_amd64.zip to /Users/joostvdg/.jx/bin/vault.zip...
Downloaded /Users/joostvdg/.jx/bin/vault.zip
export VAULT_ADDR=https://vault-jx.dev.jxgke.kearos.net
export VAULT_TOKEN=s.XXXXXXXXXXXXXXX
```

I'll ignore the downloading of the binary, which is a different topic - not expected at all, but not the current issue.
The above output is from the CJXD 10 binary:

```
jx                 2.1.46+cjxd.10
Kubernetes cluster v1.14.10-gke.36
kubectl            v1.16.7
helm client        2.16.1
git                2.21.0 (Apple Git-122.2)
Operating System   Mac OS X 10.15.5 build 19F96
```

### Actual behavior

```shell
error: unable to parse Vault configuration from ConfigMap jx-install-config: service account name cannot be empty
```

### Jx version

The output of `jx version` is:

```
**FATAL: Your current helm version v3 is not supported. Please downgrade to helm v2.**
```

Right, sorry for that.

Cluster A.

```
NAME               VERSION
jx                 2.1.62
Kubernetes cluster v1.14.10-gke.36
kubectl            v1.16.7
helm client        2.16.1
git                2.21.0 (Apple Git-122.2)
Operating System   Mac OS X 10.15.5 build 19F96


verifying packages
```

Cluster B, installed via `jx boot` with CJXD 6 or 7 I believe.

```
jx                 2.1.62
Kubernetes cluster v1.14.10-gke.36
kubectl            v1.16.7
helm client        2.16.1
git                2.21.0 (Apple Git-122.2)
Operating System   Mac OS X 10.15.5 build 19F96


verifying packages
```

### Kubernetes cluster

* Cluster A: GKE cluster, created via Jenkins X Terraform GKE
* Cluster B: GKE cluster, created via jx create (I believe)

### Operating system / Environment

```
Mac OS X 10.15.5 build 19F96
```

daveconde:
[...]

PR #7354, opened 2020-06-17
fix: attempt to make get vault-config backwardly compatible

cagiti:
fixes https://github.com/jenkins-x/jx/issues/7315

### DIFF
--- a/pkg/cmd/get/vault/config/get_vault_config.go
+++ b/pkg/cmd/get/vault/config/get_vault_config.go
@@ -2,14 +2,10 @@ package config
 
 import (
 	"fmt"
-	"os"
 	"runtime"
 
-	"github.com/jenkins-x/jx/v2/pkg/kube"
-	"github.com/pkg/errors"
-	"k8s.io/client-go/kubernetes"
-
 	"github.com/jenkins-x/jx/v2/pkg/vault"
+	"github.com/pkg/errors"
 
 	"github.com/jenkins-x/jx/v2/pkg/cmd/helper"
 
@@ -104,40 +100,12 @@ func (o *GetVaultConfigOptions) Run() error {
 }
 
 func (o *GetVaultConfigOptions) systemVaultClient() (vault.Client, error) {
-	kubeClient, devNamespace, err := o.KubeClientAndDevNamespace()
+	_, devNamespace, err := o.KubeClientAndDevNamespace()
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to create Kube client")
 	}
 
-	installValues, err := kube.ReadInstallValues(kubeClient, devNamespace)
-	if err != nil {
-		return nil, errors.Wrapf(err, "unable to retrieve ConfigMap %s", kube.ConfigMapNameJXInstallConfig)
-	}
-
-	vaultConfig, err := vault.FromMap(installValues, devNamespace)
-	if err != nil {
-		if !vaultConfig.IsExternal() {
-			fmt.Fprintf(os.Stderr, "WARNING: unable to validate internal vault configuration - attempting to default to original configuration\n")
-		} else {
-			return nil, errors.Wrapf(err, "unable to parse Vault configuration from ConfigMap %s", kube.ConfigMapNameJXInstallConfig)
-		}
-	}
-
-	if vaultConfig.IsExternal() {
-		return o.externalVaultClient(vaultConfig, kubeClient)
-	}
-
-	return o.vaultClient(vaultConfig.Name, devNamespace)
-}
-
-func (o *GetVaultConfigOptions) externalVaultClient(vaultConfig vault.Vault, kubeClient kubernetes.Interface) (vault.Client, error) {
-	factory := o.GetFactory()
-	client, err := factory.CreateExternalVaultClient(vaultConfig, kubeClient)
-	if err != nil {
-		return nil, errors.Wrap(err, "unable to create Vault client for Jenkins X managed Vault instance")
-	}
-
-	return client, nil
+	return o.SystemVaultClient(devNamespace)
 }
 
 func (o *GetVaultConfigOptions) vaultClient(name string, namespace string) (vault.Client, error) {
--- a/pkg/cmd/get/vault/config/get_vault_config.go
+++ b/pkg/cmd/get/vault/config/get_vault_config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"os"
 	"runtime"
 
 	"github.com/jenkins-x/jx/v2/pkg/kube"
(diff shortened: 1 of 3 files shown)


hferentschik:
@cagiti could you create an issue for it, explaining where and how the problem manifests itself?

pkg/vault/vault.go
```diff
@@ -92,14 +93,15 @@ func NewInternalVault(name string, serviceAccountName string, namespace string)
...
 
-	return FromMap(data, namespace)
+	vault, _, err := FromMap(data, namespace)
+	return vault, err
 }
 
 // FromMap reads the configuration of a Vault instance from a map.
 // defaultNamespace is used when there is no namespace value provided in the map (for backwards compatibility reasons).
-func FromMap(data map[string]string, defaultNamespace string) (Vault, error) {
+func FromMap(data map[string]string, defaultNamespace string) (Vault, bool, error) {
```
hferentschik, line 102: On first sight I am not a big fan of functions which return multiple concerns. Why so you need `FromMap` to return this extra boolean, if you can call `FromMap` and then `IsExternal`? This is a very slippery slope and leads imo to very intertangled APIs.

cagiti: that was an oversight, will remove that

cagiti: > @cagiti could you create an issue for it, explaining where and how the problem manifests itself? I am not quite sure atm what we are trying to fix here.

yes, does every PR need a issue?

hferentschik: > yes, does every PR need an issue?

that's what we said, yes.

pkg/vault/vault.go
```diff
@@ -127,10 +127,13 @@ func FromMap(data map[string]string, defaultNamespace string) (Vault, error) {
 	}
 
 	var err error
-	if vault.IsExternal() {
+	external := vault.IsExternal()
+	if external {
 		err = vault.validateExternalConfiguration()
 	} else {
 		err = vault.validateInternalConfiguration()
+		if err != nil {
```
hferentschik, line 135: I guess you meant to do something here?

pkg/cmd/get/vault/config/get_vault_config.go
```diff
@@ -115,7 +116,11 @@ func (o *GetVaultConfigOptions) systemVaultClient() (vault.Client, error) {
 
 	vaultConfig, err := vault.FromMap(installValues, devNamespace)
 	if err != nil {
-		return nil, errors.Wrapf(err, "unable to parse Vault configuration from ConfigMap %s", kube.ConfigMapNameJXInstallConfig)
+		if !vaultConfig.IsExternal() {
```
hferentschik, line 119: actually this is an issue we already had in _CreateSystemVaultClient_. I did not realize  this could become an issue here as well. Still would love to hear how this actually becomes an issue. 

The rub is that it is the wrong thing to do to create the Vault config instance at this point. In `factory.go` I do:

```
	if installValues[vault.URL] != "" {
		vaultConfig, err := vault.FromMap(installValues, namespace)
		if err != nil {
			return nil, errors.Wrapf(err, "error parsing Vault configuration from ConfigMap %s in namespace %s", kube.ConfigMapNameJXInstallConfig, namespace)
		}
		return f.CreateExternalVaultClient(vaultConfig, kubeClient)
	}

	name, err := f.getVaultName(namespace)
	if err != nil {
		return nil, err
	}
	return f.CreateInternalVaultClient(name, namespace)
```

Not ideal, since now I have the means on how to check whether a config is external or internal at multiple places. The best option would probably be to refactor the `Vault` code a bit and separate creation of the instance from its validation. 

Also in this particular case  I am wondering why the code is not calling `CreateSystemVaultClient` directly. That's what it is after, right?

pkg/vault/vault.go
hferentschik: the most important thing is really that there is enough context to understand the context of the issue. It just helps tremendously when reviewing the code. 

Then there is a debate between providing context on the pr or an issue. Given that the latter for example drives the changelogs (at least it should), every change should be backed by an issue. 

For me, there is another reason for it. Even though pull request and issues seem similar on GitHub they are not. Assume you create a pull request and you nicely add the stack trace which cause your trouble into the pull request which fixes the issue. If someone else get the same error and decides to search whether there this is a known issue, he would initially go to the issues page and paste the search string there. Per default, you are only searching issues though so the fact that this error occurred already once and might be fixed in a later version of `jx` is lost.

cagiti: Seems there was an existing issue: https://github.com/jenkins-x/jx/issues/7315

pkg/vault/vault.go
cagiti: yeah, not sure how that changed

pkg/cmd/get/vault/config/get_vault_config.go
cagiti: See issue https://github.com/jenkins-x/jx/issues/7315

cagiti: I didn't think it was necessary to validate the new configuration when calling `jx get vault-config`?

cagiti: > Also in this particular case I am wondering why the code is not calling `CreateSystemVaultClient` directly. That's what it is after, right?

I believe so. Would you expect to use `jx get vault-config` for an external vault setup?

patrickleet:
I came across this as well when switching between multiple kube/jx clusters on a machine - seems that my CLI was updated from one of the clusters, and then I switched to an older cluster and started getting this error.

pkg/cmd/get/vault/config/get_vault_config.go
hferentschik: >  Would you expect to use jx get vault-config for an external vault setup?

absolutely. For the Jenkins X system vault that is. It does not matter in this case whether this Vault instance is external (installed and managed independently of Jenkins X) or internal (created by Jenkins X using the Bankk-Vaults Operator). 

You need to be able to connect to your system Vault to rotate secrets for example.

Really, the best is to call CreateSystemVaultClient from _factory.go_. In fact I seem to recall that I intended to do this at some stage, but forgot about it.

pkg/vault/vault.go
```diff
@@ -172,22 +173,22 @@ func (v *Vault) validateExternalConfiguration() error {
 	}
 
 	if v.ServiceAccountName == "" {
-		err = errors.New("service account name cannot be empty")
+		err = errors.New("external vault service account name cannot be empty")
```
hferentschik, line 176: if you want to change the messages, I think "service account name for external vault instance cannot be empty" would be more appropriate. 
Same applies for the other messages.

cagiti: oh my word

pkg/cmd/get/vault/config/get_vault_config.go
cagiti: Will take a look at this one now 👍

pkg/vault/vault.go
hferentschik: ?

pkg/cmd/get/vault/config/get_vault_config.go
```diff
@@ -103,36 +100,12 @@ func (o *GetVaultConfigOptions) Run() error {
...
-
-func (o *GetVaultConfigOptions) externalVaultClient(vaultConfig vault.Vault, kubeClient kubernetes.Interface) (vault.Client, error) {
-	factory := o.GetFactory()
-	client, err := factory.CreateExternalVaultClient(vaultConfig, kubeClient)
-	if err != nil {
-		return nil, errors.Wrap(err, "unable to create Vault client for Jenkins X managed Vault instance")
-	}
-
-	return client, nil
+	return o.SystemVaultClient(devNamespace)
```
hferentschik, line 108: +1

PR #7356, opened 2020-06-17
fix: Fall back on default SA for PipelineRuns

### LINKED ISSUE #7355: Pipeline/metapipeline SA can be accidentally undefined
I haven't figured out exactly how it's happening, but when you invoke `CreateQuickstartOptions` directly rather than via `jx create quickstart`, `CommonOptions.ServiceAccount` ends up as the empty string rather than defaulting to `tekton-bot`. We should make sure we never kick off a `PipelineRun` without an SA.

abayer:
#### Submitter checklist

#### Description

Also make it a constant so that we don't have "tekton-bot" scattered randomly through the codebase.

fixes #7355

### DIFF
--- a/pkg/tekton/constants.go
+++ b/pkg/tekton/constants.go
@@ -25,4 +25,7 @@ const (
 
 	// LabelType is the label added to Tekton CRDs for the type of pipeline.
 	LabelType = "jenkins.io/pipelineType"
+
+	// DefaultPipelineSA is the default service account used for pipelines
+	DefaultPipelineSA = "tekton-bot"
 )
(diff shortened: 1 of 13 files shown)


pkg/tekton/constants.go
```diff
@@ -25,4 +25,7 @@ const (
 
 	// LabelType is the label added to Tekton CRDs for the type of pipeline.
 	LabelType = "jenkins.io/pipelineType"
+
+	// DefaultPipelineSA is the default service account used for pipelines
+	DefaultPipelineSA = "tekton-bot"
```
hferentschik, line 30: 👍 - nice one

abayer:
`tf-boot` is really hating me these days. =)

hferentschik:
> tf-boot is really hating me these days. =)

also checking back the third time, since the PR still shows on my dash ;-)

PR #7369, opened 2020-06-22
WIP: feat(import): skip draft if jenkins-x.yml exists.

mgoltzsche:
Allows to reimport projects that have previously been imported/created using Jenkins X (without explicitly specifying `--no-draft`).

#### Submitter checklist

#### Description

During repo import skip enhancing the repo with draft (buildpack) when a `jenkins-x.yml` is already present.
Previously importing a project that has been created with Jenkins X failed when `--no-draft` not set because draft attempts to manipulate the project but fails when the files it wants to write are already present.

corresponding JXUI issue linked below

### DIFF
--- a/pkg/cmd/importcmd/import.go
+++ b/pkg/cmd/importcmd/import.go
@@ -383,11 +383,21 @@ func (options *ImportOptions) Run() error {
 	options.AppName = naming.ToValidName(strings.ToLower(options.AppName))
 
 	if !options.DisableDraft {
+		// Disable draft if jenkins-x.yaml is already present
+		jenkinsxFile := filepath.Join(options.Dir, "jenkins-x.yml")
+		_, err = os.Stat(jenkinsxFile)
+		if err == nil {
+			options.DisableDraft = true
+		} else if !os.IsNotExist(err) {
+			return err
+		}
+	}
+
+	if !options.DisableDraft {
 		err = options.DraftCreate()
 		if err != nil {
 			return err
 		}
-
 	}
 	err = options.fixDockerIgnoreFile()
 	if err != nil {


mgoltzsche:
Please let me know if you have a better idea of how to do this.
Just skipping draft also doesn't seem to be ideal - especially since the `jenkins-x.yml` contains Jenkins X installation specific information. On the one hand we shouldn't overwrite changes the user may have made. On the other hand leaving it as is might not build either.

pkg/cmd/importcmd/import.go
```diff
@@ -382,12 +382,22 @@ func (options *ImportOptions) Run() error {
 	}
 	options.AppName = naming.ToValidName(strings.ToLower(options.AppName))
 
+	if !options.DisableDraft {
+		// Disable draft if jenkins-x.yaml is already present
+		jenkinsxFile := filepath.Join(options.Dir, "jenkins-x.yml")
+		_, err = os.Stat(jenkinsxFile)
```
hferentschik, line 388: For what it's worth there is a helper `FileExists` in the util package for this.

pkg/cmd/importcmd/import.go
```diff
@@ -382,12 +382,22 @@ func (options *ImportOptions) Run() error {
 	}
 	options.AppName = naming.ToValidName(strings.ToLower(options.AppName))
 
+	if !options.DisableDraft {
+		// Disable draft if jenkins-x.yaml is already present
+		jenkinsxFile := filepath.Join(options.Dir, "jenkins-x.yml")
```
hferentschik, line 387: I am a bit unsure whether the existence of `jenkins-x.yml` falls short as criteria here. There is a lot more happening in `options.DraftCreate` than the creation of `jenkins-x.yml`. I would think that one needs to integrate a bit deeper and do in general better checks and error handling along the various steps of `DraftCreate`. 

@abayer, @daveconde - wdyt?

Some tests would be nice regardless. ;-)

mgoltzsche: yeah, i realize the `jenkins-x.yml` is project or rather Jenkins X installation specific which is why it would need to be adjusted as well...

hferentschik:
> Please let me know if you have a better idea of how to do this.

I think we need to go deeper into the whole draft creation code and pry it apart. There are parts which have little to do with Draft per see and which would need to happen (I think) whether or not you skip the Draft part. 

The code needs to be more resilient to what already exists. For example, if there is an existing `jx-requirements.yml` it should read the build pack from there opposed to trying to determine the build pack by file analysis.

What do we do if Helm charts exist? There are certain requirements in terms of the naming in the Charts folders (see `DraftCreate`). Does that need to happen? I am thinking about the case where some source code gets imported, with a different app name.

> Just skipping draft also doesn't seem to be ideal 

+1, especially since there is more than the actual Draft creation. I would think at the very least one need to decompose `DraftCreate`

hferentschik:
We should start by creating an issue and describing the error in more detail. Then we can start analysing which part is failing a successive import.

mgoltzsche:
I've created a corresponding issue: #7370

ankitm123:
Probably give the user an option to decide if they want to override their existing helm charts or dockerfiles. May be a dry run flag would help?

mgoltzsche:
hm, there is a `--dry-run` option but I am not sure if it does what it is supposed to do.

mgoltzsche:
Closing this PR in favour of #7371

PR #7386, opened 2020-06-24
fix: eagerly download binary plugins

### LINKED ISSUE #7176: binary plugins can only be used in k8s clusters+namespaces that have a Plugin

jstrachan:
so that running `jx upgrade cli` will eagerly install any binary plugins defined in the current cluster.

This lets you switch clusters and be able to reuse the same binary plugins in other clusters which may not have the Plugin CRDs

fixes #7176

### DIFF
--- a/pkg/cmd/upgrade/upgrade_cli.go
+++ b/pkg/cmd/upgrade/upgrade_cli.go
@@ -4,7 +4,10 @@ import (
 	"github.com/blang/semver"
 	"github.com/jenkins-x/jx/v2/pkg/cmd/create/options"
 	"github.com/jenkins-x/jx/v2/pkg/cmd/helper"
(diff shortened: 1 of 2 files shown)


mrageh:
What's a plugin in this context, is it binaries that are needed as dependencies of `jx` or is it something completely different?

jstrachan:
@mrageh its the binaries from the Plugin CRD: https://github.com/jenkins-x/jx-api/blob/master/pkg/apis/jenkins.io/v1/types_extensions.go#L581

jstrachan:
@mrageh had to re-push the PR as there were merge issues with rebasing due to the recent logging API switch

PR #7407, opened 2020-06-29
chore: replace dependency on jx-api

garethjevans:
#### Submitter checklist

fixes #


abayer:
Codegen needs to get reworked some - failing on generating CRD ref docs currently.

garethjevans:
Yeah, I've not tackled the codegen stuff yet, am more concerned about some other parts of the generation.  It looks like the `DeepCopy` stuff has been generated slightly different, e.g.

```
3575c3575
< 	in.ListMeta.DeepCopyInto(&out.ListMeta)
---
> 	out.ListMeta = in.ListMeta
```

hferentschik:
`jx-api` only encompasses the Jenkins X CRDs atm, right?
Probably worth creating an issue in `jx` for the switch.

garethjevans:
@hferentschik its `pkg/apis`,`pkg/clients` & some of `pkg/config` to allow the install requirements to work.

Makefile
```diff
@@ -220,4 +220,3 @@ code-generate:
 	./hack/generate.sh
 
 include Makefile.docker
-include Makefile.codegen
```
abayer, line 223: So we do still need/want codegen for the deepcopy structs for `pkg/tekton/syntax`, `pkg/jenkinsfile`, and `pkg/config` (for `ProjectConfig` in the last case).

Makefile.codegen
```diff
@@ -0,0 +1,67 @@
...
+# Anything generated by this target should be checked in
+generate: build-codegen install-generate-deps generate-mocks generate-client ## Generate the Go code (crds, mocks, openapi, client)
+	@$(MAKE) fmt
+	@$(MAKE) importfmt
+	@echo "Generation complete"
+
+install-generate-deps:
+	$(GO) get github.com/petergtz/pegomock/...@$(PEGOMOCK_VERSION)
+
+generate-mocks: install-generate-deps go-generate fmt importfmt ## Generate the mocks
```
abayer: Oh yeaaaaaah - I forgot about the mocks! Good catch.

abayer:
/override tf-boot

It actually passed everything but the known destroy issue.

PR #7408, opened 2020-06-29
fix: making sure envconfig is not initialising nil structs

### LINKED ISSUE #7331: RequirementsConfig#OverrideRequirementsFromEnvironment populates nil values
Despite defining some structs in `RequirementsConfig` as pointers and setting the _omitempty_ property, the final `jx-requirement.yml` after running `jx boot` contains empty values for all possible struct even if not specified.

The cause lies in `RequirementsConfig#OverrideRequirementsFromEnvironment` which in turn calls `err := envconfig.InitWithOptions(&c, envconfig.Options{AllOptional: true})` which is responsible for the population of the empty structs.

This behaviour makes it hard to evolve the requirements in a backwards-compatible manner by introducing new pointer-based configuration options.

hferentschik:
- setting envconfig LeaveNil option to ensure nil structs are not initialised
- switching more override parameters to envconfig

fixes #7331

### DIFF
--- a/pkg/config/install_requirements.go
+++ b/pkg/config/install_requirements.go
@@ -67,6 +67,10 @@ const (
 	RequirementVaultServiceAccountName = "JX_REQUIREMENT_VAULT_SA_NAME"
 	// RequirementVeleroServiceAccountName the service account name for velero
 	RequirementVeleroServiceAccountName = "JX_REQUIREMENT_VELERO_SA_NAME"
+	// RequirementVeleroTTL defines the time to live (TTL) for the Velero backups in minutes
+	RequirementVeleroTTL = "JX_REQUIREMENT_VELERO_TTL"
+	// RequirementVeleroSchedule defines the schedule of the Velero backups in cron notation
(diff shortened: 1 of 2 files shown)


pkg/config/install_requirements.go
```diff
@@ -872,92 +876,14 @@ func (i *IngressConfig) IsAutoDNSDomain() bool {
 
 // OverrideRequirementsFromEnvironment allows properties to be overridden with environment variables
 func (c *RequirementsConfig) OverrideRequirementsFromEnvironment(gcloudFn func() gke.GClouder) {
-	//init envconfig struct tags
-	err := envconfig.InitWithOptions(&c, envconfig.Options{AllOptional: true})
+	// struct members need to use explicit 'envconfig:"<var-name>"' unless there is a match between struct member navigation
+	// path and env variable name
+	err := envconfig.InitWithOptions(&c, envconfig.Options{AllOptional: true, LeaveNil: true, Prefix: "JX_REQUIREMENT"})
```
hferentschik, line 881: @garethjevans, this is the key change for the bug, using the option `LeaveNil`

PR #7411, opened 2020-06-30
fix: explicitly empty buckets before 'terraform destroy'

### LINKED ISSUE #7406: tf-boot context frequently fails on destroy for non-empty bucket on object not found
Specifically with errors like:
```
Error: could not delete non-empty bucket due to error when deleting contents: googleapi: Error 404: No such object: vault-pr-7392-47-tf-boot-808f601781c9/sys/counters/requests/2020/06, notFound
```

Seems to be some kind of race condition/bad luck, but not sure. This is particularly annoying because it results in some non-trivial percentage of `tf-boot` runs failing on this.

hferentschik:
fixes #7406

### DIFF
--- a/jx/bdd/tf-boot/ci.sh
+++ b/jx/bdd/tf-boot/ci.sh
@@ -117,15 +117,22 @@ function authenticate() {
 function cluster_destroy()
 {
   echo "###############################################################################"
-	echo "Cleanup..."
-	pushd "$script_dir"/terraform
-	terraform destroy -auto-approve
-#	gcloud container clusters delete "$(yq r jx-requirements.yml 'cluster.clusterName')" --zone=europe-west1-c --quiet
-#	gsutil rm -r -f "$(yq r jx-requirements.yml 'storage.logs.url')"
-#	gsutil rm -r -f "$(yq r jx-requirements.yml 'storage.reports.url')"
-#	gsutil rm -r -f "$(yq r jx-requirements.yml 'storage.repository.url')"
-#	gsutil rm -r -f gs://"$(yq r jx-requirements.yml 'vault.bucket')"
-	popd
+  echo "Cleanup..."
+  pushd "$script_dir"/terraform
+
+  # stop at least the Vault instance to prevent Bank-Vault operator to interfere with deletion
+  kubectl delete vault "$(yq r jx-requirements.yml 'cluster.clusterName')"
+
+  # explicitly empty buckets (see https://github.com/jenkins-x/jx/issues/7406)
+  # see also https://github.com/GoogleCloudPlatform/gsutil/issues/417
+  gsutil -m -q rm -r -f "$(yq r jx-requirements.yml 'storage.logs.url')"/** || true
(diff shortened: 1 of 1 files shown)


hferentschik:
@abayer, explicitly emptying the buckets seem to help. Eventually, I'd like to use `jx uninstall` prior to emptying the bucket, but I don't think the existing bug in this command is properly fixed yet. For now I manually delete the Vault instance by removing the Vault CRD.

jx/bdd/tf-boot/ci.sh
```diff
@@ -117,15 +117,22 @@ function authenticate() {
...
-#	gsutil rm -r -f "$(yq r jx-requirements.yml 'storage.logs.url')"
-#	gsutil rm -r -f "$(yq r jx-requirements.yml 'storage.reports.url')"
-#	gsutil rm -r -f "$(yq r jx-requirements.yml 'storage.repository.url')"
-#	gsutil rm -r -f gs://"$(yq r jx-requirements.yml 'vault.bucket')"
-	popd
+  echo "Cleanup..."
+  pushd "$script_dir"/terraform
+
+  # stop at least the Vault instance to prevent Bank-Vault operator to interfere with deletion
+  kubectl delete vault "$(yq r jx-requirements.yml 'cluster.clusterName')"
```
hferentschik, line 124: I'd like to call `jx uninstall` here, but I don't think that's working yet.

jx/bdd/tf-boot/ci.sh
```diff
@@ -117,15 +117,22 @@ function authenticate() {
...
+
+  # explicitly empty buckets (see https://github.com/jenkins-x/jx/issues/7406)
+  # see also https://github.com/GoogleCloudPlatform/gsutil/issues/417
+  gsutil -m -q rm -r -f "$(yq r jx-requirements.yml 'storage.logs.url')"/** || true
+  gsutil -m -q rm -r -f "$(yq r jx-requirements.yml 'storage.reports.url')"/**  || true
+  gsutil -m -q rm -r -f "$(yq r jx-requirements.yml 'storage.repository.url')"/**  || true
+  gsutil -m -q rm -r -f gs://"$(yq r jx-requirements.yml 'vault.bucket')"/**  || true
```
hferentschik, line 131: That seems to do the trick regarding the vault bucket emptying. I think the `gsutil` is not even necessary, but I'll leave it for good measure.

hferentschik: @abayer - https://github.com/jenkins-x/jx/pull/7411/files#diff-0d74d27cef898d7b54f6c3c824b267a2R256-R262

This might be another reason why we see resources not getting cleaned up. We are not destroying the cluster in case a test fails. In this case, it is up to the gc. I am wondering whether the current gc jobs properly capture and delete what Terraform creates. 

I had this morning three cases where the actual BDD tests failed, so I guess that left some stuff around.

abayer: Yeah, that's a separate problem with things not getting cleaned up.

abayer:
@hferentschik I'm gonna run `tf-boot` here a few times to see if it doesn't fail.

PR #7429, opened 2020-07-07
refactor: improve build log stream

### LINKED ISSUE #7422: Stabilize & simplify build log stream
### Summary
I propose to refactor the `pkg/logs` package because its current interface and implementation comes with some problems:
* it requires the caller to implement a `LogWriter` interface with methods that accept two channels (one for the log stream and one for an error). This is cumbersome and error-prone for callers to implement (can cause concurrency issues + deadlocks potentially).
* `LogWriter` methods are called from within another goroutine (in case of JXUI this seems to be the reason for a randomly occurring panic when flushing the log stream - currently simply recovered to avoid a complete crash)
* when loading logs from buckets they are loaded into memory before they are actually streamed - causing unnecessarily high memory consumption and, when streamed concurrently, potential crashes due to memory exhaustion.
* the `LogWriter` interface mixes multiple concerns: a) writing logs, b) obfuscating secrets in logs.

Look at
* the logs package: https://github.com/jenkins-x/jx/blob/v2.1.95/pkg/logs/tekton_logging.go
* the `jx get build logs` cmd impl (caller): https://github.com/jenkins-x/jx/blob/v2.1.95/pkg/cmd/get/get_build_logs.go#L189

The JXUI log stream impl is similar to the latter and does not crash in such a case simply because the panic gets recovered now.

### Expected behavior
The logging interface should be simple, reliable and not involve the caller in the selection of the underlying log storage which depends on jx' internal state and should not be the caller's concern.

Therefore I propose to simplify the logging interface as follows:
```
type PipelineActivitySelector struct {
  Owner string
  Repository string
  Branch string
  Context string
  Build string
}
type BuildLogs interface {
  Stream(ctx context.Context, selector PipelineActivitySelector) <-chan LogLine
  Err() error
}
```
(Secret obfuscation could be implemented separately and/or as a decorator.)
The new interface could be used as follows:
```
for line := range buildLogs.Stream(ctx, activitySelector) {
  // obfuscate / write log line to destination
}
if err := buildLogs.Err(); err != nil {
  return err
}
```
Though this is a breaking change which needs to be aligned with all callers/users of the package.

### Actual behavior
Randomly occuring concurrency issues like a panic when flushing a shared writer within a `LogWriter` impl and potential deadlocks, lack of separation of concerns, redundant and cumbersome code in callers - for instance https://github.com/jenkins-x/jx/blob/v2.1.95/pkg/cmd/get/get_build_logs.go#L189

### Jx version

```
Version        2.1.79
Commit         9a5cb37
Build date     2020-06-22T16:31:09Z
Go version     1.13.8
Git tree state clean
```

### Diagnostic information

The output of `jx diagnose version` is:

```
Running in namespace: jx
Version        2.1.79
Commit         9a5cb37
Build date     2020-06-22T16:31:09Z
Go version     1.13.8
Git tree state clean
NAME                          VERSION
Kubernetes cluster            v1.14.10-gke.36
kubectl (installed in JX_BIN) v1.13.2
helm client                   2.12.2
git                           2.17.1
Operating System              Ubuntu 18.04.4 LTS
```

### Kubernetes cluster
v1.14.10-gke.36

Though when we reproduced the panic in JXUI previously it was reliably/often happening on an EKS cluster using the same jx version.

### Kubectl version
The output of `kubectl version --client` is:

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.8", GitCommit:"ec6eb119b81be488b030e849b9e64fda4caaf33c", GitTreeState:"clean", BuildDate:"2020-03-12T21:00:06Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```

Ubuntu 18.04

mgoltzsche:
Improvements:
* let build log methods return channel to let the caller
  handle streamed log lines within its own goroutine.
* stream build log to long-term storage.
* change storage provider interface to stream data.

Closes #7422

#### Submitter checklist

#### Description
improve build log stream

fixes #7422

### DIFF
--- a/pkg/cloud/amazon/storage/bucket_provider.go
+++ b/pkg/cloud/amazon/storage/bucket_provider.go
@@ -1,11 +1,11 @@
 package storage
 
 import (
-	"bufio"
-	"bytes"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/url"
+	"os"
 	"strings"
 
 	session2 "github.com/jenkins-x/jx/v2/pkg/cloud/amazon/session"
@@ -164,7 +164,7 @@ func (b *AmazonBucketProvider) UploadFileToBucket(reader io.Reader, outputName s
 }
 
 // DownloadFileFromBucket downloads a file from an S3 bucket and converts the contents to a bufio.Scanner
-func (b *AmazonBucketProvider) DownloadFileFromBucket(bucketURL string) (*bufio.Scanner, error) {
+func (b *AmazonBucketProvider) DownloadFileFromBucket(bucketURL string) (io.ReadCloser, error) {
 	downloader, err := b.s3ManagerDownloader()
 	if err != nil {
 		return nil, errors.Wrap(err, "there was a problem downloading from the bucket")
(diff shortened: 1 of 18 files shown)


go.sum
```diff
@@ -791,6 +791,7 @@ github.com/jenkins-x/go-scm v1.5.141 h1:sd0A41HYU6i9Kfl5/jH3ldgq4VkSiwuW5+KwvRxT
 github.com/jenkins-x/go-scm v1.5.141/go.mod h1:PCT338UhP/pQ0IeEeMEf/hoLTYKcH7qjGEKd7jPkeYg=
 github.com/jenkins-x/golang-jenkins v0.0.0-20180919102630-65b83ad42314 h1:kyBMx/ucSV92S+umX/V6DDaPNynlFFOM9MGJWApltoU=
 github.com/jenkins-x/golang-jenkins v0.0.0-20180919102630-65b83ad42314/go.mod h1:C6j5HgwlHGjRU27W4XCs6jXksqYFo8OdBu+p44jqQeM=
+github.com/jenkins-x/jx v1.3.1119 h1:yArkuyBqpklDSWaYT8zg0nhdVe6uVQeZRGP2jc2tjI4=
```
abayer, line 794: I'm guessing this ended up in here because you imported something as `github.com/jenkins-x/jx/pkg/...` rather than `github.com/jenkins-x/jx/v2/pkg/...` that has since been deleted?

mgoltzsche: oh, good point...

PR #7455, opened 2020-07-18
refactor: use google uuid package for generating uuids in the codebase

### LINKED ISSUE #7454: Remove satori uuid package
### Summary 
See: https://github.com/satori/go.uuid/issues/84

Checking go.mod, looks like we use three packages for generating uuid.
github.com/pborman/uuid
github.com/satori/go.uuid
github.com/google/uuid v1.1.1

Seems like only the google package has been updated recently. We should remove the other two.

ankitm123:

#### Submitter checklist

Also removed `downloadFile` function, as it was not being used anywhere. It's not an exported function, so safe to remove imo.

fixes #7454

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,8 +3,12 @@ linters-settings:
     list-type: blacklist
     packages:
       - github.com/jenkins-x/jx/v2/pkg/log/
+      - github.com/satori/go.uuid
+      - github.com/pborman/uuid
     packages-with-error-message:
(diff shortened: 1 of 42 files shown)


.golangci.yml
```diff
@@ -3,8 +3,12 @@ linters-settings:
     list-type: blacklist
     packages:
       - github.com/jenkins-x/jx/v2/pkg/log/
+      - github.com/satori/go.uuid
```
garethjevans, line 6: I really like this!

PR #7462, opened 2020-07-20
fix: Read the entire reader before writing to bucket

### LINKED ISSUE #7461: Completed build logs not properly storing in buckets
Seems to be since #7429 went in - in the build controller, we now see things like:

```
{"level":"info","msg":"storing logs for activity jenkins-x-jx-master-970 into storage at jenkins-x/logs/jenkins-x/jx/master/970.log","time":"2020-07-20T12:51:19Z"}
{"level":"error","msg":"failed to store logs for activity jenkins-x-jx-master-970 into storage at jenkins-x/logs/jenkins-x/jx/master/970.log: failed to write key jenkins-x/logs/jenkins-x/jx/master/970.log in bucket gs://logs-oss-weasel-e5201f35d164: short write","time":"2020-07-20T12:51:24Z"}
{"level":"warning","msg":"failed to write key jenkins-x/logs/jenkins-x/jx/master/970.log in bucket gs://logs-oss-weasel-e5201f35d164: short write","time":"2020-07-20T12:51:24Z"}
```

Not sure what went wrong there, but something's definitely awry.

abayer:
#### Submitter checklist

#### Description

For some reason, `io.Copy` was resulting in a "short write" error in the build controller, and the best suspicion I've got at this point is that it's a problem with the bucket writer closing early or something along those lines. So let's instead go back to earlier behavior and read the reader contents and write that directly rather than relying on `io.Copy`.

fixes #7461 (hopefully!)

### DIFF
--- a/pkg/cloud/buckets/buckets.go
+++ b/pkg/cloud/buckets/buckets.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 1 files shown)


ankitm123:
Just a note @abayer: `io.copy` is the recommended way, as readAll is inefficient (reads everything to memory), for now, it's nice to get to a working solution, but may be this can be investigated later.

abayer:
@ankitm123 Yeah, `io.Copy` errors out if number of bytes written doesn't match number of bytes read, while the `ReadAll` approach is more lenient, so my best guess is there's some set of bytes in the logs to write that the blob writer doesn't accept for some reason, causing `io.Copy` to fail. But I dunno.

PR #7468, opened 2020-07-24
fix(pencil): Fixed up local unit test failures for empty commits for …

chrismellard:
…git cherry picks vs git rebases producing different error message text

#### Submitter checklist

#### Description

Unit test fails locally because error message produced  from a cherry picked empty commit is different  from the rebase variant. Modifying tests and underlying implementation to detect this.

Unsure why this doesn't appear to reflect in a successful unit  pipeline run, but this certainly has been failing locally for a while (if I step back in git history locally). Perhaps something to do with my local git version? But having checked the git c source code latest master branch there are definitely differences in these git error messages. Are we locked in to a specific git version for jx?

### DIFF
--- a/pkg/gits/errors.go
+++ b/pkg/gits/errors.go
@@ -15,6 +15,19 @@ func IsEmptyCommitError(err error) bool {
 Otherwise, please use 'git reset'`)
 }
 
+// IsEmptyCherryPickCommitError checks if the error during git cherry-pick is caused by the commit being empty at the end of the cherry-pick
+func IsEmptyCherryPickCommitError(err error) bool {
+	if err != nil {
+		s := err.Error()
+		return strings.Contains(s, `If you wish to commit it anyway, use:
+
+    git commit --allow-empty
+
+Otherwise, please use 'git cherry-pick --skip'`)
+	}
+	return false
+}
+
 // IsRepositoryNotExportedError checks if the clone error happens because the repository is not exported
 func IsRepositoryNotExportedError(err error) bool {
 	if err == nil {
--- a/pkg/gits/helpers.go
+++ b/pkg/gits/helpers.go
@@ -680,7 +680,7 @@ func ForkAndPullRepo(gitURL string, dir string, baseRef string, branchName strin
 	for _, c := range toCherryPick {
 		err = gitter.CherryPick(dir, c.SHA)
 		if err != nil {
-			if IsEmptyCommitError(err) {
+			if IsEmptyCherryPickCommitError(err) {
 				log.Logger().Debugf("  Ignoring %s as is empty", c.OneLine())
 				err = gitter.Reset(dir, "", true)
 				if err != nil {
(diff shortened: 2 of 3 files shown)


MarckK:
/lgtm

Thank you, @chrismellard!

pkg/gits/errors.go
```diff
@@ -15,6 +15,19 @@ func IsEmptyCommitError(err error) bool {
 Otherwise, please use 'git reset'`)
 }
 
+// IsEmptyCherryPickCommitError checks if the error during git cherry-pick is caused by the commit being empty at the end of the cherry-pick
+func IsEmptyCherryPickCommitError(err error) bool {
+	if err == nil {
```
ankitm123, line 20: Personally, I prefer to use err != nil. But I see other functions have similar checks: `err == nil`, so I guess it's fine for now. May be in the future, it can be changed.

chrismellard: Ctrl+C, Ctrl+V ;)

rawlingsj:
looks like we have a network issue ATM - I'm looking at it 
```
                
error pushing image: failed to push to destination gcr.io/jenkinsxio/builder-nodejs:0.0.0-SNAPSHOT-PR-7468-5: Get https://gcr.io/v2/token?scope=repository%3Ajenkinsxio%2Fbuilder-nodejs%3Apush%2Cpull&service=gcr.io: invoking docker-credential-gcr: exit status 1; output: docker-credential-gcr/helper: could not retrieve GCR's access token: oauth2: cannot fetch token: Post https://oauth2.googleapis.com/token: dial tcp: i/o timeout
```

pkg/gits/errors.go
chrismellard: Do you think ```
func IsEmptyCherryPickCommitError(err error) bool {
 	if err != nil {
 		s := err.Error()
 		return strings.Contains(s, `If you wish to commit it anyway, use:

     git commit --allow-empty

 Otherwise, please use 'git cherry-pick --skip'`)
 	}
 	return false
 }``` is more idiomatic?

ankitm123: Yes, wdyt?

chrismellard: Toss a coin for such a small function I reckon. But return early and often as an idiom would suggest for a function testing for certain errors, if no error text is present then a check of a nil error and early return would be more appropriate.

chrismellard:
So digging further the discrepancies between local runs of `jx boot` and its unit tests vs CI is the `git` version. Locally I'm running  2.24.3, Golang docker image is 2.20.1

image

My takeaway is that I do not have a reproducible experience of `jx boot` as my `git` client version will drift compared to an immutable image such as docker. I can fix this specific PR up for this specific use case by string  matching some text that works at both version `2.20.1` and `2.24.3` but that feels a bit near-sighted. Longer sighted question is how do we ensure reproducible `jx boot` experiences with varying git client versions....

rawlingsj:
ok that makes sense, longer term we are looking at running the install as a kubernetes job so we have the same cli versions and repeatable install environments

chrismellard:
+1 for moving this to Kubernetes job. Will close the PR on the grounds that CI (and its docker/git image versions) is gospel truth.

PR #7509, opened 2020-08-07
fix: use apply instead of create when install app

nodece:


MarckK:
This looks good to me. What do you think, @ankitm123, @rawlingsj?

PR #7514, opened 2020-08-10
fix: deprecate create addon command

nodece:

#### Description

addons has been migrated to apps.


MarckK:
Thank you @nodece!

The PR to deprecate addons is very helpful; we just need to be mindful of our deprecation timetable for removing specific functionality and make sure these changes are fully communicated in advance.  @rawlingsj

PR #7552, opened 2020-09-02
fix: Bring back config refdocs generation

abayer:
This got moved to jx-api, but won't actually _work_ there, so we need to bring it back into jx.git or the docs site will never get automatically updated.


abayer:
/assign @jstrachan 
/assign @garethjevans 
/assign @MarckK 

I'm gonna have a PR up to jenkins-x-versions momentarily to go with this, but it won't actually _work_ per se until this has merged.

PR #7567, opened 2020-09-16
chore: migrate code.gitea.io/sdk/gitea to v0.13.0

6543:
just migrate to the new sdk

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/jenkins-x/jx/v2
 
 require (
-	code.gitea.io/sdk/gitea v0.12.0
+	code.gitea.io/sdk/gitea v0.13.0
 	github.com/Azure/draft v0.15.0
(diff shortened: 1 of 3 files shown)


abayer:
Failure is unrelated - it's due to an athens proxy issue. I'm gonna have a PR up for that momentarily.

abayer:
I'll retest this once #7569 has merged.

abayer:
@6543 Once #7569 merges, this is also going to need to bump the `github.com/jenkins-x/lighthouse` version to `v0.0.824` or later, since jx depends on Lighthouse and Lighthouse depends on go-scm, which itself depends on gitea's sdk too. =) Lighthouse is updated, so just need to add that bump here. I'll give it a try locally to make sure it works.

abayer:
Actually, looks like that may not be necessary - if the tests pass, then this is good on its own.

/lgtm

PR #7569, opened 2020-09-16
fix: Workarounds for go modules that fail with Athens proxy

abayer:


abayer:
Had to fix bdd-jx so we're gonna see some failures that I'll rerun.

PR #7575, opened 2020-10-02
fix: fixes #7574: import from gitlab with complicated project name

### LINKED ISSUE #7574: jx3 fails to add project hook in gitlab if project name contains spaces...
### Summary
`jx admin operator` fails on gitlab if project name looks like this: `Jx3 kubernetes` the code checks the repo name instead of the repo path against the SourceRepositorySpec.repo

`jx admin operator` from a gitlab project named as `Jx3 kubernetes` with jx3/jx-cli  
Or just `jx update webhooks` when using jx2

### Expected behavior
It should create the webhook

### Actual behavior
It fails

### Jx version

```
version: 3.0.428
```

PR coming

yelhouti:

#### Submitter checklist

#### Description
The projectId was retrieved by checking the `SourceRepositorySpec.Repo` against Project.Name instead we should check against: `Project.Path` to work also for more complicated project names with upper case spaces...

fixes #7574

### DIFF
--- a/pkg/gits/gitlab.go
+++ b/pkg/gits/gitlab.go
@@ -199,7 +199,7 @@ func (g *GitlabProvider) projectId(org, username, name string) (string, error) {
 	}
 
 	for _, repo := range repos {
-		if repo.Name == name {
+		if repo.Path == name {
 			return strconv.Itoa(repo.ID), nil
 		}
 	}


yelhouti:
Hi @daveconde , I'm upgrading to jx3 end fixing bugs on my way there. bot says "Regular contributors should join the org to skip this step." after 13 PRs in this project alone do I count as regular :p ? thanks any way ! Have a nice day

PR #7578, opened 2020-10-05
fix: set a bigger memory limit for the meta-pipeline

### LINKED ISSUE #7577: The meta-pipeline needs more memory for git operations on big repositores
### Summary

We have a git repo for which sometimes the meta-pipeline fails on a git operation: it is being OOM-killed.
The container `step-create-effective-pipeline` is OOMKilled when running `git fetch origin master`.

The default memory limit is 512M - see https://github.com/jenkins-x/jx/blob/v2.1.149/pkg/tekton/metapipeline/metapipeline.go#L149-L152

maybe we could at least set a bigger limit, for big repos? or make it configurable. It will be difficult to have it on a per-repo basis, but at least on a per-install.

### Jx version

```
2.1.149
```

vbehar:
the previous default value (512M) was not enough to run `git fetch` on big repositories

fixes #7577


vbehar:
@jstrachan what do you think? except for the "with jx3 no more meta pipelines" ;-)

PR #7632, opened 2021-04-01
chore: update code.gitea.io/sdk/gitea to v0.14.0

6543:
as title


6543:
@jstrachan anything missing?

6543:
If there is no interest ... im going to close it

PR #7867, opened 2021-07-01
fix: Install vault in image by default

wowq:
If use vault as secretstorage. need to install vault plugin every time when run jx-boot in jx-git-operator


wowq:
@jstrachan

PR #8202, opened 2022-06-07
Define maintainers

msvticket:
Define maintainers alias to be used from this and other repos

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,6 +1,6 @@
 aliases:
-- rawlingsj
-best-approvers:
-- rawlingsj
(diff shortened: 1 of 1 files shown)


msvticket:
On the other hand: it might be better to put this OWNERS_ALIASES in some repo that doesn't trigger an actual build. The question is which one...

msvticket:
Huh, I haven't noticed this repo before https://github.com/jenkins-x/issues

msvticket:
Ah, maybe https://github.com/jenkins-x/jx-community would be a good place for a centralised OWNER_ALIASES?

PR #8203, opened 2022-06-07
Use maintainers from jx-community

msvticket:
Use maintainers alias from jx-community


msvticket:
Now I’ve added Gareth and the Jameses to jx-community

PR #8223, opened 2022-06-15
fix: Upload release notes

msvticket:
This is an attempt to fix release notes

I don't really know why it doesn't work when using this GitHub Action when it works fine in the tekton pipelines. My hypothesis for this attempt is that the file name is case sensitive here and not in the tekton pipeline.


msvticket:
Bugger. That apparently didn't do the trick. I'll investigate further. Some other time...

Issue #8249, opened 2022-06-23
Commit sha and go version in version


ankitm123:
We should have a commit sha and golang version displayed for the version.

PR #8284, opened 2022-07-05
modifying jx version command

osamamagdy:
SOLVES #8249 
This is still a WIP. Writing proper unit tests for copra functions.


osamamagdy:
@ankitm123

ankitm123:
Can we have struct/map to hold all the version related information instead of strings?
Example: https://github.com/helm/helm/blob/b1e24764325135b364a7f81fb0aab683d190c546/internal/version/version.go

osamamagdy:
Okay. What about using cmd.println instead of logger?

osamamagdy:
I added the struct type and edited the Makefile accordingly

ankitm123:
> What about using cmd.println instead of logger?

I prefer using io.Writer as it's an interface. You can pass any struct which implements the writer interface for testing.

osamamagdy:
@ankitm123 
> > What about using cmd.println instead of logger?
> 
> I prefer using io.Writer as it's an interface. You can pass any struct which implements the writer interface for testing.

but how should the output buffer inside the cmd.Execute() using io.Writer() be mapped to the buffer used in compare tests?
I am thinking about creating the buffer and passing it as a parameter to the NewCmdVersion() and the Run() function ?

pkg/cmd/version/version.go
```diff
@@ -48,19 +67,51 @@ func NewCmdVersion() (*cobra.Command, *Options) {
 }
 
 // Run implements the command
-func (o *Options) Run() error {
+func (o *Options) Run(cmd *cobra.Command) error {
```
ankitm123, line 70: Instead of passing cobra.Command, just pass the io writer?
Example: https://github.com/helm/helm/blob/b1e24764325135b364a7f81fb0aab683d190c546/cmd/helm/version.go#L82

osamamagdy: This will cause the change of the root.go file as it needs to pass the buffer to it also?
https://github.com/jenkins-x/jx/blob/88d12dd14c8cb7ea0afd66a63e6bc6c425776519/pkg/cmd/root.go#L62

ankitm123: That should be fine imo.

ankitm123: or u can pass the writer as a field in the options struct?

osamamagdy: How would this affect the root.go file? Should I just create the buffer and Pass it to NewCmdVersion() and the output will be printed through fmt.Fprintf() inside Run()?

ankitm123: It wont change root.go.
You can change the options struct in version package:
```
type Options struct {
	Verbose bool
	Quiet   bool
	Out     io.Writer
}
```

and in Run function, check if o.Out is nil, if it's nil set it to os.Stdout. In your tests, you can set o.Out to a buffer and go with ur checks.

For printing you can use `fmt.Fprintf`

osamamagdy: I think it is done now

pkg/cmd/version/version.go
```diff
@@ -1,34 +1,58 @@
...
 	Revision  string
 	Branch    string
 	BuildUser string
 	BuildDate string
 	GoVersion string
-)
+}
+
+//Instance of the build info
+var buildInfo BuildInfo
```
ankitm123, line 23: any reason to have a global variable?

ankitm123: nvm, I see it.

pkg/cmd/version/version.go
```diff
@@ -1,34 +1,58 @@
...
 	"github.com/spf13/cobra"
 )
 
 // Build information. Populated at build-time.
-var (
+type BuildInfo struct {
 	Version   string
 	Revision  string
 	Branch    string
 	BuildUser string
```
ankitm123, line 17: What is build user? Also git tree state is missing ...

pkg/cmd/version/version.go
osamamagdy: Isn't this the way to set using the makefile?

pkg/cmd/version/version.go
```diff
@@ -1,34 +1,58 @@
...
+	TestBuildDate = "2020-05-31T14:51:38Z"
+
+	// TestGoVersion can be used in tests if no version is passed in the test flags
+	TestGoVersion = "1.13.8"
 )
 
 // ShowOptions the options for viewing running PRs
 type Options struct {
 	Verbose bool
 	Quiet   bool
```
ankitm123, line 54: [You can do this in a separate PR if you want] Lets deprecate Quiet (dont remove it, as it is used downstream), and add a new short field which does the exact thing as quiet. Then set a date when it's scheduled for deletion. Also when an user tries --quiet or -q, let them know that the flag is deprecated and will be removed on a date, and they should instead use --short.

Quiet does not sound correct.

osamamagdy: The reason for this is only to make naming sounds better? Also, what is the data for deletion?

ankitm123: > The reason for this is only to make naming sounds better? 

I have not seen many cli which have a quiet flag to only show version string. in helm/kubectl, it's always short (normally they have verbose to show more things). 

> Also, what is the data for deletion?

let's do 3 months - sort of a standard lifecycle to deprecate a cli flag imo.

PR #8299, opened 2022-07-12
fix: inherit go version from catalog

### LINKED ISSUE #8297: jx version may show incorrect golang version
jx version may show incorrect golang versions.
We hard code it to a value: https://github.com/jenkins-x/jx/blob/366fb476158989cf1f12d9cbe6f722bca635cda6/.github/workflows/jenkins-x/upload-binaries.sh#L16
It should instead derive the value from the image it's using to run the goreleaser step in the release pipeline.

Locally we get it from the environment: https://github.com/jenkins-x/jx/blob/366fb476158989cf1f12d9cbe6f722bca635cda6/Makefile#L16

As an additional issue, I think we should enforce some sort of conditon so that the go version in the dev machine matches the version in the catalog.

The benefit is that we can change this value in a PR and trigger a new jx build when the go version in the  catalog changes. We actually do this for most of the repos (we really need to have similar set up in all repos, this non-standard thing will make things harder to debug and maintain)

osamamagdy:
Signed-off-by: osamamagdy <osamamagdy174@gmail.com>

fixes #8297 

Here we are making the go version in the pipeline inherited from the image running gorealeaser.
In the makefile we ensure that GO_VERSION == CATALOG_GO_VERSION and error the make file if not.
To Do: How to get the CATALOG_GO_VERSION value?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,13 @@ ORG := jenkins-x
 ORG_REPO := $(ORG)/$(NAME)
 RELEASE_ORG_REPO := $(ORG_REPO)
 ROOT_PACKAGE := github.com/$(ORG_REPO)
+
+# This version is just used to trigger a new build in case we update the version of go in jx3-pipeline-catalog, and dont have new PRs which use the updated version in the catalog.
+# This does not reflect the go binary version which was used to build the jx binary, and also does not reflect the version in the catalog.
+# The sole purpose of this variable is to build a new binary if we ever need to build a new jx binary with a new go version with no code change.
+# If you notice that this version is not the same as the catalog version, please open a PR, the maintainers are happy to review it.
+DUMMY_GO_VERSION := 1.17.11
+
 GO_VERSION := $(shell $(GO) version | sed -e 's/^[^0-9.]*\([0-9.]*\).*/\1/')
 GO_DEPENDENCIES := $(call rwildcard,pkg/,*.go) $(call rwildcard,cmd/,*.go)
 
--- a/.github/workflows/jenkins-x/upload-binaries.sh
+++ b/.github/workflows/jenkins-x/upload-binaries.sh
@@ -13,7 +13,7 @@ git config --global user.email "jenkins-x@googlegroups.com"
 export BRANCH=$(git rev-parse --abbrev-ref HEAD)
 export BUILDDATE=$(date)
 export REV=$(git rev-parse HEAD)
-export GOVERSION="1.17.9"
+export GOVERSION="$(go version | sed -e 's/^[^0-9.]*\([0-9.]*\).*/\1/')"
 export ROOTPACKAGE="github.com/$REPOSITORY"
 
 goreleaser release
--- a/.github/workflows/plugins-pr.yaml
+++ b/.github/workflows/plugins-pr.yaml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v3
         with:
-          go-version: '1.17.9'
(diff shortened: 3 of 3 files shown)


osamamagdy:
Checks are failing as I am putting `1.17.8` as the Go version.

Makefile
```diff
@@ -13,6 +13,8 @@ ORG := jenkins-x
 ORG_REPO := $(ORG)/$(NAME)
 RELEASE_ORG_REPO := $(ORG_REPO)
 ROOT_PACKAGE := github.com/$(ORG_REPO)
+
+CATALOG_GO_VERSION := 1.17.8
```
ankitm123, line 17: It's not 1.17.8. This is where the value is: https://github.com/jenkins-x/jx3-pipeline-catalog/blob/f46083677cfc51f6e6e2f04af5ed2ab6b5068719/tasks/go/pullrequest.yaml#L37

osamamagdy: @ankitm123 Yes, I was trying to see if checks will fail so I made it something different. But my question is if this value should be from this file https://github.com/jenkins-x/jx3-pipeline-catalog/blob/f46083677cfc51f6e6e2f04af5ed2ab6b5068719/tasks/go/pullrequest.yaml
or it can be hardcoded in the make file

ankitm123: It's from that file. We can use updatebot to change the catalog_version in makefile.

Makefile
```diff
@@ -44,6 +46,11 @@ BUILDFLAGS :=  -ldflags \
 # Some tests expect default values for version.*, so just use the config package values there.
 TEST_BUILDFLAGS :=  -ldflags "$(BUILD_TIME_CONFIG_FLAGS)"
 
+ifneq ($(GO_VERSION),$(CATALOG_GO_VERSION))
+$(error The local version of GO doesn't match the one in pipeline catalog. Please set Go to $(CATALOG_GO_VERSION) )
```
ankitm123, line 50: Can we only warn, and not throw error?

osamamagdy: I thought error would be better to "enforce" the user to upgrade. Why you think warn is better?

ankitm123: We dont have updatebot in place yet (for triggering builds), so just failing it for now, will break things in unexpected way, warn is ok for now ...
Also updatebot may not work the way I expect it to - it may still require some manual touch ...

osamamagdy: @ankitm123 Great, done

Makefile
```diff
@@ -44,6 +46,11 @@ BUILDFLAGS :=  -ldflags \
 # Some tests expect default values for version.*, so just use the config package values there.
 TEST_BUILDFLAGS :=  -ldflags "$(BUILD_TIME_CONFIG_FLAGS)"
 
+ifneq ($(GO_VERSION),$(CATALOG_GO_VERSION))
```
ankitm123, line 49: Thinking abt this a bit more, I think this can lead to maintenance issues, and builds failing unnecessarily. So let's remove this check, but keep CATALOG_GO_VERSION (may be rename it to DUMMY_GO_VERSION to avoid any confusion) with a comment
```
This version is just used to trigger a new build in case we update the version of catalog, and dont have new PRs which use the updated version in the catalog. This does not reflect the go binary version which was used to build the jx binary 
```

osamamagdy: Done

ankitm123 (approved):

ankitm123:
can u rebase with the latest master? I think the issue is with a newer version of codecov, which I will fix soon.

ankitm123:
/override pr
The issue is fixed already in latest master

PR #8369, opened 2022-09-19
fix: Upgrade jx-helpers to show plugins in jx --help

msvticket:
Also upgrading other dependencies


ankitm123 (approved):
I am too late to the party! LGTM

PR #8400, opened 2022-09-30
fix: add missing curl to jx-go alpine image

ankitm123:

alpine does not have curl by default, this was breaking all release pipelines.


rajatgupta24 (approved):
LGTM

Issue #8411, opened 2022-10-02
Fully support running jx on arm


ankitm123:
With all major cloud providers supporting arm, I think it's important we support running Jenkins X on arm devices like AWS graviton, raspberry pi etc.
We publish arm binaries for jx cli and jx plugins. But our docker images don't fully support arm yet (they are mostly built on amd64).
Some of the third party images we use also don't support ARM atm: https://github.com/bitnami/charts/issues/7305

This issue will be used to track arm support in jx.

ankitm123:
pusher wave (used to keep pods in sync when a configuration or secret changes) also has issues:
```
k logs pusher-wave-pusher-wave-668c977679-59m5v
exec /bin/wave: exec format error
```

ba11b0y:
I'd like to contribute to this.
From what I know from Slack is we also want to migrate to ko for building images. Maybe I can start small with the one that is not supported right now and then gradually move all. Where should I be starting from?

Also some questions/comments around the same:
 - For wave, it seems that we'll have to create our own image and use it.
 - While running jx locally on my M1 machine, I found that `lighthouse-tekton-controller` also isn't supported.
 - Is https://github.com/jenkins-x/jenkins-x-arm-support related to this?

langleyd:
Arm support  has landed in https://github.com/bitnami/charts/issues/7305 now @ankitm123

msvticket:
Am I correct in assuming that it is actually arm64 that is requested in this issue? Or is arm (i.e. 32 bit) also requested? The reason I ask is that compiling for both if it actually isn't needed would be a waste of resources.

PR #8412, opened 2022-10-02
feat: create multiarch kubectl image for jx

ankitm123:

Related to #8411.

Bitnami images don't support ARM yet, and I am not sure when that will land. We can build and maintain our own kubectl image for our usecase. The image is simple to build and smaller as it uses alpine.
https://github.com/jenkins-x/jx3-versions/blob/be23d4f583b765e832a0046e0ffd4f12be3cd21d/charts/external-secrets/kubernetes-external-secrets/values.yaml.gotmpl#L46 -> This is the current image.

Rancher also publishes it's own kubectl image, but we cant use it, as it's based on scratch, and we need base64 decode in the init script: https://github.com/rancher/kubectl/blob/ecb5988682864a252e55600fcca87b6d3d3d4312/Dockerfile#L12

### DIFF
--- a/.github/workflows/jenkins-x-release.yaml
+++ b/.github/workflows/jenkins-x-release.yaml
@@ -141,6 +141,18 @@ jobs:
         registry: ghcr.io
         username: jenkins-x
         password: ${{ secrets.GHCR_TOKEN }}
+    - name: Build and push kubectl
+      uses: docker/build-push-action@v3
+      with:
+        context: .
+        file: ./Dockerfile-kubectl
+        platforms: linux/amd64,linux/arm64
+        push: true
+        build-args: |
+          VERSION=1.20.2
+        tags: |
+          ghcr.io/jenkins-x/kubectl:latest
+          ghcr.io/jenkins-x/kubectl:1.20.2
     - name: Build and push jx-go
       uses: docker/build-push-action@v3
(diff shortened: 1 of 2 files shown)


rajatgupta24 (approved):
LGTM

tomhobson (approved):

msvticket:
This would fit better in the jenkins-x-images. It's unnecessary to have to trigger a new build of jx when needing a new kubectl image or vice versa.

Now, a problem is that we don't have permissions in that organisation, right?

ankitm123:
> It's unnecessary to have to trigger a new build of jx when needing a new kubectl image or vice versa.

Agreed!

> Now, a problem is that we don't have permissions in that organisation, right?

Yes, but I have asked James for access.

I am not a big fan of having multiple orgs though, it is a pain to keep track of things (and giving access), I preferred the v2 days of having everything in a single org - there used to be repos for builder images. May be we can do that?

Alternatively, we can trigger this kubectl image build from github actions manually (from the UI), so that it's not created on every build.

msvticket:
I think it's easier to have it in a separate repo. But to me it's not necessary that the repo is in jenkins-x-images.

ankitm123:
I will revert this change, and will create a new repo today!

PR #8442, opened 2022-10-23
feat: attempt to generate provenance using slsa generator

ankitm123:
Provenance is basically where, when and how our artifacts were produced.
See: https://slsa.dev/provenance/v0.2


ankitm123:
If this fails, will most likely revert the change ...

PR #8443, opened 2022-10-24
fix: change registry org to jx

ankitm123:
- Also uses crane to get the digest sha, as cosign prefers digest sha over tags (they will remove the ability to use tags in the future)
- Removed jenkins-x-pr.yaml as it's not being used


ankitm123:
cosign warning message:
```
ARNING: Image reference ghcr.io/jenkins-x-bot-test/jx: uses a tag, not a digest, to identify the image to sign.

This can lead you to sign a different image than the intended one. Please use a
digest (example.com/ubuntu@sha256:abc123...) rather than tag
(example.com/ubuntu:latest) for the input to cosign. The ability to refer to
images by tag will be removed in a future release.
```

babadofar (approved):

PR #8465, opened 2022-11-14
docs: add instructions to verify archive using cosign with keyless signing

osamamagdy:
Signed-off-by: osamamagdy <osamamagdy174@gmail.com>


msvticket:
For the record you don't need to repeat the option -L to curl. But never mind...

/approve

PR #8496, opened 2023-01-04
feat: add yq to boot image

msvticket:
Even though a lot of juggling with yaml files are possible with `jx gitops` I keep encountering situations when yq would be solve a problem in a simpler than for example adding the functionality to `jx gitops` and a less fragile way than writing an awk or shell script.

In a pipeline I can add a step using a dedicated yq image, but in the jx boot job that isn't really possible.

### DIFF
--- a/Dockerfile-boot
+++ b/Dockerfile-boot
@@ -11,7 +11,7 @@ LABEL maintainer="jenkins-x"
 RUN addgroup -S app \
     && adduser -S -g app app \
     && apk --no-cache add \
-    ca-certificates curl git git-lfs make netcat-openbsd bash
+    ca-certificates curl git git-lfs make netcat-openbsd bash yq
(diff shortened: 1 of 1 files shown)


ankitm123 (approved):
Makes sense to use yq for this.

Dockerfile-boot
```diff
@@ -11,7 +11,7 @@ LABEL maintainer="jenkins-x"
 RUN addgroup -S app \
     && adduser -S -g app app \
     && apk --no-cache add \
-    ca-certificates curl git git-lfs make netcat-openbsd bash
+    ca-certificates curl git git-lfs make netcat-openbsd bash yq
```
ankitm123, line 14: minor NIT: May be we can have each package in a separate line to improve readability.

PR #8498, opened 2023-01-06
Fix : Function call can be replaced with helper function

nishantsikarwar:
Better to use ReplaceALL 

https://github.com/nishantsikarwar/jx/blob/10d64c7e16be9f4ad6103ea258fcf52b679584ce/cmd/docs/man_docs.go#L68

instead of using replace 

https://github.com/jenkins-x/jx/blob/01894a7dd4f29bfa5817ce7230ad98d5962da8b4/cmd/docs/man_docs.go#L68


nishantsikarwar:
@msvticket @ankitm123 

can you guys pls help in the review process of this PR 

best regards,
Nishant

msvticket (approved):

PR #8615, opened 2023-09-14
feat: update tf versions in dockerfiles

mentlak0:
- upgrade TF versions in docker images

### DIFF
--- a/Dockerfile-tfo-aws
+++ b/Dockerfile-tfo-aws
@@ -14,7 +14,7 @@ RUN curl -o aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/
     mv aws-iam-authenticator /usr/bin/aws-iam-authenticator
 
 # terraform
-ENV TERRAFORM 1.1.2
+ENV TERRAFORM 1.4.6
 RUN curl -LO https://releases.hashicorp.com/terraform/${TERRAFORM}/terraform_${TERRAFORM}_linux_amd64.zip && \
   unzip terraform_${TERRAFORM}_linux_amd64.zip && \
   chmod +x terraform && mv terraform /usr/bin && rm terraform_${TERRAFORM}_linux_amd64.zip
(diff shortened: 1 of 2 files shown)


mentlak0:
/ cc @tomhobson

msvticket (commented):
What is the purpose of this PR? As far as I can see these Dockerfiles are not in use any more.

mentlak0:
@msvticket Hi, I am upgrading terraform-jx-azure to be up to date with the azurerm provider and terraform version since the k8s 1.27+ upgrades (there are a few resource behaviour changes).

The image being changed in the PR gets pulled in to the terraform pipelineRun 

https://github.com/jenkins-x/jx3-pipeline-catalog/blob/master/tasks/terraform/pullrequest.yaml

The version is set at 1.12.0, therefore when trying to upgrade it fails.

msvticket:
But the image is not built anymore. As you can see [here](https://github.com/jenkins-x/jx-cli/pkgs/container/terraform-operator-gcp) no image has been pushed in over a year. First it was commented out from jenkins-x-release.yaml with an initial comment saying: 

    # Commenting these for now, as custom images are a bit complicated to do in the new tf operator versions

Later the block was removed entirely.

mentlak0:
@msvticket Thanks for that. I see, this may be a problem then with trying to upgrade terraform versions and trying to avoid drift etc. Do you have any suggestions?

msvticket:
Good question. One option would be to upgrade the terraform operator. New versions seem to have a design where adding custom plugins (actually jobs) with custom images is easy, so we should be able to use the default image for the terraform stuff and then use for example the jx-boot image for the equivalent of the postrunScript (which isn't supported anymore). 

See http://tf.galleybytes.com/

There is supposed to be a migration guide, but I can't find it. I created an issue regarding that: GalleyBytes/terraform-operator#163

mentlak0:
@msvticket Thanks for the reply. Especially with k8s default version being set (and upgrades made) on the reports. It is important to maintain changes in the Terrform to reflect work that was completed. Will look into the suggestions you have made in due time.

PR #8768, opened 2025-02-11
fix: plugin usage strings

msvticket:
Since cobra only use the usage string up until the first space as command name in help I change to use non-breakable space in the env variables some plugins (for example jx-gitops) use to set use.

https://github.com/spf13/cobra/blob/v1.8.1/command.go#L1503


msvticket:
A confusing drawback with this fix is that if you copy the command from the help and paste in a shell it won't work, since nbsp isn't interpreted as space by the shell either.

Instead the fix is to use a cobra annotation to set the name. See jenkins-x-plugins/jx-gitops#1001

/close

Issue #7970, opened 2021-10-14
Indicate errors in configuration when importing a project


Shiti:
I had a typo in the `triggers.yaml` file. The name of the source for pullrequest file was incorrect. 
```yaml
apiVersion: config.lighthouse.jenkins-x.io/v1alpha1
kind: TriggerConfig
spec:
  presubmits:
    - name: pr
      context: "pr"
      always_run: true
      optional: false
      source: "pull-request.yaml"  # This should have been pullrequest.yaml
```

I was able to import the project using the `jx import` command but the tests were not running on creating pull requests. It took sometime to figure out why. 

It will be helpful if the `jx import` command gave a warning or error message. One temporary solution I can think of is adding a note in docs to execute `jx pipeline lint` or something similar before doing import.

samar-syd7:
Hi, I'd like to work on this issue if it's still relevant.

I noticed the issue is still open and would like to add validation during project import so invalid pipeline references in triggers.yaml are detected early.

Before I start, could someone confirm whether this problem still exists in the current codebase and whether a contribution would be welcome?

msvticket:
Hi!

I would say that this is still valid and that it is a reasonable functionality to add. So, yes, it would be welcome. If you plan to add several validations I suggest you do it in increments to avoid unnecessary large PRs.

hiSandog:
The first increment could validate every referenced pipeline source during import by resolving it relative to `triggers.yaml`, then report all missing/unreadable files in one pass instead of failing on the first typo. A missing required presubmit source should be an import error rather than a warning, because accepting it creates a project that appears healthy but never runs checks. Unit fixtures for valid relative paths, renamed files, and path traversal outside the project root would make the rule precise.

tomatotomata:
I was thinking of starting with the low-risk increment you suggested: document that `jx import` can finish without surfacing invalid trigger sources, and recommend running `jx pipeline lint` before relying on the imported pipeline. I will keep this limited to the user-facing guidance first, since the command implementation lives behind the project plugin. Let me know what you think.

PR #9051, opened 2026-08-02
docs: explain linting after project import

tomatotomata:
## Summary

- document the `jx pipeline lint` check after `jx import`
- call out invalid trigger source references, including typos in `triggers.yaml`
- keep this first increment limited to user-facing guidance as discussed in #7970

## Verification

- `git diff --check`
- Go-based documentation generation was not run because Go is not installed in this environment

This is the documentation-first increment suggested in #7970. The import implementation is provided by the project plugin, so the note gives users a concrete validation step while the implementation work can be handled separately.

### DIFF
--- a/docs/cmd/jx_import.md
+++ b/docs/cmd/jx_import.md
@@ -12,6 +12,19 @@ jx import
 
 alias for: jx project import
 
+### Validate the imported pipeline
+
+`jx import` can finish even when a trigger configuration refers to a file
+that does not exist. Before relying on the imported pipeline, run the pipeline
+lint command from the project directory:
+
+```
+jx pipeline lint
+```
+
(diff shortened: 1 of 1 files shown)


msvticket:
The file you are trying to change is generated automatically (as it says in the bottom). 

If you would like to improve documentation that isn't about one specific command do check out the [website](https://jayex.io/) and see what you want to improve. Maybe here: https://jayex.io/v3/develop/create-project/#import-an-existing-project

msvticket:
If you want to change the output of `jx project import --help` that change would go somewhere here: https://github.com/jenkins-x-plugins/jx-project/blob/main/pkg/cmd/importcmd/import.go#L134

